PR #1, opened 2017-11-21
Remove intermediate module

flying-sheep:
I hope the way I did it has two effects:

1. File history is preserved (by deleting `__init__.py` before renaming `anndata.py`). Note that I mean [blame](https://github.com/theislab/anndata/blame/6d5cfd6cbb2e929fac0100547ebb6dab9bf0001a/anndata/__init__.py) (which works), not [history](https://github.com/theislab/anndata/commits/6d5cfd6cbb2e929fac0100547ebb6dab9bf0001a/anndata/__init__.py) (which is rather useless)
2. Autocompletion still works well (by adding `__all__`). We should keep in mind ipython/ipython#10649 here: Some people prefer seeing everything while others don’t.

   It’s configurable via <code>c.[IPCompleter.limit_to__all__](http://ipython.readthedocs.io/en/stable/config/options/terminal.html#configtrait-IPCompleter.limit_to__all__) = True</code>


falexwolf:
The thing with the history does not seem to work. Click on "history"

https://github.com/theislab/anndata/commits/6d5cfd6cbb2e929fac0100547ebb6dab9bf0001a/anndata/__init__.py

and you'll see that the changes to the structure of AnnData are not there.

Before merging: are we sure that AnnData and the helper functions around it will never become so complicated that we have to setup a more general file structure? If yes, we should maybe consider renaming to `base.py` instead of to `__init__.py`, which is the init of the whole package. I think it's nice if a package only makes those parts of the subpackges available to the user that the user is actually likely to need. In the case of the `anndata` package, this is currently only the `AnnData` class.

Still, what I have in mind is to move all the basic read-write, the basic plotting, and the basic mathematical operations from Scanpy to AnnData, as already mentioned in the README.

anndata would then have submodules `readwrite`, `plotting` and `maths` (or `preprocessing`). All of this is independent of single-cell stuff and simply practical, simple operations that you would like to have with data container. all of this also not a single (!) additional dependency and hence anndata would remain a small package that can be quickly installed to read anndata files, plot them and perform simple mathematical operations.

In the light of these plans, I would prefer to keep AnnData either in `anndata` or in `base` or something similar and use `__init__` solely for the purpose of providing the user some toplevel functions.

Btw: one could also think of adding all these simple functions as attributes to AnnData, but this could be discussed.

flying-sheep:
> The thing with the history does not seem to work.

uh, i think you skipped the third sentence i wrote.

>  If yes, we should maybe consider renaming to base.py instead of to __init__.py, which is the init of the whole package.

maybe we should think about if we can split it up right now. all the helper stuff in a module, `BoundRecArray` in another, and only leave some imports and `class AnnData` in `__init__.py`.

we can also simplify the code in AnnData, so `__init__.py` is smaller. either by putting more code into helpers, or by using a base class that contains all the implementation details, and leaves `AnnData` as just a skeleton with all the public methods.

falexwolf:
Let's do as you proposed above! I'll close this stub pull request for now.

PR #5, opened 2017-12-28
Function to read umi_tools output

dawe:
`umi_tools` is a handy set of utilities to manage scRNA-seq experiments (among other thigs). The final results of a [typical single cell analysis](https://github.com/CGATOxford/UMI-tools/blob/master/doc/Single_cell_tutorial.md) is a condensed count matrix saved in a gzipped tab-separated file. This cannot be read with `read_mtx` function, probably due to a different header. 
I've written a quick function to import umi_tools datasets. If included in the main code, please add it also in the available possibilities in `scanpy.api.read` function

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -45,6 +45,7 @@ Reading other file formats.
    read_loom
    read_mtx
    read_text
+   read_umi_tools
 
 
 Writing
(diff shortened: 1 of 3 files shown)


falexwolf:
Thank you very much!

PR #8, opened 2018-01-06
Prettify autodoc

flying-sheep:
So there’s three functions that work to create those signatures

- `autodoc.formatargspec` and `autodoc.format_annotation` create pretty python code to be displayed in the docs
- `autosummary.mangle_signature` then condenses them down for the overview table with regular expressions, but is buggy and can’t handle type annotations

since `autodoc.formatargspec` creates valid python code anyway, I just used the AST parser module to rewrite `mangle_signature` (and replaced it in the module with my version).

If we don’t want return types at all, we could remove the first three lines here (and the whole `unparse` function)

https://github.com/theislab/anndata/blob/abc1badb2ec857959f66942c5396b9249ed5f331/docs/conf.py#L277-L280

### DIFF
--- a/anndata/readwrite/read.py
+++ b/anndata/readwrite/read.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from typing import Union, Optional, Iterable, Generator, Iterator
 
 import h5py
 import numpy as np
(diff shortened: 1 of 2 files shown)


falexwolf:
Wow, cool! Very nice! :smile:

The problem is just that building the docs now fails: https://readthedocs.org/projects/anndata/builds/6508415/

flying-sheep:
damn, i thought readthedocs was python 3.6 by now.

i’ll just ~~switch it to use 3.6~~ not use f-strings.

readthedocs for some reason doesn’t support 3.6 by now, but my little f helper function is almost as pretty

PR #18, opened 2018-04-18
Fix docs

flying-sheep:
[rendered](http://anndata.readthedocs.io/en/docs_b/anndata.AnnData.html)

1. I replaced the manual specification of Attribute/Method tables with autogenerated ones

    1. This means that I changed the class.rst hack – instead of invisibly linking everything, it now does so visibly
    2. I made the code for the deprecations much more DRY, and use this to omit them in `vars(AnnData)` and therefore the autogenerated index

2. I allowed to do `AnnData()` to create a 0×0 object.
3. I switched the buggy numpydoc for sphinx.ext.napoleon

    1. This way I could also get rid of the error-prone signature mangling (just not needed)
    2. This allows to move type signatures out of the docstring into the function annotations
    3. I now use Sphinx’ own mechanism to replace `:param:` directives in order to get pretty parameter documentation (styled multiline definition lists like numpydoc instead of the default single line bullet lists)


flying-sheep:
One remaining problem: For some reason, the attributes render flawlessly locally, but are missing online. No idea why…

image

falexwolf:
Wow, this is really cool. :smile: A lot of improvements! When we find a solution for the missing attributes, which are, of course, important, we can immediately merge this.

As you're not using numpydoc anymore, you also don't have the parameters-styled attributes, right? Maybe online there is a different "napoleon" version that doesn't recognize the "attributes" keyword??

PS: Sorry about the conflict in `utils.py`... It should be trivial to resolve, though. We've spent quite some side-by-side coding here with the Andrea and Rahul from Seurat to get a converter between Scanpy and Seurat working... This required some immediate action - for some reason I needed to even put some releases on PyPI as their `reticulate` environment wasn't able to detect the GitHub based installation of `anndata`...

flying-sheep:
seems to have been a fluke, after rebasing onto master (for the utils.py changes), the attributes are there!

maybe we need to clean the build directory or so…

PR #19, opened 2018-05-02
Improve docs, remove types where they duplicate typehint info

flying-sheep:
- Adds more type hints
- Remove type hints from docs


anndata/base.py
```diff
@@ -208,9 +208,6 @@ def _fix_shapes(X, single_col=False):
     elif len(X.shape) == 1:
         n_vars = X.shape[0]
         n_obs = 1
-    elif len(X.shape) == 1 and single_col:
```
falexwolf: why don't we do flattening anymore?

falexwolf: if this doesn't work anymore, it will break a lot of code... are you here at the institute?

anndata/base.py
```diff
@@ -600,11 +601,19 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
...
-            obsm=None, varm=None, raw=None,
-            dtype='float32', single_col=False,
-            filename=None, filemode=None,
-            asview=False, oidx=None, vidx=None):
+        self,
+        X: Optional[Union[np.ndarray, sparse.spmatrix]] = None,
+        obs: Optional[Union[pd.DataFrame, Mapping[Any, Iterable[Any]], np.ndarray]] = None,
+        var: Optional[Union[pd.DataFrame, Mapping[Any, Iterable[Any]], np.ndarray]] = None,
+        uns: Optional[Mapping[Any, Any]] = None,
+        obsm: Optional[np.ndarray] = None,
```
falexwolf: since very recently - to enable interoperability with seurat - we also have dict init for obsm, varm and raw

falexwolf:
Also kind of funny: attributes are now after methods.

anndata/base.py
flying-sheep: i’m not. and this argument wasn’t actually used, that’s why i removed it. you could find out why it wasn’t and fix it if that’s a bug.

flying-sheep:
That’s the normal [order in autosummary](https://raw.githubusercontent.com/sphinx-doc/sphinx/master/sphinx/ext/autosummary/templates/autosummary/class.rst). we could override it.

anndata/base.py
flying-sheep: fixed

anndata/base.py
falexwolf: yes, it's a bug that I seem to have removed it... it should have been passed to `_fix_shapes`, otherwise, after slicing a single column, while the following statement will still hold true,
```
adata[:, 0].X == adata.X[:, 0]
```
the left hand side object should give weird effects... don't you agree? I'm astonished that there is no test that fails...

but fine, I could possibly also restore a proper behavior after merging...

falexwolf:
Yes, it would be nice to override that order... At some point, we might also consider styling attributes like parameters as in the new numpydoc, scikit-learn etc... but that's not so urgent... But they should definitely appear first.

Also, I omitted `.obs_keys()`, for instance, on purpose. It will be removed at some point in the future, people should just use `.obs.keys()`...

falexwolf:
Last thing: we now again have the ugly grey bars behind the parameters... Don't you think they're ugly? See [here](https://anndata.readthedocs.io/en/latest/anndata.AnnData.concatenate.html#anndata.AnnData.concatenate).

Your suggestion is to fix this using `css`?

anndata/base.py
flying-sheep: OK, great! that’s something pycharm caught: it highlighted the param in grey and said “isn’t used”.

flying-sheep:
did you already do sth? looks like this for me:

grafik

falexwolf:
Damn, it's a chrome issue... but as almost everyone uses chrome - my impression - it would be really cool to fix it...

flying-sheep:
sure, should be easy to fix by making the CSS more specific.

also there’s really little reason to use chome over firefox these days…

flying-sheep:
hmm @falexwolf i can’t see any problem in chrome:

grafik

falexwolf:
Right, it looks beautiful! :smile:

PR #22, opened 2018-05-13
Add gz and bz2 support for text files.

### LINKED ISSUE #21: gzip and bzip2 compression support in read()
Hey,

Currently csv/tsv files with gzip or bzip2 compression are not supported, if I'm not mistaken. There is  an issue in DCA (https://github.com/theislab/dca/issues/7) about this, so I wanted to file a tracking issue here.

I was thinking about simply adding `gzip.open()` and `bzip2.open()` calls based on the file extension, just like other functions in the implementation, but there is also an option of using pandas for that because if we add compression support to anndata `read_text()` will start to converge `pandas.read_csv()`. 

So would it make sense to use `pandas.read_csv` for all text file reading functionality? It's already a dependency of anndata, so I don't see why not.

gokceneraslan:
Added a simple filename-based gz/bz2 support to read_text() function. Fixes #21 .

### DIFF
--- a/anndata/readwrite/read.py
+++ b/anndata/readwrite/read.py
@@ -1,6 +1,8 @@
 from pathlib import Path
 from typing import Union, Optional, Iterable, Generator, Iterator
 from collections import OrderedDict
+import gzip
+import bz2
 import numpy as np
 
 from ..base import AnnData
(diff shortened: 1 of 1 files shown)


anndata/readwrite/read.py
```diff
@@ -197,8 +199,15 @@ def read_text(
         Numpy data type.
     """
     if isinstance(filename, (str, Path)):
-        with Path(filename).open() as f:
-            return _read_text(f, delimiter, first_column_names, dtype)
+        if str(filename).endswith('.gz'):
+            with gzip.open(str(filename), mode='rt') as f:
```
gokceneraslan: This is also possible without str() but requires Python >= 3.6, so I wasn't sure. https://bugs.python.org/issue28227

PR #28, opened 2018-06-05
Fix pickling for BoundRecArrays

### LINKED ISSUE #27: BoundRecArray objects don't pickle well
`BoundRecArray` objects don't keep their attributes when pickled:

```python
>>> from anndata import AnnData
/usr/local/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
>>> import pickle
>>> import numpy as np
>>> 
>>> adata = AnnData()
>>> adata.obsm._parent == adata
True
>>> adata2 = pickle.loads(pickle.dumps(adata))
>>> adata2.obsm._parent == adata2
  File "/usr/local/lib/python3.6/site-packages/numpy/core/records.py", line 450, in __getattribute__
    res = fielddict[attr][:2]
KeyError: '_parent'

  File "/usr/local/lib/python3.6/site-packages/numpy/core/records.py", line 452, in __getattribute__
    raise AttributeError("recarray has no attribute %s" % attr)
AttributeError: recarray has no attribute _parent
>>> adata2.obsm.__dict__
{}
>>> adata.obsm.__dict__
{'_parent': AnnData object with n_obs × n_vars = 0 × 0 , '_attr': 'obsm'}
```

Based on [this stackoverflow question](https://stackoverflow.com/questions/26598109/preserve-custom-attributes-when-pickling-subclass-of-numpy-array), I think the issue comes from subclassing a `numpy` object, which have custom code for pickling.

ivirshup:
Fixes #27.

BoundRecArrays were losing their attributes when pickled. This can be fixed by overriding their `__reduce__` and `__setstate__` methods. I've basically just adopted a generalized version of [this stackoverflow answer](https://stackoverflow.com/a/26599346/3841270).

I had some trouble with the python 3.6 travis build, but fixed it by setting `numpy>=1.14.0` in the requirements. With a lower version of numpy, un-pickling a `BoundRecArray` throws the error:

```
ValueError: Invalid data-type size.
```

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -84,6 +84,16 @@ class BoundRecArr(np.recarray):
         self._parent = getattr(obj, '_parent', None)
         self._attr = getattr(obj, '_attr', None)
 
+    def __reduce__(self):
+        pickled_state = super(self.__class__, self).__reduce__()
+        new_state = pickled_state[2] + (self.__dict__, )
+        return (pickled_state[0], pickled_state[1], new_state)
+    
+    def __setstate__(self, state):
+        for k, v in state[-1].items():
+            self.__setattr__(k, v)
+        super(self.__class__, self).__setstate__(state[0:-1])
+
     def copy(self, order='C'):
         new = super(BoundRecArr, self).copy()
         new._parent = self._parent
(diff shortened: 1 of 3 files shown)


falexwolf:
Oh damn, why did I forget about this... Thank you very much!

PR #38, opened 2018-07-05
Add Zarr storage

tomwhite:
This change allows Anndata to use [Zarr](http://zarr.readthedocs.io/) as the backing storage. The motivation here is to provide support for Zarr with [Scanpy](http://scanpy.readthedocs.io/) and Spark (although this patch does not have any Spark parts in it). This work can be found in [this repo](https://github.com/lasersonlab/single-cell-experiments).

The code here is based on the HDF5 codepath, although it might be possible to simplify it somewhat, and it also needs docstrings adding. I wanted to submit the PR now to see if there is general interest and if such a change would be considered.

### DIFF
--- a/anndata/readwrite/read.py
+++ b/anndata/readwrite/read.py
@@ -340,7 +340,10 @@ def _read_key_value_from_zarr(f, d, key, key_write=None):
         return
     # the '()' means 'load everything into memory' (by contrast, ':'
     # only works if not reading a scalar type)
-    value = f[key][()]
+    if key != 'X':
+        value = f[key][()]
+    else:
+        value = f[key] # don't load X into memory
 
     def postprocess_reading(key, value):
         if value.ndim == 1 and len(value) == 1:
--- a/anndata/readwrite/read.py
+++ b/anndata/readwrite/read.py
@@ -321,6 +321,52 @@ def _read_text(
                    var={'var_names': col_names})
 
 
+def read_zarr(store):
+    import zarr
+    f = zarr.open(store, mode='r')
+    d = {}
+    for key in f.keys():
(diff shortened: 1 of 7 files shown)


falexwolf:
Ah! Cool! Thanks for pointing this out...

We are currently redesigning the backed infrastructure of AnnData anyways, so this comes at the right time.

Looking at the pull request, I see some conflicts. I also see some copy and paste - I'm adding a comment in the code... It would be great to discuss this a little.

@Koncopd your opinion is also valid... how far are you with the new infrastructure? let's put your fork on a development branch here, so that we won't run into further conflicts...

anndata/readwrite/read.py
```diff
@@ -321,6 +321,55 @@ def _read_text(
...
+def read_zarr(store):
+    import zarr
+    f = zarr.open(store, mode='r')
+    d = {}
+    for key in f.keys():
+        _read_key_value_from_zarr(f, d, key)
+    return AnnData(d)
+
+
+def _read_key_value_from_zarr(f, d, key, key_write=None):
```
falexwolf: this is very redundant, instead we should have a helper that is called both by the HDF5 functions and the zarr... @Koncopd what do you think?

falexwolf:
If you grant us access to work on your pull request, we can also see if we converge in this way and add the changes that I'd like to see ourselves....

tomwhite:
@falexwolf thanks for the feedback! I agree about eliminating the redundancy with the HDF5 code.

Please go ahead and add any edits you see fit - you should have write access.

falexwolf:
Ok, sorry for the long wait. I'm merging this now... we're updating several aspects of the functionality and will also address the questions that I raised above.

PR #54, opened 2018-08-27
include test for `.write_loom`

VolkerBergen:


flying-sheep:
please just put all changes into the other PR/branch, then we can see if the test actually passes :wink:

PR #55, opened 2018-08-27
Switch back to 2D-only arrays (WIP)

### LINKED ISSUE #42: Failure reading single observation h5ads
AnnData doesn't seem to like h5ad files with a single observation:

```
In [54]: anndata.__version__
Out[54]: '0.6.6'

In [55]: df.shape
Out[55]: (23465, 1)

In [56]: adata = anndata.AnnData(df.values.T, {"cell_names": df.columns.values}, {"gene_names": df.index.values})

In [57]: adata.write("test.h5ad")

In [58]: bdata = anndata.read_h5ad("test.h5ad")
<ipython-input-58-222d2ae8c407> in <module>()
----> 1 bdata = anndata.read_h5ad("test.h5ad")

~/.local/lib/python3.5/site-packages/anndata/readwrite/read.py in read_h5ad(filename, backed)
    342         # load everything into memory
    343         d = _read_h5ad(filename=filename)
--> 344         return AnnData(d)
    345
    346

~/.local/lib/python3.5/site-packages/anndata/base.py in __init__(self, X, obs, var, uns, obsm, varm, raw, dtype, shape, filename, filemode, asview, oidx, vidx)
    639                 obsm=obsm, varm=varm, raw=raw,
    640                 dtype=dtype, shape=shape,
--> 641                 filename=filename, filemode=filemode)
    642
    643     def _init_as_view(self, adata_ref, oidx, vidx):

~/.local/lib/python3.5/site-packages/anndata/base.py in _init_as_actual(self, X, obs, var, uns, obsm, varm, raw, dtype, shape, filename, filemode)
    758                 raise ValueError(
    759                     'If `X` is a dict no further arguments must be provided.')
--> 760             X, obs, var, uns, obsm, varm, raw = self._from_dict(X)
    761
    762         # init from AnnData

~/.local/lib/python3.5/site-packages/anndata/base.py in _from_dict(ddata)
   1920                     if key in d_true_keys[true_key].dtype.names:
   1921                         d_true_keys[true_key] = pd.DataFrame.from_records(
-> 1922                             d_true_keys[true_key], index=key)
   1923                         break
   1924                 d_true_keys[true_key].index = d_true_keys[true_key].index.astype('U')

~/.local/lib/python3.5/site-packages/pandas/core/frame.py in from_records(cls, data, index, exclude, columns, coerce_float, nrows)
   1267         else:
   1268             arrays, arr_columns = _to_arrays(data, columns,
-> 1269                                              coerce_float=coerce_float)
   1270
   1271             arr_columns = _ensure_index(arr_columns)

~/.local/lib/python3.5/site-packages/pandas/core/frame.py in _to_arrays(data, columns, coerce_float, dtype)
   7493     else:
   7494         # last ditch effort
-> 7495         data = lmap(tuple, data)
   7496         return _list_to_arrays(data, columns, coerce_float=coerce_float,
   7497                                dtype=dtype)

~/.local/lib/python3.5/site-packages/pandas/compat/__init__.py in lmap(*args, **kwargs)
    129
    130     def lmap(*args, **kwargs):
--> 131         return list(map(*args, **kwargs))
    132
    133     def lfilter(*args, **kwargs):

TypeError: 'numpy.int64' object is not iterable
```

But faking another cell works fine:

```
In [59]: df2 = pandas.concat([df, df], axis=1)

In [60]: df2.shape
Out[60]: (23465, 2)

In [61]: adata = anndata.AnnData(df2.values.T, {"cell_names": df2.columns.values}, {"gene_names": df2.index.values})

In [62]: adata.write("test.h5ad")

In [63]: bdata = anndata.read_h5ad("test.h5ad")

In [64]: bdata.n_obs, bdata.n_vars
Out[64]: (2, 23465)
```

I'm guessing this has something to do with `_fix_shapes`

### LINKED ISSUE #56: Can’t concat single-observation anndatas
Fixed by retaining 2D storage via #55

flying-sheep:
We have always either no data or 2D data. Representing this super flexibly as scalar or 1D array is error-prone, and making it one explicitly is one `.flatten()` away for the user.

Fixes #42, fixes #56

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -34,6 +34,7 @@ from . import h5py
 from .layers import AnnDataLayers
 
 from . import utils
+from .utils import Index, get_n_items_idx
 from .compat import PathLike
 
 FORMAT = '%(message)s'
@@ -67,9 +68,6 @@ class StorageType(Enum):
     def classes(cls):
         print(ZarrArray)
         return tuple(c.value for c in cls.__members__.values())
-
-
-Index = Union[slice, int, np.int64, np.ndarray, Sized]
 
 
 class BoundRecArr(np.recarray):
@@ -230,26 +228,16 @@ def df_to_records_fixed_width(df):
 def _fix_shapes(X):
     """Fix shapes of array or sparse matrix.
 
-    Flatten 2d array with a single row or column to emulate numpys
-    behavior of slicing.
+    Assure that X is always 2D: Unlike numpy we always deal with 2D arrays.
     """
     if X.dtype.names is None and len(X.shape) not in {0, 1, 2}:
         raise ValueError('X needs to be 2-dimensional, not '
                          '{}-dimensional.'.format(len(X.shape)))
-    if len(X.shape) == 2:
-        n_obs, n_vars = X.shape
-        if n_obs == 1 and n_vars == 1:
-            X = X[0, 0]
-        elif n_obs == 1 or n_vars == 1:
-            if issparse(X): X = X.toarray()
-            X = X.flatten()
-    elif len(X.shape) == 1:
-        n_vars = X.shape[0]
-        n_obs = 1
-    else:
-        n_vars = 1
-        n_obs = 1
-    return X, n_obs, n_vars
+    if len(X.shape) == 1:
+        X.shape = (1, X.shape[0])
+    elif len(X.shape) == 0:
+        X.shape = (1, 1)
+    return X
 
 
 def _normalize_index(index, names):
@@ -681,12 +669,6 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
                 filename=filename, filemode=filemode)
 
     def _init_as_view(self, adata_ref: 'AnnData', oidx: Index, vidx: Index):
-        def get_n_items_idx(idx: Index):
-            if isinstance(idx, np.ndarray) and idx.dtype == bool:
-                return idx.sum()
-            else:
-                return len(idx)
-
         self._isview = True
         self._adata_ref = adata_ref
         self._oidx = oidx
@@ -708,19 +690,19 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
         self._slice_uns_sparse_matrices_inplace(uns_new, self._oidx)
         # fix _n_obs, _n_vars
         if isinstance(oidx, slice):
-            self._n_obs = get_n_items_idx(obs_sub.index)
+            self._n_obs = get_n_items_idx(obs_sub.index, adata_ref.n_obs)
         elif isinstance(oidx, (int, np.int64)):
             self._n_obs = 1
         elif isinstance(oidx, Sized):
-            self._n_obs = get_n_items_idx(oidx)
+            self._n_obs = get_n_items_idx(oidx, adata_ref.n_obs)
         else:
             raise KeyError('Unknown Index type')
         if isinstance(vidx, slice):
-            self._n_vars = get_n_items_idx(var_sub.index)
+            self._n_vars = get_n_items_idx(var_sub.index, adata_ref.n_vars)
         elif isinstance(vidx, (int, np.int64)):
             self._n_vars = 1
         elif isinstance(vidx, Sized):
-            self._n_vars = get_n_items_idx(vidx)
+            self._n_vars = get_n_items_idx(vidx, adata_ref.n_vars)
         else:
             raise KeyError('Unknown Index type')
         # fix categories
@@ -748,13 +730,6 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
             self._X = None
             return
         X = self._adata_ref.X[self._oidx, self._vidx]
-        if len(X.shape) == 2:
-            n_obs, n_vars = X.shape
-            if n_obs == 1 and n_vars == 1:
-                X = X[0, 0]
-            elif n_obs == 1 or n_vars == 1:
-                if issparse(X): X = X.toarray()
-                X = X.flatten()
         if isinstance(X, sparse.csr_matrix):
             self._X = SparseCSRView(X, view_args=(self, 'X'))
         elif isinstance(X, sparse.csc_matrix):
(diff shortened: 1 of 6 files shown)


falexwolf:
I'll have time again during the next couple of days - am just returning from my wedding weekend.

I'm opposing this change. Automatic flattening wasn't a mistake but made a lot of code a lot simpler. This is why numpy arrays do it. If we go back to 2d arrays many people's code will break.

falexwolf:
Sure, I'm glad you came! :smile:
---
I made the change already more than a year ago when anndata was still a part of Scanpy: https://github.com/theislab/anndata/commit/d1f28b84780ab782c788772031ac019beaadf7eb

It's been in the docs for more than a year:
> A data matrix is flattened if either n_obs or n_vars is 1, so that numpy’s slicing behavior is reproduced:
```
adata = AnnData(np.ones((2, 2)))
adata[:, 0].X == adata.X[:, 0]
```

Scanpy 1.0 was released in April. It's unfortunate that you weren't aware of the change.

---

> i very deliberately decided to always leave the backing storage 2D when I implemented it.

I know, we even talked about it! This concerns a discussion we had in March 2017. I had many discussions about this in April and May 2017 with other users and made the change probably around that time.

The consequence was that a lot of `.flattens()` disappeared from the user's codes. The prototypical call of
```
adata[:, 'Gata1'].X
```
always returned a dense one-dimensional vector and no longer a 2d array. By that, AnnData became essentially as convenient as a dataframe, where the call is
```
df['Gata1']
```
and returns a Series object.

I acknowledge that one might have better introduced another access operation for retrieving a 1d array column, but before getting into that, there are also further arguments.

---

AnnData is a container object of data and annodations. To me, data is always a tensor where the first axis labels the observations. The second axis starts labeling the dimensions of the observations. For structured data, these are scalars and a simple label (like a gene name) is enough. But now consider spatial transcriptomics, where you have say a 2d array of measurements for each gene. Then `.X` will be a 4d tensor.

Having multiple observations of a scalar is most conventiently represented as a column vector. And not a 2d array. As the whole point of AnnData is convenient access to data and annotations, I'm strongly advocating to keep things as they are.

---

I have not yet tried to fix the bug @mckinsel discovered (https://github.com/theislab/anndata/issues/42). But I'm sure it's easy to fix. I don't know which problems @M0hammadL had, but I'm sure they are easy to fix, too.

---

Why are you using the notion "backing storage"? Let's stick with the notion data matrix. In the future, this will be a data tensor. Even more so, as people use anndata with tensorflow, which is already the case.

---

What do you think?

flying-sheep:
The core of my argument was that this behavior is confusing and error-prone (as evidenced by the two mentioned bugs in scanpy: #42 & #56). They’re both caused because subsetting AnnData objects can result in the mentioned change of shape, which proves my point: People (not even the authors of scanpy) can’t remember this edge-case behavior and are bitten by it in a weird and hard-to-trace way: `TypeError: 'numpy.int64' object is not iterable` doesn’t help at all!

Experienced users might learn what the weird error message means after a while (oh, that one again!), but newbies will always have to waste (their and/or our) time when encountering it.

Storing higher-dimensional data is besides the point: If we have a 3D tensor, it should stay 3D and not degrade to 2D or 1D when subsetting observations and/or variables (actually that’s a great argument for my side: things will get even more confusing once we’d start having a 2D tensor only be interpretable when cross-referencing its shape with the AnnData’s shape).

Therefore I think we should have gone with the option you mentioned: Another API for getting rid of one or the other dimension.

falexwolf:
> People (evidently even the authors of scanpy) can’t remember this edge-case behavior and are bitten by it in a weird and hard-to-trace way: TypeError: 'numpy.int64' object is not iterable doesn’t help at all!

I'm sorry that you missed the change 1.5 years ago. It astonishes me that you weren't bitten by it earlier.

> Storing higher-dimensional data is besides the point, let me generalize my argument: If we have a ND tensor, it should stay ND and not degrade to N-1 or N-2 dimensions when subsetting observations and/or variables. Actually I think that’s a great argument for my opinion: Things will get even more confusing once we’d start having e.g. a 2D tensor only be interpretable when cross-referencing its shape with the AnnData’s shape.

Array's were designed automatically flatten upon slicing because many people consider this highly intuitive behavior, also me. Not only for 2d-1d but also for 4d>3d (slicing a gene in the the spatial case, for instance). Also, in the high-dimensional case, `.obs` and `.var` won't have that simple correspondence to the axes of `.X` anymore: `.obs` will match the first index and `.var` will correspond to the "interpretable" variables or to a collection of variable axes but for sure not to pixels.

> You also didn’t address my other argument: that only ndarrays degrade, not sparse matrices. They always stay 2D. This introduces further inconsistency, and limits the utility of not having to use “flatten” to dense arrays.

That's not true. Slicing a variable from an AnnData that has a sparse `.X` has been returning a dense (1d) vector for more than one 1 year. Don't know why you missed it. This feature has added a lot of convenience for myself when working with AnnData.

----

I completely agree that the code becomes simpler if one doesn't cast `.X` to different dimensions. And probably a "convenience accessor" instead of casting `.X` would have been a better solution. But it's way too late for the change and also with the present solution, it's definitely not a lot of work to fix the remaining bugs. Also for the future, I don't see fundamental problems, just a bit more bookkeeping work to be done. After 1.5 years of bug reports, I don't think there are any problems for the users: both #42 and #56 absolutely are non-canonical use cases. That's why I'm taking to so much time with fixing them.

flying-sheep:
OK, you do the lion’s share of all the coding and bug fixing, so of course you have the last word.

falexwolf:
I'm adapting this such that `._X` actually always remains 2d, but `.X` still get's flattened upon slicing. Then the whole thing is backwards compat.

We can then go and see how we move from there.

One option would be to replace to an accessor `.values` that would always be 2d and deprecate `.X` in the long run. But this will be another pull request...

falexwolf:
Ok, I'm merging this now. `._X` is now always 2d. `.X` is presented to the user flattened if slicing a single dimension.

We can address the issues https://github.com/theislab/anndata/issues/42 and https://github.com/theislab/anndata/issues/56 in a new pull request. They should be easy to address now.

Same for the discussion https://github.com/theislab/anndata/issues/60#issuecomment-422749492 and the comment about `.values` above.

PR #58, opened 2018-09-05
Don’t modify logging.root

flying-sheep:
Creates out own root logger with a INFO loglevel and very basic format.


falexwolf:
Awesome! :smile: Why did I not see this at the time... maybe because it was during the first couple of confused days here...

PR #66, opened 2018-10-05
Support running on distributed compute engines like Spark via Zap.

tomwhite:
This change is needed for https://github.com/theislab/scanpy/pull/283


falexwolf:
Thank you! That's an easy one. ;)

PR #68, opened 2018-10-08
Method to_df() to transform anndata in pandas DF

fbrundu:
It should implement the transformation only to a shallow dataframe (X, with obs_names and var_names as indexes).
It only bugs me that the dataframe returned has a slightly different precision than the one used to initialize the AnnData, eg:

```python
> import pandas as pd
> import anndata as an
> df = pd.DataFrame(pd.np.random.rand(10,10), index=range(10), columns=range(10,20))
> df2 = an.AnnData(df, obs=df.index, var=df.columns).to_df()
> (df - df2).abs().max().max()
2.943521026921303e-08
```


fbrundu:
Sounds good, thanks!

PR #80, opened 2018-11-01
Fix numpy compatibility operator in requirements.txt

gokceneraslan:
Current requirement is incompatible with numpy 1.15, which is not we want. See https://github.com/theislab/anndata/commit/8d331cfcebb06ed202c7db3a1690b8bf7e6453d1#r30515166 for the discussion.

What do you think @flying-sheep ?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pandas>=0.21.0
-numpy~=1.14.0
+numpy~=1.14
 scipy~=1.0
 h5py
 natsort


flying-sheep:
You’re right, I didn’t know that the definition of `~=` is so fucked up. It would make sense for it to be the same as semver’s `^=`, but it isn’t.

gokceneraslan:
No problem, thanks!

falexwolf:
Great, thanks!

PR #81, opened 2018-11-05
add dtype for read functions

huidongchen:
This change is needed for [https://github.com/theislab/anndata/issues/75](url)

### DIFF
--- a/anndata/readwrite/read.py
+++ b/anndata/readwrite/read.py
@@ -40,6 +40,7 @@ def read_csv(
 def read_excel(
     filename: PathLike,
     sheet: Union[str, int],
+    dtype: str='float32',
 ) -> AnnData:
     """Read ``.xlsx`` (Excel) file.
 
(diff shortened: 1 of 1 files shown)


anndata/readwrite/read.py
```diff
@@ -196,7 +198,7 @@ def read_mtx(filename: PathLike, dtype: str='float32') -> AnnData:
     X = mmread(fspath(filename)).astype(dtype)
     from scipy.sparse import csr_matrix
     X = csr_matrix(X)
-    return AnnData(X)
+    return AnnData(X,dtype=dtype)
```
falexwolf: no space behind commas, please

anndata/readwrite/read.py
```diff
@@ -55,14 +56,14 @@ def read_excel(
     """
     # rely on pandas for reading an excel file
     from pandas import read_excel
-    df = read_excel(fspath(filename), sheet)
-    X = df.values[:, 1:].astype(float)
+    df = read_excel(fspath(filename), sheet, dtype=dtype)
+    X = df.values[:, 1:].astype(dtype)
```
falexwolf: why do we need the dtype in both lines? it should not be necessary in the second line anymore, should it?

falexwolf:
Do the tests run through on your end? Can you add a test, please?

I believe that the fact that one test currently fails is not due to your PR; are you sure you rebased properly? It went through on master.

falexwolf:
Oh damn, the test failing due to the "open file" is my fault. It went through for me but I'll look into this, ignore this comment.

falexwolf:
OK, I just fixed the test on master... nothing to do with anything you'd do.

Thank you for your PR! :smile:

anndata/readwrite/read.py
huidongchen: Agreed. Will remove it.

huidongchen:
hmm, it passed all the local testing on my end. Have no idea why it still failed..

falexwolf:
Thanks, it's fine!

Issue #76, opened 2018-10-26
numpy/scipy version dependencies


Hoeze:
Hi, would it be possible to change the numpy/scipy dependencies to numy >= 1.14 and scipy >= 1.0?
anndata otherwise causes problems with other packages requiring newer versions of numpy/scipy.

falexwolf:
Yes, I guess that won't harm anyone. PR welcome!

flying-sheep:
Hi, what problems are you talking about? we just fixed a problem regarding this syntax in #80.

I’m confident that we understand everything now, so #83 won’t be needed, and will actually cause problems once numpy 2.0 or scipy 2.0 are released.

Hoeze:
Ah sorry I did not see that change ^^

I usually just use '>=' to keep pip from downgrading packages :)

PR #83, opened 2018-11-07
update project dependencies: require numpy >= 1.14 and scipy >= 1.0

Hoeze:
See also https://github.com/theislab/anndata/issues/76

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pandas>=0.21.0
-numpy~=1.14
-scipy~=1.0
+numpy>=1.14
+scipy>=1.0
 h5py
 natsort


Hoeze:
```python
from anndata import AnnData
E   ImportError: No module named 'anndata'
```
hmm.. Travis configuration problem?

flying-sheep:
can we discuss this in #76 before doing anything?

Hoeze:
Ah sorry I did not see that change ^^

I usually just use '>=' to keep pip from downgrading packages :)

flying-sheep:
yeah, `~=` is pretty unintuitive… it should work like semver’s `^=` but doesn’t.

maybe we should just go explicit and do `numpy>=1.4.0, numpy<2.0.0` and so on…

PR #85, opened 2018-12-02
use read_direct in read_h5ad

Koncopd:
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/memory_issue_huge.ipynb
here added the same changes as in the test code for `read_h5ad('1M_neurons.h5ad', rd=True)` from the benchmark above.
solves https://github.com/theislab/scanpy/issues/146


falexwolf:
As mentioned, this is absolutely great, thank you for actually doing the benchmarks! :smile:

Issue #63, opened 2018-09-26
Concatenate and layers


Marius1311:
The concatenate function does not take care of layers of an anndata object yet.

falexwolf:
Yes, you currently need to work around by concatenating manually. 
```
unspliceds = [ad.layers['unspliced'] for ad in adatas]
unspliced  = unspliceds[0]
for m in unspliceds[1:]:
    unspliced = sp.sparse.vstack([unspliced, m])

adata = adatas[0].concatenate(
    adatas[1:], batch_key='timepoints', batch_categories=timepoints)

adata.layers['spliced'] = adata.X
adata.layers['unspliced'] = unspliced
```

@Koncopd We need to be careful not to mess up the concatenate function, it's already way too long. Can you check whether you can essentially a one-line edit to bring in the functionality, write a test for it and make PR?
```
X_type = ['X', 'key1', 'key2', ...]
Xs = [adata.X if X_type == 'X' else adata.layers[key] for for adata in adatas]
```

Mid-term, we want to split all of this so that it also works in backed mode.

Koncopd:
@falexwolf 
I'll look.

ijayden-lung:
> Yes, you currently need to work around by concatenating manually.
> 
> ```
> unspliceds = [ad.layers['unspliced'] for ad in adatas]
> unspliced  = unspliceds[0]
> for m in unspliceds[1:]:
>     unspliced = sp.sparse.vstack([unspliced, m])
> 
> adata = adatas[0].concatenate(
>     adatas[1:], batch_key='timepoints', batch_categories=timepoints)
> 
> adata.layers['spliced'] = adata.X
> adata.layers['unspliced'] = unspliced
> ```
> 
> @Koncopd We need to be careful not to mess up the concatenate function, it's already way too long. Can you check whether you can essentially a one-line edit to bring in the functionality, write a test for it and make PR?
> 
> ```
> X_type = ['X', 'key1', 'key2', ...]
> for key in X_type:
>      Xs = [adata.X if key == 'X' else adata.layers[key] for adata in adatas]
> ```
> 
> Mid-term, we want to split all of this so that it also works in backed mode.

---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-75-eb3640cb5ec2> in <module>()
      3 unspliced  = unspliceds[0]
      4 for m in unspliceds[1:]:
----> 5     unspliced = sp.sparse.vstack([unspliced, m])
      6 
[...]

PR #94, opened 2019-01-07
enable concat layers

VolkerBergen:
Addressing issue #63 . So far, only join=='inner' is considered. A test has not been written yet.

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -1831,14 +1831,14 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
 
         # check whether tries to do 'outer' join and layers is non_empty.
         if join == 'outer' and len(shared_layers) > 0:
-            logger.info('layers concatenation is not yet available for '
-                        '\'outer\' intersection and will be ignored.')
+            logger.info(
+                'layers concatenation is not yet available for \'outer\' intersection and will be ignored.')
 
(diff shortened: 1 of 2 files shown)


falexwolf:
Thank you! :smile:

Looks good, except that one should output a warning if one tries to do an "outer" join and layers is non-empty in one of the passed AnnDatas. One should also output a warning if the layers are not consistently set in all the passed AnnDatas.

Finally, a mini-test would be nice. Just take a chunk of this test https://github.com/theislab/anndata/blob/9dce27428d5403713e41ce89a0a10da97ee18616/anndata/tests/test_base.py#L266 and adapt it with a layers attribute `test_concatenate_layers`.

VolkerBergen:
Thanks for your comments. I added a few lines in test_concatenate (imho better than having a separate test_concatenate_layers). Further, two warnings (logger.info) are included for 1) 'outer' join and non-empty layers, and 2) inconsistent layers among AnnDatas.

falexwolf:
This is a lot better! I don't really see how come that the coverage still went down by 1%... The only non-covered lines are the logging lines... Hm.

I'm fixing the conflict and merging this into master.

Thank you! 🙂

flying-sheep:
Maybe our codecov setup is wrong, and the bot didn’t update the overview.

When you look at the site, everything is green: https://codecov.io/gh/theislab/anndata/pull/94?src=pr&el=continue

Issue #100, opened 2019-01-20
Read option `backed=True` throws error


ivirshup:
Something I'd seen before and mentioned in [a Scanpy issue](https://github.com/theislab/scanpy/issues/263), but am running into again. Here's a quick example from the base of this repo:

```python
In [1]: import anndata                                                                              

In [2]: anndata.read("./test.h5ad", backed=True)    
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-d3b620ce7a6c> in <module>
----> 1 anndata.read("./test.h5ad", backed=True)

...
/usr/local/lib/python3.7/site-packages/h5py/_hl/files.py in make_fid(name, mode, userblock_size, fapl, fcpl, swmr)
    195                 fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
    196     else:
--> 197         raise ValueError("Invalid mode; must be one of r, r+, w, w-, x, a")
    198 
    199     try:

ValueError: Invalid mode; must be one of r, r+, w, w-, x, a
```

<details>
<summary>Full traceback:</summary>

```python                              
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-d3b620ce7a6c> in <module>
----> 1 anndata.read("./test.h5ad", backed=True)

~/github/anndata/anndata/readwrite/read.py in read_h5ad(filename, backed, chunk_size)
    430     if backed:
    431         # open in backed-mode
--> 432         return AnnData(filename=filename, filemode=backed)
    433     else:
    434         # load everything into memory

~/github/anndata/anndata/base.py in __init__(self, X, obs, var, uns, obsm, varm, layers, raw, dtype, shape, filename, filemode, asview, oidx, vidx)
    652                 layers=layers,
    653                 dtype=dtype, shape=shape,
--> 654                 filename=filename, filemode=filemode)
    655 
    656     def _init_as_view(self, adata_ref: 'AnnData', oidx: Index, vidx: Index):

~/github/anndata/anndata/base.py in _init_as_actual(self, X, obs, var, uns, obsm, varm, raw, layers, dtype, shape, filename, filemode)
    764                     'If initializing from `filename`, '
    765                     'no further arguments may be passed.')
--> 766             self.file = AnnDataFileManager(self, filename, filemode)
    767             # will read from backing file what is returned is, at this
    768             # stage, a dictionary that needs to be processed in the

[...]

PR #101, opened 2019-01-20
Make documentation for `read_h5ad` consistent with behaviour

ivirshup:
Proposed fix for #100. Updated the documentation to limit arguments which return a backed `AnnData` to those that work. Additionally added a deprecation warning for people passing in `bool`, though that should only affect people using an explicit `False`, since `True` doesn't work.


falexwolf:
Thank you! Is better! 😄

PR #103, opened 2019-01-25
fix condition for backed

Koncopd:
Fixes https://github.com/theislab/scanpy/issues/434
Incorrect condition forced _read_h5ad to load X and then dismiss it.


Koncopd:
Why does this decrease code coverage?

flying-sheep:
You deleted a line that was covered. Technically that decreases total coverage unless we had 100%.

Very silly of codecov to interpret it that way.

PR #107, opened 2019-01-29
Don't convert to categorical if it's already categorical

ivirshup:
Quick fix for https://github.com/theislab/scanpy/issues/448

Fun fact (for at least pandas `v0.24.0` & `v0.22.0`):

```python
>>> import pandas as pd
>>> from pandas.api.types import is_string_dtype
>>> is_string_dtype(pd.Categorical([1,2,3]))
True
```


flying-sheep:
hahaha I JUST figured out the same thing!

PR #119, opened 2019-02-25
Add better subclassing support

pavlin-policar:
##### Problem

Consider the following use-case: I want to subclass `AnnData` and create my own class called `MyData`, where I implement some additional methods. However, whenever I index or copy an instance of `MyData`, I get an instance `AnnData`:

```python
In [3]: class MyData(AnnData): pass

In [4]: adata = MyData(np.array([[1, 2, 3], [4, 5, 6]]))

In [5]: type(adata)
Out[5]: __main__.MyData

In [6]: type(adata[:, 0])
Out[6]: anndata.base.AnnData

In [7]: type(adata.copy())
Out[7]: anndata.base.AnnData
```

##### Changes

Instead of returning an AnnData instance for indexing or copy methods, we can determine the current class using `type(self)` and use that to construct new instances (when subclassing and copying) to preserve the class type.

I think I've changed this in all the appropriate places (somebody more familliar with the codebase should confirm), and this does fix the described problem. If anybody knows of a more elegant solution, please let me know.

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -1300,7 +1300,7 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
 
     def _getitem_view(self, index: Index) -> 'AnnData':
         oidx, vidx = self._normalize_indices(index)
-        return AnnData(self, oidx=oidx, vidx=vidx, asview=True)
+        return type(self)(self, oidx=oidx, vidx=vidx, asview=True)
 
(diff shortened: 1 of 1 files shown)


pavlin-policar:
Thinking about this a bit more, I'm not sure this is actually a good idea, since then it's unclear how reading/writing the data sets would work e.g. I write my subclassed file into a loom file but when I read it, I get back an `AnnData` instance.

This use case may not make much sense.

PR #124, opened 2019-03-11
Add access to loompy.connect arguments.

gokceneraslan:
There are some useful arguments in `loompy.connect` that we might want to access, like `validate: bool = True,  spec_version: str = '2.0.1'`.


falexwolf:
Makes sense! Thanks!

PR #127, opened 2019-03-23
Write obsm/varm to loom file on request

gokceneraslan:
It'd be nice to have a way to export embeddings into loom files. Since this is ugly and likely to create many keys in col and row attributes, we can keep it disabled by default. But the users should be able to do it if they want, I think.

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -2021,7 +2021,7 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
         from .readwrite.write import write_csvs
         write_csvs(dirname, self, skip_data=skip_data, sep=sep)
 
-    def write_loom(self, filename: PathLike, write_embeddings: bool = False):
+    def write_loom(self, filename: PathLike, write_obsm_varm: bool = False):
         """Write ``.loom``-formatted hdf5 file.
 
         Parameters
@@ -2030,7 +2030,7 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
             The filename.
         """
         from .readwrite.write import write_loom
-        write_loom(filename, self, write_embeddings = write_embeddings)
+        write_loom(filename, self, write_obsm_varm = write_obsm_varm)
 
     def write_zarr(
         self,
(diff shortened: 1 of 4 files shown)


falexwolf:
Cool! That's actually how AnnData worked in the very beginning when there was no multi-dimensional annotation... :)

The only thing reminiscent of this are some backwards compat functions at the very bottom of `base.py`... but all of that was ugly and unstable at the time...

Just one request: can we call it `write_obsm_varm` or `write_m_annotations` where `m` stands for multi-dimensional? Or anything other that does not use `embedding`? In `.varm` these aren't even embeddings; anything in `.obsm` might qualify as a representation if a mapping exists, but I wouldn't use the notion _embedding_ for 50-dimensional PCA or autoencoder representation.

gokceneraslan:
Agreed, I was hesitant about the parameter name anyway. How about `write_multidim_annotations` to make it more explicit? Is it too long?

Instead of one-dimension-at-a-time approach, I think we can also store the full matrix at once as row/col attributes similar to how Seurat stores: 
https://github.com/satijalab/seurat/blob/master/R/conversion.R#L114. That would be without for loop and just `col_attrs['obsm_{}'.format(key)] = adata.obsm[key]`. 

I'm not sure which way would be more useful.

falexwolf:
I'd store the full matrices, if loom allows that. :)

`write_multidim_annotations` is pretty long, I'd say. `ignore_obsm_varm` or `write_obsm_varm` is very clear (I think) and quite a bit shorter. So, I'd go for that.

gokceneraslan:
Done. Actually, it's a lot less ugly now since we store matrices. Shall I change the default value of the option to True?

falexwolf:
I'd leave it at `False` as, I think, we don't account for multi-dimensional entries in the loom files when reading from loom files... We'd need to fix that, too.

I guess, you're adding the option mostly for interoperability between some tools... So, I think, for now it's fine to have it `False`.

What do you think?

gokceneraslan:
OK, I'm fine with both :) Reading is already fixed, by the way. Have a look at the test  🎊 😄

gokceneraslan:
Shall I merge this :) @falexwolf

falexwolf:
Sorry, Gökcen! Yes, we can merge this!

PR #135, opened 2019-04-10
Better error for _normalize_index

ivirshup:
Based on https://github.com/theislab/scanpy/issues/593, the error now tells you which values were not valid.


falexwolf:
Great! Thank you!

PR #139, opened 2019-04-15
use obs and var when creating adata from df

### LINKED ISSUE #137: obs is ignored when initializing AnnData with DataFrame
Minimal example:
```python
import anndata
import numpy as np
import pandas as pd

X = pd.DataFrame(np.empty((2,100)))
obs = pd.DataFrame(np.empty((2, 5)))

anndata.AnnData(X=X, obs=obs)
```
Output:
```python
AnnData object with n_obs × n_vars = 2 × 100 
```
Please note, that this `AnnData` has no entries in `obs`, so `obs` was ignored. In contrast, `obs` is set correctly, when I pass an array to X:
```python
anndata.AnnData(X=X.values, obs=obs)
```
Output:
```python
AnnData object with n_obs × n_vars = 2 × 100 
    obs: 0, 1, 2, 3, 4
```

fabianrost84:
fixes #137

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -797,8 +797,18 @@ class AnnData(IndexMixin, metaclass=utils.DeprecationMixinMeta):
 
             # init from DataFrame
             elif isinstance(X, pd.DataFrame):
-                obs = pd.DataFrame(index=X.index)
-                var = pd.DataFrame(index=X.columns)
+                if obs is None:
+                    obs = pd.DataFrame(index=X.index)
+                else:
+                    if not X.index.equals(obs.index):
+                        raise ValueError(
+                            'Index of obs must match index of X.')
+                if var is None:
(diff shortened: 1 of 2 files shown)


flying-sheep:
Look great! Could you please also test error cases using `with pytest.raises(ValueError, match=r'Index of ...'): ...`

fabianrost84:
Thanks! Sure, I will add the error cases. I was about to ask how do that, because I did not know. So, thanks for letting me know

flying-sheep:
Sure! For the future you should bear in mind that the argument passed to “match” is a regex. But your error messages don’t contain special regex characters so that’s no problem here.

fabianrost84:
great to know, thanks! I just made another correction because the tests failed.

flying-sheep:
Great, thank you very much!

PR #144, opened 2019-04-30
Add obs_array and var_array functions

ivirshup:
Convenience functions for retrieving an array of values along either the `obs` or `var` dimension.

A proposed update to `_get_{obs,var}_array` as discussed with @falexwolf in theislab/scanpy#619.

As I was writing it, it got more complicated that I expected (of course). I think `use_raw` may be poorly posed here, unless it's meant to only refer to `.raw.X`. For `var_array(..., use_raw=True)`, how many values should be returned? `adata.shape[1]` or `adata.raw.shape[1]`? Should we only check `adata.raw.var` for keys? This would change `_get_var_array`'s current behavior: https://github.com/theislab/anndata/blob/34f4eb63710628fbc15e7050e5efcac1d7806062/anndata/base.py#L1488-L1510

### DIFF
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -9,6 +9,7 @@ from typing import Any, Union, Optional
 from typing import Iterable, Sized, Sequence, Mapping, MutableMapping
 from typing import Tuple, List, Dict, KeysView
 from copy import deepcopy
+import warnings
 
 import numpy as np
 from numpy import ma
@@ -1611,7 +1612,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             if "X" in self.layers:
                 pass
             else:
-                # TODO: Raise deprecation warning
+                warnings.warn(
+                    "In a future version of AnnData, access to `.X` by passing"
+                    " `layer='X'` will be removed. Instead pass `layer=None`.",
+                    FutureWarning
+                )
                 layer = None
 
         if k in self.obs:
@@ -1650,7 +1655,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             if "X" in self.layers:
                 pass
             else:
-                # TODO: Raise deprecation warning
+                warnings.warn(
+                    "In a future version of AnnData, access to `.X` by passing"
+                    " `layer='X'` will be removed. Instead pass `layer=None`.",
+                    FutureWarning
+                )
                 layer = None
 
         if k in self.var:
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -1663,7 +1663,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 layer = None
 
         if k in self.var:
-            a = self.var[k].values
+            return self.var[k].values
         else:
             idx = self._normalize_indices((k, slice(None)))
             a = self._get_X(layer=layer)[idx]
@@ -1675,13 +1675,19 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def _get_obs_array(self, k, use_raw=False, layer=None):
         """Get an array from the layer (default layer='X') along the observation dimension by first looking up
         obs.keys and then var.index."""
-        return self.obs_vector(k=k, use_raw=use_raw, layer=layer)
+        if use_raw:
+            return self.raw.obs_vector(k)
+        else:
+            return self.obs_vector(k=k, layer=layer)
 
     @utils.deprecated("var_vector")
-    def _get_var_array(self, k, use_raw=False, layer='X'):
+    def _get_var_array(self, k, use_raw=False, layer=None):
         """Get an array from the layer (default layer='X') along the variables dimension by first looking up
         ``var.keys`` and then ``obs.index``."""
-        return self.var_vector(k=k, use_raw=use_raw, layer=layer)
+        if use_raw:
+            return self.raw.var_vector(k)
+        else:
+            return self.var_vector(k=k, layer=layer)
 
     def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
         """Full copy, optionally on disk."""
--- a/anndata/base.py
+++ b/anndata/base.py
@@ -1464,54 +1464,6 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         adata_subset = self[index].copy()
         self._init_as_actual(adata_subset, dtype=self._X.dtype)
 
-    def _get_obs_array(self, k, use_raw=False, layer='X'):
-        """Get an array from the layer (default layer='X') along the observation dimension by first looking up
-        obs.keys and then var.index."""
-        in_raw_var_names = k in self.raw.var_names if self.raw is not None else False
-
-        if use_raw and self.raw is None:
-            raise ValueError('.raw doesn\'t exist')
-
-        if k in self.obs.keys():
-            x = self._obs[k]
-        elif in_raw_var_names and use_raw and layer == 'X':
-            x = self.raw[:, k].X
-        elif k in self.var_names and not use_raw and (layer == 'X' or layer in self.layers.keys()):
-            x = self[:, k].X if layer=='X' else self[:, k].layers[layer]
-        elif use_raw and layer != 'X':
-            raise ValueError('No layers in .raw')
-        elif layer != 'X' and layer not in self.layers.keys():
-            raise ValueError('Did not find {} in layers.keys.'
-                             .format(layer))
-        else:
-            raise ValueError('Did not find {} in obs.keys or var_names.'
-                             .format(k))
-        return x
-
-    def _get_var_array(self, k, use_raw=False, layer='X'):
-        """Get an array from the layer (default layer='X') along the variables dimension by first looking up
-        ``var.keys`` and then ``obs.index``."""
-        in_raw_obs_names = k in self.raw.obs_names if self.raw is not None else False
-
-        if use_raw and self.raw is None:
-            raise ValueError('.raw doesn\'t exist')
-
-        if k in self.var.keys():
-            x = self._var[k]
-        elif in_raw_obs_names and use_raw and layer == 'X':
-            x = self.raw[k].X
-        elif k in self.obs_names and not use_raw and (layer == 'X' or layer in self.layers.keys()):
-            x = self[k].X if layer=='X' else self[k].layers[layer]
-        elif use_raw and layer != 'X':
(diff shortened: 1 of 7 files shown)


anndata/base.py
```diff
@@ -1561,6 +1513,123 @@ def to_df(self) -> pd.DataFrame:
...
+        -------
+        A one dimensional nd array, with values for each var in the same order
+        as `.var_names`.
+        """
+        if layer == "X":
+            if "X" in self.layers:
+                pass
+            else:
+                # TODO: Raise deprecation warning
+                layer = None
```
falexwolf: I just checked the previous implementations of this, which got really complicated by several people iterating on it... This here is much better.

I would do:
```
if use_raw:
     obj = self.raw
else:
     obj = self
```
So that if `use_raw==True` this looks into `self.raw.var`.

anndata/base.py
```diff
@@ -1561,6 +1513,123 @@ def to_df(self) -> pd.DataFrame:
...
+            a = self.raw[:, k].X
+        else:
+            idx = self._normalize_indices((slice(None), k))
+            a = self._get_expression(layer=layer)[idx]
+        if issparse(a):
+            a = a.toarray().reshape(-1)
+        return a
+
+
+    def var_array(
```
falexwolf: I don't think the naming is optimal and I'm also not sure whether we really want this as a public method. If public, I'd think that something like
```
adata.get_var_1dslice()
```
would make it more clear that this is not another data storage attribute but a convenience function that gives access to existing data.

anndata/base.py
```diff
@@ -1561,6 +1513,123 @@ def to_df(self) -> pd.DataFrame:
             X = self._X
         return pd.DataFrame(X, index=self.obs_names, columns=self.var_names)
 
+    def _get_expression(self, use_raw=False, layer=None):
```
falexwolf: AnnData is not meant for expression data only. Hence this should be `_get_matrix` or `_get_X`, I'd say. :)

ivirshup: Totally agree. Though I might just get rid of this function if I go with `obj = self.raw`

anndata/base.py
ivirshup: I'd find it useful enough to have as a public function, since it'd be good to advertise that there's a solution to having:

```python
if issparse(a):
    a.toarray().reshape(-1)
```

which is all over the place in my analysis code.

I agree there's a better name. `get_1d_obs` or `obs_array_1d`. I'll think about it a bit.

anndata/base.py
ivirshup: Hmm. I don't think I like that the `var` accessor could return something with a different shape that `adata.shape[1]`. What if you have to ask `adata.raw` for those values? Like `adata.raw.var_array`?

falexwolf: Your alternative solution is 100% fine, too!

anndata/base.py
falexwolf: Good! :smile:

How come you have this all over your analysis code and I don't? A sparse matrix slice is cast to a 1d array automatically: https://github.com/theislab/anndata/blob/34f4eb63710628fbc15e7050e5efcac1d7806062/anndata/base.py#L965

ivirshup: I get this:

```python
from scipy.sparse import random, issparse
issparse(random(10, 10, format="csr")[0, :])
# True
issparse(random(10, 10, format="csr")[:, 0])
# True
```

But I might be working more directly with the sparse matrices.

ivirshup: This does remind me of something. Are there any benchmarks for the overhead of subsetting AnnData? It'd be nice if subsetting was lazy enough to have low overhead, but I'm not sure if this is the case.

ivirshup:
Implemented methods for `Raw` and changed naming scheme to the more precise but verbose: `get_{obs,var}_slice_1d`. I'd be happy to make these private.

falexwolf:
I'd leave them as public!

`get_{obs,var}_slice_1d` is very long. In particular, if we foresee it to become the default access for 1d slices along the obs and var axes, then this will seriously start cluttering code. I agree we should be verbose in general, but for highly accessed stuff, a short name greatly helps readability. That's why `.obs` is not called `.annotations_of_observations` or something. :)

You don't like this other suggestion we had: `get_1d_obs` (when reading, one mentally completes this with "slice", I think)? I don't think there is any chance of confusion. Every user knows that `.obs` is not 1d but a DataFrame, so that it's implicitly clear that it must refer to a slice-retrieval method.

But you're doing all the work, so, you should decide!

anndata/base.py
falexwolf: > But I might be working more directly with the sparse matrices.
Scipy sparse matrices obviously maintain their type, `.X` does the cast for you. That's why I linked to `.toarray()` above. :)

falexwolf: > This does remind me of something. Are there any benchmarks for the overhead of subsetting AnnData? It'd be nice if subsetting was lazy enough to have low overhead, but I'm not sure if this is the case.

No, there are no benchmarks and you are absolutely right. It should be lazier. I think the subsets on all the arrays, dataframes etc. are fine because they are themselves lazy (that's at least what I expect). But the `deepcopy(adata.uns)` is awful...

ivirshup:
When I put `get_{obs,var}_slice_1d` I was thinking these would be private, in which case niceness of name is less important.

I think if this is going to be part of the API, I'd like a shorter name (ideally two words), especially since it's a pretty simple operation. `get_1d_obs` is pretty good, but I'm not totally happy with it. I think it's partially because of the numeric value and partially because it kinda looks like it will return all one dimensional observations.

Just some brainstorming:

* `vector` or possibly `vec` to replace need to specify `1d`

```python
adata.get_obs_vector("leiden")
adata.get_obs_vec("leiden")
adata.get_var_vector("pct_dropouts")
```

* Merge it into one function, where the dimension being sliced across is specified by a keyword argument replacement for `k`:

```python
adata.get_vec(obs="Oct4", layer="counts")
adata.get_vec(var="pct_dropout")
adata.get_vector(obs="leiden")
```

* Following functions like `obs_keys` we could drop the `get` verb?

```python
adata.obs_vector("Oct4", layer="counts")
for k in adata.obs_keys():
    adata.obs_vector(k)
```

ivirshup:
I think I like `{obs,var}_vector` the best, but I'd be happy with `get_1d_{obs,var}` if you think there is ambiguity here. Mostly I think this is ready to go, and will help me merge a bunch of stuff over on scanpy.

ivirshup:
@falexwolf, thoughts on this?

falexwolf:
As mentioned before, I'd vote for prepending a `get_` to distinguish from an attribute that can be set: `get_obs_1d_...` or `get_obs_vector_...`, whatever you like better.

Don't you think there might be confusion otherwise? The `get_...` can be a natural place for people to look up convenience accessors.

ivirshup:
I'm resistant to a three word name for such a short method, but could be swayed. I also like leading with the dimension name, since I think it fits better with the existing (though small) anndata api.

```python
adata.obs_names
adata.obs_keys()
adata.obs_names_make_unique()
```

To me, `.obs_vector()` 'rhymes' well with `.obs_keys()`. I could settle for `adata.obs_get_vector`, though I think `.obs_vector` reads equivalently in use.

If we were to start a `.get_` naming convention, what other convenience accessors could fit there? Could you suggest other APIs I could look at that use this convention? I only found a couple xarray methods (`get_index`, `get_axis_num`). If this method was meant to reflect other python array libraries, I'd probably make it work something like:

```python
adata.get_vector(key, dim="obs")
# Or the equivalent
adata.get_vector(key, axis=0)
```

Of course, this doesn't have to live here or be a public method (it could always become a public method in the future). I do want this functionality somewhere, so I can use it in and merge https://github.com/theislab/scanpy/pull/619 followed by https://github.com/theislab/scanpy/pull/467.

falexwolf:
OK. I'm convinced. :smile: Please make it `.obs_vector()`. I think it should be public... It could become the lightweight way of what is `adata[:, 'var`'].X` right now. I'd say...

PS: I had always thought that, e.g., `adata.obs_keys()` should be deprecated over `.obs.keys()` at some point, but probably it's not going to happen...

ivirshup:
I think those should be the names right now.

Any idea what codecov is trying to tell me? I'm not sure how I've increased coverage but also decreased coverage.

falexwolf:
Hm, I think that _69.35% of diff hit (target 73.61%)_ means that in the changes you made, only 69% are covered with tests and that in the previous corresponding code pieces, this fraction was 73%. But as you reduced the total lines of untested code, the total coverage even slightly increased. [If one replaces 100 lines with coverage 80% with 3 lines with coverage 30%, one still reduced the total number of uncovered lines from 20 to 2.]

Don't worry... Should we merge, now?

flying-sheep:
The issue codacy reports is valid:

> Using type() instead of isinstance() for a typecheck.
> ```py
> assert type(r1) == type(r2)
> ```

This should be

```py
assert type(r1) is type(r2)
```

ivirshup:
@flying-sheep, I thought it wasn't going to like either since it recommended `isinstance`. I've changed it anyways.

@falexwolf so what's the other passing codecov check telling me?

And yeah, I think this is good to go.

falexwolf:
Awesome! :smile:

falexwolf:
Despite telling me that merging in master (rebasing on the command line would also have been an option), it tells me that I can only squash and merge. So, I'd go ahead and do that. But, after all your work here and in the other PR, @ivirshup, you should be an admin (also on the author list) of anndata. I'll quickly make you one and then you can merge this in yourself.

falexwolf:
You now have write access and can merge yourself, Isaac.

ivirshup:
Thanks for the access!

It wouldn't let me do any sort of merging without getting that codecov check to pass.

It now tells me that I can only squash and merge, and that both "Rebase and Merge" and "Create a merge commit" are not enabled for this repo. Is that what we want to do? If not I think the repository owner can modify this at `Settings > Options > Merge button`.

falexwolf:
OK, now all options are available. "Rebase and Merge" is definitely fine. I don't know why this wasn't enabled by default.

ivirshup:
Since tests weren't passing on master (an f-string was added in a recent commit) I've gone ahead and bumped the minimum python version to 3.6. This was previously discussed in https://github.com/theislab/anndata/pull/140#issuecomment-493032332.

ivirshup:
Thanks for changing those settings!

Issue #148, opened 2019-05-09
Can't assign to obs in AnnData view


joshua-gould:
<table>
<tr>
<td><code>In [1]:</code></td>
<td>

```py
import anndata
import pandas as pd
adata = anndata.AnnData(
    X=np.array([[1,2], [1,2], [1,2]]),
    obs=pd.DataFrame(index=['a','b','c']),
    var=pd.DataFrame(index=['a','b']),
)
s = adata[:, ['b','a']]
df = pd.DataFrame(index=['a', 'b'], data={'c1':[1,2]})
df2 = s.obs.join(df)
df2['c1']
```

</td>
</tr>
<tr>
<td><code>Out[1]:</code></td>
<td>

```
a    1.0
b    2.0
c    NaN
Name: c1, dtype: float64
```

</td>
</tr>
<tr>
<td><code>In [2]:</code></td>
<td>

```py
s.obs = df2
s.obs['c1']
```

</td>
</tr>
<tr>
<td><code>Out[2]:</code></td>
<td>

```
KeyError: 'c1'
```

</td>
</table>
</tr>

ivirshup:
@joshua-gould, that seems wrong, thanks for the report! (btw: I think you're missing an `import numpy as np`)

@falexwolf or @flying-sheep: this should be creating a copy of s then assigning `df2` to that, right? The setter looks like it's trying to do the modification in-place:

https://github.com/theislab/anndata/blob/fc7f3cb06c191aff4fcad6af88f0c3012d7d1a65/anndata/base.py#L1090-L1098

(ditto with var)

flying-sheep:
In the commit that introduced the view functionality,  542edccc29edf0b9d43b3a912c213cf18b599a6d, the `AnnDataAttributeView` and `ArrayView` classes already had copy-on-write behavior (their `__setitem__` methods called `adata_view._init_as_actual(…)`)

That was also the last commit that touched the last two lines of `obs.setter` and so on, not introducing COW behavior for those.

I also think that the only sound behavior would be to be consistent here,

ivirshup:
Agreed.

I just keep running into code that's definitely trying to make modifications in-place.

Aside on the formatting above:

* I think you might want `In[2]:` 
* Is that just a table, or is there nice notebook formatting now? I can't look at the source markdown on this repo :(

flying-sheep:
Sadly just a table. It was an experiment to see how much work it is, and the answer is: too much (I had to do the HTML markup manually)

Issue #142, opened 2019-04-21
Failing to set value on view mutates anndata (makes it not a view)


ivirshup:
Attributes which derive from the `_SetItemMixin` mutate the anndata object even when setting the values fails. Here's an example:

```python
import scanpy as sc

pbmc = sc.datasets.pbmc68k_reduced()
subset = pbmc[pbmc.obs["bulk_labels"] == "Dendritic"]

subset.isview  # True
subset.obs["new_val"] = [1, 1]  # Raises ValueError
subset.isview  # False
```

I think the cause is line 391 from here: https://github.com/theislab/anndata/blob/34f4eb63710628fbc15e7050e5efcac1d7806062/anndata/base.py#L383-L392

The anndata is being modified before the value has been set. This could be averted by making a copy of the attribute to be modified, modifying that, then modifying the anndata object.

ivirshup:
Fixed!

Issue #136, opened 2019-04-10
Backwards breaking change: Sparse observation and variables annotation


falexwolf:
We had this discussion several times already, last time here: https://github.com/theislab/anndata/pull/133

It would be great to generate a new attribute for sparse annotations of observations and variables. Presumably `.obss` and `.vars`, where `s` stands for sparse.

I'd assume that @Koncopd could be the one to implement it, let's talk about it soon! If anyone else would feel there is bandwidth, please let me know.

flying-sheep:
why not just into obsm?

Issue #45, opened 2018-08-02
[bug] ValueError: could not broadcast input array from shape (1000,1) into shape (1000)


Hoeze:
I'd like to store some design matrix inside adata.obsm.
However, there are cases where this design matrix has only one column, i.e. it has shape (adata.n_obs, 1).

In this case I get the following error message:
```python3
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 2961, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-28-86462c4012f3>", line 1, in <module>
    adata.obsm["asdf"] = np.reshape(np.arange(1000), (1000,1))
  File "/usr/lib/python3.6/site-packages/anndata/base.py", line 126, in __setitem__
    new[name] = arr
ValueError: could not broadcast input array from shape (1000,1) into shape (1000)
```

Example to reproduce:
```python3
adata.obsm["asdf"] = np.reshape(np.arange(adata.n_obs), (adata.n_obs,1))```

PR #155, opened 2019-06-05
AlignedMapping (obsm, varm, layers refactor)

ivirshup:
Supersedes #140.

This work is now ready for a final review and merging. The history is a bit unwieldy, but I'm happy to cut it down if requested.

This PR refactors `obsm`, `varm`, and `layers` to have a shared root class and backend. Now they are all `Mappings` whose contents have restrictions on their `shape`. This means they're much more flexible, and can contain array-like objects of multiple types. Also this centralizes the code for taking views of these objects and makes their behavior consistent. Currently, this works best with `scipy.sparse.spmatrix` and `np.ndarray` type objects. `pd.DataFrames` work alright, but need to have reading and writing implemented. That will follow in another pull request, since I think this has diverged far enough from master and is otherwise self contained.

This PR can close the following issues:

* #142 This no longer occurs.
* #148 A user can now modify `.obs` of a view, a copy of the view is made first.
* #136 obsm and varm now support sparse matrices.
* #45 Since obsm and varm values are only expected to align on their first axis, more shapes are allowed.

Work to follow from this PR includes:

* Modifying the scanpy docs to make clear the copy on modify nature of views of anndata objects
* Allowing dataframes in to be stored in an AnnData's mappings (mostly as issue of IO)
* Add `obsp`, `varp` attributes for storing pairwise measurements (i.e. cell-cell knn graphs, or gene correlation matrices). This is currently commented out pending test and IO implementation.
* (further in the future) Allow lazy and backed loading for arrays in aligned mapping.


LuckyMD:
Is this also something that is/will go towards being able to write pandas dataframes from `.uns` to .h5ad?

ivirshup:
Yes, though a little tangentially. I want to be able to store dataframes anywhere in an AnnData, so `.uns` is included in that. This requires defining an on disk representation for these dataframes. Since it'll be written to disk, it's probably worth making sure we're happy with how it's done, or at least having some versioning info.

My draft of how to do this keeps the current "basically a recarray" format, but adds a tag so we know to decode it as a data frame. It also stores the categorical variables with the hdf dataset, so we don't need to put them in `.uns`. This is probably a non-optimal solution, but would be very easy to implement.

<details>
<summary> Draft implementation </summary>

```python
import numpy as np
import pandas as pd
from pandas.api.types import (
    is_string_dtype,
    is_categorical
)
from anndata import h5py
import h5py as _h5py

# This is basically a modified `df_to_records_fixed_width`
def _df_to_hdf5_recarray(df, var_len_str=True):
    """Convert a pandas dataframe into a recarray h5py can write."""
    if is_string_dtype(df.index):
        if var_len_str:
            index = df.index.values.astype(h5py.special_dtype(vlen=str))
        else:
            max_len_index = 0 if 0 in df.shape else df.index.map(len).max()
            index = df.index.values.astype('S{}'.format(max_len_index))
    else:
        index = df.index.values
    names = ['index']
    arrays = [index]
    formats = [index.dtype]
    for k in df.columns:
        names.append(k)
        if is_string_dtype(df[k]) and not is_categorical(df[k]):
            if var_len_str:
                arrays.append(df[k].values)
                formats.append(h5py.special_dtype(vlen=str))
            else:
                lengths = df[k].map(len)
                if is_categorical(lengths):
                    lengths = lengths.cat.as_ordered()
                arrays.append(df[k].values.astype('S{}'.format(lengths.max())))
                formats.append(arrays[-1].dtype)
        elif is_categorical(df[k]):
            arrays.append(df[k].cat.codes)
            formats.append(_make_h5_cat_dtype(df[k].values))
        else:
            arrays.append(df[k].values)
            formats.append(arrays[-1].dtype)
    assert len(formats) == len(arrays)
    return np.rec.fromarrays(
        arrays,
        dtype={'names': names, 'formats': formats})

def _make_h5_cat_dtype(s):
    """This creates a hdf5 enum dtype based on a pandas categorical array."""
    dt = h5py.special_dtype(enum=(
        s.codes.dtype,
        dict(zip(s.categories, np.unique(s.codes)))
    ))
    return dt

def write_dataframe(f, key, value, dataset_kwargs={}):
    recs = _df_to_hdf5_recarray(value)
    f.create_dataset(key, data=recs, **dataset_kwargs)
    f[key].attrs["source_type"] = "dataframe"  # Add a tag so we know how to decode it

def read_dataframe(dataset):
    df = pd.DataFrame(dataset[:])
    df.set_index(df.columns[0], inplace=True)
    dt = dataset.dtype
    for col in dt.names[1:]:
        check = _h5py.check_dtype(enum=dt[col])
        if not isinstance(check, dict):
            continue
        mapper = {v: k for k, v in check.items()}
        df[col] = pd.Categorical(df[col].map(mapper), ordered=False)
    return df
```

</details>

LuckyMD:
Okay, so once this is merged, I will change the `sc.tl.marker_gene_overlap()` function to allow storing the DataFrames inplace in `adata.uns`.

ivirshup:
Adding IO for dataframes will be in a separate pull request. I'm aiming for these to be in the same release, but I'd hold off for just a bit more.

ivirshup:
@falexwolf, did the questions you mentioned with `read_loom` (https://github.com/theislab/anndata/pull/140#issuecomment-499036092) get resolved?

falexwolf:
Yes, just waiting 2 min for the tests to complete. Would you add all info about `obs_vector` etc. to the release notes? And then we make the last `0.6` sub release (it's been more than a year of 0.6 now). 🙂

falexwolf:
Made `v0.6.21`. We can move forward with merging, in my opinion.

ivirshup:
I think it's good to go. Unless anyone else would like to review? @flying-sheep @Koncopd?

flying-sheep:
Things are still incorrectly named “views”.

PR #157, opened 2019-06-05
fix write_zarr and write_h5ad

Koncopd:
https://github.com/theislab/anndata/issues/156

### DIFF
--- a/anndata/readwrite/write.py
+++ b/anndata/readwrite/write.py
@@ -269,7 +269,7 @@ def _write_key_value_to_h5(f, key, value, **kwargs):
     if value is None or not value.dtype.descr:
         return
     try:
-        if _set_matching_array(f, key, value, h5py.Dataset, 'sparse_format'):
+        if _set_matching_array(f, key, value, h5py.Dataset, ['sparse_format']):
             return
         f.create_dataset(key, data=value, **kwargs)
     except TypeError:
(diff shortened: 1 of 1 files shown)


flying-sheep:
Hi, I deduplicated this a bit. My IDE just broke and I need to leave soon so I didn’t test it.

Koncopd:
Actually it is pointless because of this line
https://github.com/Koncopd/anndata/blob/4c6dc4ca1566994cbe6ecc5f0e72c29c249348f2/anndata/readwrite/write.py#L266

Koncopd:
We need to delete [this line](https://github.com/Koncopd/anndata/blob/4c6dc4ca1566994cbe6ecc5f0e72c29c249348f2/anndata/readwrite/write.py#L266) if we want all this commits to works. I'm not sure why it is there, but if it is really needed then we can just remove all these conditions as with this line they will never be active.

flying-sheep:
> By the way, the previous version also worked with lists, i.e allowed multiple excluded attrs.

Huh? How? You had `{excluded_attrs}`, which creates a set containing “excluded_attrs” as its single element. `excluded_attrs = ['a', 'b']; {excluded_attrs}` will result in `Error: A list object is not hashable` and `excluded_attrs = ('a', 'b'); {excluded_attrs}` will be a `Set[Tuple[str, str]]`: `{('a','b')}`.

> Need to test if this line is really required.

Yes. The line was introduced by @tomwhite implementing zarr storage in 2102629fbbc13c5948485c051bb6888680a681bf. Maybe he can help?

Koncopd:
Yeah, you are absolutely right about the problem with `{excluded_attrs}`, i was so sure about this for some reason that didn’t even check. My mistake, thank you. `{*excluded_attrs}` also seems to work, but it will split a string into characters.

Koncopd:
Hm, i see that these line was before that commit and i think it was just copied from write_h5ad to write_zarr.

flying-sheep:
Superseded by @ivirshup’s IO refector

Issue #159, opened 2019-06-06
anndata in scanpy seems to break scipy sparse data


reemagit:
Whenever I import anndata or scanpy it seems that scipy sparse matrix assignment is broken. The minimal example is:

```python
import numpy as np
from scipy.sparse import csr_matrix
import anndata
x = csr_matrix(np.eye(10))
x[0,3] = int(4)
```

which for which I get the error

```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-36-2e1809373207> in <module>
----> 1 x[0,3] = int(4)

~/anaconda2/envs/macrophage/lib/python3.6/site-packages/scipy/sparse/_index.py in __setitem__(self, key, x)
     67             if x.size != 1:
     68                 raise ValueError('Trying to assign a sequence to an item')
---> 69             self._set_intXint(row, col, x.flat[0])
     70             return
     71 

~/anaconda2/envs/macrophage/lib/python3.6/site-packages/scipy/sparse/compressed.py in _set_intXint(self, row, col, x)
    795     def _set_intXint(self, row, col, x):
    796         i, j = self._swap((row, col))
--> 797         self._set_many(i, j, x)
    798 
    799     def _set_arrayXarray(self, row, col, x):

~/anaconda2/envs/macrophage/lib/python3.6/site-packages/anndata/h5py/h5sparse.py in _set_many(self, i, j, x)
    176     i, j, M, N = self._prepare_indices(i, j)
    177 
--> 178     n_samples = len(x)
    179     offsets = np.empty(n_samples, dtype=self.indices.dtype)
    180     ret = _sparsetools.csr_sample_offsets(M, N, self.indptr, self.indices,

TypeError: object of type 'numpy.float64' has no len()
---------------------------------------------------------------------------
```

I am using anndata 0.6.21. Am I doing something wrong?

EDIT: just as a clarification, this error doesn't occur if I don't import anndata

ivirshup:
I think I just caught this myself.

Are you using scipy `v1.3`?

reemagit:
No, I am using the last version 1.4.3. 

As I have understood, the __setitem__ function of scipy CSR matrices ([…]/scipy/sparse/_index.py) first converts the value to be set to a scalar and then calls _set_many. 
[...]

PR #160, opened 2019-06-08
Allow changing sparsity structure in backed mode

ivirshup:
Allows modifying backed AnnData in scipy 1.3.0 (I think the cause of #159) and giving more helpful errors when trying to modify a backed AnnData's sparse `.X`.


ivirshup:
@Koncopd, I've reenabled the error and commented out the code for changing the sparsity structure of a backed anndata https://github.com/theislab/anndata/blob/8bed6fa201ac2f3df72e1ac6aa50959f23969022/anndata/h5py/h5sparse.py#L219-L233. I did this because when I tried testing it, it threw an error under scipy 1.3.0, and silently failed to modify the sparse matrix with scipy 1.2.1. To your knowledge, should that work? If so, could you send me specs for the environment it worked in?

Koncopd:
@ivirshup Actually, it just could not work, so it didn't. I added some changes and it seems to work properly now, but i'm not sure if it is a good solution as it requires loading indptr, indices and data into memory.

ivirshup:
Yeah, that was my impression. I think to allow changes to sparsity structure in any efficient way would require the sparse dataset to be resizable, which isn't how we initialize those objects.

I suggest we just don't allow changes in sparsity structure for backed objects (throw an error if it's attempted), but allow inplace modification otherwise. At least for now, this solves the issue of scipy 1.3.0 compatibility and we can test for.

Koncopd:
@ivirshup actually it is resizable. 
I think it is better to leave as i did - allow to change sparsity structure but throw a warning. You can check my commits.

ivirshup:
Ah, I had been searching for `maxsize` in the repo, which definitely wouldn't have found uses of `maxshape`. Probably thought that cause its `dataset.resize`.

I'm not sure I think we should allow people to do something with such poor performance. To me, if someone set the mode to be backed, they didn't intend to load the whole matrix into memory.

Koncopd:
The matrix is in the compressed form, so it is not that bad. If there is no need in changing the sparsity structure, it will not load anything. I think it is better to allow changing the sparsity structure with a warning even if it is expensive than just forbid it.

Koncopd:
@falexwolf , and what do you think?

falexwolf:
First thought: I'd tend to forbid changing the sparsity structure in backed mode. If one wants to change it, one can go to memory mode and change it. Typically, if in backed mode, you don't have the memory to load in things.

Second thought: Is there a lazy backed mode, meaning, that if calling `.X[:, :]` we load the whole sparse matrix into memory but keep the rest of the object backed. If we do it, we should be able to change the sparsity structure as the whole thing is in memory. I know, it's probably hard to solve this "lazy-loading" of just the attributes that we actually want in memory. But it could be a very general feature...

ivirshup:
On your second thought, that's what I've been thinking of doing – once it's reasonably easy to deal with each of the attributes separately. I'm thinking this could be implemented as a set of [descriptor](https://docs.python.org/3.7/howto/descriptor.html) classes. For example, a `Lazy` attribute would only get load the data when accessed, at which point it would become a `Memory` attribute with the current behavior.

flying-sheep:
Is this ready to merge? We need this to fix #159, #165, and theislab/anndata2ri#18

ivirshup:
I think @falexwolf and I agree that we shouldn't load the whole expression matrix into memory while in backed mode for now.

@Koncopd would you be alright with me reverting your changes? At least until we can keep memory usage low, potentially by keeping the rest of the object backed?

falexwolf:
Agreed with @ivirshup !

Koncopd:
@ivirshup yes, of course.

flying-sheep:
I already merged the other commits, so this PR can stay open until we can merge it too!

flying-sheep:
@ivirshup in the light of the IO refactor: Can we merge this now or does it still load the whole thing into memory?

ivirshup:
This still loads everything into memory. Not sure how close we are to having changes in sparsity structure (without loading everything) working in backed mode.

flying-sheep:
If this is in hiatus I’ll go ahead and reformat the last file as well.

ivirshup:
I've been thinking about this recently (along with general operations on backed data) and had an idea which I think could help simplify this, inspired by [pybedtools](https://daler.github.io/pybedtools/topical-design-principles.html#principle-1-temporary-files-are-created-and-deleted-automatically).

So that we don't have to deal with updating a sparse matrix on-disk in-place, we instead make a new "temporary" object and write the new data there. The new on disk object will now be what adata.X refers to. Advantages here:

* Don't have to worry about failed in-place operations corrupting original data
* Don't have to bother with shuffling the data as we resize the original arrays

This does separate the idea of one file: one AnnData object, but I think that increase in complexity is easier to deal with than the above two points. I think this can be summarized to users as "We don't promise that `adata.filename` is up to date with all changes unless you've called `adata.write`".

Another advantage here is that we could allow `backed.copy()` without having to provide a filename. This should make it much easier to make backed mode work with existing code.

PR #161, opened 2019-06-11
[Suggestion] Run travis with scipy 1.2 and 1.3

ivirshup:
Until we can set scanpy to work with scipy 1.3, I think we should test against scipy 1.2 and 1.3. Especially since sparse matrices changed so much.


flying-sheep:
superseded by #166

PR #162, opened 2019-06-11
No concatinate duplicates in var

Koncopd:
This leaves only one var column after concatination if this column is the same (same name, same values) in all anndata objects.

### DIFF
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -174,7 +174,7 @@ def _gen_dataframe(anno, length, index_names):
         anno = anno.copy()
         if not is_string_dtype(anno.index):
             logger.warning('Transforming to str index.')
-            anno.index = anno.index.astype(str)        
+            anno.index = anno.index.astype(str)
         return anno
     if anno is None or len(anno) == 0:
         _anno = pd.DataFrame(index=RangeIndex(0, length, name=None).astype(str))
@@ -1829,6 +1829,17 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
         var = pd.DataFrame(index=var_names)
 
+        if join == 'inner':
+            ad_ref = all_adatas[0]
+            cols_intersect = set(ad_ref.var.columns)
+            for ad in all_adatas[1:]:
+                cols_intersect &= set(ad.var.columns)
+                cols_intersect = {
+                    col for col in cols_intersect
+                    if ad_ref.var.loc[var_names, col].equals(ad.var.loc[var_names, col])
+                }
+                if not cols_intersect: break
+
         obs_i = 0  # start of next adata’s observations in X
         out_obss = []
         for i, ad in enumerate(all_adatas):
@@ -1867,6 +1878,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
             # var
             for c in ad.var.columns:
+                if join == 'inner' and c in cols_intersect:
+                    if c not in var.columns:
+                        var.loc[vars_intersect, c] = ad.var.loc[vars_intersect, c]
+                    continue
                 new_c = c + (index_unique if index_unique is not None else '-') + categories[i]
                 var.loc[vars_intersect, new_c] = ad.var.loc[vars_intersect, c]
 
--- a/anndata/tests/test_base.py
+++ b/anndata/tests/test_base.py
@@ -77,7 +77,7 @@ def test_create_from_df_with_obs_and_var():
     assert df.index.tolist() == ad.obs_names.tolist()
     assert obs.equals(ad.obs)
     assert var.equals(ad.var)
-    
+
     from pytest import raises
     with raises(ValueError, match=r'Index of obs must match index of X.'):
         AnnData(df, obs=obs.reset_index())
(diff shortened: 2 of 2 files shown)


anndata/core/anndata.py
```diff
@@ -1829,6 +1829,30 @@ def concatenate(
 
         var = pd.DataFrame(index=var_names)
 
+        if join == 'inner':
```
falexwolf: @Koncopd, great! But does it really have to be that long? Can you just use `.drop_replicates` (`adata.var.drop_duplicates`) on the joint object and simplify this heavily? I would have expected a 2 line solution with no indendation... :wink:

This is way too much code and way too complicated. 🙂

Koncopd: I can do as you say, but i thought we need not just remove duplicates, but remove duplicates and leave the original column name (without "-0") only if  **all** anndatas have the same var column.
This is the reason i wrote it in this way. 
If we don't need it, i willjust add (however it will require to copy the whole var i believe)
```
var = var.T
var.drop_duplicates(inplace=True)
var = var.T
```

falexwolf: Copying the whole var is not a problem. Minimizing code complexity is more important, here. And yes, we need to leave the original column name (without "-0"). :)

falexwolf: Thank you!

Koncopd: @falexwolf 
It is still unclear to me what are the right conditions for removing "duplicated" columns.
What if columns have different names but the same values? drop_duplicates drops such columns, but is it desired?
For example,

```
    adata1 = AnnData(X1,
                     {'obs_names': ['s1', 's2'],
                      'anno1': ['c1', 'c2']},
                     {'var_names': ['a', 'b', 'c'],
                      'annoA': [0, 1, 2]},
                     layers={'Xs': X1})
    adata2 = AnnData(X2,
                     {'obs_names': ['s3', 's4'],
                      'anno1': ['c3', 'c4']},
                     {'var_names': ['d', 'c', 'b'],
                      'annoA': [0, 1, 2]},
                     layers={'Xs': X2})
    adata3 = AnnData(X3,
                     {'obs_names': ['s1', 's2'],
                      'anno2': ['d3', 'd4']},
                     {'var_names': ['d', 'c', 'b'],
                      'annoB': [0, 1, 2]},
                     layers={'Xs': X3})

    # inner join
    adata = adata1.concatenate(adata2, adata3)
```
The result for `adata.var_keys()` will be `['annoA-0', 'annoA-1']`, it removes `'annoB'`.
Is this approach correct?

Also, it is not clear when exactly we should leave the original column name, and where we shouldn't.

flying-sheep: Yeah, we should rather fix the code that creates the duplicated columns.

When concatenating and there’s identical columns with identical names, they should be merged, I think.

Why do we copy columns manually instead of using `var1.join(var2)`?

Koncopd: Yes, thanks, this seems like the right approach. 

Still unsure about naming of duplicates and non-duplicates as there are batch categories and so on in the function. If we need to leave the original name, then what to do with columns that have the same name but different content? Like the former will have the original name, the latter will have "name+number of the batch". This is not systematic.

flying-sheep: I think renaming them all and throwing a warning would be a good approach.

falexwolf: It's really just about the rows of `.var` that have exactly the same column names and exactly the same values. This is a big use case. You could even simplify to making this reduction only if all rows in `.var` meet this condition.

If not, it makes sense to be as verbose as we are in the current case. All intermediate cases add complexity.

Koncopd: @falexwolf , 

> rows of .var that have exactly the same column names and exactly the same values

> all rows in .var meet this condition

Do you mean columns instead of rows?

falexwolf: Yes, sorry, I mean columns of the `var` DataFrame (when saying "rows", I had the whole AnnData object in mind). `gene_ids-0`, `gene_ids-1`, etc. doesn't make sense if they store exactly the same content. This is the main use case.

Koncopd:
@falexwolf 
Now this should be consistent with you last conditions

> You could even simplify to making this reduction only if all rows in .var meet this condition.

anndata/tests/test_base.py
```diff
@@ -375,6 +375,34 @@ def test_concatenate():
...
+    adata3 = AnnData(X3,
+                     {'obs_names': ['s1', 's2'],
+                      'anno2': ['d3', 'd4']},
+                     {'var_names': ['a', 'b', 'c'],
+                      'annoA': [0, 1, 2],
+                      'annoB': [1.1, 1.0, 2.0],
+                      'annoD': [2.1, 2.0, 3.1]})
+
+    adata = adata1.concatenate(adata2, adata3)
+    assert adata.var_keys() == ['annoA', 'annoB', 'annoC-0', 'annoD-0', 'annoC-1', 'annoD-1', 'annoD-2']
```
falexwolf: Perfect, @Koncopd ! :smile:

falexwolf:
Merging this now. :)

anndata/tests/test_base.py
flying-sheep: That looks great!

LuckyMD:
I guess this is backwards breaking for my tutorial where I remove the duplicates. Will adapt that then.

Issue #62, opened 2018-09-14
Views take around as much memory as object


ivirshup:
I was playing around with some visualization on a large dataset, when I noticed some surprisingly high memory usage. I think I've narrowed it down to unexpected memory growth from taking views:

```python
In [1]: import scanpy.api as sc

In [2]: %load_ext memory_profiler

In [3]: %memit adata = sc.read("bm.h5ad")
peak memory: 5317.54 MiB, increment: 5187.05 MiB

In [4]: %memit
peak memory: 2624.52 MiB, increment: 0.00 MiB

In [5]: %memit view = adata[:, (adata.var["n_cells_by_counts"] > 10000)]
peak memory: 5299.68 MiB, increment: 2675.16 MiB

In [6]: %memit
peak memory: 5080.07 MiB, increment: 0.00 MiB
``` 

My assumption here being: taking a view shouldn't cause noticeable growth in memory usage. I'm pretty sure it's not just how `memory_profiler` is counting objects, since `top` and ActivityMonitor pick this up as well.

falexwolf:
Thank you for pointing this out! It's surprising to me. I extensively memory-profiled when introducing the functionality about a year ago.

Right now, I can only think of one thing that would cause this behavior as a result of the design of AnnData being not mature enough: the `.uns` annotations are deep copied... everything else uses numpy's views in the background.

The other reason for views, besides saving memory (if they work properly) is that one should be able to write to the data matrix of the underlying object, especially in backed mode, that is
```
adata[batch].X = function(adata[batch].X)
```
should modify the original data (see [here](https://falexwolf.de/blog/171223_AnnData_indexing_views_HDF5-backing/)).

Anyways, this is a severe issue and I'll definitely fix it soon if it persists.

ivirshup:
In the example I gave, I don't think there was anything in the `uns` field. Just to be sure, I've run it again with a AnnData object that definitely doesn't have anything in the `uns` field and gotten similar results.

Question about the syntax for views. What should happen in the following code?

```python3
view = adata[adata.obs["total_counts"] > 500, :]
view.X[view.X != 0] = 0
print(view.X.sum())
print(adata[adata.obs["total_counts"] > 500, :].X.sum())
```

My assumption is the change on the view should be replicated in the base object. That is, both print statements should return 0.

grst:
Observed the same when filtering an adata object. For demonstration purposies, here, the same filter is applied over and over again: 

```python
import numpy as np
import scanpy.api as sc
[...]

Issue #145, opened 2019-04-30
inconsistent slicing of .X vs .layer


fidelram:
This is related to #74 for for slices and is behind the problem reported in https://github.com/theislab/scanpy/pull/555

```PYTHON
from  scipy.sparse import csr_matrix

adata = sc.AnnData(np.ones((3,3)))
adata.layers['test'] = adata.X.copy()
adata.layers['sparse'] = csr_matrix(adata.X)

print(adata[:, ['0']].X.shape, 
      adata[:, ['0']].layers['test'].shape,
      adata[:, ['0']].layers['sparse'].shape)
print(adata[:, '0'].X.shape, 
     adata[:, '0'].layers['test'].shape, 
     adata[:, '0'].layers['sparse'].shape)

print(type(adata[:, ['0']].X), 
      type(adata[:, ['0']].layers['test']), 
      type(adata[:, ['0']].layers['sparse']))

print(type(adata[:, '0'].X), 
      type(adata[:, '0'].layers['test']), 
      type(adata[:, '0'].layers['sparse']))
```

The output is:
```
(3,) (3, 1) (3, 1)
(3,) (3,) (3, 1)
<class 'anndata.base.ArrayView'> <class 'numpy.ndarray'> <class 'scipy.sparse.csr.csr_matrix'>
<class 'anndata.base.ArrayView'> <class 'numpy.ndarray'> <class 'scipy.sparse.csr.csr_matrix'>
```

Ideally, while slicing .layers, the output should be the same as for .X

falexwolf:
@Koncopd, any easy fix for this?

This whole thing relates back to the early issue of not flattening `.X` when effectively 1d (what Phil always wanted) and instead using another convenience accessor (@ivirshup's `.get_obs_array` or `.get_obs_1dslice` or `.get_obs_1d` or `.get_1d_obs` (I like the last best) or a similar name https://github.com/theislab/anndata/pull/144).

The problem is that this is a backwards breaking change and many people got used to `adata.[, 'gene_name'].X` giving something 1d. But I do think we should move there and make this change and point people to `.get_1d_obs`. We could make it anndata 2.0 and put a big warning in the docs. Replacing all occurances in Scanpy is no problem at all.

fidelram:
If I remember correctly, in some past discussions it was proposed that the slicing should to mimic the behavior of numpy, pandas etc. which return a 1d array in the case of a single slice. Why don't we stick with this behaviour?

ivirshup:
[...]

PR #164, opened 2019-06-16
Views of views

ivirshup:
Addressing #62.

This PR aims to make it so when you subset an AnnDataView, it makes a reference to the original AnnData.

Things left to do:

- [x] The test is a bit flaky, for certain combinations of subsetting techniques it generates incompatible indexers.
- [x] I should make sure AlignedMappings work well with this, I suspect they may not due to problems getting `.X` working right.

-----------------------------

Update:

A few more features of this pull request:

* Memory usage from taking a view are significantly reduced, since X is now lazy for views (You only subset X if the user accesses the attribute)
* More extensive testing of subsetting anndata objects

### DIFF
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -819,13 +819,14 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             )
         else:
             X = self._X
-        if self.n_obs == 1 and self.n_vars == 1:
-            return X[0, 0]
-        elif self.n_obs == 1 or self.n_vars == 1:
-            if issparse(X): X = X.toarray()
-            return X.flatten()
-        else:
-            return X
+        return X
+        # if self.n_obs == 1 and self.n_vars == 1:
+        #     return X[0, 0]
+        # elif self.n_obs == 1 or self.n_vars == 1:
+        #     if issparse(X): X = X.toarray()
+        #     return X.flatten()
+        # else:
+        #     return X
 
 
     @X.setter
@@ -1298,6 +1299,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             dfs = [self.obs, self.var]
             if self.isview:
                 if not self.isbacked:
+                    warnings.warn("Initializing view as actual.")
                     self._init_as_actual(self.copy())
                 else:
                     dont_modify = True
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -1568,11 +1568,21 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
         """Full copy, optionally on disk."""
         if not self.isbacked:
-            X = self.X
+            if self.isview:
+                # TODO: How do I unambiguously check if this is a copy?
+                # Subsetting this way means we don't have to have a view type
+                # defined for the matrix, which is needed for some of the current
+                # distributed backend.
+                X = _subset(self._adata_ref.X, (self._oidx, self._vidx)).copy()
+            else:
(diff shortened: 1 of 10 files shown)


ivirshup:
I've just pushed an update which significantly speeds up indexing. Instead of creating a new pd.Series with it's own index (a fairly expensive computation, I think this has to do with making a new index) I've just relied on the existing index.

falexwolf:
This looks great, @ivirshup !

anndata/core/anndata.py
```diff
@@ -571,23 +575,29 @@ def _init_as_view(self, adata_ref: 'AnnData', oidx: Index, vidx: Index):
                 'Currently, you cannot index repeatedly into a backed AnnData, '
                 'that is, you cannot make a view of a view.')
         self._isview = True
+        if isinstance(oidx, (int, np.integer)): oidx = slice(oidx, oidx+1, 1)
+        if isinstance(vidx, (int, np.integer)): vidx = slice(vidx, vidx+1, 1)
```
ivirshup: This touches on #145, since it effectively does what I said I didn't want and maintains the dimensionality of all constituent objects when the container object is subset. For example: `adata[1, :].layers["counts"]` is now a two dimensional ndarray.

This did make it much easier to get everything working, since I know which axis of all the contained array aligns to which axis of the AnnData object without having to check either's dimensions.

However, this makes some tests fail, since there are effectively checks on the dimensionality of `X`. To get around this I modified the accessor for `X` to return a one dimensional array if any dimension of `X` had a length of 1. This has opened a new can of worms with inconsistencies between `X` and `layers` as well as `adata[0, :].X.shape == adata[[0], :].X.shape` (which previously was not the case).

I'm trying to figure out what the right way forward here is. It makes our lives easier if contained arrays match up with the shape of their `AnnData`, but there's code that relies on this not being the case (`sc.pl.umap` for example, though this currently doesn't work with sparse layers anyways).

ivirshup:
@falexwolf I'm thinking I'd like to merge this soon, mostly so that we can have the always-2d and the improved test suite on master.

Potential issues

* Some scanpy plotting tests will fail without https://github.com/theislab/scanpy/pull/730 being merged
* Distributed tests will fail, but I'm pretty sure they were silently doing the wrong thing before https://github.com/theislab/scanpy/pull/733.

Some things to note:

* I've updated the docs for AnnData, but this could definitely use another pass.
* I've made sure the behaviour for `adata[idx1].X[idx2] = val` doesn't modify the original `adata` for an in memory dataset. This was the case in earlier versions of `anndata` as well (at least `0.6.19`, `0.6.20`). I feel like this wasn't the intended behavior, but I don't think we currently have the code to deal with this case otherwise.

Just to be clear on current behaviour:

```python
from anndata import AnnData
import numpy as np

a1 = AnnData(np.zeros((10, 10)))
a2 = a1.copy()

a1[0, 0].X = 1
assert a1.X[0, 0] == 1

a2[:5, :5].X[0, 0] = 1 # Prints warning about making copy
assert a2.X[0, 0] == 0
```

anndata/tests/test_hdf5_backing.py
```diff
@@ -172,3 +172,26 @@ def test_backed_modification_sparse(adata, backing_h5ad, sparse_format):
     assert np.all(adata.X[0, :] == np.array([10, 0, 10]))
     assert np.all(adata.X[1, :] == np.array([11, 0, 12]))
     assert np.all(adata.X[2, :] == np.array([7, 0, 9]))
+
+
+# TODO: Work around h5py not supporting this
```
ivirshup: I'm putting this off so it can be fixed with other changes to IO.

falexwolf:
Congrats on the merge! Fantastic work, @ivirshup ! :D

PR #167, opened 2019-06-20
[WIP] IO refactor

### LINKED ISSUE #200: reading categorical with only one category
Hi @flying-sheep 

rather a rare occasion, though maybe important to mention:

```py
import anndata, scanpy, pandas, numpy
adata = anndata.AnnData(np.array([[2, 1], [2, 1]]))
adata.obs['a'] = pandas.Categorical(numpy.ones(2))
adata.write('filename.h5ad')
scanpy.read('filename.h5ad')
```

raises an error failing to read categoricals when only consisting of one category (both in .var and .obs).

ivirshup:
This is a major reworking of the IO which is meant to make it easier to modify and figure out how each element gets written to a file by factoring them out. This makes it easier to read one object at a time from an h5ad file. Additionally, this PR makes allows reading and writing of dataframes to locations other than `obs` and `var`.

TODO:

- [x] Chunked zarr reading/ writing
- [x] Get zarr reading and writing working
- [x] Cleanup old code
- [x] H5AD dataframes as columnar
- [x] Tests for reading legacy files

Fixes #200


Koncopd:
Hi, the important thing is to use read_direct where possible, it is really important for performance.
Like [here]( https://github.com/theislab/anndata/blob/c20ed1876b480ba43ba28d6b58eda54a18fbf850/anndata/readwrite/read.py#L520) or [here](https://github.com/theislab/anndata/blob/c20ed1876b480ba43ba28d6b58eda54a18fbf850/anndata/h5py/h5sparse.py#L311).

ivirshup:
> the important thing is to use read_direct where possible, it is really important for performance.

Is this only for cases when we would have an array to overwrite anyways, or others as well? I don't see a change if performance for this simple case:

<details>
<summary> simple hdf5 benchmark </summary>

```python
import h5py
import numpy as np

with h5py.File("tmp.h5") as f:
    f["x"] = np.random.random_sample((10000, 10000))

def read_direct(d):
    a = np.empty(d.shape, dtype=d.dtype)
    d.read_direct(a)

def read_parens(d):
    a = d[()]

def read_colon(d):
    a = d[:]

# Benchmark:
f = h5py.File("tmp.h5", "r")
d = f["x"]

# Time:

%timeit read_colon(d)
# 474 ms ± 5.51 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit read_parens(d)
# 467 ms ± 11.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit read_direct(d)
# 461 ms ± 4.23 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Memory (all measures fluctuated around 20 MiB):

%memit read_colon(d)
# peak memory: 1576.33 MiB, increment: 740.04 MiB

%memit read_parens(d)
# peak memory: 1558.64 MiB, increment: 722.46 MiB

%memit read_direct(d)
# peak memory: 1583.15 MiB, increment: 746.96 MiB
```

</details>

Koncopd:
Yeah, for small examples it doesn't show any difference, i also faced this fact when tried to fix the memory issue.
Maybe you remember - https://github.com/theislab/scanpy/issues/146
For really big datasets it does have a difference.

ivirshup:
I think the h5ad io code is pretty close to done. Here's some benchmark results I've just gotten using the benchmark suite I'm working on ([ivirshup/anndata-benchmarks](https://github.com/ivirshup/anndata-benchmarks)), comparing io speed and memory usage from the current commit on this branch to master:

<details>
<summary> Benchmark results </summary>

```
$ asv compare c20ed18 18c730f8f


       before           after         ratio
     [c20ed187]       [18c730f8]
     <0.6.22rc1>       <io_refactor>
            6.07M            6.07M     1.00  readwrite.H5ADReadSuite.mem_readfull_object('pbmc3k_raw.h5ad')
            2.32M            2.32M     1.00  readwrite.H5ADReadSuite.mem_readfull_object('10x_pbmc68k_reduced.h5ad')
-            124M            95.5M     0.77  readwrite.H5ADReadSuite.peakmem_read_full('pbmc3k_raw.h5ad')
              77M            72.9M     0.95  readwrite.H5ADReadSuite.peakmem_read_full('10x_pbmc68k_reduced.h5ad')
-         166±3ms       77.2±0.9ms     0.46  readwrite.H5ADReadSuite.time_read_full('pbmc3k_raw.h5ad')
-         117±1ms         61.1±2ms     0.52  readwrite.H5ADReadSuite.time_read_full('10x_pbmc68k_reduced.h5ad')
  1.1476142424679256  1.2287729571466595     1.07  readwrite.H5ADReadSuite.track_read_full_memratio('pbmc3k_raw.h5ad')
  1.0346641166626822  1.0333333333333334     1.00  readwrite.H5ADReadSuite.track_read_full_memratio('10x_pbmc68k_reduced.h5ad')
-            125M             104M     0.83  readwrite.H5ADWriteSuite.peakmem_write_compressed('pbmc3k_raw.h5ad')
            79.3M            76.8M     0.97  readwrite.H5ADWriteSuite.peakmem_write_compressed('10x_pbmc68k_reduced.h5ad')
-            125M             104M     0.83  readwrite.H5ADWriteSuite.peakmem_write_full('pbmc3k_raw.h5ad')
            80.6M            74.5M     0.92  readwrite.H5ADWriteSuite.peakmem_write_full('10x_pbmc68k_reduced.h5ad')
         693±20ms         680±10ms     0.98  readwrite.H5ADWriteSuite.time_write_compressed('pbmc3k_raw.h5ad')
          139±2ms          136±2ms     0.97  readwrite.H5ADWriteSuite.time_write_compressed('10x_pbmc68k_reduced.h5ad')
         406±10ms          363±8ms    ~0.89  readwrite.H5ADWriteSuite.time_write_full('pbmc3k_raw.h5ad')
-        61.5±2ms         55.3±1ms     0.90  readwrite.H5ADWriteSuite.time_write_full('10x_pbmc68k_reduced.h5ad')
-         20.1875          13.0625     0.65  readwrite.H5ADWriteSuite.track_peakmem_write_compressed('pbmc3k_raw.h5ad')
+      3.30859375       4.91015625     1.48  readwrite.H5ADWriteSuite.track_peakmem_write_compressed('10x_pbmc68k_reduced.h5ad')
-      20.6953125         13.84375     0.67  readwrite.H5ADWriteSuite.track_peakmem_write_full('pbmc3k_raw.h5ad')
+        3.296875       4.08203125     1.24  readwrite.H5ADWriteSuite.track_peakmem_write_full('10x_pbmc68k_reduced.h5ad')
```

The `+` and `-` indicate whether a significant increase or decrease (respectively) in measured value occurred.

</details>

I still need to get set up on a better machine for benchmarks that my laptop, but I think these results would carry over. By the way, I'd appreciate any input you had on the choices of things to benchmark!

Koncopd:
If it is almost ready, i can check this on the same huge datasets.

falexwolf:
This makes everything so much better! What a major improvement again! Awesome

Benchmarks: Great! And thank you @Koncopd, for pointing out the large-data memory issues!

ivirshup:
# Update

I think this is very close, and could really use a looking over. The big remaining thing to be implemented is chunked reading and writing.

Some questions I have:

* How should I include datasets? I'd like to include a few to show that backwards compatibility works. `git-lfs` maybe?
* I think the option of reading a dense array as sparse should happen when the file is read, not when it's written. Does anyone mind if that changes? (see test `test_readwrite_sparse_as_dense` for example)
* I've changed the internal representation of dataframes from recarrays to columnar (i.e. a dataframe is a group, with every series being a dataset) since this was just so much easier. I'd like to hear thoughts on this. My main reasons for doing this are:
  * It's much easier to add a column this way, and similarly difficult to add rows.
  * Per column metadata makes it easier to deal with categoricals.
  * My impression is that columnar formats are now more popular for tabular data than the "list of records" format.

## Current benchmarks:

As of https://github.com/ivirshup/anndata-benchmarks/commit/28df5d6315a11a55fbfbadafa202d09e2dee9c20. They fluctuate a bit, but the general trends seem robust:

<details>
<summary> Results </summary>

```
$ asv compare 75296bb 58f6834c9

All benchmarks:

       before           after         ratio
     [75296bb2]       [58f6834c]
     <0.6.22.post1>       <io_refactor>
            6.07M            6.07M     1.00  readwrite.H5ADReadSuite.mem_readfull_object('/Users/isaac/data/pbmc3k_raw.h5ad')
            2.32M            2.32M     1.00  readwrite.H5ADReadSuite.mem_readfull_object('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-            127M            99.3M     0.78  readwrite.H5ADReadSuite.peakmem_read_full('/Users/isaac/data/pbmc3k_raw.h5ad')
            80.6M            75.1M     0.93  readwrite.H5ADReadSuite.peakmem_read_full('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-         161±3ms       76.9±0.7ms     0.48  readwrite.H5ADReadSuite.time_read_full('/Users/isaac/data/pbmc3k_raw.h5ad')
-         113±2ms       58.2±0.9ms     0.52  readwrite.H5ADReadSuite.time_read_full('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
  1.1544843370047204  1.2359829177572668     1.07  readwrite.H5ADReadSuite.track_read_full_memratio('/Users/isaac/data/pbmc3k_raw.h5ad')
  1.0357833655705997  1.0360161705255422     1.00  readwrite.H5ADReadSuite.track_read_full_memratio('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-            127M             107M     0.84  readwrite.H5ADWriteSuite.peakmem_write_compressed('/Users/isaac/data/pbmc3k_raw.h5ad')
            85.1M            80.3M     0.94  readwrite.H5ADWriteSuite.peakmem_write_compressed('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-            129M             108M     0.84  readwrite.H5ADWriteSuite.peakmem_write_full('/Users/isaac/data/pbmc3k_raw.h5ad')
            85.2M            79.3M     0.93  readwrite.H5ADWriteSuite.peakmem_write_full('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
         723±20ms         640±10ms    ~0.89  readwrite.H5ADWriteSuite.time_write_compressed('/Users/isaac/data/pbmc3k_raw.h5ad')
          138±2ms          135±4ms     0.98  readwrite.H5ADWriteSuite.time_write_compressed('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
        464±100ms         340±10ms    ~0.73  readwrite.H5ADWriteSuite.time_write_full('/Users/isaac/data/pbmc3k_raw.h5ad')
        71.0±30ms         58.3±1ms    ~0.82  readwrite.H5ADWriteSuite.time_write_full('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-         20.1875       12.2265625     0.61  readwrite.H5ADWriteSuite.track_peakmem_write_compressed('/Users/isaac/data/pbmc3k_raw.h5ad')
+      3.98046875            5.375     1.35  readwrite.H5ADWriteSuite.track_peakmem_write_compressed('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')
-     20.19140625      12.46484375     0.62  readwrite.H5ADWriteSuite.track_peakmem_write_full('/Users/isaac/data/pbmc3k_raw.h5ad')
       3.19140625       3.44140625     1.08  readwrite.H5ADWriteSuite.track_peakmem_write_full('/Users/isaac/github/scanpy/scanpy/datasets/10x_pbmc68k_reduced.h5ad')

```

</details>

ivirshup:
Minor updates:

Potential consideration for categorical series. In the h5ad backend I used hdf5 enums for categories. In the zarr backend a categorical array stores the codes as the values and the categories as a list in the attributes. While there isn't an enum type for zarr, the h5 backend would work just as well using the zarr representation (which is pretty close to what hdf5 does with enum types anyways).

`obsp` and `varp` are now working in a separate branch.

anndata/core/anndata.py
```diff
@@ -1,24 +1,25 @@
...
 from functools import reduce
 from pathlib import Path
-from typing import Any, Union, Optional
-from typing import Iterable, Sized, Sequence, Mapping
-from typing import Tuple, List
-from copy import deepcopy
+from typing import (
+    Any, Union, Optional,
+    Iterable, Sequence, Mapping, Tuple, List
+)
```
flying-sheep: This was deliberate, but I should have added comments:

```py
from typing import Any, Union, Optional                # Meta
from typing import Iterable, Sized, Sequence, Mapping  # Abstract collections
from typing import Tuple, List                         # Concrete types
```

And on separate lines because that’ll survive black formatting once we switch to it.

anndata/core/anndata.py
```diff
@@ -296,17 +297,20 @@ def __init__(
...
             self._var = adata.var.copy()
-            self._varm = adata.varm.copy()
+            self._varm = AxisArrays(self, 1, adata.varm.copy())
 
     @property
     def X(self):
         if self._adata.isbacked:
             if not self._adata.file.isopen: self._adata.file.open()
-            X = self._adata.file['raw.X']
+            if "raw/X" in self._adata.file._file.h5f:
```
flying-sheep: Please implement `AnnDataFileManager.__contains__` instead, then this could be:

```py
x_key = 'raw/X' if 'raw/X' in self._adata.file else 'raw.X'
X = self._adata.file[x_key]
```

or you could also implement `AnnDataFileManager.get` and do

```py
X = self._adata.file.get('raw/X') or self._adata.file['raw.X']
```

anndata/core/anndata.py
```diff
@@ -794,13 +789,20 @@ def _init_as_actual(
                 self._raw = raw
             else:
                 # is dictionary from reading the file, nothing that is meant for a user
-                shape = self.file['raw.X'].shape if self.isbacked else raw['X'].shape
+                if self.isbacked:
+                    if "raw/X" in self.file._file.h5f:
```
flying-sheep: same as above

anndata/tests/test_deprecations.py
```diff
@@ -34,32 +34,33 @@ def test_get_obsvar_array_warn(adata):
...
-    )
-    assert np.allclose(
-        adata._get_var_array("s1", layer="x2"),
-        adata.var_vector("s1", layer="x2")
-    )
-    assert np.allclose(
-        adata._get_var_array("s1", use_raw=True),
-        adata.raw.var_vector("s1")
-    )
+    with pytest.warns(DeprecationWarning):  # Just to hide warnings
```
flying-sheep: What warnings are thrown there anyway?

ivirshup: Deprecation warnings. But they're supposed to be thrown, and I can't get the pytest marker which for ignoring them to work.

ivirshup:
Thanks for the review!

Its a lot, so I know a complete look over would be too much, theres just a few things I wanted some input on/ to make sure someone else looked at.

* I would like to have more legacy files present for testing against. Should we use git-lfs to manage these?
* Should I use the hdf5 `enum` dtype for categoricals, or just handle them the same way I do it for zarr, where the categories are stored in the attributes? This is mainly a question of internal consistency vs consistency with the storage format.
* Dataframes on disk changed the most. Could you give that code a look over? Any cases you think might not work?

anndata/core/anndata.py
ivirshup: This is a little nitpicky on my part, but aren't the `Tuple` and `List` types generic, not concrete?

anndata/core/anndata.py
ivirshup: I've implemented `__contains__` but I've kept the multiline conditionals since I've added error messages if neither is present.

ivirshup:
@flying-sheep, do you think you'd be able to review those three parts sometime soon?

flying-sheep:
Hi! I’ve been thesis writing / on holiday, but back now :smile:

anndata/readwrite/h5ad.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+@report_key_on_error
+def read_dataframe_legacy(dataset) -> pd.DataFrame:
+    """
+    Read pre-anndata 0.7 dataframes.
+    """
+    df = pd.DataFrame(_from_fixed_length_strings(dataset[()]))
+    df.set_index(df.columns[0], inplace=True)
+    # TODO: Decide whether to keep next few lines, since I don't this ever made a release
```
flying-sheep: Nah, get rid of it if it wasn’t in a release. Less code is better.

anndata/readwrite/utils.py
```diff
@@ -51,3 +56,43 @@ def convert_string(string):
...
     else:
         return string
+
+
+class AnnDataReadError(OSError):
+    """Error caused while trying to read in AnnData."""
+    pass
+
+
+def report_key_on_error(func):
```
flying-sheep: Nice!

anndata/tests/test_deprecations.py
flying-sheep: I meant: By what package and/or line?

anndata/core/anndata.py
flying-sheep: Sure, I meant “concrete” in the sense of “non-abstract”. Is there a better word?

anndata/core/anndata.py
```diff
@@ -297,34 +269,38 @@ def __init__(
...
+                self._adata.file.open()
+            # Handle legacy file formats:
+            if "raw/X" in self._adata.file:
+                X = self._adata.file["raw/X"]
+            elif "raw.X" in self._adata.file:
+                X = self._adata.file['raw.X']  # Backwards compat
+            else:
+                raise AttributeError(
+                    f"Could not find dataset for raw X in file: {self._adata.file.filename}."
+                )
```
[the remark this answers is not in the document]
ivirshup: I added this here since we needed a condition to throw an error if some accessed `adata.raw.X` if it didn't exist. `adata.X` should already throw an error the way it's accessed.

As to this existing in `_init_as_actual`, this was based on us checking the size of `raw` via it's `X`. This is different than `adata.X`. It's possible to read an `h5ad` even it doesn't have `X` defined, but should this be the case?

anndata/readwrite/h5ad.py
```diff
@@ -0,0 +1,412 @@
...
+    elif value.dtype.names is not None:
+        value = _to_hdf5_vlen_strings(value)
+    f.create_dataset(key, data=value, **dataset_kwargs)
+
+
+def write_dataframe(f, key, df, dataset_kwargs={}):
+    if isinstance(f, adh5py.File):  # If it's the patched one
+        group = f.h5f.create_group(key)
+    else:
+        group = f.create_group(key)
```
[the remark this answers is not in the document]
ivirshup: Added. In general, I would like to remove the `anndata.h5py` file, since I'm pretty sure we can handle sparse matrices without also having our own `File`, `Group`, and `Dataset`. Since we already have to deal with indexing ourselves, I think it'll be easier to share code between backends (like `zarr`) as well. Do have any thoughts on this?

Thanks! Now I just need to actually use it.

anndata/readwrite/h5ad.py
```diff
@@ -0,0 +1,387 @@
...
+    # Convert unicode to fixed length strings
+    if value.dtype.kind in {'U', 'O'}:
+        value = value.astype(adh5py.special_dtype(vlen=str))
+    elif value.dtype.names is not None:
+        value = _to_hdf5_vlen_strings(value)
+    f.create_dataset(key, data=value, **dataset_kwargs)
+
+
+def write_dataframe(f, key, df, dataset_kwargs={}):
+    group = f.h5f.create_group(key)
```
flying-sheep: Didn’t you retain the wrong branch? This should be `f.create_group(key)`, right?

flying-sheep:
> I would like to remove the `anndata.h5py` file, since I'm pretty sure we can handle sparse matrices without also having our own `File`, `Group`, and `Dataset`. Since we already have to deal with indexing ourselves, I think it'll be easier to share code between backends (like `zarr`) as well. Do have any thoughts on this?

Sounds good. And idk if we can get rid of `anndata.h5py` so easily, don’t we have to fix have code so `__getitem__` on a file or group gives us a `SparseDataset`?

anndata/readwrite/h5ad.py
ivirshup: Sadly, in this case no. For some reason writing the categories doesn't work if I directly pass an `int8` array as data, but does if I allocate then assign values. `anndata.h5py`'s `create_dataset` only supports passing data directly at dataset instantiation. It seemed like a lot more work to add support for instantiating an empty dataset than just accessing the underlying `h5py` group.

ivirshup:
Thesis writing and holiday? Shouldn't those be opposites?

> I’m for Git LFS if we can detect its absence (will there be dummy files?), and we skip those tests when it isn’t installed.

I think these might be important enough tests that they should always run. Why would you want it to be optional? That said, it should be easy to see if the file looks like a stub.

> I’d say we should go native enum: Looks like it’s easy to do with h5py, and I don’t see much reason to stay similar to zarr, as there’s already enough difference.

Haha, I only saw the comments on the code when I was making changes. I ended up representing it the same way for zarr and hdf5. It wasn't actually that easy to do with `h5py`. Statements I thought would be identical caused segfaults plus it's not well documented.

ivirshup:
> Sounds good. And idk if we can get rid of anndata.h5py so easily, don’t we have to fix have code so __getitem__ on a file or group gives us a SparseDataset?

We already handle the file directly in the `X` accessor, so I don't think this will add much complexity. I'm probably not going to make a PR just to remove that module, but I think it'll come about as I try to make our data management more flexible.

anndata/readwrite/h5ad.py
flying-sheep: Oh, OK. maybe add a comment then.

flying-sheep:
> I ended up representing it the same way for zarr and hdf5. It wasn't actually that easy to do with h5py. Statements I thought would be identical caused segfaults plus it's not well documented.

Sounds like a good reason for going native enum after all: Subtle differences are harder to maintain than two separate, straightforward solutions, right?

Your call. Otherwise LGTM!

ivirshup:
> Sounds like a good reason for going native enum after all:

Oh, sorry, I wasn't clear before. I meant to say using `h5py`s enum caused segfaults and is poorly documented.

flying-sheep:
Ah! OK, then let’s get this merged finally :smile:

flying-sheep:
Congrats and thank you for this huge effort!

flying-sheep:
I slimmed down the sub-commit messages in a9b8b035980e505e43d09ba0edb6c47e8c37163d a bit, please tell me if I missed something

ivirshup:
Oh, I had totally missed that you rebased the commits here.

I don't think squashing this much together makes sense for a PR of this size. For instance, I was just trying to blame some lines in h5sparse that I commented ambiguously, but now it's hard to tell when this code was written, or what commit message was intended to go with it.

I'm all for attempting to keep a cleaner git history for the repo. It'd make sense to squash commits which modified the same code. In this case a bunch of commits to different parts of the code were squashed together, which makes it very difficult to tell which part of the commit message relates to the code I blamed. I also think the author should be the one to clean the history, unless there's a very thorough review of all parts of the code. They're in the best place to know what documentation (in the form of commit messages) should be kept, and what code it's related to.

flying-sheep:
You’re right, I should probably have waited for you to rebase this into something clean.

It would have been hard though, a lot of fixes in not necessarily the order where they could be squased into preceding commits, you’d have had to reorder a lot, with potential for merge conflicts during rebase :fearful:

PR #168, opened 2019-06-25
fixed index bug in _remove_unused_categories

milescsmith:
I had an issue with subsetting AnnData objects created in scVelo where `AnnData._remove_unused_categories()` was attempting to, I suppose, incorrectly subset some `.uns` color values.  

In particular, I had a `.obs` key named `cluster`, representing the cluster number from `scanpy.tl.leiden` and a corresponding value in `.uns['cluster_color']`.  When subsetting the object for any cell that was in a cluster between 2 and 9, `_remove_unused_categories()` tried to use that cluster number to subset the one member `.uns['cluster_color']` Series, giving rise to an `index 12 is out of bounds for axis 0 with size 11` error.

I've added a check to `_remove_unused_categories()` so that it now first checks to see if any subsetting is of a `_color` uns member is necessary to prevent such an error.

### DIFF
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -1235,12 +1235,13 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 df_sub[k].cat.remove_unused_categories(inplace=True)
                 # also correct the colors...
                 if k + '_colors' in uns:
-                    # this is a strange hack...
+                    # this is a strange hack...cd wor   
                     if np.array(uns[k + '_colors']).ndim == 0:
                         uns[k + '_colors'] = np.array(uns[k + '_colors'])[None]
-                    uns[k + '_colors'] = np.array(uns[k + '_colors'])[
-                        np.where(np.in1d(
-                            all_categories, df_sub[k].cat.categories))[0]]
+                    elif np.array(uns[k + '_colors']).ndim > 1:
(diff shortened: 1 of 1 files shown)


anndata/core/anndata.py
```diff
@@ -1260,9 +1260,10 @@ def _remove_unused_categories(self, df_full, df_sub, uns):
                     # this is a strange hack...
                     if np.array(uns[k + '_colors']).ndim == 0:
                         uns[k + '_colors'] = np.array(uns[k + '_colors'])[None]
-                    uns[k + '_colors'] = np.array(uns[k + '_colors'])[
-                        np.where(np.in1d(
-                            all_categories, df_sub[k].cat.categories))[0]]
+                    elif np.array(uns[k + '_colors']).ndim > 1:
```
flying-sheep: why not `else`? `ndim` can hardly be `<0`, can it?

milescsmith: Sure?  I no longer remember why I didn't use an `else` here.

flying-sheep: How would an array with -1 dimensions look like?

flying-sheep: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.ndim.html

PR #172, opened 2019-06-30
Warn on flattening X

ivirshup:
In response to https://github.com/theislab/anndata/issues/171#issuecomment-506988177

Throw a warning if X is flattened prior to returning, since that behavior will change in `v0.7`.


falexwolf:
Perfect! :smile:

Issue #174, opened 2019-07-04
What do we export?


ivirshup:
I just found out that [`batchglm` expects to be able to import `anndata.base.Raw`](https://github.com/theislab/batchglm/blob/dc07cf5773c7624a232c2b2487738b3e9fbc937f/batchglm/data.py#L59), which is an issue for upcoming releases on `anndata`.

What should we export from where in `v0.7`? My instinct is that any class a user could come in contact with should be exported at a relatively shallow depth. I'm thinking the main `anndata` namespace could be a good option.

Advantages:

* No reliance on module structure for the rest of the code
* Cleaner import syntax

```python
from Module import Class1, Class2, func1, func2
# vs
from Module import func1, func2
from Module.core.submodule import Class1, Class2
```

Disadvantages:

* Requires modules defining these objects to be imported, possibly leading to slow overall import times
* Potentially cluttered main namespace
* I feel like main-namespace can be a strong API commitment

falexwolf:
I think that any class or function from a private submodule that a user gets in contact with should be exported in the main namespace. I think, `.core` should be private and hence better named `._core`.

We can simply point @davidsebfischer to the fact that `anndata.base.Raw` will be gone from that location. `anndata.base` was never meant as a user module, and it should have been prefixed by an underscore `anndata._base`, but I wasn't that aware of these issues at the time and never renamed it in order to maintain backwards compat. As we're breaking it now, anyway, it would be a good time to also go for `._core`.

What do you think?

ivirshup:
I agree get the reasoning behind naming it `._core` but I'd have to think about why others use `.core` (`xarray`, `pandas`, `dask` (also has `base`), and `zarr`). Possibly due to [PEP8](https://www.python.org/dev/peps/pep-0008/#id40) conventions?

I think it's also fair to say undocumented stuff has no guarantees. `Raw` was documented, so it's reasonable to rely on it.

falexwolf:
I think that recently, quite a few packages went away from a ton of public modules (like `.core`), with no real public use, to having dedicated private modules like `._core`. I think that in Scanpy, it made a lot of sense. Some while ago, @flying-sheep also shared the API design of a package he liked a lot; and they also went down this road. But I don't remember the name.

falexwolf:
So, to answer your question more directly: Maybe they have it for backward compat reasons as well. It just doesn't make sense to have a non-user module in the high-level lookup menus, in my opinion. The lookup should direct the user to the functions and classes he/she needs.
[...]

PR #177, opened 2019-07-05
Export Raw

ivirshup:
#174

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -1,4 +1,4 @@
-from .core.anndata import AnnData
+from .core.anndata import AnnData, Raw
 from .readwrite import (
     read_h5ad, read_loom, read_hdf,
     read_excel, read_umi_tools,


falexwolf:
Fine!

flying-sheep:
It should be linked here, right? https://anndata.readthedocs.io/en/latest/api.html

ivirshup:
I guess... I'm not a fan of `Raw` since it requires implementing a bunch of things twice, and can mostly be replaced by `layers`. I personally don't want to advertise it too much just in case we're able to deprecate it. I do, however, think we should export from a consistent location, mostly so it can be used for type annotations and dispatch.

If other people like it, I'd be happy to hear that point of view. It's just something which I don't use, but have to code around.

falexwolf:
I agree with you, Isaac. :) I'm essentially entirely stopped using it myself in the past year. I would also not advertise too much but rather work towards more consistent solutions.

`.layers` completely works as a replacement if one never actually filters on variables, but replaces every filtering stepped by a purely annotation based solution. It wouldn't be hard at all to make this the default in Scanpy, generally; it's just that I still lack bandwidth...

LuckyMD:
I currently use it to store uncorrected data in the single cell tutorial. Could change that to layers in the next update though. Is there a layers equivalent to `use_raw` for plotting functions?

ivirshup:
@LuckyMD Mostly the `layer` key word argument.

@falexwolf no worries. I think `Raw` would take a while to fully excise since there's so much logic dealing with it. The main argument I can see for `Raw` is that a boolean filter (like we have with highly variable genes) has to be re-applied each time it's used, so there's a performance cost. There are ways around this, but it'd take some consideration to choose one.

Right now, some plotting functions have `use_raw` as `None`, which turns to `True` if raw is defined. This would be a big change in behavior if raw is deprecated, since there's no convention for what "raw" is in layers. I also think the current behavior could be more consistently documented.

falexwolf:
@LuckyMD It's perfectly fine that your tutorial uses it; the Seurat-based tutorial also uses it and people can continue. I just agreed with Isaac that we shouldn't advertise beyond because I also don't like it. Thinking a bit more, though, adding`Raw` to the docs is more a clarification than an advertisement as long as the tutorials legitly keep using it.

@ivirshup All that logic should be maintained and would be replaced with a special slot in `.layers` called 'raw'. Then it really boils down to merely a more elegant implementation. For instance, it would be nice to be able to generate an AnnData from any `layer` (something that people have asked for `Raw`)... But, it's not a priority...

PR #178, opened 2019-07-07
Fix access to raw.X by .raw.{obs,var}_vector

ivirshup:
Previously had relied on subsetting the entire object to get vector of X. Now just normalizes index.
Also stops throwing warning about changing behaviour. Whoops. https://github.com/theislab/anndata/issues/171#issuecomment-508619952


falexwolf:
Looks good! :)

PR #183, opened 2019-07-15
ipython tab completion for layers, obsm, varm

ivirshup:
I was procrastinating and found an easy win.

`layers`, `obsm`, `varm` and soon `obsp`, `varp` will now all support tab completion for their keys through ipython.


ivirshup:
I'm not really sure there's a good way to test this. But I also can't merge things that don't fit the `Required` marks.

flying-sheep:
Eh, I’ll losen this requirement!

PR #189, opened 2019-08-07
Fix r+ mode for old h5ad files

joshua-gould:


flying-sheep:
Any reason you don’t use `if '/' in key`?

joshua-gould:
Nope, that will work too

flying-sheep:
OK! Thank you!

PR #198, opened 2019-08-27
Update docs/require.txt to work with setuptools_scm

ivirshup:
@flying-sheep, readthedocs was complaining about not having setuptools_scm or importlib_metadata installed. `setuptools_scm` recommends against this approach, but I'm not sure how to do this without also having to load `authors` another way.

I'm also going to go apply this to `scanpy`.


flying-sheep:
Thank you!

PR #207, opened 2019-09-06
Add obsp and varp

### LINKED ISSUE #136: Backwards breaking change: Sparse observation and variables annotation

ivirshup:
Fixes #136

This introduces attributes for pairwise measurements on the observations and variables. Currently mostly for cell-cell graphs, but could also be used for gene correlation graphs.

Thoughts @falexwolf @flying-sheep @Koncopd?

### DIFF
--- a/anndata/core/alignedmapping.py
+++ b/anndata/core/alignedmapping.py
@@ -292,30 +292,28 @@ class PairwiseArraysBase(AlignedMapping):
         return self._dimnames[self._axis]
 
 
-
-# TODO: Allow these to be written, and add them back to anndata
-# class PairwiseArrays(AlignedActualMixin, PairwiseArraysBase):
-#     def __init__(self, parent: "AnnData", axis: int, vals: Optional[Mapping] = None):
-#         self._parent = parent
-#         if axis not in (0, 1):
-#             raise ValueError()
-#         self._axis = axis
-#         self.dim_names = (parent.obs_names, parent.var_names)[self._axis]
-#         self._data = dict()
-#         if vals is not None:
-#             self.update(vals)
+class PairwiseArrays(AlignedActualMixin, PairwiseArraysBase):
+    def __init__(self, parent: "AnnData", axis: int, vals: Optional[Mapping] = None):
+        self._parent = parent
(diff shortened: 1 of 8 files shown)


flying-sheep:
Great! We will need code in scanpy that automatically checks for square sparse matrices in `.uns` which will be superfluous after this.

anndata/core/anndata.py
```diff
@@ -655,12 +655,12 @@ def _init_as_view(self, adata_ref: 'AnnData', oidx: Index, vidx: Index):
...
-        # varp=None, obsp=None,
-        raw=None, layers=None,
-        dtype='float32', shape=None,
-        filename=None, filemode=None,
+            self, X=None, obs=None, var=None, uns=None,
+            obsm=None, varm=None,
+            varp=None, obsp=None,
+            raw=None, layers=None,
+            dtype='float32', shape=None,
+            filename=None, filemode=None
```
flying-sheep: Please rebase on master so my style edits don’t get reverted

anndata/core/alignedmapping.py
```diff
@@ -292,30 +292,29 @@ def dim(self):
         return self._dimnames[self._axis]
 
 
-
 # TODO: Allow these to be written, and add them back to anndata
```
flying-sheep: The TODO won’t be necessary right?

ivirshup:
Which renaming?

flying-sheep:
How the abstract classes and the concrete/view classes are named, as I said in the other PR. This PR introduces more of those which is why I mentioned it.

ivirshup:
Gotcha. I agree some renaming and decisions on exports need to be made. I'm still considering my thoughts on this, but I'll add those to the renaming issue soon.

Issue #201, opened 2019-09-02
adata.copy() doesn't deepcopy layers


gokceneraslan:
After copying an anndata via `adata.copy()`, layers of the resulting object are still pointing to the layers of the original object:

```python
import scanpy as sc

adata = sc.datasets.paul15()
adata.layers['L'] = adata.X.copy()
adata.layers['L'].sum()
```

Output: 6083492.0

```python
bdata = adata.copy()
adata.layers['L'] *= 2.0
bdata.layers['L'].sum()
```

Output: 12166984.0

I realized that when I saw scvelo normalizing the layers of the original object even though I passed a copy to scv.pp.filter_and_normalize...

ivirshup:
What version are you on? I'm not able to reproduce on `master`.

VolkerBergen:
Hi Gokcen, I just run this minimal example and couldn't reproduce that either.

gokceneraslan:
Can anybody try with the stable anndata package?

![image](https://user-images.githubusercontent.com/1140359/64431019-8d390f00-d087-11e9-8874-9655b4e79fa9.png)

VolkerBergen:
With your version ('0.6.22.post1') I get the same error, with the latest ('0.6.22.post1+29.g16615d7') it's fixed.

gokceneraslan:
ok, thanks. Adding a test would be useful then.

PR #208, opened 2019-09-06
Add test for layer deepcopy

gokceneraslan:
See #201.


ivirshup:
Definitely good to have this test. Any chance you could add it for the other attributes as well?

PR #211, opened 2019-09-10
Fix docs for layers.

ivirshup:
Remove out of date warning in layers doc.

I also changed the wording in the reference to loom since I think it's more straightforward to just have the behavior described as Dictionary-like.


flying-sheep:
Thank you!

PR #214, opened 2019-09-10
Simplify h5sparse

flying-sheep:


gokceneraslan:
Any ideas why this bot is spamming us that much? 😀

flying-sheep:
damn, no idea. Actually we can probably get rid of it, because codacy AFAIK also supports coverage. If we get a badge that should work.

PR #216, opened 2019-09-11
Fix builds and export warning type

ivirshup:
Fixes build bug introduced by 2c970263fc6daefddc3f7a34bb4e9bf60dcabeab.


ivirshup:
Looks like @flying-sheep caught it.

In general, I like to make most changes through PRs so I can see if the master build breaks. Plus it's generally easier to see changes made via PRs vs direct commits since we're not too strict about clean commit histories.

flying-sheep:
You’re right, there’s always something slipping through :sweat_smile:

Issue #219, opened 2019-09-12
allow backed mode for zarr format


joshua-gould:

flying-sheep:
We do:

https://github.com/theislab/anndata/blob/92b6791c05b1e2f54dac5d6090ecc991ff6a50b4/anndata/core/anndata.py#L48-L58

joshua-gould:
There is an implementation for zarr, but backed_mode is not supported.
Since the APIs for h5py and zarr are almost identical, I think it makes
sense to abstract much of the h5 functionality into base classes in which a
few methods are implemented by h5 subclasses and zarr subclasses and can
easily be extended to future implementations that implement APIs similar to
h5py and zarr.

ivirshup:
+1, that's pretty much my current plan.

joshua-gould:
Any update of support for backed mode using zarr? Thanks.

tmchartrand:
Is this still an active direction?

LucaMarconato:
@ilan-gold I saw some recent PRs involving backed Zarr objects, is there any updated on this? Thanks 😊

ivirshup:
Support for backed sparse zarr arrays is in anndata 0.10. You can use them like:

```python
import scanpy as sc, anndata as ad
import zarr

adata = sc.datasets.pbmc3k_processed().raw.to_adata()
adata.write_zarr("pbmc3k.zarr")

g = zarr.open("pbmc3k.zarr")

def read_backed(group):
    return ad.AnnData(
        ad.experimental.sparse_dataset(group["X"]),
        **{
            k: ad.experimental.read_elem(group[k]) if k in group else {}
            for k in ["layers", "obs", "var", "obsm", "varm", "uns", "obsp", "varp"]
        }
    )


adata = read_backed(g)
adata.X
```

You can see more example usage in this gist: https://gist.github.com/ivirshup/c29c9fb0b5b21a9c290cf621e4e68b18

It's not quite the same as "backed mode", but is actually better since this backed sparse array can be any matrix in the anndata object. Not just `X`.

ivirshup:
There are some performance issues at the moment, see: 

* https://github.com/scverse/anndata/issues/1224

LucaMarconato:
Fantastic! Thanks for the update! Currently we have some performance bottlenecks in `spatialdata` when large tables are fully loaded into memory. We can experiment with this, it should work 😊

PR #222, opened 2019-09-19
Unifiy zarr and h5ad backends

joshua-gould:
This commit unifies the zarr and h5ad backends for reading. It supports reading zarr files in backed mode by doing a few instanceof checks that need to be fixed.

Remaining to dos:
Unify backends for writing
Update AnnDataFileManager to work with zarr files and other future formats that support the mapping interface.

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -1,18 +1,16 @@
 from .core.anndata import AnnData, Raw
+from .h5py import read_h5ad
+from .h5py import read_h5ad as read  # backwards compat / shortcut for default format
 from .readwrite import (
-    read_h5ad,
     read_loom,
     read_hdf,
     read_excel,
     read_umi_tools,
     read_csv,
     read_text,
-    read_mtx,
-    read_zarr,
+    read_mtx
 )
-from .readwrite import (
-    read_h5ad as read,
-)  # backwards compat / shortcut for default format
+from .zarr import read_zarr
 
 __doc__ = """\
 API
@@ -89,9 +87,9 @@ __author__ = ', '.join(
 )
 __email__ = ', '.join(
     [
-        'philipp.angerer@helmholtz-muenchen.de',
-        'f.alex.wolf@gmx.de',
-        # We don’t need all, the main authors are sufficient.
+            'philipp.angerer@helmholtz-muenchen.de',
+            'f.alex.wolf@gmx.de',
+            # We don’t need all, the main authors are sufficient.
     ]
 )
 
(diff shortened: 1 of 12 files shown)


flying-sheep:
Nice approach!

Stylistically, please use black to format the code, use relative imports instead of `import anndata.thing` and delete code instead of commenting it out.

Why did you choose name `mapping` for the HDF5 module?

joshua-gould:
Thanks. I will reformat and make the changes you requested.
The base classes are in the mapping package because h5py and zarr
groups both implement most of the Python dictionary convention.

flying-sheep:
Hmm, I think it should be named in a way that reflects that this is about disk-backed persistence using hdf5.

joshua-gould:
Is there a use case for changing to/from backed mode after instantiating an AnnData object? It might make the code simpler to remove this ability.

flying-sheep:
How do we shift back and forth?

joshua-gould:
By setting AnnData.filename to a string (on) or None (off)

flying-sheep:
Ah! Well, I’ll let @ivirshup take over, he’s been working on this

ivirshup:
Thanks for making this PR! It's nice to see people contributing the features they want to see. However, my inclination would be to not accept it, since it replicates existing work.

As previous mentioned to you in #219, improving backed mode is something I'm already working on. As you've noted, the zarr and h5ad backends are almost identical. The main reason I've been doing that, along with other efforts to make storage consistent (like the contents of `alignedmappings.py`), was to make a more useful/ flexible backing ability.

Like you've noticed with `adata.filename = path/to/file.h5ad`, there's some limiting behaviour for backed mode at the moment. Another issue to me is that we support sparse and dense matrices all over the place, but in backed mode only `.X` is kept on disk. I think these behaviors should change, but I think the changes should be bundled with a larger overhaul, with many tests, and plenty of warning.

My current plans involve setting an abstraction layer between the `AnnData` object and the actual data, so that it's easier to improve and experiment with data management. I think this may be similar to what you were thinking with the updates to the `AnnDataFileManager`. The work is still in draft stages, but accomplishes much of what is covered here.

If you're wondering why you haven't seen this yet, it's because I would like to get 0.7 out and release before adding more new features. There's already a fair amount of the current backed mode that doesn't work as expected (modification largely, access time frequently), and I think it's going to require some more time to work out the kinks. However, there are already large performance and consistency improvements on master which are worth getting out to users.

Again, thanks for putting in the work on this PR. Sorry if this is disappointing. If you'd like I could let you know when I think we could use more contributions in this area. Seeing as it looks like your TODOs and implemented code are quite similar to what I've been doing, it could also useful to get your input as these features are added – if you'd be interested in that.

joshua-gould:
Makes sense. I’d be happy to contribute to your refactoring efforts. Thanks.

PR #227, opened 2019-09-21
[WIP] Faster sparse backed access

ivirshup:
* Very preliminary benchmarks pretty good speed up (~1500X for a large dataset)
* Please forgive the stylistic sins
* This could be expanded, currently just addresses simple access along compressed axes

Edit: This has expanded a bit to include the last big item on my checklist for 0.7: re-enabling writing sparse arrays to dense hdf5 datasets. I was touching a lot of the same code on the two branches and figured I'd save myself the trouble and just combine them.


anndata/h5py/__init__.py
```diff
@@ -21,9 +21,4 @@
    Dataset
    SparseDataset
 """
-from .h5sparse import File, Group, SparseDataset, _load_h5_dataset_as_sparse
-from h5py import Dataset, special_dtype
-
-# Problem: the H5py intersphinx is broken, and only contains e.g. `Dataset` directly.
-# So we can’t possibly link to it using :class:`Dataset`, since that will always find our version.
-Dataset.__doc__ = """Equivalent to :class:`h5py.Dataset <h5py:Dataset>`."""
+from .h5sparse import SparseDataset, _load_h5_dataset_as_sparse
```
flying-sheep: The docs still mention this exporting Group, Dataset, … which you should fix.

anndata/tests/test_h5py.py
```diff
@@ -4,48 +4,48 @@
...
+#         [[0, 1, 0], [0, 0, 1], [0, 0, 0], [1, 1, 0]], dtype=np.float64
+#     )
+#     with h5py.File(from_h5_path) as from_h5f:
+#         from_dset = from_h5f.create_dataset('sparse/matrix', data=sparse_matrix)
+
+#         with h5py.File(to_h5_path) as to_h5f:
+#             to_h5f.create_dataset('sparse/matrix', data=from_dset)
+#             assert (to_h5f['sparse/matrix'].value != sparse_matrix).size == 0
+
+# TODO: Add this one back
```
flying-sheep: Do it.

flying-sheep:
Great! Deduplication is always good.

I think we should have made clear that the `h5py` module is internal.

We should probably soon create a “For AnnData developers” section where the docs for this and potentially other internal stuff is moved.

anndata/tests/test_h5py.py
ivirshup: I may rewrite the append function so it generalizes to csc first.

ivirshup:
> I think we should have made clear that the h5py module is internal.

Yeah, that would be good. It was fairly public and is even documented. I'm considering even adding back the module with a lot of deprecation and `SparseDataset` stuff moved out – then re-imported.

> We should probably soon create a “For AnnData developers” section where the docs for this and potentially other internal stuff is moved.

👍

anndata/h5py/__init__.py
ivirshup: This file has moved to `anndata/core/sparsedataset/__init__.py` and those references are removed. The docs there could be expanded though.

anndata/core/sparsedataset/__init__.py
```diff
@@ -60,21 +60,21 @@ class backed_csc_matrix(BackedSparseMatrixMixin, ss.csc_matrix):
 
 
 def get_format_str(data):
-    for fmt, backed_class, memory_class in FORMATS:
+    for fmt, _, memory_class in FORMATS:  # To appease codacy
```
flying-sheep: This is just regular pythonic code, nothing about codacy :wink:

anndata/core/anndata.py
```diff
@@ -39,7 +40,8 @@
     asview,
     _resolve_idxs,
 )
-from .. import h5py, utils
+from .sparsedataset import SparseDataset
```
flying-sheep: `sparse_dataset`, only stdlib names are `nounderscore`

ivirshup:
# State of the PR

## Implemented

* Much faster access of on disk sparse data along their compressed axes (mentioned above)
* `SparseDataset.append` now works for CSC
* Completely removed `anndata.h5py`
  * This has meant re-doing the spec for hdf5 sparse datasets, they now use the same convention as zarr.
* Reimplemented reading dense datasets as sparse and writing sparse datasets as dense. This is done in chunks along the compressed axis (not necessarily the case before) which should speed things up. Initial benchmarks show much better memory usage (peak memory of ~600mb as opposed to 1.4 gb for entire process when writing 43k cell dataset from in-memory csr sparse to dense, though equivalent times).

## Remaining questions

### `SparseDataset.append` may be too permissive

Should we allow appending sparse matrices when they don't match in size on the non-extending axis? I.e., does it make sense to allow appending two CSR matrices of sizes `(50, 100)` and (50, 200)` resulting in a `(100, 200)` matrix? Neither `scipy.sparse` or `pydata/sparse` allow this, but `SparseDataset` does. I don't think it should.

### API for reading dense as sparse, writing sparse as dense

I don't like the current API for `force_dense`. I think the user should have some control over which elements are written as dense/ read as sparse. This should also not be set on the file, but set by the arguments used to read it. I'm thinking this could be accomplished by adding `as_sparse` as an argument to `read_h5ad` and `as_dense` to `write_h5ad`. These would both be of type `List[Union[Tuple[str, str], str]]`.

Usage would look like:

```python
adata.write_h5ad("path.h5ad", as_dense=["X", ("layers", "counts")])

backed = sc.read_h5ad("path.h5ad", backed="r+", as_sparse=[("layers", "counts")])
```

## Style

I can't bring myself to turn a readable line of length 82 into multiple lines. 90 characters could maybe be an error,  but shorter lines should be a warning at most. [There are plenty of ways to set this up](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length).

anndata/core/anndata.py
ivirshup: > only stdlib names are nounderscore

I'm not sure that's a common convention. [PEP-8 recommends against underscores in module names](https://www.python.org/dev/peps/pep-0008/#id40).

> Or maybe hdf5_sparse

I'm pretty sure you can currently just pass a zarr group to SparseDataset and it'd work. This works:

```python
import anndata as ad
import scanpy as sc
import zarr

pbmc = sc.datasets.pbmc3k()
pbmc.write_zarr("pbmc.zarr")
zx = ad.core.sparsedataset.SparseDataset(zarr.open("pbmc.zarr/X"))
assert not np.any((zx[50:100, :] != pbmc[50:100, :].X).toarray())
```

> no reason the module has to be called the same as the class

I am wondering why we have separate classes for `SparseDataset` and `backed_csc_matrix`/ `backed_csr_matrix`. We could just have `CSRDataset` and `CSCDataset`.

anndata/core/sparsedataset/__init__.py
ivirshup: I just don't think it's worth a  ❌.

On a related note, what's up with test coverage?

anndata/core/anndata.py
flying-sheep: > PEP-8 recommends against underscores in module names.

Yes, and PEP 8 is a style guide … for the stdlib, not 3rd party modules. Common misconception.

> I'm pretty sure you can currently just pass a zarr group to SparseDataset and it'd work.

Right, zarr isn’t HDF5-based. Is there a nice concise word for what they both are? They’re both supposed to be an efficient memory layout for binary data, with focus on performance and compressability, right? But that’d make a shitty module name :wink:

ivirshup: > Yes, and PEP 8 is a style guide … for the stdlib, not 3rd party modules.

I mean, it's a style guide which the standard library uses, but so do many packages. Just going off convention from here:

* `readwrite` not `read_write`
* `h5sparse` not `h5_sparse`
* `scanpydoc` not `scanpy_doc`

I also prefer `from sparsedataset import SparseDataset` to `from sparse_dataset import SparseDataset`.

Ultimately, it's an internal module so I guess it doesn't matter much.

> Is there a nice concise word for what they both are? 

Hierarchical array store?

> Or maybe (don’t know how feasible), remove mentions of rows and columns there and make other code use SparseDataset in terms of major_axis and minor_axis.

I think I might just wait and watch for what pydata/sparse does here. I do like having separate classes for dispatch – since we want to do different things depending on layout. If we could dispatch on generic types, that might be different.

Also I'd want a usability/ performance reason to do any major overhauls of this code.

flying-sheep: Let’s call it it “storage” then!

flying-sheep:
> Should we allow appending sparse matrices when they don't match in size on the non-extending axis?

We shouldn’t, you’re right. I don’t like broadcasting, it’s too surprising if you didn’t intend to to it.

> I think the user should have some control over which elements are written as dense/ read as sparse. This should also not be set on the file, but set by the arguments used to read it.

Agreed that this usage makes more sense but why is any of this needed at all for h5ad? I can see it making sense for reading formats that don’t support sparse matrices, but for h6ad:

Reading dense as sparse means someone didn’t use the right format for storing their matrix in the AnnData object before writing. Writing sparse as dense seems like it could be useful for compatibility with readers from other programming languages. But in that case, writing a reader that supports our sparse representation seems like the better idea.

> These would both be of type `List[Union[Tuple[str, str], str]]`

Never `List`, in a parameter type, but I’m also not a fan of the whole signature. This is basically paths right? I think the signature should be `Mapping[str, Sequence[Hashable]]`, with `{'layers': None | 'X'}` referring to `adata.X`

This being said, we should ban calling a layer `'X'` or `''` or `None` because I want to unify layers and `X`.

ivirshup:
> We shouldn’t, you’re right. I don’t like broadcasting, it’s too surprising if you didn’t intend to to it.

Agreed this is confusing. But I'm not sure this is broadcasting, since reasonable languages (i.e. not `R`) throw an error if you try to broadcast with mismatched dimensions.

> Agreed that this usage makes more sense but why is any of this needed at all for h5ad?

To me, this fits the same reason you'd have a backed mode. 

**Reading dense as sparse:** If you have a dataset (probably generated somewhere else) stored in dense format, I don't think you should have to have enough memory to store the dense matrix to read it in as sparse.

**Writing sparse as dense:** This was for the case where you might not even have the resources to read in the sparse dataset, but would like to use backed mode (sparse backed mode is lacking many features at the moment). This allows you to convert a sparse on-disk matrix to a dense one without needing to be able to have either completely in memory.

> I’m also not a fan of the whole signature.

I'm flexible on the `List` typing, I think I've just been running into scanpy functions which mainly allow `list`s (to the exclusion of `np.ndarray`s or `pd.Series`).

This branches from a more general thoughts I've been having for having paths within anndata, and having consistent nested element access. While mapping works in this case (since order doesn't matter), there are cases like this (where you want to access multiple nested elements) where order does matter. You can't really represent [("layers", "a"), ("obsm", "b"), ("layers", "c")] with a dict. This also matches arguments like `obsm_keys` of `sc.get.obs_df`.

I'd like to codify this a bit more, so cases like `[("layers", ["a", "b"])]` would work. I've played around with the idea of having something like `[anndata.Acc("layers", "scvelo*"), ("obsm", "clusters")]` work too.

flying-sheep:
OK, in the light of backed mode, dense-as-sparse and sparse-as-dense makes sense.

>  think I've just been running into scanpy functions which mainly allow lists (to the exclusion of np.ndarrays or pd.Series)

I got rid of `isinstance(thing, list)` everywhere in favor of `isinstance(thing, collection.abc.Sequence)`, but didn’t realize that’s only slightly better as `ndarray` isn’t a `Sequence`. I think Python made a mistake with half of its collection base classes being sensible in their requirements, and half of them requiring you to implement the kitchen sink. We should fix that.

Regarding your thoughts about paths: Sounds sensible, but where *would* we need a “sequence of paths to arbitrary anndata members”?

ivirshup:
> Regarding your thoughts about paths: Sounds sensible, but where would we need a “Ordered sequence of paths to arbitrary anndata members”?

`sc.get.obs_df` and `sc.pl` (for any plotting method with subplots) come to mind for me. This also came up with the ViewArgs type, where it was useful to have a data structure for multiple levels of indexing. The Accessor class I was thinking of would be pretty similar to that class.

On sparse-as-dense and dense-as-sparse. I think it'll actually take a bit of doing to get these working right for all elements. I think this should happen post 0.7 since I expect adding some more structure to reading and writing will make that easier.  Also merging this should otherwise get us to feature completion for the release.

In the meantime, I'm thinking I'll only allow `"X"` and `"raw/X"` / `("raw", "X")` as values in `as_sparse` or `as_dense`, with other values throwing a `NotImplementedError`.

ivirshup:
I've been working on trying to get access along the minor access sped up a bit (or at least not requiring the entire matrix to be loaded into memory), but this takes some work. I'll layout where I'm at, though I think this PR could be done and the next release could occur before I implement these improvements. 

I've managed translate/ modify the scipy.sparse code for access along the uncompressed dimension to numba, and it's going pretty fast. I've run into some issues when I try to integrate this with h5py.

First, there's how the `numba.objmode` context manager currently works in numba. Right now there isn't really a way to pass a reference to a python object into a numba function. This means my current code has to either have a global reference to the h5py dataset, or be recompiled on every call. There's a solution to this here: https://github.com/numba/numba/pull/3640.

Second is an issue with h5py. It turns out there is quite a bit of overhead to indexing into a h5py dataset. So much so that it eclipses reading a chunk from disk https://github.com/h5py/h5py/issues/1391. Since a single default chunk can hold multiple sets of indices, this can take a while. I'm planning on working around this by keeping our own LRU cache for chunks in python. I'm aiming to implement this caching layer in numba to reduce overhead. This will be made easier with some improvements I'm trying to make to slice object behaviour in numba.

anndata/core/anndata.py
ivirshup: I'd rather keep the name specific to the data structure until it's diverged enough we can just get rid of the old licensing.

anndata/readwrite/utils.py
```diff
@@ -1,10 +1,38 @@
...
-from .. import h5py as patched_h5py
+from functools import wraps, singledispatch
+from ..core.sparsedataset import SparseDataset
 
 # -------------------------------------------------------------------------------
 # Type conversion
 # -------------------------------------------------------------------------------
 
+# Could be numba'd if it returned tuples instead of slices
+def idx_chunks_along_axis(shape: tuple, axis: int, chunk_size: int):
```
flying-sheep: Where is this used?

ivirshup: It's used in all the dense <-> sparse io functions. Github hides those diffs by default due to length (probably since a lot of signatures got modified).

flying-sheep: Ah, I missed that hidden diff as it came after the diffs hidden for deleted files.

ivirshup:
I could remove the latest two fix commits, but I think trying to rebase on the merge commits would do more harm than good. Maybe if I had `git rerere` enabled? The only reason I merged from master instead of rebasing in the first place was due to how complicated the rebase got.

flying-sheep:
Well, you could try enabling it and then rebasing …

ivirshup:
My impression was `git rerere` records how you resolved merge conflicts when you resolved them, instead of figuring it out posthoc. Right now I'm wondering why this isn't on by default.

flying-sheep:
No idea, I just enabled it for me …

ivirshup:
I think this is good to merge. I think there's a problem with the black setup, since there's no way  this is a reasonable edit:

```python
-    adata1 = adata[:2,]
+    adata1 = adata[
+        :2,
+    ]
```

I'm also pretty sure codacy is wrong about that analysis error, since it's typically poor form to modify a list while iterating over it's elements. It also didn't care about that line on the last commit?

flying-sheep:
Yeah, but the commit history is really ugly. I’ll rebase, you check if I did it right, OK?

PR #228, opened 2019-09-23
Improve alignedmapping typing

flying-sheep:
See 62c488658fe7f3a3898d35c576ce308e69e78de8

Let’s continue discussion here. I seem to misunderstand a few things about the new `AlignedMapping`s:

- I thought `obsm` and `varm` (`AxisArrays`) are the only ones where heterogenous `pd.DataFrame`s can be put in?
- I thought `layers` (`Layers`) is the onle one that is aligned to both axes instead of just one?
- I thought all of those are indexed only with `str`s? I mean technically there’s no need to, any hashable should work…

### DIFF
--- a/anndata/core/alignedmapping.py
+++ b/anndata/core/alignedmapping.py
@@ -5,6 +5,7 @@ from typing import (
     Mapping,
     Optional,
     Tuple,
+    Type,
     Iterable,
     Sequence,
     Union,
@@ -22,11 +23,12 @@ from ..utils import deprecated
 from .views import asview, ViewArgs
 
 
-OneDIdx = Sequence[Union[int, bool]]
+OneDIdx = Union[Sequence[int], Sequence[bool]]
 TwoDIdx = Tuple[OneDIdx, OneDIdx]
 
-I = TypeVar("I", OneDIdx, TwoDIdx)
-V = TypeVar("V", pd.DataFrame, spmatrix, np.ndarray)
+I = TypeVar("I", OneDIdx, TwoDIdx, covariant=True)
+# TODO: pd.DataFrame only allowed in AxisArrays?
+V = Union(pd.DataFrame, spmatrix, np.ndarray)
 
 
 @singledispatch
@@ -44,19 +46,11 @@ def _subset_df(df: pd.DataFrame, subset_idx: OneDIdx):
 
 
 class AlignedMapping(cabc.MutableMapping, ABC):
-    """
-    Attributes
-    ----------
-    axes
-        Which axes is this aligned along?
-    _view_class
-        The view class for this aligned mapping.
-    _actual_class
-        The actual class (which has it's own data) for this aligned mapping.
-    """
+    _view_class: ClassVar[Type["AlignedViewMixin"]]
+    """The view class for this aligned mapping."""
 
-    _view_class: ClassVar["AlignedViewMixin"]
-    _actual_class: ClassVar["AlignedActualMixin"]
+    _actual_class: ClassVar[Type["AlignedActualMixin"]]
+    """The actual class (which has it's own data) for this aligned mapping."""
 
     def __repr__(self):
         return f"{self.__class__.__name__,} with keys: {self.keys()}"
--- a/anndata/core/alignedmapping.py
+++ b/anndata/core/alignedmapping.py
@@ -1,4 +1,3 @@
-from warnings import warn
 from abc import ABC, abstractmethod
 from collections import abc as cabc
 from functools import singledispatch
--- a/anndata/core/alignedmapping.py
+++ b/anndata/core/alignedmapping.py
@@ -1,19 +1,11 @@
+from warnings import warn
 from abc import ABC, abstractmethod
-import collections.abc as cabc
+from collections import abc as cabc
 from functools import singledispatch
-from typing import (
-    Mapping,
-    Optional,
-    Tuple,
-    Type,
-    Iterable,
-    Sequence,
-    Union,
-    List,
-    TypeVar,
-    Iterator,
-    ClassVar,
-)
+from typing import TYPE_CHECKING
+from typing import Union, Optional, Type, ClassVar, TypeVar  # Special types
+from typing import Hashable, Iterable, Iterator, Mapping, Sequence  # ABCs
+from typing import Tuple, List, Dict  # Generic base types
 
 import numpy as np
 import pandas as pd
@@ -21,6 +13,9 @@ from scipy.sparse import spmatrix
 
 from ..utils import deprecated
 from .views import asview, ViewArgs
+
+if TYPE_CHECKING:
+    from .anndata import AnnData
 
 
 OneDIdx = Union[Sequence[int], Sequence[bool], slice]
@@ -46,6 +41,14 @@ def _subset_df(df: pd.DataFrame, subset_idx: OneDIdx):
 
 
 class AlignedMapping(cabc.MutableMapping, ABC):
+    """\
+    An abstract base class for Mappings containing array-like values aligned
+    to either one or both AnnData axes.
+    """
+
+    _allow_df: bool
+    """If this mapping supports heterogeneous DataFrames"""
+
     _view_class: ClassVar[Type["AlignedViewMixin"]]
     """The view class for this aligned mapping."""
 
@@ -53,12 +56,12 @@ class AlignedMapping(cabc.MutableMapping, ABC):
     """The actual class (which has it's own data) for this aligned mapping."""
 
     def __repr__(self):
-        return f"{self.__class__.__name__,} with keys: {self.keys()}"
+        return f"{type(self).__name__} with keys: {self.keys()}"
 
-    def _ipython_key_completions_(self) -> List[str]:
+    def _ipython_key_completions_(self) -> List[Hashable]:
         return list(self.keys())
 
-    def _validate_value(self, val: V, key: str) -> None:
+    def _validate_value(self, val: V, key: Hashable) -> V:
         """Raises an error if value is invalid"""
         for i, axis in enumerate(self.axes):
             if self.parent.shape[axis] != val.shape[i]:
@@ -75,6 +78,15 @@ class AlignedMapping(cabc.MutableMapping, ABC):
                 raise IndexError()  # Maybe not index error
         except AttributeError:
             pass
+        if not self._allow_df and isinstance(val, pd.DataFrame):
+            was_heterogeneous = val.dtypes.nunique() != 1
+            val = val.to_numpy()
+            if was_heterogeneous:
+                warn(
+                    f'Layer {key!r} converted to numpy array '
+                    f'with dtype {val.dtype}'
+                )
+        return val
 
     @property
     @abstractmethod
@@ -113,43 +125,41 @@ class AlignedMapping(cabc.MutableMapping, ABC):
 
 
 class AlignedViewMixin:
-    """
-    Attributes
-    ----------
-    parent
-        Reference to parent AnnData view
-    attrname
-        What attribute in the parent is this?
-    parent_mapping
-        The object this is a view of.
-    """
+    parent: "AnnData"
+    """Reference to parent AnnData view"""
+
+    attrname: str
+    """What attribute in the parent is this?"""
+
+    parent_mapping: Mapping[Hashable, V]
+    """The object this is a view of."""
 
     isview = True
 
-    def __getitem__(self, key: str) -> V:
+    def __getitem__(self, key: Hashable) -> V:
         return asview(
             _subset(self.parent_mapping[key], self.subset_idx),
             ViewArgs(self.parent, self.attrname, (key,)),
         )
 
-    def __setitem__(self, key: str, value: V):
-        self._validate_value(value, key)  # Validate before mutating
+    def __setitem__(self, key: Hashable, value: V):
+        value = self._validate_value(value, key)  # Validate before mutating
         adata = self.parent.copy()
         new_mapping = getattr(adata, self.attrname)
         new_mapping[key] = value
         self.parent._init_as_actual(adata)
 
-    def __delitem__(self, key: str):
+    def __delitem__(self, key: Hashable):
         self[key]  # Make sure it exists before bothering with a copy
         adata = self.parent.copy()
         new_mapping = getattr(adata, self.attrname)
         del new_mapping[key]
         self.parent._init_as_actual(adata)
 
-    def __contains__(self, key: str) -> bool:
+    def __contains__(self, key: Hashable) -> bool:
         return key in self.parent_mapping
 
-    def __iter__(self) -> Iterator[str]:
+    def __iter__(self) -> Iterator[Hashable]:
         return iter(self.parent_mapping)
 
     def __len__(self) -> int:
@@ -157,28 +167,25 @@ class AlignedViewMixin:
 
 
 class AlignedActualMixin:
-    """
-    Attributes
-    _data: dict
-        Underlying mapping to the data
-    """
+    _data: Dict[Hashable, V]
+    """Underlying mapping to the data"""
 
     isview = False
 
-    def __getitem__(self, key: str) -> V:
+    def __getitem__(self, key: Hashable) -> V:
         return self._data[key]
 
-    def __setitem__(self, key: str, value: V):
-        self._validate_value(value, key)
+    def __setitem__(self, key: Hashable, value: V):
+        value = self._validate_value(value, key)
         self._data[key] = value
 
-    def __contains__(self, key: str) -> bool:
+    def __contains__(self, key: Hashable) -> bool:
         return key in self._data
 
-    def __delitem__(self, key: str):
+    def __delitem__(self, key: Hashable):
         del self._data[key]
 
-    def __iter__(self) -> Iterator[str]:
+    def __iter__(self) -> Iterator[Hashable]:
         return iter(self._data)
 
     def __len__(self) -> int:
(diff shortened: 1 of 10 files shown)


ivirshup:
> I thought obsm and varm (AxisArrays) are the only ones where heterogenous pd.DataFrames can be put in?

All of the aligned mappings can hold dataframes. We test this anywhere that uses the `gen_adata` helper (which is a lot of the view and io tests)

https://github.com/theislab/anndata/blob/914d7375a93ca617bd6dee1565097656e6e1d359/anndata/tests/helpers.py#L130-L134

As well as for `obsp` and `varp`:

https://github.com/theislab/anndata/blob/914d7375a93ca617bd6dee1565097656e6e1d359/anndata/tests/test_obspvarp.py#L25-L45

> I thought layers (Layers) is the onle one that is aligned to both axes instead of just one?

Does the fact that all of the aligned mappings can hold dataframes resolve this? I'm not sure if this is about something else.

> I thought all of those are indexed only with strs? I mean technically there’s no need to, any hashable should work…

They should only be indexed with `str` if the user wants to write them to hdf5 or zarr, but I don't think that's enforced by these types. In general, I lean towards keeping the typing as broad as possible.

anndata/core/alignedmapping.py
```diff
@@ -22,11 +23,12 @@
 from .views import asview, ViewArgs
 
 
-OneDIdx = Sequence[Union[int, bool]]
+OneDIdx = Union[Sequence[int], Sequence[bool]]
```
ivirshup: Slices too!

```suggestion
OneDIdx = Union[Sequence[int], Sequence[bool], slice]
```

anndata/core/alignedmapping.py
```diff
@@ -22,11 +23,12 @@
...
 TwoDIdx = Tuple[OneDIdx, OneDIdx]
 
-I = TypeVar("I", OneDIdx, TwoDIdx)
-V = TypeVar("V", pd.DataFrame, spmatrix, np.ndarray)
+I = TypeVar("I", OneDIdx, TwoDIdx, covariant=True)
+# TODO: pd.DataFrame only allowed in AxisArrays?
+V = Union(pd.DataFrame, spmatrix, np.ndarray)
```
ivirshup:
```suggestion
V = Union[pd.DataFrame, spmatrix, np.ndarray]
```

flying-sheep:
I think it’s weird to allow dataframes with heterogeneous columns in a pairwise matrix (those should hold symmetric or asymmetric transitions right? Homogeneous data), and slightly less weird when aligned to genes (Why should gene A have a `str` column and gene B a float column?)

About layers being the only one aligned to both axes: I meant that it is the only `AlignedMapping` whose members gets sliced when either AnnData axis is sliced, i.e. the only one needing to remember a 2D index.

ivirshup:
> I think it’s weird to allow dataframes with heterogeneous columns in a pairwise matrix

The main reason I allowed this is to test that we can handle arrays with labels (in this case, `index` and `column`) in these fields. I would like to also include sparse dataframes here, as well as xarray data arrays. Plus dataframes don't have to have heterogenous element types.

> About layers being the only one aligned to both axes

`obsp` and `varp` will also get indexed on two axes.

ivirshup:
Oh, but the only place I though the usage of `OneDIdx` should change was `_subset_df`

flying-sheep:
> Plus dataframes don't have to have heterogenous element types.

They’re designed to have a heterogeneous axis and a homogeneous one. AnnData is designed to have homogeneous 2D layers and 2× heterogeneous 2D annotations per axis.

I don’t like `varp`/`obsp` and `layers` to be able to contain something heterogeneous. Users will be confused and it doesn’t fit AnnData’s design. We could runtime check if a DataFrame is homogeneous, in `Layer.__setitem__`, but it’s probably easier to not support DataFrames there and test with another class.

> Oh, but the only place I though the usage of `OneDIdx` should change was `_subset_df`

See above: With `DataFrame` layers that would make sense, but I don’t think `DataFrame` layers should exist.

> `obsp` and `varp` will also get indexed on two axes.

Yes, _indexed_ with `(subset_idx, subset_idx)`. Both axes of an `obsp` matrix are _aligned_ to one `AnnData` axis.

ivirshup:
> the only place I though the usage of OneDIdx should change was _subset_df

To me,  the `_subset` is meant to be an interface for array-like objects that subsets them consistently with respect to the index provided. I think it should be considered separately from the `alignedmapping` module, and probably moved to a `utils` file or maybe `views`.

> We could runtime check if a DataFrame is homogeneous, in Layer.__setitem__, but it’s probably easier to not support DataFrames there and test with another class.

My perspective with allowing dataframes in `layers` was that it's pretty common for people have data (especially expression matrices) in dataframes already. I also like the simplicity of non-defensive code, and not having to special case too much as long as we can provide consistent behavior. It seems like more effort to define behavior specific to dataframes, rather than to say "we accept them, it's up to you whether that's useful".

I guess it would be fine to define some behavior for dataframes, we'd just have to decide what that is. If we're going to require they're homogenous, then I'd probably say we should just convert them to a ndarray.

flying-sheep:
OK, how about

```py
def __setitem__(self, key, layer):
    if isinstance(layer, pd.DataFrame):
        layer = layer.to_numpy()
        warn(f'Layer {key!r} converted to numpy array with dtype {layer.dtype}')
    ...
```

flying-sheep:
@ivirshup what do you say now?

ivirshup:
@flying-sheep I've been thinking about it, and want to go back on my previous statement (that homogenous dataframes should be converted to ndarrays) for a couple reasons.

1. Dataframes have very useful methods.
2. Dataframes have an extension dtype mechanism.

Both of these let you accomplish things you wouldn't be able to with just numpy arrays. For 1, the pandas `.rank` method is much more powerful than numpys. For 2, you could have categories of cell interactions in `obsp`.

I'm still dubious about the proposition objects in layers, obsp, and varp need have homogenous dtypes, but even if they do – I think there's value in allowing the containers to be dataframes.

flying-sheep:
Isn’t `pd.DataFrame(adata.X).do_stuff()` easy enough? I’m pretty sure it creates a slim view over the array, so it should be pretty much for free to do this on the fly.

Another thing I want to enforce: No layer called ``X``, `''` or `None`, as I want to unify layers and `X` down the road.

gokceneraslan:
I don't understand 90% of the discussion above but I have some practical questions/comments :) 

- What happens if a DataFrame layer has an index different than `adata.obs.index`?
- What happens if a DataFrame layer has column names different than `adata.var_names`? Isn't it too much flexibility given that the initial idea for layers was to support coupled features like unspliced/spliced counts of genes.
- Same questions for obsp/varp.
- One advantage of having a DF in obsm/varm (even for homogeneous data) is the ability to store column names (seem trivial but can be useful) That'd be great for storing HTO barcodes as I wrote in #237. This is also a case where just doing pd.DataFrame(adata.obsm[bla]) doesn't solve the issue. Image a `sc.pp.read_hashtags()` method that reads a csv file and stores such barcodes in obsm (see https://github.com/theislab/scanpy/issues/351). Then a sc.tl. method can simply read adata.obsm[hashtag] field that's created by `sc.pp.read_hashtags()` and use both the values and column names to produce things.

ivirshup:
Sorry for the big delay on feedback! I've been distracted with other work.

@flying-sheep 

> Isn’t pd.DataFrame(adata.layers['foo']).do_stuff() easy enough? I’m pretty sure it creates a slim view over the array, so it should be pretty much for free to do this on the fly.

Does pandas promise that it will always be slim view? I'd expect other dataframe libraries (especially cudf) wouldn't be. I'd also note that pandas extension dtypes won't convert to numpy arrays, and there's going to be performance changes around C vs Fortran array ordering.

> Another thing I want to enforce: No layer called X, '' or None, as I want to unify layers and X down the road.

I don't think I feel strongly about this. This was probably more "banned" before we stopped allowing `use_rep="X"`.

@gokceneraslan 

> What happens if a DataFrame layer has an index different than adata.obs.index
> What happens if a DataFrame layer has column names different than adata.var_names? Isn't it too much flexibility given that the initial idea for layers was to support coupled features like unspliced/spliced counts of genes.

We probably shouldn't allow either of these. I'd thought they wouldn't be, but it looks like they are at the moment. Any axis aligned to the one of the main axes should have the same labels. This is similar to netCDF and xarray.

> One advantage of having a DF in obsm/varm (even for homogeneous data) is the ability to store column names

👍 , I've found this useful for CITE seq data.

flying-sheep:
> Does pandas promise that it will always be slim view?

[Only the code says so](https://github.com/pandas-dev/pandas/blob/cce0b374fcdce7d5aa6f646a3020564ab7ead6f5/pandas/core/internals/construction.py#L122). A guarantee exists the other way round though: If a df is homogeneous, `to_numpy` will return a view.

> We probably shouldn't allow either of these

That’s exactly my problem: data frames are so complex that supporting them in more places means opening a can of worms, and either having to introduce checks everywhere or risk inconsistency.

> :+1:, I've found this useful for CITE seq data.

As above: Allowing dataframes in more places has more implications than just “neat, we get colnames”.

The multimodal usecase will be covered more thoroughly by #237. Let’s get that going!

flying-sheep:
So yeah, my vision:

- modes allow having different sets of layers with different variables and the same observations as described in #237
- `layers` becomes unified with X.
- layers/obsm/varm/obsp/varp entries are each homogeneous in type, and a dense or sparse array (possibly disk-backed), but not something complex like a DataFrame

Do you think any of this is unrealistic/impractical for daily use? What operations on DataFrames did you guys find useful that aren’t available for arrays? When would someone ever need a pandas extension dtype anywhere but in obs/var?

ivirshup:
> layers/obsm/varm/obsp/varp entries are each homogeneous in type, and are realized as dense or sparse arrays (possibly disk-backed).

We should definitely have data frames in `obsm` and `varm`. `varm` is probably where differential expression results (as dataframes) should go, but would also make it easier to organize data than just putting it all in `.obs`, `.var`. This might be pretty similar to your thinking about `X` just being an elements of `layers`. Here, `obs`/`var` would just be elements of `obsm`/`varm`.

I agree use cases for dataframes in `.layers` and `.obsp`, `.varp` aren't as obvious. That's mainly me not seeing any huge drawbacks.

> That’s exactly my problem: data frames are so complex that supporting them in more places means opening a can of worms, and either having to introduce checks everywhere or risk inconsistency.

I was making more of a general statement on labelled arrays. We already do throw an error for the case I thought we shouldn't allow:

```python
import scanpy as sc
import pandas as pd
import numpy as np

adata = sc.AnnData(
    X=np.ones((10, 10)),
    obs=pd.DataFrame([], index=[f"cell{i}" for i in range(10)]),
    var=pd.DataFrame([], index=[f"gene{i}" for i in range(10)])
)
newdf = pd.DataFrame([np.arange(10)], index=[f"obs{i}" for i in range(10)])

adata.obsm["new"] = newdf
IndexError                                Traceback (most recent call last)
<ipython-input-6-5bd9adc00a4f> in <module>
----> 1 adata.obsm["new"] = newdf

~/github/anndata/anndata/core/alignedmapping.py in __setitem__(self, key, value)
    176 
    177     def __setitem__(self, key: str, value: V):
--> 178         self._validate_value(value, key)
    179         self._data[key] = value
    180 

~/github/anndata/anndata/core/alignedmapping.py in _validate_value(self, val, key)
     79             # Could probably also re-order index if it's contained
     80             if not (val.index == self.dim_names).all():
---> 81                 raise IndexError()  # Maybe not index error
     82         except AttributeError:
     83             pass
```

> Do you think any of this is unrealistic/impractical for daily use? What operations on DataFrames did you guys find useful that aren’t available for arrays?

* Labelled columns in obsm, varm are really useful
* Not having to put all "annotation"-like info in .obs, .var is very useful for keeping my objects organized. For example, I might have multiple versions of qc metrics based on different pipelines. `SingleCellExperiment` achieves something similar with nested dataframes.

> When would someone ever need a pandas extension dtype anywhere but in obs/var?

Besides the cases where it's useful to just store a dataframe, I think arrays of user defined datatypes are a very powerful tool. Numpy doesn't make defining methods on these particularly easy, but I think pandas has (i.e. categoricals, geo-pandas). That said, this did get me thinking about how similar things could be done by using numba (https://github.com/numba/numba/issues/4740).

flying-sheep:
OK, decision time:

- This PR’s code for {obs,var}p and layers will stay: dataframes will be converted to arrays there, and there’ll be a warning if the conversion changed the dtype.
- I’ll allow {obs,var}m to be dataframes again

If I missed something I’

ivirshup:
That sounds reasonable, I'll take another look at the code now.

anndata/tests/test_layers.py
```diff
@@ -41,21 +41,30 @@ def test_views():
     assert not adata_view.isview
 
 
+def test_set_dataframe():
+    adata = AnnData(X)
+    df = gen_typed_df_t2_size(*X.shape)
+    with pytest.warns(UserWarning, match=r"Layer 'df'.*dtype object"):
+        adata.layers["df"] = df
+    assert isinstance(adata.layers["df"], np.ndarray)
+    assert np.issubdtype(adata.layers["df"].dtype, object)
```
ivirshup: Please add cases for: `obsp`, `varp`, and a homogenous dataframe.

Also, does this do the same thing as `adata.X = df`?

anndata/core/alignedmapping.py
```diff
@@ -44,27 +41,27 @@ def _subset_df(df: pd.DataFrame, subset_idx: OneDIdx):
...
-        Which axes is this aligned along?
-    _view_class
-        The view class for this aligned mapping.
-    _actual_class
-        The actual class (which has it's own data) for this aligned mapping.
-    """
 
-    _view_class: ClassVar["AlignedViewMixin"]
-    _actual_class: ClassVar["AlignedActualMixin"]
+    _allow_df: bool
```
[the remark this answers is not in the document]
flying-sheep: yes!

anndata/core/anndata.py
```diff
@@ -1101,7 +1101,7 @@ def uns(self, value: MutableMapping):
         self._uns = value
 
     @property
-    def obsm(self) -> AxisArraysBase:
+    def obsm(self) -> Union[AxisArrays, AxisArraysView]:
```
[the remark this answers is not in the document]
flying-sheep: Python doesn’t have a notion of closed classes with a fixed number of subtypes. I thought this way it’s cleaner that we only have two possible classes here.

anndata/core/anndata.py
```diff
@@ -1030,16 +1030,16 @@ def raw(self) -> Raw:
...
             self._raw = None
-        else:
+        elif isinstance(value, AnnData):
             if self.isview:
                 self._init_as_actual(self.copy())
             self._raw = Raw(value)
+        else:
+            raise ValueError(
+                'Can only init raw attribute with an AnnData object or `None`.'
+            )
```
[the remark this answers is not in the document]
flying-sheep: Right!

anndata/tests/test_layers.py
flying-sheep: will do

anndata/core/alignedmapping.py
```diff
@@ -81,6 +78,15 @@ def _validate_value(self, val: V, key: str) -> None:
                 raise IndexError()  # Maybe not index error
         except AttributeError:
             pass
+        if not self._allow_df and isinstance(val, pd.DataFrame):
+            was_heterogeneous = val.dtypes.nunique() != 1
+            val = val.to_numpy()
```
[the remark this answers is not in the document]
ivirshup: Hmm. I think I'd say go for a reference and if it's confusing someone will open an issue. I think that'd feel more consistent to me.

More generally, when we do `adata.raw = adata` we make a copy. Is that counter intuitive if we otherwise assign references?

anndata/core/anndata.py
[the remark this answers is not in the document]
ivirshup: I think I've seen docs with `adata.raw = None`. We should probably check on that, maybe just add the deleter but keep the behavior of `.raw = None`?

anndata/core/alignedmapping.py
```diff
@@ -75,12 +71,10 @@ def _validate_value(self, val: V, key: str) -> None:
                     f"{self.axes} of parent. Value had shape {val.shape} while "
                     f"it should have had {right_shape}."
                 )
-        try:  # TODO: Handle objects with indices
-            # Could probably also re-order index if it's contained
-            if not (val.index == self.dim_names).all():
-                raise IndexError()  # Maybe not index error
-        except AttributeError:
-            pass
```
ivirshup: @flying-sheep, why did this get removed?

flying-sheep: It wasn’t, ctrl+f “val.index == self.dim_names”:

https://github.com/theislab/anndata/pull/228/files#diff-2a3b9403444574f42d160b2f5ea71bc4R221

ivirshup: But why not keep it here? Arguably the same `super()` dispatch could also happen with the homogenous df. In either case, I think the shape check error is more helpful when it applies, and should happen first.

flying-sheep: Agreed about the second part, but I didn’t understand the first: Where could it maybe happen?

ivirshup: For the first part, the Layers and Pairwise classes could get a method like this:

```python
def _validate_value(self, val: V, key: Hashable) -> V:
    if not self._allow_df and isinstance(val, pd.DataFrame):
        name = self.attrname.title().rstrip('s')
        val = ensure_df_homogeneous(val, f'{name} {key!r}')
    return super()._validate_value(val, key)
```

Not that I think this is the way to go. This feels a little "aspect-y", which is not a common paradigm in our codebase.

PR #230, opened 2019-09-25
Support for ordered categoricals.

ivirshup:
* IO supported via attribute on categories
* Tests now check order of categoricals

Were there other reasons we didn't support these other than having an on-disk representation?

### DIFF
--- a/anndata/tests/helpers.py
+++ b/anndata/tests/helpers.py
@@ -38,6 +38,7 @@ def gen_typed_df(n, index=None):
     return pd.DataFrame(
         dict(
             cat=pd.Categorical(np.random.choice(letters, n)),
+            cat_ordered=pd.Categorical(np.random.choice(letters, n), ordered=True),
             int64=np.random.randint(-50, 50, n),
             float64=np.random.random(n),
             uint8=np.random.randint(255, size=n, dtype="uint8"),
@@ -297,9 +298,6 @@ def are_equal_dataframe(a, b, exact=False, elem_name=None):
         report_name(pd.testing.assert_frame_equal)(
             a,
             b,
-            check_names=False,
-            check_categorical=False,  # check encoded values, but not codes
-            # Should different orderings be allowed?
             _elem_name=elem_name,
         )
     else:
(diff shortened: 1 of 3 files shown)


LuckyMD:
There were some issues in between with all categoricals being ordered. That created problems for integrating other DE methods with scanpy, as the DE methods treat ordered categoricals differently to unordered categoricals. I wonder if people may accidentally store categoricals as ordered and then get strange results in DE testing if this is supported. What do you need ordered categoricals for in sc analysis?

anndata/tests/helpers.py
```diff
@@ -299,11 +300,6 @@ def are_equal_dataframe(a, b, exact=False, elem_name=None):
         b,
         check_index_type=exact,
         check_exact=exact,
-        # If False: check encoded values, but not codes
-        check_categorical=exact,
-        # Should different orderings be allowed?
-        # check_like=exact,
-        check_names=False,
```
flying-sheep: @ivirshup can you check if my merge is correct? I simplified this code simulataneously to you changing it.

ivirshup:
@LuckyMD, this doesn't remove support for unordered categorical or change the behaviour for automatic categorical conversion. Now it just means that if you do something like:

```python
adata.obs["time"] = pd.Categorical(times, categories=["1hr", "5hr", "10hr"], ordered=True)
```

That will be maintained on roundtrips to disk, and will be available for downstream use by libraries like `seaborn`. This is mostly inspired by some figures I've seen at talks where people "faked" ordered categories with `scanpy` by using numerical values and a continuous colormap. Real ugly stuff.

anndata/tests/helpers.py
ivirshup: Looks fine to me!

LuckyMD:
I guess seaborn just orders the legend differently if you pass ordered categoricals? Or is there another effect?

ivirshup:
Looks like it's now running into this https://github.com/linnarsson-lab/loompy/issues/113

ivirshup:
Is anyone strongly against merging this?

flying-sheep:
I’m happy with this!

PR #252, opened 2019-11-18
Fix structured array IO

### LINKED ISSUE #249: Zarr structured arrays with strings
AFAIK the last blocking bug for 0.7. Originally reported in https://github.com/theislab/scanpy/issues/832.

Minimal reproducer:

```python
import scanpy as sc
pbmc = sc.datasets.pbmc68k_reduced()
pbmc.write("tmp.h5ad")
fromdisk = sc.read("tmp.h5ad")  # Do we read okay
fromdisk.write(pbmc)  # Can we round trip
```

The issue here is with structured numpy arrays and the variety of string types, and didn't get caught earlier because these are a bit of pain to actually instantiate... A brief summary of the conflict (copied from the earlier issue:

* h5py doesn't do fixed length unicode strings
* h5py does do variable length unicode strings, pretty much anywhere
* zarr doesn't do variable length strings in structured arrays
* We probably don't actually want to use fixed length unicode strings much. Bytestrings, more likely.
* We can probably just add another element type to allow special handling for these. I think it'd be fine to not do np.str_ type arrays.

This is pretty easy to fix for hdf5 if we just say all unicode strings are variable length. Zarr has an open pull request to support this https://github.com/zarr-developers/zarr-python/pull/422.
 The question is whether we wait for a zarr release to keep consistency between the formats. This is the simplest solution, and probably what we should go with once it's available. The problem is we end up with some intermediary solution if it's not available yet, which adds complexity to backwards compatibility.

ivirshup:
Fixes #249 and fixes theislab/scanpy#832 by saying we'll always read unicode strings in as variable length. This still needs an editing pass, and another look over in the morning, but I'd like to get feedback as soon as possible. @flying-sheep, thoughts?

### DIFF
--- a/anndata/compat.py
+++ b/anndata/compat.py
@@ -75,7 +75,7 @@ def _from_fixed_length_strings(value):
         # https://github.com/h5py/h5py/issues/1307
         if issubclass(np.dtype(dt_type).type, np.string_):
             dt_list[1] = f"U{int(dt_type[2:])}"
-        elif is_annotated or issubclass(np.dtype(dt_type).type, np.str_):
+        elif is_annotated or np.issubdtype(np.dtype(dt_type), np.str_):
             dt_list[1] = "O"  # Assumption that it's a vlen str
         new_dtype.append(tuple(dt_list))
     return value.astype(new_dtype)
@@ -91,9 +91,9 @@ def _to_fixed_length_strings(value: np.ndarray) -> np.ndarray:
     new_dtype = []
     for dt_name, (dt_type, dt_offset) in value.dtype.fields.items():
         if dt_type.kind == "O":
-            #  Asumming the objects are str
+            #  Assuming the objects are str
             size = max(len(x.encode()) for x in value.getfield("O", dt_offset))
-            new_dtype.append((dt_name, f"U{size}"))
+            new_dtype.append((dt_name, ("U", size)))
         else:
             new_dtype.append((dt_name, dt_type))
     return value.astype(new_dtype)
--- a/anndata/compat.py
+++ b/anndata/compat.py
@@ -89,9 +89,11 @@ def _to_fixed_length_strings(value: np.ndarray) -> np.ndarray:
     https://github.com/zarr-developers/zarr-python/pull/422
     """
     new_dtype = []
-    for dt_name, (dt_type, _) in value.dtype.fields.items():
-        if dt_type.str[1] in ("U", "O"):
-            new_dtype.append((dt_name, "U200"))
+    for dt_name, (dt_type, dt_offset) in value.dtype.fields.items():
+        if dt_type.kind == "O":
+            #  Asumming the objects are str
+            size = max(len(x.encode()) for x in value.getfield("O", dt_offset))
+            new_dtype.append((dt_name, f"U{size}"))
         else:
             new_dtype.append((dt_name, dt_type))
     return value.astype(new_dtype)
--- a/anndata/compat.py
+++ b/anndata/compat.py
@@ -75,9 +75,25 @@ def _from_fixed_length_strings(value):
         # https://github.com/h5py/h5py/issues/1307
         if issubclass(np.dtype(dt_type).type, np.string_):
             dt_list[1] = f"U{int(dt_type[2:])}"
-        elif is_annotated:
+        elif is_annotated or issubclass(np.dtype(dt_type).type, np.str_):
             dt_list[1] = "O"  # Assumption that it's a vlen str
         new_dtype.append(tuple(dt_list))
+    return value.astype(new_dtype)
+
+
+def _to_fixed_length_strings(value: np.ndarray) -> np.ndarray:
+    """
+    Convert variable length strings to fixed length.
+
+    Currently a workaround for
+    https://github.com/zarr-developers/zarr-python/pull/422
+    """
+    new_dtype = []
+    for dt_name, (dt_type, _) in value.dtype.fields.items():
+        if dt_type.str[1] in ("U", "O"):
+            new_dtype.append((dt_name, "U200"))
+        else:
+            new_dtype.append((dt_name, dt_type))
     return value.astype(new_dtype)
 
 
--- a/anndata/readwrite/h5ad.py
+++ b/anndata/readwrite/h5ad.py
@@ -34,8 +34,8 @@ def _to_hdf5_vlen_strings(value: np.ndarray) -> np.ndarray:
     This corrects compound dtypes to work with hdf5 files.
     """
     new_dtype = []
-    for dt_name, dt_type in value.dtype.descr:
-        if dt_type[1] in ("U", "O"):
+    for dt_name, (dt_type, _) in value.dtype.fields.items():
+        if dt_type.kind in ("U", "O"):
             new_dtype.append((dt_name, h5py.special_dtype(vlen=str)))
         else:
             new_dtype.append((dt_name, dt_type))
--- a/anndata/readwrite/h5ad.py
+++ b/anndata/readwrite/h5ad.py
@@ -15,7 +15,13 @@ from scipy import sparse
 from ..core.sparsedataset import SparseDataset
 from ..core.anndata import AnnData, Raw
 from ..compat import _from_fixed_length_strings, _clean_uns, Literal
-from .utils import report_key_on_error, idx_chunks_along_axis, write_attribute
+from .utils import (
+    report_read_key_on_error,
+    report_write_key_on_error,
+    idx_chunks_along_axis,
+    write_attribute,
+    read_attribute,
+)
 
 
 H5Group = Union[h5py.Group, h5py.File]
@@ -138,6 +144,7 @@ def write_raw(f, key, value, dataset_kwargs=MappingProxyType({})):
     write_attribute(f, "raw/varm", value.varm, dataset_kwargs=dataset_kwargs)
 
 
+@report_write_key_on_error
 def write_not_implemented(f, key, value, dataset_kwargs=MappingProxyType({})):
     # If it's not an array, try and make it an array. If that fails, pickle it.
     # Maybe rethink that, maybe this should just pickle,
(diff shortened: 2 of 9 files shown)


anndata/compat.py
```diff
@@ -75,12 +75,28 @@ def _from_fixed_length_strings(value):
...
+    """
+    Convert variable length strings to fixed length.
+
+    Currently a workaround for
+    https://github.com/zarr-developers/zarr-python/pull/422
+    """
+    new_dtype = []
+    for dt_name, (dt_type, _) in value.dtype.fields.items():
+        if dt_type.str[1] in ("U", "O"):
+            new_dtype.append((dt_name, "U200"))
```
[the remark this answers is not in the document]
ivirshup: Zarr is not currently able to do that. All this function is for is working around that limitation until it's solved.

ivirshup:
> I think we should either allow any length of strings or warn the user if their strings are too long.

Totally agree with that being the eventual goal. This is easy to do by just keeping all values as variable length strings. Until those are supported in structured arrays by zarr, I'm thinking we just keep the old behavior of using an arbitrary fixed length that works for most gene names (since that's the only time we use structured arrays). Previously this was fixed to 50 bytes for zarr, but 200 for `hdf5` (based on the pbmc68k dataset).

anndata/compat.py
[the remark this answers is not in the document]
ivirshup: Ah, turns out that was making the values have length 0.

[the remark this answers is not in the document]
flying-sheep: OK, it’s different on array creation, so there seems to be a way to make it automatically infer it:

```py
>>> np.array(['abc', 'a', 'abcd'], dtype=np.unicode).dtype                                                                                                                                                                               
dtype('<U4')
```

anndata/compat.py
```diff
@@ -75,12 +75,30 @@ def _from_fixed_length_strings(value):
         # https://github.com/h5py/h5py/issues/1307
         if issubclass(np.dtype(dt_type).type, np.string_):
             dt_list[1] = f"U{int(dt_type[2:])}"
-        elif is_annotated:
+        elif is_annotated or issubclass(np.dtype(dt_type).type, np.str_):
```
flying-sheep: Isn’t there `issubdtype` for this?

anndata/compat.py
```diff
@@ -75,12 +75,30 @@ def _from_fixed_length_strings(value):
...
+
+    Currently a workaround for
+    https://github.com/zarr-developers/zarr-python/pull/422
+    """
+    new_dtype = []
+    for dt_name, (dt_type, dt_offset) in value.dtype.fields.items():
+        if dt_type.kind == "O":
+            #  Asumming the objects are str
+            size = max(len(x.encode()) for x in value.getfield("O", dt_offset))
+            new_dtype.append((dt_name, f"U{size}"))
```
flying-sheep:
```suggestion
            new_dtype.append((dt_name, ("U", size)))
```

anndata/readwrite/h5ad.py
```diff
@@ -29,7 +35,7 @@ def _to_hdf5_vlen_strings(value: np.ndarray) -> np.ndarray:
     """
     new_dtype = []
     for dt_name, dt_type in value.dtype.descr:
-        if dt_type[1] == "U":
+        if dt_type[1] in ("U", "O"):
```
flying-sheep:
```suggestion
        if dt_type.kind in ("U", "O"):
```

anndata/readwrite/h5ad.py
```diff
@@ -29,7 +35,7 @@ def _to_hdf5_vlen_strings(value: np.ndarray) -> np.ndarray:
     """
     new_dtype = []
     for dt_name, dt_type in value.dtype.descr:
```
flying-sheep:
```suggestion
    for dt_name, (dt_type, _offset) in value.dtype.fields:
```

anndata/compat.py
```diff
@@ -75,12 +75,30 @@ def _from_fixed_length_strings(value):
...
+    """
+    Convert variable length strings to fixed length.
+
+    Currently a workaround for
+    https://github.com/zarr-developers/zarr-python/pull/422
+    """
+    new_dtype = []
+    for dt_name, (dt_type, dt_offset) in value.dtype.fields.items():
+        if dt_type.kind == "O":
+            #  Asumming the objects are str
```
flying-sheep:
```suggestion
            #  Assuming the objects are str
```

anndata/readwrite/utils.py
```diff
@@ -150,3 +161,51 @@ def func_wrapper(elem, *args, **kwargs):
                 )
 
     return func_wrapper
+
+
+def report_write_key_on_error(func):
```
flying-sheep: duplicated code, please don’t copy&paste code.

anndata/compat.py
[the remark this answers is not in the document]
ivirshup: Because arrays with structured dtypes are poorly supported?

anndata/compat.py
ivirshup: For this one, why?

anndata/compat.py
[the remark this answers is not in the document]
flying-sheep: A pity, but what we have now should work! Thank you.

anndata/compat.py
flying-sheep: I think it’s more elegant not to construct strings if there’s another API.

anndata/readwrite/utils.py
```diff
@@ -196,9 +196,6 @@ def func_wrapper(elem, key, val, *args, **kwargs):
                 raise type(e)("\n\n".join((
                     str(e),
                     f"Above error raised while writing key '{key}' of type {type(val)} from {parent}."
-                )))
-                # raise AnnDataReadError(
-                #     f"Above error raised while writing key '{key}' of type {type(val)} from {parent}."
-                # )
+                ))).with_traceback(sys.exc_info()[2])
```
flying-sheep: Unclear what this is supposed to do. “Thrown twice”? If you mean you don’t want to see “this exception was caused by”, you should use [PEP 3134](https://www.python.org/dev/peps/pep-3134/#explicit-exception-chaining): `raise ... from e` or `raise ... from e.__cause__` or so.

flying-sheep: <table>
<tr>
<td>

```pytb
In [12]: try: 
    ...:     try: 
    ...:         raise Exception('test') 
    ...:     except Exception: 
    ...:         raise Exception('test2') 
    ...: except Exception as e: 
    ...:     raise Exception(f'{e} 3') from e 
    ...:      
    ...:                                                                                                                                                                                                                                     
Exception                                 Traceback (most recent call last)
<ipython-input-12-6e2b8f9e70be> in <module>
      2     try:
----> 3         raise Exception('test')
      4     except Exception:

Exception: test

Exception                                 Traceback (most recent call last)
<ipython-input-12-6e2b8f9e70be> in <module>
      4     except Exception:
----> 5         raise Exception('test2')
      6 except Exception as e:

Exception: test2

Exception                                 Traceback (most recent call last)
<ipython-input-12-6e2b8f9e70be> in <module>
      5         raise Exception('test2')
      6 except Exception as e:
----> 7     raise Exception(f'{e} 3') from e
      8 
      9 

Exception: test2 3
```

</td><td>

```pytb
In [13]: try: 
    ...:     try: 
    ...:         raise Exception('test') 
    ...:     except Exception: 
    ...:         raise Exception('test2') 
    ...: except Exception as e: 
    ...:     raise Exception(f'{e} 3').with_traceback(sys.exc_info()[2]) 
    ...:                                                                                                                                                                                                                                     
Exception                                 Traceback (most recent call last)
<ipython-input-13-8dde79ff88d6> in <module>
      2     try:
----> 3         raise Exception('test')
      4     except Exception:

Exception: test

Exception                                 Traceback (most recent call last)
<ipython-input-13-8dde79ff88d6> in <module>
      4     except Exception:
----> 5         raise Exception('test2')
      6 except Exception as e:

Exception: test2

Exception                                 Traceback (most recent call last)
<ipython-input-13-8dde79ff88d6> in <module>
      5         raise Exception('test2')
      6 except Exception as e:
----> 7     raise Exception(f'{e} 3').with_traceback(sys.exc_info()[2])
      8 

<ipython-input-13-8dde79ff88d6> in <module>
      3         raise Exception('test')
      4     except Exception:
----> 5         raise Exception('test2')
      6 except Exception as e:
      7     raise Exception(f'{e} 3').with_traceback(sys.exc_info()[2])

Exception: test2 3
```

</td></tr></table>

ivirshup:
This definitely shouldn't be squashed. The granularity of the commits will be useful in the future. While writing this I wished I still had the commit history from previous implementation so I could see what things I had tried before. This is especially the case for working with poorly documented interfaces like structured dtypes.

I could potentially reorder the commits, and squash some formatting ones, but since I typically navigate git history through blames I'm not sure how useful this would be.

anndata/readwrite/utils.py
ivirshup: When the traceback is longer I think it doesn't look as strange for the `with_traceback` version, but sure. `from` also seems to do some reordering, but either is easy enough to parse.

What I would really like to do, is just modify the existing error to have some extra information in the message. I don't think there's a perfect solution to this since exceptions can define their own string representation.

flying-sheep: yes, `type(e)(msg)` is flaky. The exception type’s `__init__` might not accept a message as its first argument.

If you understand the `with_taceback` output, please explain it to me because I don’t. The modified exception has 2 locations? huh?

flying-sheep:
Of course blame should be maximally useful, so there’s two options:

1. Squash everything (low effort, low granularity)
2. Manually rebase: Every commit done to fix tests or formatting or bugs you introduced in earlier commits of the same PR need to be squashed. (high effort, perfect granularity)

Keeping “Formatting” and “whoops” commits are detrimental to squash being useful.

ivirshup:
@flying-sheep, I've opened a new issue about managing git history so we can split that discussion off. For this PR, I'll follow the strategy I propose in my top comment. Once this is ready to merge, I'll remove formatting commits, and consider squashing other ones. We can figure out a different strategy, but I don't think it's worth holding up this PR for.

---------------------------------

Back on topic of this PR, is there code that should change?

flying-sheep:
No, looks good to go!

ivirshup:
@flying-sheep, I've removed the formatting commits. I think I want to test this more thoroughly, but I'll do that in a separate PR. Ready to merge.

flying-sheep:
Great, thank you! Now let's get this released!

PR #256, opened 2019-11-21
to_adata for .raw

falexwolf:
Allows to replace ugly code like
```
adata_new = AnnData(adata.raw.X, var=adata.raw.var, obs=adata.obs, obsm=adata.obsm, uns=adata.uns)
```
with 
```
adata_new = adata.raw.to_adata()
```

### DIFF
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -369,7 +369,7 @@ class Raw:
             varm=None if self._varm is None else self._varm.copy(),
             obs=self._adata.obs.copy(),
             obsm=self._adata.obsm.copy(),
-            uns=self._adata.uns.copy()
+            uns=self._adata.uns.copy(),
         )
 
     def _normalize_indices(self, packed_index):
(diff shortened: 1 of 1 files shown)


falexwolf:
I don't think there is anything to discuss, here, is there?

gokceneraslan:
Oh I also needed this so much, thanks!

fidelram:
Appreciated. I am doing the ugly way frequently.

gokceneraslan:
Are adata.raw.obsm and adata.obsm pointing to  the same thing? Same goes for uns. If not the same, then the ugly code and the new code are not equivalent.

ivirshup:
`adata.raw.obsm` isn't defined, same for `.raw.uns`, `.raw.obs`, `.raw.obsp`.

gokceneraslan:
ok, that's what I suspect, too. I missed the _adata part in `self._adata.obs.copy `. thanks.

gokceneraslan:
Do we need to add `obsp` to this code, now that neighbors and distances are stored in `obsp`?

PR #259, opened 2019-11-23
Color handling in uns fix

### LINKED ISSUE #257: AnnData exception in sc.pp.pca
When I subset an anndata and run PCA, I get this exception sometimes:

image

Any ideas?

ivirshup:
@gokceneraslan

Fixes #257.

When the number of colors was less than the number of categories, subsetting would throw an error. This just resets the colors instead.

### DIFF
--- a/anndata/core/anndata.py
+++ b/anndata/core/anndata.py
@@ -1416,16 +1416,22 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             all_categories = df_full[k].cat.categories
             df_sub[k].cat.remove_unused_categories(inplace=True)
             # also correct the colors...
-            if f'{k}_colors' not in uns:
+            color_key = f"{k}_colors"
+            if color_key not in uns:
                 continue
+            color_vec = uns[color_key]
             # this is a strange hack...
-            if np.array(uns[f'{k}_colors']).ndim == 0:
-                idx = None
+            # What case is this? Does this happen?
+            if np.array(color_vec).ndim == 0:
+                uns[color_key] = np.array(color_vec)[(None,)]
+            elif len(color_vec) < len(all_categories):
+                # Reset colors
+                del uns[color_key]
             else:
                 idx = np.where(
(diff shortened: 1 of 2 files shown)


anndata/core/anndata.py
```diff
@@ -1416,16 +1416,22 @@ def _remove_unused_categories(self, df_full, df_sub, uns):
...
+            if color_key not in uns:
                 continue
+            color_vec = uns[color_key]
             # this is a strange hack...
-            if np.array(uns[f'{k}_colors']).ndim == 0:
-                idx = None
+            # What case is this? Does this happen?
+            if np.array(color_vec).ndim == 0:
+                uns[color_key] = np.array(color_vec)[(None,)]
+            elif len(color_vec) < len(all_categories):
```
gokceneraslan: Maybe `elif len(color_vec) != len(all_categories):` instead?

ivirshup: The current behavior is to re-use the colors in that case, but I've never been sure if that's intentional. It preserves user supplied color schemes, but it doesn't necessarily preserve the mapping between cluster identifier and color.

flying-sheep: You two are right: It doesn’t makes sense that way. They should just be redone.

anndata/core/anndata.py
```diff
@@ -1420,9 +1420,8 @@ def _remove_unused_categories(self, df_full, df_sub, uns):
             if color_key not in uns:
                 continue
             color_vec = uns[color_key]
-            # this is a strange hack...
-            # What case is this? Does this happen?
             if np.array(color_vec).ndim == 0:
+                # Make 0D arrays into 1D ones
```
ivirshup: Do we want this to happen?

ivirshup: I would assume not. If it's not a one dimensional array, I don't think we put it there.

flying-sheep: @falexwolf implemented that specifically in ac08c617458dc500bc5a308f368bb28c132d5d26.

Alex, can you help here?

PR #268, opened 2019-12-06
Restructure slim

flying-sheep:
Slimmed down version of #260

- No aligned_mapping restructure
- No isbacked restructure

Rebasing done, should be almost final!

- Remaining question: We should be consistent with `is_` and `as_` (except for `asarray`, which is a numpy convention). We’ll therefore get rid of `isbacked` soon: Should we rename it to `is_backed` before it vanishes completely in favor of `backing: Union[Path, None, ...]`? Seems as too much churn…
- Also needs more Raw tests


anndata/core.py
```diff
@@ -0,0 +1,5 @@
+from warnings import warn
+
+warn("Please only import from anndata, not anndata.core", DeprecationWarning)
```
flying-sheep: @falexwolf Is that the warning you had in mind?

anndata/tests/test_raw.py
```diff
@@ -31,37 +31,49 @@
...
+    with pytest.warns(
+        ImplicitModificationWarning, match="Initializing view as actual"
+    ):  # TODO: don’t modify adata just to write it
+        adata_raw.filename = backing_h5ad
 
-    del adata.raw
-    assert adata.raw is None
+    assert adata_raw.var_names.tolist() == ['var1', 'var2']
+    assert adata_raw.raw.var_names.tolist() == ['var1', 'var2', 'var3']
+    assert adata_raw.raw[:, 0].X[:].tolist() == [[1], [4], [7]]
```
flying-sheep: Well, this is broken because of

https://github.com/theislab/anndata/blob/4baf74bed7bf120176d9fa1eb3e59c646ca16a0d/anndata/_core/raw.py#L102-L105

And has been since the beginning. Wat do.

ivirshup: How is this broken exactly?

flying-sheep: `adata_raw.raw[:, 0].X.shape != (3, 1)`

ivirshup: Ah, I think that just need to be fixed here:
https://github.com/theislab/anndata/blob/4baf74bed7bf120176d9fa1eb3e59c646ca16a0d/anndata/_core/raw.py#L37-L63

I think this is the need for "view of raw" that pops up in all the comments.

ivirshup: If it's always been broken you could probably just open an issue and not touch it for now.

anndata/_core/anndata.py
```diff
@@ -1796,10 +1430,13 @@ def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
...
                 obsp=self.obsp.copy(),
                 varp=self.varp.copy(),
-                raw=None if self._raw is None else self._raw.copy(),
                 layers=self.layers.copy(),
                 dtype=dtype,
             )
+            if self._raw is not None:
+                adata._raw = self._raw.copy()
+                adata._raw._adata = adata
+            return adata
```
ivirshup: This should still go through the AnnData constructor.

Was there a reason you didn't want to do that?

anndata/_core/anndata.py
```diff
@@ -832,34 +488,27 @@ def _init_as_actual(
...
+                raw, Raw
+            ), "got raw from other adata but also filename?"
+            from ..readwrite.h5ad import _read_raw, _read_legacy_raw_h5ad
+
+            # Read raw dict from file
+            raw, raw_override = {}, raw
+            if "raw" in self.file:
+                raw = _read_raw(self.file)
+            _read_legacy_raw_h5ad(self.file, raw, self.filename)
+
```
ivirshup: Doesn't this read `raw/X`? This is what we want to avoid in backed mode.

anndata/_core/anndata.py
```diff
@@ -832,34 +488,27 @@ def _init_as_actual(
...
+            # override retrieved entries with specified ones
+            if raw_override:
+                raw.update(raw_override)
+        if raw and isinstance(raw, cabc.Mapping):
```
ivirshup: I still think if you're checking if the value is `None` that should be explicit.

As an appeal to authority, the [google python style guide](http://google.github.io/styleguide/pyguide.html#2144-decision) and [pep8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) say this too.

anndata/tests/test_raw.py
flying-sheep: definitely out of scope for this PR

anndata/_core/anndata.py
flying-sheep: Fixed.

anndata/_core/anndata.py
flying-sheep: Yes, e.g. `test_sparse_to_dense_inplace` will fail, as  `assert isinstance(backed.raw.X, h5py.Dataset)`:

```pytb
anndata/tests/test_io_conversion.py:62 (test_sparse_to_dense_inplace[scipy-csr])
tmp_path = PosixPath('/tmp/pytest-of-phil/pytest-12/test_sparse_to_dense_inplace_s0')
spmtx_format = <class 'scipy.sparse.csr.csr_matrix'>

    def test_sparse_to_dense_inplace(tmp_path, spmtx_format):
        pth = tmp_path / "adata.h5ad"
        orig = gen_adata((50, 50), spmtx_format)
        orig.raw = orig
        orig.write(pth)
        backed = ad.read_h5ad(pth, backed="r+")
        backed.write(as_dense=("X", "raw/X"))
        new = ad.read_h5ad(pth)
    
        assert_equal(orig, new)
        assert_equal(backed, new)
    
        assert isinstance(new.X, np.ndarray)
        assert isinstance(new.raw.X, np.ndarray)
        assert isinstance(orig.X, spmtx_format)
        assert isinstance(orig.raw.X, spmtx_format)
        assert isinstance(backed.X, h5py.Dataset)
>       assert isinstance(backed.raw.X, h5py.Dataset)
E       AssertionError: assert False
E        +  where False = isinstance(<50x50 sparse matrix of type '<class 'numpy.float32'>'\n	with 1010 stored elements in Compressed Sparse Row format>, <class 'h5py._hl.dataset.Dataset'>)
E        +    where <50x50 sparse matrix of type '<class 'numpy.float32'>'\n	with 1010 stored elements in Compressed Sparse Row format> = <anndata._core.raw.Raw object at 0x7efbf1cd64f0>.X
E        +      where <anndata._core.raw.Raw object at 0x7efbf1cd64f0> = AnnData object with n_obs × n_vars = 50 × 50 backed at '/tmp/pytest-of-phil/pytest-12/test_sparse_to_dense_inplace_s0/...    varm: 'array', 'df', 'sparse'\n    layers: 'array', 'sparse'\n    obsp: 'array', 'sparse'\n    varp: 'array', 'sparse'.raw
E        +    and   <class 'h5py._hl.dataset.Dataset'> = h5py.Dataset

anndata/tests/test_io_conversion.py:80: AssertionError
```

anndata/_core/anndata.py
flying-sheep: What you linked agrees with me:

> Google: Use the “implicit” false if possible, e.g., `if foo`: rather than `if foo != []`

> PEP8: Also, beware of writing `if x` when you really mean `if x is not None` – e.g. when testing whether a variable or argument that defaults to None was set to some other value.

I precisely mean “if raw is not empty”, elegantly expressed in Python as `if raw`: `raw` can either be `{something}` or `{}` or `None` and I want to treat the last two cases in the same way.

Using `if x` is like using `if x == y`: It (`__bool__`/`__eq__`) relies on user-specified behavior and it’s pythonic to use it. If you propose to not rely on `__bool__` on the grounds of numpy overriding it, we also need to get rid of our use of `__eq__` in our code base. I say we keep using both, because we usually know if something might be a numpy array or not.

Being specific about `is None` is useful for one purpose: If you want to treat the `None` case differently from each other empty case. That’s common if you have one of those (overlapping) use cases:

1. have `param=None` as default value and want some custom, smart behavior.
2. want to treat `None` differently from other falsy values

In this case, `raw={}` and `raw=None` mean exactly the same, so we use `if raw`.

anndata/_core/anndata.py
ivirshup: That makes it sound like there's something very broken going on. When is `copy` even being called?

I think this needs addressing, since `raw.X` should not be in-memory or sparse here.

flying-sheep: Out of scope for this PR. If someone want to further fix Raw afterwards, we should do it afterwards.

ivirshup: That commit looks right to me. These tests:

```python
assert isinstance(backed.X, h5py.Dataset)
assert isinstance(backed.raw.X, h5py.Dataset)
```

definitely shouldn't fail. I don't see how changing `copy` could make them fail unless something else is going wrong.

setup.py
```diff
@@ -26,7 +26,7 @@
         for l in Path('requirements.txt').read_text('utf-8').splitlines()
     ],
     python_requires='>=3.6',
-    packages=find_packages(),
+    packages=find_namespace_packages(include=['anndata.*']),
```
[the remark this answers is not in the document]
flying-sheep: I already did. It just finds all directories in there, which e.g. means that the tests of an installed anndata can actually be run because the directory with the necessary test data gets installed.

anndata/readwrite/h5ad.py
```diff
@@ -340,19 +341,8 @@ def read_h5ad_backed(
...
-            raw["varm"] = read_attribute(f["raw/varm"])
-    else:  # Legacy case
-        if "raw.var" in f:
-            raw["var"] = read_dataframe(f["raw.var"])  # Backwards compat
-        if "raw.varm" in f:
-            raw["varm"] = read_attribute(f["raw.varm"])
-
-    if len(raw) > 0:
-        d["raw"] = raw
+    d["raw"] = _read_raw(f, filename=filename)
```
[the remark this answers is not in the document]
flying-sheep: fixed.

anndata/readwrite/h5ad.py
```diff
@@ -455,25 +445,30 @@ def read_h5ad(
...
-            ), f"File {filename} has both legacy and current raw formats."
-            d["raw"] = raw
-
     _clean_uns(d)  # backwards compat
 
     return AnnData(**d)
 
 
+def _read_raw(
+    f, as_sparse=(), rdasp=None, *, attrs=("X", 'var', "varm"), filename=None
```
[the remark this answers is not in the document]
flying-sheep: filename is only used for an error message. can I just take it from `f`?

anndata/readwrite/h5ad.py
```diff
@@ -455,25 +445,30 @@ def read_h5ad(
...
+    if "X" in attrs and "raw/X" in f:
+        read_x = rdasp if "raw/X" in as_sparse else read_attribute
+        raw["X"] = read_x(f["raw/X"])
+    if "var" in attrs and "raw/var" in f:
+        raw["var"] = read_dataframe(f["raw/var"])
+    if "varm" in attrs and "raw/varm" in f:
+        raw["varm"] = read_attribute(f["raw/varm"])
+    _read_legacy_raw_into(
+        f, raw, read_dataframe, read_attribute, attrs=attrs, filename=filename
+    )
```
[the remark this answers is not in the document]
flying-sheep: !

This is a function called from zarr and from h5ad, so we need to pass in the functions it uses. If you wanted to know something else, please be more specific :wink:

anndata/_core/anndata.py
flying-sheep: If I do `self.raw = AnnData(**raw)`, it fails in this way.

anndata/readwrite/h5ad.py
```diff
@@ -455,25 +445,30 @@ def read_h5ad(
...
+    f, as_sparse=(), rdasp=None, *, attrs=("X", 'var', "varm"), filename=None
+):
+    if as_sparse:
+        assert rdasp is not None, 'must supply rdasp if as_sparse is supplied'
+    raw = {}
+    if "X" in attrs and "raw/X" in f:
+        read_x = rdasp if "raw/X" in as_sparse else read_attribute
+        raw["X"] = read_x(f["raw/X"])
+    if "var" in attrs and "raw/var" in f:
+        raw["var"] = read_dataframe(f["raw/var"])
```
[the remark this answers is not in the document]
flying-sheep: I took this from here:

https://github.com/theislab/anndata/blob/5a6459cbb2de947f8ff56507a0819d58dcdf7e40/anndata/readwrite/h5ad.py#L444

anndata/readwrite/h5ad.py
```diff
@@ -455,25 +445,30 @@ def read_h5ad(
...
 
+def _read_raw(
+    f, as_sparse=(), rdasp=None, *, attrs=("X", 'var', "varm"), filename=None
+):
+    if as_sparse:
+        assert rdasp is not None, 'must supply rdasp if as_sparse is supplied'
+    raw = {}
+    if "X" in attrs and "raw/X" in f:
+        read_x = rdasp if "raw/X" in as_sparse else read_attribute
+        raw["X"] = read_x(f["raw/X"])
```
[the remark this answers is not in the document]
flying-sheep: I like it better the way it is :smile:

anndata/readwrite/h5ad.py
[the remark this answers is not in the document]
ivirshup: Yes. If it's a `File` `f.filename`, if it's a group: `f.file.filename`.

anndata/_core/anndata.py
```diff
@@ -832,34 +488,26 @@ def _init_as_actual(
...
-                    raw['var'], shape[1], ['var_names', 'col_names']
+            raw = {
+                # Read raw dict from file
+                **_read_raw(
+                    self.file, attrs={"var", "varm"}, filename=self.filename
                 ),
-                varm=raw['varm'] if 'varm' in raw else None,
-            )
+                # override retrieved entries with specified ones
+                **convert_to_dict(raw),
```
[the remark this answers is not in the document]
ivirshup: I think that would be good, as long as nothing breaks. `read_h5ad_backed` should have no value for raw if it's not there, right?

anndata/readwrite/h5ad.py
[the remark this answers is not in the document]
ivirshup: Why does a function for reading current raw files call `_read_legacy_raw_into` when there's a another function which is specifically for reading legacy raw files.

anndata/_core/anndata.py
ivirshup: It makes sense to me that calling copy would make this go into memory. What doesn't make sense to me is that copy is even being called here, and somehow it only effects `.raw`?

Also, how would that even effect these lines, since these are all under a `if not self.isbacked`?

ivirshup: > If I do self.raw = AnnData(**raw) as you suggested in the init function, it fails in this way.

Hmm. I don't recall recommending that for the `AnnData.__init__` function, but definitely see how it wouldn't work for backed mode. I'm also unsure of what `**raw` would contain if it were coming from `copy`.

---------------------------

I just tried reverting this block to:

```python
return AnnData(
    X=X,
    obs=self.obs.copy(),
    var=self.var.copy(),
    # deepcopy on DictView does not work and is unnecessary
    # as uns was copied already before
    uns=self.uns.copy()
    if isinstance(self.uns, DictView)
    else deepcopy(self.uns),
    obsm=self.obsm.copy(),
    varm=self.varm.copy(),
    obsp=self.obsp.copy(),
    varp=self.varp.copy(),
    raw=None if self._raw is None else self._raw.copy(),
    layers=self.layers.copy(),
    dtype=dtype,
)
```

on the current commit, and do not get an error.

setup.py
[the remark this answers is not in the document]
ivirshup: While it seems to work fine for me, `asv` just totally fails to install this now. I'm pretty unclear on why.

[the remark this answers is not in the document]
ivirshup: But it does work if I revert this to `packages=find_packages(),`, and add `anndata/core/__init__.py`

anndata/readwrite/h5ad.py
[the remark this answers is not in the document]
flying-sheep: OK, fixing!

setup.py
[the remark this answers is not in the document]
flying-sheep: I got asv to run in the meantime using the virtualenv backend. It works fine for me. What’s your error?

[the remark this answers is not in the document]
ivirshup: I'm using the conda backend. It says it installs fine, but just fails to import anndata:

```python
File "/usr/local/lib/python3.7/site-packages/asv/benchmark.py", line 1315, in <module>
main()
File "/usr/local/lib/python3.7/site-packages/asv/benchmark.py", line 1308, in main
commands[mode](args)
File "/usr/local/lib/python3.7/site-packages/asv/benchmark.py", line 1044, in main_run
benchmark_dir, benchmark_id, extra_params=extra_params)
File "/usr/local/lib/python3.7/site-packages/asv/benchmark.py", line 939, in get_benchmark_from_name
module = import_module(modname)

File "/Users/isaac/github/anndata-benchmarks/.asv/env/8cb4204c4acf92f37ebde6042bb3eb94/lib/python3.7/importlib/__init__.py", line 127, in import_module
return _bootstrap._gcd_import(name[level:], package, level)
File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
File "<frozen importlib._bootstrap>", line 983, in _find_and_load
File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
File "<frozen importlib._bootstrap>", line 677, in _load_unlocked

File "<frozen importlib._bootstrap_external>", line 728, in exec_module
File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
File "/Users/isaac/github/anndata-benchmarks/benchmarks/readwrite.py", line 32, in <module>
    from .utils import (
File "/Users/isaac/github/anndata-benchmarks/benchmarks/utils.py", line 10, in <module>
    from anndata import AnnData
ImportError: cannot import name 'AnnData' from 'anndata' (unknown location)
```

anndata/_core/anndata.py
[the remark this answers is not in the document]
flying-sheep: Also nothing makes sense. `Raw.__init__` has `if X is not None` initialize from arguments `else` take var and varm from the `adata` argument.

So there’s no way to initialize `Raw` without X but with custom `var` and `varm`. What.

anndata/readwrite/h5ad.py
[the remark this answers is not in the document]
flying-sheep: I don’t know what you mean.

anndata/_core/anndata.py
flying-sheep: “This block” being `AnnData.copy`?

setup.py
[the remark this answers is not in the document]
flying-sheep: Pretty sure that means you have another version of anndata somewhere else on your PYTHONPATH. If you fix that it should work.

anndata/_core/anndata.py
ivirshup: Here is a version of copy that works and uses the initializer:

```python
    def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
        """Full copy, optionally on disk."""
        if not self.isbacked:
            if self.is_view:
                # TODO: How do I unambiguously check if this is a copy?
                # Subsetting this way means we don't have to have a view type
                # defined for the matrix, which is needed for some of the
                # current distributed backend.
                X = _subset(self._adata_ref.X, (self._oidx, self._vidx)).copy()
            else:
                X = self.X.copy()
            # TODO: Figure out what case this is:
            if X is not None:
                dtype = X.dtype
                if X.shape != self.shape:
                    X = X.reshape(self.shape)
            else:
                dtype = "float32"
            return AnnData(
                X=X,
                obs=self.obs.copy(),
                var=self.var.copy(),
                # deepcopy on DictView does not work and is unnecessary
                # as uns was copied already before
                uns=self.uns.copy()
                if isinstance(self.uns, DictView)
                else deepcopy(self.uns),
                obsm=self.obsm.copy(),
                varm=self.varm.copy(),
                obsp=self.obsp.copy(),
                varp=self.varp.copy(),
                raw=None if self._raw is None else self._raw.copy(),
                layers=self.layers.copy(),
                dtype=dtype,
            )
        else:
            from ..readwrite import read_h5ad

            if filename is None:
                raise ValueError(
                    'To copy an AnnData object in backed mode, '
                    'pass a filename: `.copy(filename=\'myfilename.h5ad\')`.'
                )
            mode = self.file._filemode
            self.write(filename)
            return read_h5ad(filename, backed=mode)
```

flying-sheep: gotcha!

anndata/tests/test_views.py
```diff
@@ -393,9 +393,17 @@ def test_view_of_view(matrix_type, subset_func, subset_func2):
...
+    try:
+        view_of_actual_copy = (
+            adata[:, var_s1].copy()[obs_s1, :].copy()[:, var_s2].copy()
+        )
+    except ValueError as e:
+        # TODO: fix
+        if "while it should have had (15,)" in str(e):
+            pytest.xfail("Raw slicing is broken")
+        else:
+            raise
```
flying-sheep: OK, everything looks good now except for this. If you can confirm that this problem also exists (but expresses itself differently) in master, I’ll merge this PR.

setup.py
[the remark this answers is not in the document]
ivirshup: There isn't. However, anndata (at a011d7efaa6) that get's installed by asv only has these files:

<details>
<summary> anndata installed by asv files </summary>


```bash
# The version installed by asv
$ cd ~/github/anndata-benchmarks/.asv/env/8cb4204c4acf92f37ebde6042bb3eb94/lib/python3.7/site-packages/anndata
# Listing the none `__pycache__` files
$ find . | grep -v "__pycache__" 
.
./readwrite
./readwrite/write.py
./readwrite/__init__.py
./readwrite/utils.py
./readwrite/zarr.py
./readwrite/h5ad.py
./readwrite/read.py
./_core
./_core/index.py
./_core/alignedmapping.py
./_core/raw.py
./_core/anndata.py
./_core/sparse_dataset.py
./_core/file_backing.py
./_core/views.py
./tests
./tests/test_utils.py
./tests/conftest.py
./tests/test_obsmvarm.py
./tests/test_inplace_subset.py
./tests/test_readwrite.py
./tests/test_layers.py
./tests/test_deprecations.py
./tests/test_io_conversion.py
./tests/test_views.py
./tests/test_raw.py
./tests/test_hdf5_backing.py
./tests/test_repr.py
./tests/test_helpers.py
./tests/test_base.py
./tests/helpers.py
./tests/test_uns.py
./tests/test_backed_sparse.py
./tests/test_obspvarp.py
```

</details>

anndata/tests/test_views.py
ivirshup: Slicing `adata.raw[:, idx]` is broken on master if raw has varm, but this shouldn't trigger that.

What do you mean "expresses itself differently"?

anndata/_core/anndata.py
```diff
@@ -1422,13 +1422,10 @@ def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
                 varm=self.varm.copy(),
                 obsp=self.obsp.copy(),
                 varp=self.varp.copy(),
+                raw=self.raw,  # copies automatically
```
ivirshup: I don't think the constructor should copy automatically. That leads to redundant memory usage which isn't always needed. This was part of the rework of IO and why reading is faster and uses much less memory now.

anndata/_core/anndata.py
```diff
@@ -492,22 +492,15 @@ def _init_as_actual(
...
-        if raw and isinstance(raw, cabc.Mapping):
-            raw = Raw(self, **raw)
-        elif raw:  # is a Raw from another AnnData
-            raw = raw.copy()
-            raw._adata = self
-        self._raw = raw if raw else None
+            if {"raw", "raw.X"} & set(self.file):
+                raw = dict(X=None, **raw)
+        if not raw:
+            del self.raw
```
ivirshup: What case is this?

anndata/_core/raw.py
```diff
@@ -25,14 +25,17 @@ def __init__(
...
+            raise ValueError("Cannot specify X if adata is backed")
+        self._X = X if adata.isbacked or X is not None else adata.X.copy()
+        self._var = (
+            adata.var.copy()
+            if var is None
+            else _gen_dataframe(var, self.X.shape[1], ['var_names'])
+        )
+        self._varm = AxisArrays(
+            self, 1, adata.varm.copy() if varm is None else varm
+        )
```
ivirshup: Can you explain your reasoning for this solution?

I don't think raw should ever copy `var`, `varm` or `X` from it's parent. I think it might be okay to break this at the moment, since the previous version made no sense, and backed mode didn't work.

anndata/tests/test_views.py
flying-sheep: I built in more checks, so I think there’s just nonsensical behavior happening on master that doesn’t trigger an error.

setup.py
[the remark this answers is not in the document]
flying-sheep: Ah, I see. Fixed!

anndata/_core/anndata.py
flying-sheep: It only copies when passed another AnnData’s Raw. So what you say is that it should only do a flat copy?

anndata/_core/anndata.py
flying-sheep: raw is None or raw == {} → `self._raw = None`, which I’m doing through the `fdel` API.

anndata/_core/raw.py
flying-sheep: I didn’t change that. `self._var = adata.var.copy()` is on master.

anndata/_core/anndata.py
ivirshup: Yes

anndata/tests/test_views.py
ivirshup: What are the relevant checks, and what is the wrong behavior?

flying-sheep: Ah, I think now I got it. check the newest state.

anndata/_core/anndata.py
ivirshup: But how would this be triggered? What is the case where raw has already been defined on the object, and this is how we remove it?

anndata/readwrite/utils.py
```diff
@@ -195,3 +196,19 @@ def func_wrapper(elem, key, val, *args, **kwargs):
             ) from e
 
     return func_wrapper
+
+
+def _read_legacy_raw_into(
+    f, raw, read_df, read_attr, *, attrs=("X", "var", "varm")
+):
```
ivirshup: Why does this modify a dict instead of creating one?

anndata/_core/anndata.py
```diff
@@ -1422,7 +1421,7 @@ def copy(self, filename: Optional[PathLike] = None) -> 'AnnData':
                 varm=self.varm.copy(),
                 obsp=self.obsp.copy(),
                 varp=self.varp.copy(),
-                raw=self.raw,  # copies automatically
+                raw=self.raw.copy(),
```
ivirshup: I think this should be:

```python
raw=self.raw.copy() if self.raw is not None else None,
```

ivirshup:
If my [suggested change](https://github.com/theislab/anndata/pull/268#discussion_r357956634) gets the tests to pass, I think this should be good to go. Does that sound right to you?

anndata/_core/anndata.py
flying-sheep: It’s just `self._raw = None`.

anndata/_core/anndata.py
flying-sheep: Oh. We never test `.copy()`? I’ll add a test

anndata/readwrite/utils.py
flying-sheep: temporary, internal, needs to make sure the dict it receives is empty anyway.

But sure, it’s still possible to do this pythonically

ivirshup:
Did you add a backwards compatibility module for `readwrite`?

flying-sheep:
Sure!

PR #270, opened 2019-12-09
Initial update of release notes

ivirshup:
Starting release notes updates for 0.7. There's a fair amount of backlog.


flying-sheep:
Great, thank you. Yes, there’s a lot of backlog.

Could you please do versions like in theislab/scanpy#951? Nobody calls “1.4.1” a “Post-1.4” release so we shouldn’t either. I always found it confusing that there isn’t a “Version 1.4” heading with 1.4.0, 1.4.1, … below.

Issue #271, opened 2019-12-09
Should AnnData define __eq__?


chris-rands:
While unit testing in #265, I've found that `AnnData` does not define `__eq__` which leads to some surprises (I think because `Nan` is not equal to `Nan`), e.g.

```py
from anndata import AnnData
import numpy as np

# dense data
X1 = np.array([[1, 2, 3], [4, 6, 7]])
X2 = np.array([[1, 2, 3], [4, 5, 6]])
X3 = np.array([[1, 2, 3], [4, 5, 6]])

adata1 = AnnData(
        X1,
        dict(obs_names=['s1', 's2'], anno1=['c1', 'c2']),
        dict(var_names=['a', 'b', 'c'], annoA=[0, 1, 2]),
        layers=dict(Xs=X1),
    )

adata2 = AnnData(
        X2,
        dict(obs_names=['s3', 's4'], anno1=['c3', 'c4']),
        dict(var_names=['d', 'c', 'b'], annoA=[0, 1, 2]),
        layers={'Xs': X2},
    )

adata3 = AnnData(
        X3,
        dict(obs_names=['s1', 's2'], anno2=['d3', 'd4']),
        dict(var_names=['d', 'c', 'b'], annoB=[0, 1, 2]),
        layers=dict(Xs=X3),
    )

a1 = AnnData.concatenate(adata1, adata2, adata3)
a2 = AnnData.concatenate(adata1, adata2, adata3)

print(a1 == a2)
print(a1.obs == a2.obs)

print(a1.obs,'\n', a2.obs)
``` 

Outputs:
```py
False
      anno1  anno2  batch
s1-0   True  False   True
s2-0   True  False   True
s3-1   True  False   True
s4-1   True  False   True
s1-2  False   True   True
s2-2  False   True   True
     anno1 anno2 batch
s1-0    c1   NaN     0
s2-0    c2   NaN     0
s3-1    c3   NaN     1
s4-1    c4   NaN     1
s1-2   NaN    d3     2
s2-2   NaN    d4     2 
      anno1 anno2 batch
s1-0    c1   NaN     0
s2-0    c2   NaN     0
s3-1    c3   NaN     1
s4-1    c4   NaN     1
s1-2   NaN    d3     2
s2-2   NaN    d4     2
```

[...]

PR #272, opened 2019-12-10
Anndata.__eq__ now raises NotImplementedError

chris-rands:
see issue #271

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -547,10 +547,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def __eq__(self, other):
         """Equality testing"""
         raise NotImplementedError(
-            (
-                'Equality comparisons are not supported for AnnData objects, '
-                'instead compare the desired attributes.'
-            )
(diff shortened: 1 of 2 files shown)


ivirshup:
Hmm. I'm wondering if this should be:

```python
def __eq__(self, other):
    if isinstance(other, AnnData):
        raise NotImplementedError(...)
    return NotImplemented
```

So that other objects can decide if they are or aren't equal to the anndata object. For instance, I think it's fair to say `"foo" != adata`. @flying-sheep may know more about conventions for this. Then the error message should be something about "Equality has not been defined for anndata objects".

Also could you split the error message onto multiple lines, so it isn't over 80 characters?

chris-rands:
I would say that anyone doing `"foo" != adata` is making a mistake pointing at a bug in their code, I can't imagine a valid use case for such a comparison, but I'm happy to implement whatever you decide is best of course

The Codacy issues appear false positives to me, since these statements are valid unit tests

flying-sheep:
I agree, let’s do it like this for now.

PR #284, opened 2019-12-27
Concatenate obsm

Koncopd:
Adds obsm concatenation in adata.concatenate.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1646,6 +1646,16 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         for key in shared_layers:
             layers[key] = []
 
+        # create obsm dict that contains obsm keys shared among all AnnDatas
+        obsm = OrderedDict()
+        shared_obsm_keys = [
+            key
+            for key in all_adatas[0].obsm
+            if all([key in ad.obsm for ad in all_adatas])
+        ]
+        for key in shared_obsm_keys:
+            obsm[key] = []
+
         # check whether tries to do “outer join” and layers is non_empty.
         if join == "outer" and len(shared_layers) > 0:
             logger.info(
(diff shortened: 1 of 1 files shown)


anndata/_core/anndata.py
```diff
@@ -1701,6 +1711,11 @@ def concatenate(
                 for key in shared_layers:
                     layers[key].append(ad[:, vars_intersect].layers[key])
 
+            # obsm
+            if join == "inner":
```
falexwolf: Great, but I don't think it should be restricted to merely "inner" as this only relates to the variables. I'll fix that.

gokceneraslan:
@Koncopd @falexwolf 

There is an issue with the obsm concatenation. When we run `sc.tl.diffmap` with different anndata objects, concatenate them and run sc.pp.neighbors, we get the following exception. The reason is that  `obsm`s are concatenated but `.uns['diffmap_evals']` are not available.

```pytb
KeyError                                  Traceback (most recent call last)
<timed exec> in <module>

/opt/conda/lib/python3.7/site-packages/scanpy/neighbors/__init__.py in neighbors(adata, n_neighbors, n_pcs, use_rep, knn, random_state, method, metric, metric_kwds, copy)
    104     if adata.isview:  # we shouldn't need this here...
    105         adata._init_as_actual(adata.copy())
--> 106     neighbors = Neighbors(adata)
    107     neighbors.compute_neighbors(
    108         n_neighbors=n_neighbors, knn=knn, n_pcs=n_pcs, use_rep=use_rep,

/opt/conda/lib/python3.7/site-packages/scanpy/neighbors/__init__.py in __init__(self, adata, n_dcs)
    527                 self._number_connected_components = self._connected_components[0]
    528         if 'X_diffmap' in adata.obsm_keys():
--> 529             self._eigen_values = _backwards_compat_get_full_eval(adata)
    530             self._eigen_basis = _backwards_compat_get_full_X_diffmap(adata)
    531             if n_dcs is not None:

/opt/conda/lib/python3.7/site-packages/scanpy/neighbors/__init__.py in _backwards_compat_get_full_eval(adata)
    395         return np.r_[1, adata.uns['diffmap_evals']]
    396     else:
--> 397         return adata.uns['diffmap_evals']
    398 
    399 

KeyError: 'diffmap_evals'
```

Doesn't it make more sense to make `obsm` concatenation False by default, by the way? Should concatenating `obsm` be the default behaviour?

PR #289, opened 2020-01-10
Centralize encoding versions and test them

### LINKED ISSUE #287: Breaking change in 0.7 : a bug?
When I save a `.h5ad` file using `adata.write()` in `anndata` `0.7rc2`, and then try to open it using `scanpy.read_h5ad()` with `anndata` `0.6.22.post1`, I get an error.

flying-sheep:
Since we have versions we should also check them. Fixes #287

### DIFF
--- a/anndata/_io/zarr.py
+++ b/anndata/_io/zarr.py
@@ -24,6 +24,7 @@ from .utils import (
     report_write_key_on_error,
     write_attribute,
     _read_legacy_raw,
+    EncodingVersions,
 )
 from . import WriteWarning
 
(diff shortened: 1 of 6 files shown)


anndata/_io/zarr.py
```diff
@@ -191,7 +192,7 @@ def write_csr(f, key, value, dataset_kwargs=MappingProxyType({})):
 def write_csc(f, key, value, dataset_kwargs=MappingProxyType({})):
     group = f.create_group(key)
     group.attrs["encoding-type"] = "csc_matrix"
-    group.attrs["enocding-version"] = "0.1.0"
+    group.attrs["encoding-version"] = EncodingVersions.csc_matrix.value
```
flying-sheep: Hey @falexwolf @ivirshup: This is why I’m so anal about reusing code.

flying-sheep: What do you mean? They aren’t in sync.

PR #291, opened 2020-01-10
Avoid int-named obs_names and var_names.

### LINKED ISSUE #286: Writing error
I upgraded to `scanpy==1.4.5`, which resulted in `anndata==0.7rc1 `.

Now, when I write an anndata object using

```
dataset_111.write("data/data.h5ad", compression="lzf")
```
I receive the errors below.

Downgrading to `0.6.22rc1` and `h5py==2.10.0` fixed the issue, though I'm not sure it's due to the h5py change.

```
AttributeError: 'numpy.int64' object has no attribute 'encode'

Above error raised while writing key 0 of <class 'h5py._hl.group.Group'> from /.

Above error raised while writing key 'obs' of <class 'h5py._hl.files.File'> from /.
```

flying-sheep:
obs_names & var_names are instances of pandas.Index, and can have a .name.
We try to write that name to disk as zarr/hdf5. At least HDF5 doesn’t support ints as names.

Fixes #286

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -749,13 +749,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @obs.setter
     def obs(self, value: pd.DataFrame):
-        if not isinstance(value, pd.DataFrame):
-            raise ValueError("Can only assign pd.DataFrame.")
-        if len(value) != self.n_obs:
-            raise ValueError("Length does not match.")
-        utils.warn_no_string_index(value.index)
-        if self.is_view:
-            self._init_as_actual(self.copy())
+        self._check_dim_df_and_actualize(value, self.n_obs)
         self._obs = value
 
     @obs.deleter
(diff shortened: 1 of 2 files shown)


anndata/_core/anndata.py
```diff
@@ -914,8 +910,14 @@ def var_names(self) -> pd.Index:
...
         utils.warn_no_string_index(names)
         self._var.index = names
-        if not self._var.index.is_unique:
-            utils.warn_names_duplicates("var")
+        self._normalize_index(self._var.index, "var")
+
+    @staticmethod
+    def _normalize_index(idx: pd.Index, idx_name: str):
+        if isinstance(idx.name, (int, np.integer)):
+            idx.name = None
```
ivirshup: @flying-sheep, I'm not sure I like this. This lets the expression `adata.var_names = foo` mutate the RHS.

flying-sheep: No, because I pass `self._var.index`/`self._obs.index`

flying-sheep: No, because I pass `self._var.index`/`self._obs.index`

ivirshup: I think that still mutates the original, since no copy has been made. Here's an example:

```python
import pandas as pd
a = pd.DataFrame({"a": [1,2,3,4]})
b = a.copy()
b.index = a.index
b.index.name = "b" 
print(a.index.name)
"b"
```

flying-sheep: Oh, so no copy is created implicitly when setting a DataFrame’s index, whoops. Will fix.

flying-sheep: Fixed in #294

PR #294, opened 2020-01-13
Error trying to set obs/var_names to an index with no-string name

flying-sheep:
See https://github.com/theislab/anndata/pull/291#discussion_r365782837

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -749,7 +749,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @obs.setter
     def obs(self, value: pd.DataFrame):
-        self._check_dim_df_and_actualize(value, self.n_obs)
+        self._check_dim_df_and_actualize(value, self.n_obs, "obs")
         self._obs = value
 
     @obs.deleter
@@ -763,19 +763,19 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @var.setter
     def var(self, value: pd.DataFrame):
-        self._check_dim_df_and_actualize(value, self.n_vars)
+        self._check_dim_df_and_actualize(value, self.n_vars, "var")
         self._var = value
 
     @var.deleter
     def var(self):
         self.var = pd.DataFrame(index=self.var_names)
 
-    def _check_dim_df_and_actualize(self, value: pd.DataFrame, n: int):
+    def _check_dim_df_and_actualize(self, value: pd.DataFrame, n: int, attr: str):
         if not isinstance(value, pd.DataFrame):
             raise ValueError("Can only assign pd.DataFrame.")
         if len(value) != n:
             raise ValueError("Length does not match.")
-        utils.warn_no_string_index(value.index)
+        utils.check_index_is_strings(value.index, attr)
         if self.is_view:
             self._init_as_actual(self.copy())
 
@@ -897,8 +897,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @obs_names.setter
     def obs_names(self, names: Sequence[str]):
-        utils.warn_no_string_index(names)
-        self._obs.index = names.copy() if isinstance(names, pd.Index) else names
+        utils.check_index_is_strings(names, "obs")
+        self.obs.index = names
         self._normalize_index(self._obs.index, "obs")
 
     @property
@@ -908,8 +908,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @var_names.setter
     def var_names(self, names: Sequence[str]):
-        utils.warn_no_string_index(names)
-        self._var.index = names.copy() if isinstance(names, pd.Index) else names
+        utils.check_index_is_strings(names, "var")
+        self.var.index = names
         self._normalize_index(self._var.index, "var")
 
     @staticmethod
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -898,7 +898,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     @obs_names.setter
     def obs_names(self, names: Sequence[str]):
         utils.warn_no_string_index(names)
-        self._obs.index = names
+        self._obs.index = names.copy() if isinstance(names, pd.Index) else names
         self._normalize_index(self._obs.index, "obs")
 
     @property
@@ -909,7 +909,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     @var_names.setter
     def var_names(self, names: Sequence[str]):
         utils.warn_no_string_index(names)
-        self._var.index = names
+        self._var.index = names.copy() if isinstance(names, pd.Index) else names
         self._normalize_index(self._var.index, "var")
 
     @staticmethod
--- a/anndata/tests/test_base.py
+++ b/anndata/tests/test_base.py
@@ -117,19 +117,31 @@ def test_names():
 @pytest.mark.parametrize(
     "names,after",
     [
-        (["a", "b"], None),
-        (pd.Series(["AAD", "CCA"], name="barcodes"), "barcodes"),
-        (pd.Series(["x", "y"], name=0), None),
-        (pd.Index(["x", "y"], name=0), None),
+        pytest.param(["a", "b"], None, id="list"),
+        pytest.param(
+            pd.Series(["AAD", "CCA"], name="barcodes"), "barcodes", id="Series-str"
+        ),
+        pytest.param(pd.Series(["x", "y"], name=0), None, id="Series-int"),
     ],
 )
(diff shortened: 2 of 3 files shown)


flying-sheep:
You’re right. Since I propose the same in the #297, this would be more consistent and less implicit magic.

anndata/utils.py
```diff
@@ -82,11 +82,16 @@ def warn_names_duplicates(attr: str):
...
+            f"AnnData expects .{attr}.index.name to be a string, "
+            f"but you passed a name of type {type(names.name).__name__!r}"
+        )
     if not isinstance(names[0], str):
         logger.warning(
-            f"AnnData expects string indices for some functionality, "
-            f"but your first two indices are: {names[:2]}. "
+            f"AnnData expects .{attr}.index to contain strings, "
+            f"but your first indices are: {names[:2]}, …"
         )
```
flying-sheep: Can we make this an error by now?

ivirshup: Don't we automatically convert anyways?

anndata/tests/test_base.py
```diff
@@ -114,13 +114,34 @@ def test_names():
...
-    assert adata.obs_names.name is None
-    adata.obs_names = pd.Series(["AAD", "CCA"], name="barcodes")
-    assert adata.obs_names.name == "barcodes"
-    adata.obs_names = pd.Series(["x", "y"], name=0)  # happens, but we strip it
-    assert adata.obs_names.name is None
+    assert getattr(adata, attr).name is None
+    with pytest.raises(ValueError, match=fr"AnnData expects \.{attr[:3]}\.index\.name"):
+        setattr(adata, attr, pd.Index(["x", "y"], name=0))
+    assert getattr(adata, attr).tolist() != ["x", "y"]
+    assert getattr(adata, attr).tolist() == getattr(adata_dense, attr).tolist()
```
ivirshup: I'm not sure if I'm following this right, but does `pd.Index(["x", "y"], name=0)` error, while `pd.Series(["x", "y"], name=0)` doesn't? My assumption would be that they are equivalent.

anndata/utils.py
ivirshup: Probably worth checking if it'd break anything.

anndata/_core/anndata.py
```diff
@@ -763,19 +763,19 @@ def var(self) -> pd.DataFrame:
...
-    def _check_dim_df_and_actualize(self, value: pd.DataFrame, n: int):
+    def _check_dim_df_and_actualize(self, value: pd.DataFrame, n: int, attr: str):
         if not isinstance(value, pd.DataFrame):
             raise ValueError("Can only assign pd.DataFrame.")
         if len(value) != n:
             raise ValueError("Length does not match.")
-        utils.warn_no_string_index(value.index)
+        utils.check_index_is_strings(value.index, attr)
         if self.is_view:
             self._init_as_actual(self.copy())
```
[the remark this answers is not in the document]
flying-sheep: What method? `_init_as_actual`? I’m not touching it in this PR.

anndata/tests/test_base.py
flying-sheep: When you do `some_df.index = some_series`, then a new Index is created with the same name and values, so setting `.some_df.index.name = something` afterwards won’t mutate `some_series`.

anndata/utils.py
flying-sheep: `test_set_obs` currently throws that warning, which it shouldn’t. We could do the following to handle this case:

```py
    if not isinstance(names, pd.RangeIndex) and not isinstance(names[0], str):
        raise ValueError(...)
```

I don’t know where we handle a `RangeIndex` being written to disk or used in slicing, but I guess it has always worked and is supposed to work.

ivirshup: Looks like you can write it but you can't read it. I think it would be fine for this to be an error, since it generally doesn't work. I don't know if you could even use index using numeric indices.

I just wonder if it'll break something in someone else's code.

anndata/_core/anndata.py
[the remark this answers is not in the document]
ivirshup: Oops: meant to comment on the range of code. `_check_dim_df_and_actualize` is the function in question.

anndata/utils.py
flying-sheep: Hmm, actutally, slicing a RangeIndex will create an Int64Index…

```py
>>> import pandas as pd
>>> pd.DataFrame(dict(a=[1,2,3])).iloc[[0,2], :].index                                                                                                                                                                                   
Out[6]: Int64Index([0, 2], dtype='int64')
```

anndata/_core/anndata.py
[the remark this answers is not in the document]
flying-sheep: Well, it doesn’t actually *set* the attribute though.

anndata/utils.py
flying-sheep: Hmm, actutally, slicing a RangeIndex will create an Int64Index…

```py
>>> import pandas as pd
>>> pd.DataFrame(dict(a=[1,2,3])).iloc[[0,2], :].index                                                                                                                                                                                   
Out[6]: Int64Index([0, 2], dtype='int64')
```

flying-sheep:
> If adata.var_names = new_var_names is set, shouldn't that update the index of a dataframe in varm? Should we try to enforce that these are literally the same index? I.e assert adata.var_names is adata.varm["df"].index?

I’d say this is out of scope for this PR, but we should do it. I’ll file an issue.

> Could you make sure there are tests which check that failing to set obs, var, var_names, or obs_names will not modify the original object?

Way ahead of you :smile: https://github.com/theislab/anndata/blob/6dbcc493cf7f6d3cd34e01180fb0539c2bbd454f/anndata/tests/test_base.py#L141-L144

anndata/_core/anndata.py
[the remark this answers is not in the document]
ivirshup: Is there another use case?

If not, I think just setting it in this function would make sense. I'd also add the logic for making sure the var names change. I can do this a PR for checking the indices of `varm`, `obsm` and `raw.obs_names`.

anndata/tests/test_base.py
ivirshup: Ah, that makes sense.

anndata/utils.py
ivirshup: I think it will only keep being a range index if you index with `slice`s.

```python
pd.Series(np.ones(10)).iloc[range(4)].index
# RangeIndex(start=0, stop=4, step=1)
```

Interestingly, while you can index a RangeIndex with a range, it results in a Int64Index:

```python
pd.Series(np.ones(10)).iloc[range(4)].index 
# Int64Index([0, 1, 2, 3], dtype='int64')
```

I'm still not quite sure why python has `slice`s and `range`s as separate objects.

ivirshup:
> but we should do it.

Definitely, I think this counts as a bug:

<details>
<summary> example </summary>

```python
import pandas as pd

a = gen_adata((10, 10))
a.obs_names = pd.Index([f"Cell0{i}" for i in range(10)])
a[:4].copy()
# ---------------------------------------------------------------------------
# ValueError                                Traceback (most recent call last)
# <ipython-input-7-8b58a4f5968f> in <module>
# ----> 1 a[:4].copy()

# ~/github/anndata/anndata/_core/anndata.py in copy(self, filename)
#    1413                 if isinstance(self.uns, DictView)
#    1414                 else deepcopy(self.uns),
# -> 1415                 obsm=self.obsm.copy(),
#    1416                 varm=self.varm.copy(),
#    1417                 obsp=self.obsp.copy(),

# ~/github/anndata/anndata/_core/aligned_mapping.py in copy(self)
#      84         d = self._actual_class(self.parent, self._axis)
#      85         for k, v in self.items():
# ---> 86             d[k] = v.copy()
#      87         return d
#      88 

# ~/github/anndata/anndata/_core/aligned_mapping.py in __setitem__(self, key, value)
#     148 
#     149     def __setitem__(self, key: str, value: V):
# --> 150         value = self._validate_value(value, key)
#     151         self._data[key] = value
#     152 

# ~/github/anndata/anndata/_core/aligned_mapping.py in _validate_value(self, val, key)
#     210             # Could probably also re-order index if it’s contained
#     211             raise ValueError(
# --> 212                 f"value.index does not match parent’s axis {self.axes[0]} names"
#     213             )
#     214         return super()._validate_value(val, key)

# ValueError: value.index does not match parent’s axis 0 names
```

</details>

> Way ahead of you 😄

I was thinking of being a bit more thorough on this. Like using the `assert_equals` function and checking that if it were a view it'd stay a view if setting failed.

anndata/_core/anndata.py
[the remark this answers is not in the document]
flying-sheep: There isn’t, it’s just to deduplicate code between obs.setter and var.setter.

And sure, go ahead! :smiley:

PR #297, opened 2020-01-15
Varseries

jeffhsu3:
Closes theislab/anndata#296

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -91,13 +91,18 @@ def _check_2d_shape(X):
 
 
 def _gen_dataframe(anno, length, index_names):
-    if isinstance(anno, pd.DataFrame) or isinstance(anno, pd.Series):
+    if isinstance(anno, pd.Series):
+        if anno.name is None:
+            raise ValueError(
+                    "anno needs to be a pd.DataFrame(value) or be a"
+                    "pd.Series(value) with a name"
+            )
+        anno = pd.DataFrame(anno)
+    if isinstance(anno, pd.DataFrame):
         anno = anno.copy()
         if not is_string_dtype(anno.index):
             logger.warning("Transforming to str index.")
             anno.index = anno.index.astype(str)
-        if isinstance(anno, pd.Series):
-            anno = pd.DataFrame(anno)
         return anno
     if anno is None or len(anno) == 0:
         _anno = pd.DataFrame(index=RangeIndex(0, length, name=None).astype(str))
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -91,11 +91,13 @@ def _check_2d_shape(X):
 
 
 def _gen_dataframe(anno, length, index_names):
-    if isinstance(anno, pd.DataFrame):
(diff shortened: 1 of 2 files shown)


anndata/_core/anndata.py
```diff
@@ -91,11 +91,13 @@ def _check_2d_shape(X):
...
 
 def _gen_dataframe(anno, length, index_names):
-    if isinstance(anno, pd.DataFrame):
+    if isinstance(anno, pd.DataFrame) or isinstance(anno, pd.Series):
         anno = anno.copy()
         if not is_string_dtype(anno.index):
             logger.warning("Transforming to str index.")
             anno.index = anno.index.astype(str)
+        if isinstance(anno, pd.Series):
+            anno = pd.DataFrame(anno)
```
flying-sheep: Hmm, why not

```py
if isinstance(anno, pd.Series):
    anno = pd.DataFrame(anno)
if isinstance(anno, pd.DataFrame):
    anno = anno.copy()
    ...
```

anndata/_core/anndata.py
```diff
@@ -91,11 +91,13 @@ def _check_2d_shape(X):
 
 
 def _gen_dataframe(anno, length, index_names):
-    if isinstance(anno, pd.DataFrame):
+    if isinstance(anno, pd.DataFrame) or isinstance(anno, pd.Series):
```
flying-sheep: If you want to keep this instead of going for the idea in the other comment, do
```suggestion
    if isinstance(anno, (pd.DataFrame, pd.Series)):
```

anndata/tests/test_base.py
```diff
@@ -82,6 +82,15 @@ def test_create_from_df_with_obs_and_var():
         AnnData(df, var=var.reset_index())
 
 
+def test_create_from_df_with_obs_and_var_series():
+    df = pd.DataFrame(np.ones((3, 2)), index=["a", "b", "c"], columns=["A", "B"])
+    obs = pd.DataFrame(np.ones((3, 1)), index=df.index, columns=["C"])
+    var = pd.Series(np.ones(2), index=df.columns, name="D")
+    ad = AnnData(df, obs=obs, var=var)
+
+    assert var.tolist() == ad.var.D.tolist()
```
flying-sheep: Maybe also check that the original series isn’t modified when you modify ad.var

ivirshup:
What happens if the series has no name?

flying-sheep:
… right, the data frame gets 0 as column index.

I think we should just throw an error in that case, there’s no good guess to make.

flying-sheep:
Hey @jeffhsu3, still interested in seeing this through?

jeffhsu3:
I guess making it raise an error if a pd.Series is passed makes more sense.  What do you guys think?

anndata/_core/anndata.py
jeffhsu3: Yeah makes sense.

flying-sheep:
People certainly can call `series.to_frame()` themselves.

The question is if it’s likely that they do `ad.var = some_series` and meant to do `ad.var['thing'] = some_series`, and we need to protect them from it, or not.

ivirshup:
I think it should just be an error

PR #300, opened 2020-01-21
[WIP] Fix up setting of indices

### LINKED ISSUE #299: Unify obs_names/var_names with DataFrame indices in obsm/varm
`obs_names` is an alias for `.obs.index`, but `.obsm` can contain pd.DataFrames with any index. We should think about how to tackle this

ivirshup:
Eventually fixes #299

Needs tests for:

* Failure to modify should leave object in original state, in particular I'm not sure about the case where indexes are set
* Setting dimension index should update indices of dataframes in obsm/ varm
* Setting obs_names should be reflected in `raw`

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -742,6 +742,51 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         """Number of variables/features."""
         return self._n_vars
 
+    def _set_dim_df(self, value: pd.DataFrame, attr: str):
+        if not isinstance(value, pd.DataFrame):
+            raise ValueError(f"Can only assign pd.DataFrame to {attr}.")
+        value_idx = self._prep_dim_index(value.index, attr)
+        if self.is_view:
+            self._init_as_actual(self.copy())
+        setattr(self, f"_{attr}", value)
+        self._set_dim_index(value_idx, attr)
+
+    def _prep_dim_index(self, value, attr: str) -> pd.Index:
+        """Prepares index to be uses as obs_names or var_names for AnnData object.AssertionError
+
+        If a pd.Index is passed, this will use a reference, otherwise a new index object is created.
+        """
+        if self.shape[attr == "var"] != len(value):
+            raise ValueError(
+                f"Length of passed value for {attr}_names is {len(value)}, but this AnnData has shape: {self.shape}"
+            )
+        if isinstance(value, pd.Index) and not isinstance(
+            value.name, (str, type(None))
+        ):
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


ivirshup:
Coming across a strange issue:

```python
import joblib

a = gen_adata((10, 10))
hash1 = joblib.hash(a)
a[:, :]  # Create a view, I don't think it matters how
hash2 = joblib.hash(a)
assert hash1 != hash2
```

This only seems to occur if a dataframe in a has categorical values. It's the dataframe itself that's different, but pandas testing functions don't pick this up:

```python
import pickle
import pandas as pd

a = gen_adata((10, 10))
df = a.obs
a[:, :]
assert pickle.dumps(df) != pickle.dumps(df.copy())
assert joblib.hash(df) != joblib.hash(df.copy())
pd.testing.assert_frame_equal(df, df.copy())
```

My current thinking is that pandas doesn't pick this up because it's not actually important. It would be nice to use hashing as a "gold standard" comparison (which I'd been doing before). I haven't been able to replicate this with just pandas yet, so it looks like we're doing something strange. I would like to figure out what's going on here, but I'll work around for now.

flying-sheep:
Does joblib offer to create a tree of objects it uses for hashing? Would be very helpful to debug where a comparison fails.

ivirshup:
I don't think so, but I thought something like that might be available for pandas.

ivirshup:
@flying-sheep, I think this is good to go if you want to take a look. It covers a few not-great bugs (we weren't carrying over `obsp` or `varp` when calling `_init_as_actual` on a view, for example).

Currently this makes the 10x reader functions fail on scanpy, but they pass indexes with [integer](https://github.com/theislab/scanpy/blob/40271f20209f0b15306543711a69512bf2c2ac77/scanpy/readwrite.py#L333) [names](https://github.com/theislab/scanpy/blob/40271f20209f0b15306543711a69512bf2c2ac77/scanpy/readwrite.py#L365). It's an easy fix though, and I'll add it to https://github.com/theislab/scanpy/pull/989

flying-sheep:
> It covers a few not-great bugs (we weren't carrying over obsp or varp when calling _init_as_actual on a view, for example).

Aw, I should have waited a few hours longer to release 0.7. I thought this PR would only cover non-essential things…

ivirshup:
Oh! Sorry, I should have been more clear. I thought the obs_names, obsm bug was bad enough to hold on.

ivirshup:
Great!

Also turns out I don't need to update theislab/scanpy#989, it can merge as is. My editor an shell were out of sync about what branch I was on, and my changes in that PR already fix the issue.

PR #310, opened 2020-01-29
[WIP] Proof of concept for not converting objects to strings

### LINKED ISSUE #115: Consider changing detections of strings from `is_string_dtype`
@flying-sheep 

Just found this issue for pandas https://github.com/pandas-dev/pandas/issues/21653, and looked at the code for `is_string_dtype` and think it probably shouldn't be relied on. Anything whose `dtype.kind` is `O` (anything pandas considers an `object`) returns `True`. For example:

```python
>>> is_string_dtype(pd.Series([(0,), (1,)]))
True
```

It's probably not reasonable to assume any column that has objects should be converted to categorical.

Side note: I'm unaware of the rationale for converting all string columns to categorical. Could someone point me towards the reasoning behind this?

ivirshup:
Potentially fixes #115. This just preserves objects during `strings_to_categoricals`.

This will change behaviour for cases where there we mixed types in a `Series` (like: `["abc", 0.6]`), but I'm not sure if that's a common case. It will still convert if there are missing values.

This still converts things to strings for writing, where it should probably error or at least warn.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -19,7 +19,7 @@ import numpy as np
 from numpy import ma
 import pandas as pd
 from pandas.core.index import RangeIndex
-from pandas.api.types import is_string_dtype, is_categorical
+from pandas.api.types import is_string_dtype, is_categorical, infer_dtype
 from scipy import sparse
 from scipy.sparse import issparse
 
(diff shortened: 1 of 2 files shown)


LuckyMD:
I remember having issues when having `object` as dtype in `obs`. I'm not sure what it was... but I remember there was a long back and forth with me being clueless as to why sth didn't work. Not very helpful atm... but just a heads up.

ilan-gold:
@flying-sheep what is the context of this arising again?

flying-sheep:
Sorry for the brevity: The idea is to revisit in general how we approach autoconversion: to strings, to categoricals, …

ilan-gold:
> Sorry for the brevity: The idea is to revisit in general how we approach autoconversion: to strings, to categoricals, …

Ok! What's the idea here? Make everything settings?

PR #313, opened 2020-01-31
Makes DataFrameView.drop actualize its parent AnnData

flying-sheep:
Required for theislab/scanpy#1015

### DIFF
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -1,5 +1,6 @@
+from contextlib import contextmanager
 from copy import deepcopy
-from functools import reduce, singledispatch
+from functools import reduce, singledispatch, wraps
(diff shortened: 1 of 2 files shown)


anndata/_core/views.py
```diff
@@ -100,6 +108,13 @@ class DictView(_ViewMixin, dict):
 class DataFrameView(_ViewMixin, pd.DataFrame):
     _metadata = ["_view_args"]
 
+    @wraps(pd.DataFrame.drop)
+    def drop(self, *args, inplace: bool = False, **kw):
+        if not inplace:
+            return super().drop(*args, **kw)
```
flying-sheep: The test for this case still breaks … no idea how to elegantly and correctly do this.

PR #314, opened 2020-02-05
Export ImplicitModificationWarning

### LINKED ISSUE #296: Initializing AnnData with var as a pd.Series
Initializing AnnData with var=pd.Series results in the following AnnData.vars:

|  | gene_symbol |
| ------------- | ------------- |
| 0  | NaN  |
| 1  | NaN  |

Coercing the series into a DataFrame fixes it, but throwing an error or altering _gen_dataframe to handle a Series might be helpful.

### LINKED ISSUE #311: Is logging "Transforming to str index." necessary?
I do a lot of moving back and forth between DataFrames and AnnData objects, and the "Transforming to str index." warning can get annoying, especially if I'm running things in a loop:

image

In these cases, I'm creating a DataFrame using something like:

```python
data = np.random.normal(size=(1000,100))
data = pd.DataFrame(data)
... # code that expects data as a DataFrame
adata = sc.AnnData(data)
```

Does this need to be logged in a manner that is printed by default? Is there an easy way to disable this?

It looks like this is the offending line
https://github.com/theislab/anndata/blob/master/anndata/_core/anndata.py#L97

flying-sheep:
Fixes #311, fixes #296, closes #297

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -6,7 +6,7 @@ import collections.abc as cabc
 from collections import OrderedDict
 from copy import deepcopy
 from enum import Enum
-from functools import reduce
+from functools import reduce, singledispatch
 from pathlib import Path
 from os import PathLike
 from typing import Any, Union, Optional  # Meta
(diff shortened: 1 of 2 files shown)


anndata/_core/anndata.py
```diff
@@ -90,32 +90,47 @@ def _check_2d_shape(X):
...
 
+@singledispatch
 def _gen_dataframe(anno, length, index_names):
-    if isinstance(anno, pd.DataFrame):
-        anno = anno.copy()
-        if not is_string_dtype(anno.index):
-            logger.warning("Transforming to str index.")
-            anno.index = anno.index.astype(str)
-        return anno
     if anno is None or len(anno) == 0:
```
flying-sheep: could have registered for `type(None)` here, but since this case also applies to zero length sequences, I left it this way.

PR #315, opened 2020-02-06
Fix version type

flying-sheep:
Discovered in chanzuckerberg/cellxgene#1132


bkmartinjr:
Suggest a test case be added, to prevent future regression.

flying-sheep:
If we change this in the future, it will be to another build system entirely.

PR #316, opened 2020-02-07
Fix index setting when initializing AnnData with DataFrames

flying-sheep:
The logic was dodgy, see e.g. the added test case which would throw an exception, as we don’t check if the `obs` parameter is a DataFrame before accessing its `index`:

https://github.com/theislab/anndata/blob/fa7d1bf4c3b51aa310b6cf62410443aa8bf4f521/anndata/_core/anndata.py#L424-L428

Now we take the indices from any DataFrame provided. When two indices are present for an axis, they will be compared (In both steps, “RangeIndex” counts as “no index”). It works like this:

1. If obs/var is a thing that could gain an index in conversion later (DataFrame, dict) and X.index/columns isn’t an (probably autogenerated) RangeIndex we mark it for comparison.

   https://github.com/theislab/anndata/blob/b4008ed747d64d7cc5bdfe6ddaf7f40543cd81e2/anndata/_core/anndata.py#L429-L432

2. After conversion, we set the index if there was none before, or compare the two possibly clashing indices for the same axis:

   https://github.com/theislab/anndata/blob/b4008ed747d64d7cc5bdfe6ddaf7f40543cd81e2/anndata/_core/anndata.py#L491-L496

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -395,6 +395,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         # various ways of initializing the data
         # ----------------------------------------------------------------------
 
+        # If X is a data frame, we store its indices for verification
+        x_indices = []
+
         # init from file
         if filename is not None:
             self.file = AnnDataFileManager(self, filename, filemode)
@@ -422,16 +425,15 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
             # init from DataFrame
             elif isinstance(X, pd.DataFrame):
+                # to verify index matching, we wait until obs and var are DataFrames
                 if obs is None:
                     obs = pd.DataFrame(index=X.index)
-                else:
-                    if not X.index.equals(obs.index):
-                        raise ValueError("Index of obs must match index of X.")
+                elif not isinstance(X.index, pd.RangeIndex):
+                    x_indices.append(("obs", "index", X.index))
                 if var is None:
                     var = pd.DataFrame(index=X.columns)
-                else:
-                    if not X.columns.equals(var.index):
-                        raise ValueError("Index of var must match columns of X.")
+                elif not isinstance(X.columns, pd.RangeIndex):
+                    x_indices.append(("var", "columns", X.columns))
                 X = ensure_df_homogeneous(X, "X")
 
         # ----------------------------------------------------------------------
@@ -482,10 +484,16 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                         raise ValueError("`shape` is inconsistent with `var`")
 
         # annotations
-        self._obs = _gen_dataframe(
-            obs, self._n_obs, ["obs_names", "row_names", "smp_names"]
-        )
+        self._obs = _gen_dataframe(obs, self._n_obs, ["obs_names", "row_names"],)
         self._var = _gen_dataframe(var, self._n_vars, ["var_names", "col_names"])
+
+        # now we can verify if indices match!
+        for attr_name, x_name, idx in x_indices:
+            attr = getattr(self, attr_name)
+            if isinstance(attr.index, pd.RangeIndex):
+                attr.index = idx
+            elif not idx.equals(attr.index):
+                raise ValueError(f"Index of {attr_name} must match {x_name} of X.")
 
         # unstructured annotations
         self._uns = uns or OrderedDict()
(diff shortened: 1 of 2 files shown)


anndata/_core/anndata.py
```diff
@@ -395,6 +395,9 @@ def _init_as_actual(
         # various ways of initializing the data
         # ----------------------------------------------------------------------
 
+        # If X is a data frame, we store its indices for verification
+        x_indices = []
```
ivirshup: Is there a way to write this without having to remember a variable 90 lines later?

anndata/_core/anndata.py
```diff
@@ -422,16 +425,15 @@ def _init_as_actual(
 
             # init from DataFrame
             elif isinstance(X, pd.DataFrame):
+                # to verify index matching, we wait until obs and var are DataFrames
                 if obs is None:
                     obs = pd.DataFrame(index=X.index)
-                else:
-                    if not X.index.equals(obs.index):
-                        raise ValueError("Index of obs must match index of X.")
+                elif not isinstance(X.index, pd.RangeIndex):
```
ivirshup: Why just `RangeIndex`? What about an integer valued index?

anndata/_core/anndata.py
flying-sheep: I think a different way of doing this could be that we convert obs and var first, and then do our checks, but I’m happy enough with this as it is.

anndata/_core/anndata.py
flying-sheep: Do we have a consistent behavior here over the whole codebase? If not we could of course introduce a function to use in these cases like `is_ignorable_index`.

Since that would touch many more places though, I feel like it’s a thing for another PR.

flying-sheep:
> I'm also not sure if I'm getting what the full behaviour is. What you're doing is figuring out what `obs_names`/ `var_names` should be if `X` was a dataframe. Could you succinctly describe what the desired behaviour is?

Both `X` and `obs`/`var` can be sources for `obs_names`/`var_names`. If for a given axis only one of them provides an explicitly set* index, we use that. Else we check if they match.

*the various RangeIndex checks we do are a heuristic to check if it was implicitly created or explicitly set.

> I'm also wondering what happens if obs and X have indices which aren't strings, but they agree. What if they don't?

What kind of non-string indices do we support anyway?

ivirshup:
Sorry for not responding to this earlier! I thought I had, but looks like I didn't hit comment.

At the moment I'd lean towards not defining any special behaviour for range/ integer valued indexes beyond throwing an error. This is going off [this discussion over on scanpy](https://github.com/theislab/scanpy/issues/1030#issuecomment-584238178) where interest in having integer valued indices was brought up again.

I think it's more feasible to support this now than it was last time it got brought up, since the machinery is a more organized now.

flying-sheep:
can you make a suggestion on what to change in this PR then to keep it working while not treating RangeIndex specially?

I think doing this should work, as it’s pretty clear where the names come from while both DataFrames *technically* have an index for the same axis.

```py
my_data = load_some_df()
assert my_data.index[:3].tolist() == ['kitten', 'hamster', 'fluffy_rabbit']

ad = AnnData(my_data, obs=pd.DataFrame(dict(cuteness=[4, 3, 5, ...]))
assert ad.obs_names[:3].tolist() == ['kitten', 'hamster', 'fluffy_rabbit']
```

ivirshup:
I think some safe behavior would be something like:

```python
if isinstance(X, pd.DataFrame):
    if isinstance(obs, pd.DataFrame):
        assert (obs.index == X.index).all()
    obs_names = X.index
    if isinstance(var, pd.DataFrame):
        assert (var.index == X.columns).all()
    var_names = X.columns
else:
    if isinstance(obs, pd.DataFrame):
        obs_names = obs.index
    else:
        obs_names = pd.RangeIndex(X.shape[0])
    if isinstance(var, pd.DataFrame):
        var_names = var.index
    else:
        var_names = pd.RangeIndex(var.shape[1])
```

Later, separately, the conversion to string happens.

I haven't checked if this conflicts with existing behavior, but hopefully this logic is mostly compatible.

flying-sheep:
nope, won’t work. The following would succeed while silently overwriting obs’ index:

```py
ad = AnnData(pd.DataFrame(..., index=list('abc')), obs=dict(obs_names=list('def')))
```

That’s the part where your code goes wrong:

```py
if isinstance(obs, pd.DataFrame):
    assert (obs.index == X.index).all()
```

This can’t only trigger if a DataFrame is passed as `obs`, it has be checked after `obs` has been converted to a DataFrame.

flying-sheep:
Well, this is still better than what we have and I need it. Let’s tackle remaining concerns in the sprint

PR #320, opened 2020-02-10
[WIP] Add copy argument to obs/var_vector

ivirshup:
Per some benchmarking results in https://github.com/theislab/anndata/issues/317#issuecomment-583963717, it looks like we can speed up getting a vector out of a dense array with the `obs_vector` methods. Since it always returned a copy before, I figure this was worth a keyword argument.

This gives a 1000x speed up when the array which values are being taken from is dense.

Needs tests.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1327,7 +1327,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         else:
             return self.X
 
-    def obs_vector(self, k: str, *, layer: Optional[str] = None) -> np.ndarray:
+    def obs_vector(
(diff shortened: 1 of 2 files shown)


flying-sheep:
Why do we have 4 almost identical blocks of code and why do they differ?

ivirshup:
For raw, it's a different set of attributes. Then there's `obs` and `var`, and at the time I was less comfortable with doing `getattr` instead of `obj.attr`.

flying-sheep:
The last 6 lines each are identical, yet people trying to modify it have to be aware they have to modify 4 places. Please change. Case in point: Why does `var_vector` grab stuff from `.var` but `obs_vector` doesn’t grab stuff from `.parent.obs` or so? Please make it all work identically.

Issue #326, opened 2020-02-20
request: reliable means to interrogate sparsity


bkmartinjr:
It would be very helpful if the API had a standard way to query the anndata X sparsity, and sparse format, independent of the backed mode.

In other words, a safe / reliable / performant equivalent equivalent of the following on memory-backed adata:
* `scipy.sparse.issparse(adata.X)` -- bool indicates if is sparse
* `adata.X.get_format()` -- return scipy sparse format, eg, 'csr'

Is there a way to do this now?  Or a hack to work around the lack of API today?

One possible hack-around relies on the weak test for the format_str internal to spmatrix (these are overly wordy for clarity):

```
from scipy import sparse

def issparse(adata):
    if adata.isbacked:
        return hasattr(adata.X, 'format_str')
    else:
        return sparse.issparse(adata.X) 

def get_sparse_format(adata):
    if issparse(adata):
        if adata.isbacked:
            return getattr(adata.X, 'format_str', 'und')
        else:
            return adata.X.getformat()
    return False
```

flying-sheep:
If you just want to test sparsity, and not if `X` is some other kind of array-like thing, you’d try `scipy.sparse.issparse(adata.X)` or `isinstance(anndata.X, scipy.sparse.spmatrix)`. For the sparse format, you’d use `scipy.sparse.isspmatrix_csr` or `isinstance(..., ...csr_matrix)`, right?

This should actually work for `BackedSparseMatrix`, as all its subclasses actually inherit from the `spmatrix` subclasses:

https://github.com/theislab/anndata/blob/4440b90ff3dff213b4c512478e21426cf210368d/anndata/_core/sparse_dataset.py#L117

https://github.com/theislab/anndata/blob/4440b90ff3dff213b4c512478e21426cf210368d/anndata/_core/sparse_dataset.py#L143

bkmartinjr:
> If you just had plain scipy.sparse classes and wanted to test sparsity, you’d use scipy.sparse.issparse(adata.X) or isinstance(anndata.X, scipy.sparse.spmatrix), right? 

Yes, that would be perfect.

> For the sparse format, you’d use scipy.sparse.isspmatrix_csr or isinstance(..., ...csr_matrix), right?

I would probably use `spmatrix.getformat()`, but that is equivalent to scipy.sparse.isspmatrix_csr(), and both should work if `adata.X` is a subclass of the appropriate spmatrix.

[...]

PR #327, opened 2020-02-22
Export SparseDataset

ivirshup:
Exporting the SparseDataset class since we expose it to users anyways. Re: #326

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -1,5 +1,6 @@
 from ._core.anndata import AnnData, ImplicitModificationWarning
 from ._core.raw import Raw
+from ._core.sparse_dataset import SparseDataset
 from ._io import (
     read_h5ad,
     read_loom,
--- a/anndata/_core/sparse_dataset.py
+++ b/anndata/_core/sparse_dataset.py
@@ -278,12 +278,12 @@ class SparseDataset:
 
     def __getitem__(self, index: Union[Index, Tuple[()]]) -> Union[float, ss.spmatrix]:
         row, col = self._normalize_index(index)
-        mtx = self.to_backed()
+        mtx = self._to_backed()
         return mtx[row, col]
 
     def __setitem__(self, index: Union[Index, Tuple[()]], value):
         row, col = self._normalize_index(index)
-        mock_matrix = self.to_backed()
+        mock_matrix = self._to_backed()
         mock_matrix[row, col] = value
 
     def _normalize_index(
@@ -300,7 +300,7 @@ class SparseDataset:
         # Prep variables
         shape = self.shape
         if isinstance(sparse_matrix, SparseDataset):
-            sparse_matrix = sparse_matrix.to_backed()
+            sparse_matrix = sparse_matrix._to_backed()
 
         # Check input
         if not ss.isspmatrix(sparse_matrix):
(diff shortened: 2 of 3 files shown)


ivirshup:
But`SparseDataset` does have `shape` and `__getitem__`?

flying-sheep:
Right, it does have that, sorry. What I meant: I don’t think exporting `SparseDataset` with its current API and in its current role is wise, as we lock ourselves down to `X` of a backed AnnData having an API that’s both too specific (tied to HDF5) and not powerful enough (isn’t a real matrix/array)

My ideal API would be:

```py
if adata.backing is not None:
    # do backed things
if scipy.sparse.issparse(adata.X):
    # do sparse things
else:
    # do dense things
```

ivirshup:
I don't think `SparseDataset` is tied to hdf5, IIRC it works with a zarr group on the backend too. I'm not sure what you mean by "not powerful enough".

Isn't that API what we have right now?

flying-sheep:
It’s not, as `not issubclass(SparseDataset, spmatrix)`. And in case of a backed dense matrix, we have a straight `h5py.Dataset` as `X`. If we document this as the truth, we’ll always be tied to having a `Dataset` with its full API there. I was thinking more of something like this:

ivirshup:
> It’s not, as not issubclass(SparseDataset, spmatrix)

Ah, thought there was an `elif` in there.

Does subclassing go in reverse order of the arrows?

We definitely should not be subclassing the scipy sparse matrix types at all. This was done before and is why backed sparse arrays were broken for quite a while (pre-`0.7`). They are specialized for in memory data, and even specific memory layout, and we can't expect their methods to work on backed arrays.

On the `h5py.Dataset`, I think that has become the standard convention for backed dense arrays in the ecosystem. It's that api that make working with `zarr` and `h5py` pretty interchangeable. The `h5py.Dataset` and `zarr.Array` also increase `dask` compatibility so I think those are valuable to keep.

ivirshup:
Regardless of what the ideal interface is here, can we agree on keeping the current type interface until `0.8`?

flying-sheep:
Sure, but we can’t export it until we converge on something we want to support for a long time, which isn’t this.

ivirshup:
I think we shouldn't change this before 0.8, as people use it and open issues for it. I also think exporting is pretty light weight in terms of support. There's nothing that prevents us from deprecating it. I would like to have a solution for #326, and I think the simplest solution is:

```python
if isinstance(adata.X, (spmatrix, SparseDataset)):
    # then it's sparse
```

It's not clear to me what you don't like about the current set up. Could you elaborate on that?

* What is lacking from the current implementation? What would you like to be able to do?
* What is the alternative implementation you're proposing? How would this be implemented, and how does usage change?

flying-sheep:
Our docs represent our public, supported interface. Supporting something means we update the the type info to include the newly supported thing.

AnnData.X is documented to have type `Optional[Union[np.ndarray, sparse.spmatrix, ArrayView]]`. Supporting something else there means:

1. The class has to be public
2. We’ll support X to have that class for a long time.

By now we actually don’t really support backed mode, as we lie about `AnnData.X`’s type. To stop lying we need to define the class that can to go there that we want to support.

And that class can’t be `h5py.Dataset`, as `h5py.Dataset` has h5py specific APIs. If we add a class of our design, we have to export and document that class’s APIs.

ivirshup:
Superseded by #765

PR #328, opened 2020-02-22
add comment asking for maintaining the location of the deprecated decorator

falexwolf:
I started using @flying-sheep's awesome decorator in a few other packages. Can we leave it where it is?

Or shall we move it somewhere more sensible?

For instance, we have `utils` be replaced by `_utils` so that it doesn't appear in the auto-suggest anymore? The old story...

### DIFF
--- a/anndata/utils.py
+++ b/anndata/utils.py
@@ -125,12 +125,12 @@ def convert_dictionary_to_structured_array(source: Mapping[str, Sequence[Any]]):
 
     return arr
 
-
+# this decorator is used by external packages - please refrain from simply
+# moving it around and try to ensure backward compat
 def deprecated(new_name: str):
     """\
-    This is a decorator which can be used to mark functions
-    as deprecated. It will result in a warning being emitted
-    when the function is used.
+    This decorator marks functions as deprecated. It will result in a warning
+    being emitted when the function is used.
     """
 
     def decorator(func):


flying-sheep:
Hi Alex! A suggestion so your expectations are aligned with what people do: You should generally never use things that aren’t documented. In most Python projects, “privateness” not indicated by the name starting with `_`, but by the fact that it appears at that location in the docs.

The best idea for something small like this here is IMHO:

1. Ask people to export it to officially document, export and support it.
   If they do that and you don’t want to wait for it to be released you can do the following safely now:

   ```py
   try:
       from new_location import thing
   except ImportError:
       from old_undocumented_location import thing
   ```

2. If they don’t and it’s feasible to not have the thing, you should do

   ```py
   try:
       from undocumented_location import thing
   except ImportError:
       thing = None  # or something that doesn’t do anything but doesn’t break your code
   ```

3. If they don’t and you need the thing, you should copy it to a location where you can depend on it existing. E.g. a new package you maintain if you need it in multiple locations.

I happen to maintain https://pypi.org/project/legacy-api-wrap/

Maybe that’s a good spot for it?

flying-sheep (changes requested):
Hi Alex! A suggestion so your expectations are aligned with what people do: You should generally never use things that aren’t documented. In most Python projects, “privateness” not indicated by the name starting with `_`, but by the fact that it appears at that location in the docs.

And that means that if you use something private, it can and will break without notice, and nobody did something wrong.

The best idea for something small like this here is IMHO (I’m speaking generally, I’m aware we’re AnnData’s maintainers and can control this here):

1. Ask people to export it to officially document, export and support it.
   If they do that and you don’t want to wait for it to be released you can do the following safely now:

   ```py
   try:
       from new_location import thing
   except ImportError:
       from old_undocumented_location import thing
   ```

2. If they won’t export it and it’s feasible to not have the thing, you should add a test to get notified if they changed it, and create a dummy so people’s code depending on your project doesn’t break:

   ```py
   try:
       from undocumented_location import thing
   except ImportError:
       thing = None  # or something else that is quick to define and doesn’t break your code
   ```

3. If they won’t export it and you need the thing, you should copy it to a location where you can depend on it existing. E.g. a new package you maintain if you need it in multiple locations.

I happen to maintain https://pypi.org/project/legacy-api-wrap/

Maybe that’s a good spot for it?

Issue #295, opened 2020-01-13
Concatenate enhancements


chris-rands:
I know that the `AnnData.concatenate()` docstring says "The uns, varm and obsm attributes are ignored." , but I still wondered if there is any light-weight way to do this?

My use case here is I'm running bbknn on several datasets separately and then I want to merge the neighbourhood graphs after. Thanks!

flying-sheep:
Well, we’ll migrate to that to `.obsp` very soon, and therefore we might* concatenate `.obsp` but definitely not stuff in `.uns`. We might merge `.uns` instead…

*might because graphs should probably be recalculated anyway: If you concatenate graphs, you get a graphs with disconnected parts, represented by a block diagonal matrix:

![](https://i.stack.imgur.com/3E6A8.jpg)

ivirshup:
I think we'll allow merging `obsp`, since there are cases where it's valid, but we should make it easy to exclude items from merges.

`uns` is more difficult. I think it'll be difficult to tell which things are okay to merge, so I think I'd like to just drop it all/ have the user select a canonical version. Colors may be an exception to this.

flying-sheep:
I think we could offer

```py
def concatenate(
	...,
	merge_uns: Literal[
		'if_clean', # the default, throws an error on key collision, else just merges
		'always', # ignores key collision, just overwrite earlier with later items
	] = None):
	...
```

Maybe you can come up with more?

flying-sheep:
Hi @giovp, would such an option in concatenate work for your spatial use case?

Similarly: You need to concat `obsm['X_spatial']`, so we should just generally concatenate `.obsm` entries

ivirshup:
@flying-sheep, re - your last comment. I'm thinking of just providing the same options as xarray.

@giovp, could you go over how you store images in the Anndata object currently?

flying-sheep:
Link please

ivirshup:
Here's the [narrative docs](http://xarray.pydata.org/en/stable/combining.html), but the mainly the options for [`xr.concat`](http://xarray.pydata.org/en/stable/generated/xarray.concat.html) are what I'm thinking of.

flying-sheep:
Ha, that’s exactly the same idea as I had independently. Seems like indication that it’s a good one.

giovp:
@ivirshup , images and scale factors are stored in Anndata as:
* `adata.uns['images']`
* `adata.uns['scalefactors']`. 

[...]

PR #329, opened 2020-02-26
multi-tissue functionality

giovp:
This implementation modifies the `concatenate` function and allows to work with multiple tissues in the same anndata object.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1894,6 +1894,76 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 UserWarning,
             )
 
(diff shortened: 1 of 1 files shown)


giovp:
Connected to theislab/scanpy#1073 and https://github.com/giovp/spatial-scripts/blob/master/multiple_slices_functionality.ipynb

LuckyMD:
Do you want to update `requirements.txt` already for latest anndata in the PR from scanpy?

flying-sheep:
Sorry @giovp, but we need to make this generic or add something in scanpy. And I think making it generic will benefit us more. Please let’s continue discussion in #295

PR #335, opened 2020-03-06
Fix ArrayView and numpy subclassing

michalk8:
Hi,
I have a `numpy.ndarray` subclass with some column annotations (also `ndarrays`), including a
view-like object. Problem is that taking view of `adata` object first converts this to `np.ndarray`,
effectively removing my subclass (`arr = np.asarray(input_array)`). This PR fixes this such that only non-arrays are converted to arrays, rest is left the same.

### DIFF
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -69,9 +69,8 @@ class ArrayView(_SetItemMixin, np.ndarray):
         input_array: Sequence[Any],
         view_args: Tuple["anndata.AnnData", str, Tuple[str, ...]] = None,
     ):
-        if not isinstance(input_array, np.ndarray):
-            input_array = np.array(input_array)
-        arr = input_array.view(cls)
+        arr = np.asanyarray(input_array).view(cls)
+
         if view_args is not None:
             view_args = ViewArgs(*view_args)
         arr._view_args = view_args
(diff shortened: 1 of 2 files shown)


anndata/_core/views.py
```diff
@@ -69,7 +69,9 @@ def __new__(
         input_array: Sequence[Any],
         view_args: Tuple["anndata.AnnData", str, Tuple[str, ...]] = None,
     ):
-        arr = np.asarray(input_array).view(cls)
+        if not isinstance(input_array, np.ndarray):
+            input_array = np.array(input_array)
+        arr = input_array.view(cls)
```
ivirshup:
```suggestion
        arr = np.asanyarray(input_array).view(cls)
```

Could this work instead? I think it should if your class is literally a np.ndarray subclass.

michalk8: Thanks for the suggestion! I've added a test and it works.

michalk8:
Any news on this? I wanted to fix the code-quality check (just for the test - either get overriding builtin when using `type` as an argument or parameters differ...).
@ivirshup would be great if you think everything is on point and then merge this.

ivirshup:
Sorry, just been busy! I think this looks good, thanks!

PR #337, opened 2020-03-16
Use `obsp` instead of `uns["neighbors"]`

### LINKED ISSUE #336: Switch to using obsp by default
Integrate obsp within Scanpy neighbors in a backwards compat way:

**NOTE** This will probably require making scanpy 1.5.0 dependent on the anndata release this is fixed in

* Upon reading an anndata object from disk, move .uns[`neighbors`] to `.obsp`
    * Issue warning upon that
    * Point out that the old 
* How do we make the transition easy for users?
    * Do lookups in uns redirect with a warning!
    * And have .uns[‘neighbors’] do a lookup in `.obsp`

# Current questions:

* Are these equivalent? Or should removing the value from uns just remove the value the deprecated dict? Would I be able to track that?

```python
del adata.uns["neighbors"]["connectivities"]
del adata.obsp["connectivities"]
```

* Slicing sparse matrices in uns has to be deprecated, not removed, right? Since the whole thing was searched. I think we can only really move the ones we know about.
* Should the deprecated keys show up in things like `uns["neighbors"].keys()`? If they show up when the value hasn't been placed there, I feel like this is sending mixed messages.

ivirshup:
Fixes #336

This PR switches us from using sparse matrices in uns to using obsp.

## Currently implemented

* OverloadedDict, a dict which allows special keys for elements that have moved.
* Deprecation warnings for accessing, adding, or modifying "neighbors.connectivities" or "neighbors.distances"
* Soooo many warnings

## TODO:

- [x] Move elements when they're read in
- [x] What to do if the user places a sparse matrix in uns?
  - [x] If it's not in neighbors, not much we can do about it. We still search uns, but only copy if there is a sparse matrix to subset. We also warn for this.
- [x] How do we refer to these elements later? I.e. does `neighbors` have keys for related `obsp` entries? Yes, as implemented by @Koncopd there are `"connectivities_key"` and `"distances_key"` entries in `.uns["neighbors"]` now.
- [x] Get all tests passing
- [x] Simplify (especially mapping)


anndata/tests/test_base.py
```diff
@@ -298,6 +298,7 @@ def test_slicing_strings():
         adata[["A", "B", "not_in_obs"], :]
 
 
+# TODO: Decide what this should do
```
ivirshup: Should these matrices be moved from .uns["neighbors"] at read time, or at instantiation time?

ivirshup:
@Koncopd, I think this is ready for a look over. I think you should be able to start working on neighbors stuff on top of this. We should talk design of that. I'd like to make sure we end up with the ability to have multiple "neighbors" representations.

ivirshup:
Ran benchmarks on this ([results](https://gist.github.com/ivirshup/5eb48627ffac40a8ea3fba1692117fd0)). Mostly positive (which is expected since we don't need to copy `uns` for every subset anymore). There are a few cases which got worse though. I'm not sure I'm seeing a pattern, other than the regressions mostly being memory based.

anndata/_core/anndata.py
```diff
@@ -854,7 +865,24 @@ def var_names(self, names: Sequence[str]):
     @property
     def uns(self) -> MutableMapping:
         """Unstructured annotation (ordered dictionary)."""
-        return self._uns
+        uns = self._uns
+        # Special case for old behaviour of neighbors
+        if "neighbors" in uns:
```
Koncopd: So this is created every time when` .uns` is accessed. Is there any way to create it once and then just return on every access?

ivirshup: Two reasons for this:

There's a reference to the anndata object. This makes sure it's pointing to the right one. If it was always the same we could have trouble with statements like:

```python
a: AnnData

b = a[:100].copy()
b.uns["neighbors"]["connectivities"]  # This should get b.obsp["connectivities"]
```

Second is that copy and deep copy are already weird for DictViews. This simplifies it a bit.

Koncopd: Ok, thats reasonable.

ivirshup:
This was failing some scanpy tests due to the case:

```python
adata.uns["neighbors"] = dict(connectivities=...)
```

I don't think my first implementation was going to be able to deal with this case. I've played around with an alternative implementation to work with this. It's basically descriptors for keys, so allowing each key for a dict to be overloaded with `getter`, `setter`, and `deleter` methods.

This class wraps an underlying dict, with overloaded keys registered and providing methods. These methods default to basic dict behaviour. Overall, I think this new implementation is much more elegant, and I'm actually quite proud of how it's turned out. All tests now pass for `scanpy` and `anndata` with this implementation.

ivirshup:
@Koncopd, I think this is ready to go. Would you like to review it?

Koncopd:
Hi, this looks great.

PR #338, opened 2020-03-17
add concatenate incomplete obsm

giovp:
As discussed in #295 and theislab/scanpy#1105, this PR addresses the need for concatenating obsm even if incomplete (not shared). This is useful in the case of concatenating anndata-visium object with anndata-chromium object. 

I've introduced an additional parameter for `anndata.concatenate` that resembles the one for `uns` proposed by @flying-sheep in #295 
```python
merge_obs: Literal["if_clean", "always"] = None,
```

What is concatenated if the `obsm` is missing? 
As discussed in theislab/scanpy#1105 w/ @ivirshup , an array filled with zeros is concatenated. I decided to leave out the `fill_value` argument as originally proposed. This is because I though priority would be to preserve the sparse format if this is available, and indeed if that's the case an empty `scipy.csr_matrix` is added (not sure if this is the best way to instantiate an empty sparse matrix with scipy though).

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1481,6 +1481,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         join: str = "inner",
         batch_key: str = "batch",
         batch_categories: Sequence[Any] = None,
+        merge_obs: Literal["if_clean", "always"] = None,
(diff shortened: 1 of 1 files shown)


giovp:
Hi @ivirshup , do you mind having a look at this? It's related to #343 . Thank you!

PR #342, opened 2020-03-19
Ref paths

flying-sheep:
Continuation of theislab/scanpy-tutorials#15 in the proper repo.

This is already pretty usable, I’ll extend and refine it. API:

- [x] `RefPath(attr: str, *path: str|int)` creates and validates a path
- [x] `RefPath.parse(str|(str|int)s|RefPath)` does too, allowing shortcuts like `"o/foo"` → `("obs", "foo")`
- [x] `RefPath.get_vector(adata, alias_col)` fetches a 1D `np.ndarray`
- [x] `AnnData.get_vector(*path: str|RefPathLike, dim: "obs"|"var", layer: str, …)` allows fallback. E.g. `ad.get_vector('foo', dim='obs')` finds foo in `X` or `obs` while `ad.get_vector('X_pca', 0)` finds it in `obsm`. Or so.
- [x] `AnnData.get_df(paths: (str|RefPathLike)s, dim: "obs"|"var", *, layer: str, …)` multiplexed `AnnData.get_vector`. Returns a data frame with unique column names. Colname builder resolves collisions: `"obsm/X_pca/0"` becomes `"X_pca1"` while `"obsm/protein/CD11b_TotalSeqB"` becomes `"CD11b_TotalSeqB"`

Docs:

- [x] [`RefPath`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.RefPath.html)
- [x] [`AnnData.get_vector`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.AnnData.get_vector.html)
- [x] [`AnnData.get_df`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.AnnData.get_df.html)
- [x] [`AnnData.resolve_path`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.AnnData.resolve_path.html)

Checklist:

- [x] `RefPath.get_vector` supports paths as shorthand `"X/Actb"` or long `("obs", "A/B")`
- [x] add RefPath.dim which is e.g. useful in plotting to check if all RefPaths will return matching stuff
- handle numeric key situation.

  - [x] specifically for varp/obsp
  - [ ] generically for varm/obsm: **But how**? always converting strings to ints when possible is bad since DataFrame ambiguities, but `"obsm/X_pca/0"` is useful if X_pca is an array.

- `get_vector` and `get_df` methods on AnnData that allow leaving things out there when not ambiguous

  - [ ] unambiguous subpaths or keys should resolve (e.g. `("X_pca", 0)` → `("obsm", "X_pca", 0)`), use the usual `layer`, … params as fallback
  - [x] implement `get_df`
    - [x] allow multipaths: `[("obsm", "X_pca", [0, 1])]` would expand to `[("obsm", "X_pca", 0), ("obsm", "X_pca", 1)]`
    - [ ] allow globs: **How**? just allow `"X/var/*"` or so? How to specify in tuples?
      - options: “Could we use either fspath or regex matching?”
    - [ ] maybe allow specifying colnames instead of deriving from paths (maybe by passing `Mapping[str, RefPathLike]`?)
  - [x] implement `get_vector`

- add more tests:

  - [ ] get_df
  - [x] parsing
  - [x] multipath splitting
  - [ ] alias_col
  - [ ] AnnData method

### DIFF
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -8,7 +8,7 @@ import pandas as pd
 from scipy.sparse import spmatrix, issparse
 
 from ..compat import Literal
-from . import anndata
+from . import anndata, raw
 
 
 Index1D = Union[slice, int, str, np.int64, np.ndarray]
@@ -141,35 +141,46 @@ def make_slice(idx, dimidx, n=2):
     return tuple(mut)
 
 
-def get_vector(
-    adata: "anndata.AnnData",
-    k: str,
-    coldim: Literal["obs", "var"],
-    idxdim: Literal["obs", "var"],
-    layer: Optional[str] = None,
-):
+def find_vector(
+    adata: Union["anndata.AnnData", "raw.Raw"], k: str, idxdim: Literal["obs", "var"]
+) -> bool:
+    coldim = "obs" if idxdim == "var" else "var"
     # adata could be self if Raw and AnnData shared a parent
     col = getattr(adata, coldim).columns
     idx = getattr(adata, f"{idxdim}_names")
-
     in_col = k in col
     in_idx = k in idx
-
-    if (in_col + in_idx) == 2:
+    if in_col and in_idx:
         raise ValueError(
             f"Key {k} could be found in both .{idxdim}_names and .{coldim}.columns"
         )
-    elif (in_col + in_idx) == 0:
+    elif not in_col and not in_idx:
         raise KeyError(
             f"Could not find key {k} in .{idxdim}_names or .{coldim}.columns."
         )
-    elif in_col:
+    return in_col
+
+
+def get_vector(
+    adata: Union["anndata.AnnData", "raw.Raw"],
+    k: str,
+    idxdim: Literal["obs", "var"],
+    layer: Optional[str] = None,
+):
+    in_col = find_vector(adata, k, idxdim)
+    if in_col:
+        coldim = "obs" if idxdim == "var" else "var"
         return getattr(adata, coldim)[k].values
-    elif in_idx:
+    else:
         return get_x_vector(adata, idxdim, k, layer)
 
 
-def get_x_vector(adata, idxdim: Literal["obs", "var"], k: str, layer: str = None):
+def get_x_vector(
+    adata: Union["anndata.AnnData", "raw.Raw"],
+    idxdim: Literal["obs", "var"],
+    k: str,
+    layer: str = None,
+):
     selected_dim = ("obs", "var").index(idxdim)
     idx = adata._normalize_indices(make_slice(k, selected_dim))
     a = adata._get_X(layer=layer)[idx]
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -1,11 +1,13 @@
 import collections.abc as cabc
 from functools import singledispatch
 from itertools import repeat
-from typing import Union, Sequence, Optional, Tuple, Literal, TYPE_CHECKING
+from typing import Union, Sequence, Optional, Tuple, TYPE_CHECKING
 
 import numpy as np
 import pandas as pd
 from scipy.sparse import spmatrix, issparse
+
+from ..compat import Literal
 
 if TYPE_CHECKING:
     from .anndata import AnnData
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -1,11 +1,14 @@
 import collections.abc as cabc
 from functools import singledispatch
 from itertools import repeat
-from typing import Union, Sequence, Optional, Tuple
+from typing import Union, Sequence, Optional, Tuple, Literal, TYPE_CHECKING
 
 import numpy as np
 import pandas as pd
 from scipy.sparse import spmatrix, issparse
+
+if TYPE_CHECKING:
+    from .anndata import AnnData
 
 
 Index1D = Union[slice, int, str, np.int64, np.ndarray]
@@ -138,9 +141,14 @@ def make_slice(idx, dimidx, n=2):
     return tuple(mut)
 
 
-def get_vector(adata, k, coldim, idxdim, layer=None):
+def get_vector(
+    adata: "AnnData",
+    k: str,
+    coldim: Literal["obs", "var"],
+    idxdim: Literal["obs", "var"],
+    layer: Optional[str] = None,
+):
     # adata could be self if Raw and AnnData shared a parent
-    dims = ("obs", "var")
     col = getattr(adata, coldim).columns
     idx = getattr(adata, f"{idxdim}_names")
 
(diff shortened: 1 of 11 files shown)


ivirshup:
This looks really awesome!

A few comments and questions

> allow globs: How? just allow "X/var/*" or so? How to specify in tuples?

Could we use either `fspath` or regex matching?

Also, what does `"X/var"` mean?

---------------------

Couple questions about the current scope:

What about getting things from `var` aligned arrays? How about getting rows from `obsm` arrays

anndata/_core/index.py
```diff
@@ -138,9 +141,14 @@ def make_slice(idx, dimidx, n=2):
     return tuple(mut)
 
 
-def get_vector(adata, k, coldim, idxdim, layer=None):
+def get_vector(
+    adata: "anndata.AnnData",
+    k: str,
+    coldim: Literal["obs", "var"],
+    idxdim: Literal["obs", "var"],
+    layer: Optional[str] = None,
```
falexwolf: How about adding an `axis=None` param that defaults to 0 for `obsp` and to ` for `obsm`?

falexwolf: To resolve the ambiguity of indexing into `obsp` (defaulting to rows), and potentially allows indexing into rows of `obsm` (where it defaults to columns).

falexwolf:
Looks great, Phil! 😄

anndata/tests/test_ref_path.py
```diff
@@ -0,0 +1,124 @@
+import pytest
+import numpy as np
+import pandas as pd
+import scipy.sparse as ssp
+
+from anndata import AnnData
+from anndata._core.ref_path import RefPath, split_paths
+
```
ivirshup: Could there also be tests for ambiguities throwing errors here?

anndata/_core/index.py
flying-sheep: To what for `obsm`?

anndata/tests/test_ref_path.py
flying-sheep: you mean ambiguities in `AnnData.get_vector`

flying-sheep:
> Also, what does "X/var" mean?

It’s because they’re unambiguous, so the axis has to be specified:

`("X", "var", "Foo")`means “get the `n_obs` length vector at `np.where(var_names == 'Foo')[0]`”.

> What about getting things from var aligned arrays? How about getting rows from obsm arrays?

That doesn’t fit with the [`dim`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.RefPath.dim.html) attribute, but if you can come up with a good syntax and a solution for that we can do it!

anndata/tests/test_ref_path.py
ivirshup: Yes

flying-sheep:
Can you help me come up with shortcut paths that [`AnnData.resolve_path`](https://icb-anndata.readthedocs-hosted.com/en/refpaths/anndata.AnnData.resolve_path.html) should support?

Do you agree with the following? Would you add more?

```py
@pytest.mark.parametrize("short_path,resolved", [
    # single strings should be found in {obs,var}{.columns,_names}
    # but not in {obs,var}m DataFrame columns (too deep and obscure)
    ("Cell1", ("layers", "X", "obs", "Cell1")),
    ("group", ("obs", "group")),
    # keys with subpaths should be found in layers, {obs,var}{p,m}
    ("X_pca/1", ("obsm", "X_pca", 1)),
    ("unspliced/GeneY", ("layers", "unspliced", "var", "GeneY")),
    # {obs,var}p should default to axis 0
    (("neighbors_distances", "Cell2"), ("obsp", "neighbors_distances", "Cell2", 0)),
    ...?
])
def test_resolve(adata, short_path, resolved):
    assert adata.resolve_path(short_path) == RefPath(*resolved)
```

ivirshup:
I don't think I like the axis being specified the same way a nested element would be. I think it's mixing metaphors, and namespaces. I would prefer an `axis` or `dim` key word argument for this. 

It's also something that won't need to be specified by the user in the contexts we've been discussing, since it's implied. For example, in `obs_df`, all elements should be aligned to `obs_names`.  For `sc.pl.embedding`, similar.

> That doesn’t fit with the dim attribute, but if you can come up with a good syntax and a solution for that we can do it!

Alright. I don't think this is too important for now.

flying-sheep:
The `dim` keyword exists. This is for cases where it’s `None`

ivirshup:
> The `dim` keyword exists. This is for cases where it’s `None`

I don't think it should be the second argument like this. There could maybe be a different delimiter for it, but otherwise I think it's a keyword argument to the RefDim constructor.

ivirshup:
Recap of our call:

* I don't like specifying the dimension as a sub-element. I think it would be good to be able to specify the dimension in a string representation, but I don't think it's critical right now. Can we make it so cases like `"X/var/elem"` don't work?
* The internals you built look interesting, and I'd like a chance to play around with them to find edge cases and see how they can be extended. So we have a chance to do this, and can have a release soon, could we make them private for now? I'd like to make the surface of this minimal until we've had time to play with it more. Also the really important thing this does is allow stuff like `sc.pl.embedding(adata, colors="obsm/X_pca/0")`, which doesn't require `adata.get_vector` to be public right now.

flying-sheep:
I added an “Internal API” page because all those docs don’t write themselves.

anndata/_core/ref_path.py
```diff
@@ -0,0 +1,387 @@
...
+                        raise ValueError(
+                            f"{err_prefix} is not of one of the types {check!r}"
+                        )
+                elif not any(elem == c for c in check):
+                    raise ValueError(f"{err_prefix} is not of one of {check!r}")
+            elif isinstance(check, type):
+                if not isinstance(elem, check):
+                    raise ValueError(f"{err_prefix} is non of type {check}")
+            else:
+                assert False, f"Unhandled check {check!r} for {self}"
```
ivirshup: Could you post some examples of what how these error could come up and what they'd look like? Mostly for my interest.

flying-sheep: It’s just when you try to construct an invalid path (too long, elements of the wrong type or value)

See `test_parse_failures`.

flying-sheep:
`resolve_path` progress:


- [x] finds single strings in `{obs,var}{.columns,_names}`
      but not in {obs,var}m DataFrame columns (too deep and obscure)
      - `"Cell1"` → `RefPath("layers", "X", "obs", "Cell1")`
      - `"group"` → `RefPath("obs", "group")`
- [ ] finds keys with subpaths in `layers`, `{obs,var}{p,m}`
      - `"X_pca/1"` → `("obsm", "X_pca", 1)`
      - `"unspliced/GeneY"` → `("layers", "unspliced", "var", "GeneY")`
- [ ] `{obs,var}p` default to axis 0
      - `("neighbors_distances", "Cell2")` → `RefPath("obsp", "neighbors_distances", "Cell2", 0)`

flying-sheep:
docs TODO:

- remove `adata` from docstring
- examples

PR #343, opened 2020-03-20
Concatenate uns

giovp:
See https://github.com/theislab/anndata/issues/295#issuecomment-600014248 for context.

I've added the `concatenate_uns` function in `utils`, not sure if it is correct.

This follows #338 to support multiple visium anndata.


ivirshup:
Sorry for taking so long on the review!

Could this function be defined so that it checks keys at every level, not just the first two?

giovp:
No problem at all, thank you for all the help w/ all these PRs ! Really appreciate !

Docs are still missing, will work on them as soon as you think such approach makes sense.

giovp:
Hi @ivirshup this would also be ready for review. Thank you!!!

ivirshup:
Superseded by #350

PR #345, opened 2020-03-22
Ensure that names in make_index_unique are unique

mweiden:
Fixes https://github.com/theislab/anndata/issues/344


mweiden:
**Alternative solution**: raise an exception when we encounter a duplicate name and suggest that the user choose a different `join` character.

* Pro: The API would encourage users to make more clearly interpretable indices for their data. The user would know when a suffix indicates duplicates and when it was part of the original string.
* Con: Will (more often) force programs to crash or developers to handle an exception

Thoughts?

mweiden:
@ivirshup after thinking about this a bit, I decided to try a "tweener" solution: issuing a warning. What do you think? Hopefully, this captures the best of both worlds.

anndata/utils.py
```diff
@@ -65,10 +69,25 @@ def make_index_unique(index: pd.Index, join: str = "-"):
     values = index.values
     indices_dup = index.duplicated(keep="first")
     values_dup = values[indices_dup]
+    values_set = set(values)
     counter = defaultdict(lambda: 0)
```
ivirshup: While you're at it, would you mind making this a real `collections.Counter`?

anndata/utils.py
```diff
@@ -65,10 +69,25 @@ def make_index_unique(index: pd.Index, join: str = "-"):
...
+            issue_interpretation_warning = True
+
+    if issue_interpretation_warning:
+        warnings.warn(
+            f"Suffix used ({join}[0-9]+) to deduplicate index values may make index "
+            + "values difficult to interpret. There values with a similar suffixes in "
+            + "the index. Consider using a different join parameter."
+        )
     values[indices_dup] = values_dup
     index = pd.Index(values)
```
ivirshup: Could the original index name also be passed here?

anndata/utils.py
```diff
@@ -65,10 +69,25 @@ def make_index_unique(index: pd.Index, join: str = "-"):
     values = index.values
```
ivirshup: Could you make a copy of the values here? I was testing this and saw:

```python
b = a.copy()
b.obs_names_make_unique()
```

Could modify `a.obs_names`. This is because pandas assumes `pd.Index` objects are immutable, even though their underlying array is mutable. Another way to indirectly modify the index would be:

```python
obs_copy = a.obs.copy()
obs_copy.index.values[0] = "newval"
assert a.obs_names[0] == "newval"
```

mweiden:
@ivirshup Good feedback. Made the changes you requested. For the warning, I sampled the first 5 collisions for the warning to reduce memory footprint and to keep the message from flooding the screen.

anndata/utils.py
ivirshup: Ref for posterity: https://github.com/pandas-dev/pandas/issues/33001

anndata/utils.py
```diff
@@ -60,17 +64,37 @@ def make_index_unique(index: pd.Index, join: str = "-"):
...
+                break
+            issue_interpretation_warning = True
+            if len(example_colliding_values) < 5:
+                example_colliding_values.append(tentative_new_name)
+
+    if issue_interpretation_warning:
+        warnings.warn(
+            f"Suffix used ({join}[0-9]+) to deduplicate index values may make index "
+            + "values difficult to interpret. There values with a similar suffixes in "
+            + "the index. Consider using a different join parameter."
```
ivirshup: I like our errors and warnings to provide as explicit of a solution as possible. Could this be more like:

```suggestion
            + "the index. Consider using a different delimiter by passing `join={delimiter}`."
```

mweiden:
@ivirshup we have a bit of code that uses this that is intended for a less computational audience that may not be able to self-solve if the code throws an error. Further, the index values are paired with ontological metadata, which should serve as the ground truth for the semantic meaning of the value.

PR #346, opened 2020-03-25
[draft] html repr

ivirshup:
@flying-sheep @falexwolf

Draft implementation of the html repr I had been talking about. It's an expandable tree that lets you look into your anndata object. If you have some time, I'd love to get some feedback on how this looks and what information should be presented.

Here's a static representation, but I'd recommend checking out this branch and trying it:

image

This is still very WIP, so a few key things are still TODO:

- [ ] Efficient generation for views (currently copies all elements due to access)
- [ ] Support for backed arrays (should be pretty easy)
- [ ] Fallback if ipytree isn't installed
- [ ] Fallback for unrecognized types

I'd also like to consider not using `ipytree`, and just returning some plain html. `ipytree` seems like it was built for a different usecase, and I don't think this needs to be a widget. Ideally this would be a bit lighter weight.

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -73,6 +73,11 @@ class AlignedMapping(cabc.MutableMapping, ABC):
         pass
 
     @property
+    def dims(self) -> Tuple[str, ...]:
+        """Which dimensions of the parent is this aligned to?"""
+        return tuple(("obs", "var")[i] for i in self.axes)
+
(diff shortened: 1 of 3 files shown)


ivirshup:
`mudata` has [this implemented](https://muon.readthedocs.io/en/latest/notebooks/quickstart_mudata.html#Rich-representation), and we have permission to steal it, so I think that's a better way forward here.

PR #349, opened 2020-03-28
Report `n_vars` of `.raw` in `__repr__`

falexwolf:
It'd be useful to have the shape of the raw data matrix be reported, as that's often what one is interested in. See this example:
image

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -549,7 +549,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             backed_at = f"backed at {str(self.filename)!r}"
         else:
             backed_at = ""
-        descr = f"AnnData object with n_obs × n_vars = {n_obs} × {n_vars} {backed_at}"
+        if self.raw is not None:
+            n_vars_raw = f"({self.raw.shape[1]})"
+        else:
+            n_vars_raw = ""
+        descr = f"AnnData object with n_obs × n_vars = {n_obs} × {n_vars}{n_vars_raw} {backed_at}"
         for attr in [
             "obs",
             "var",


ivirshup:
Could raw be a sub element? I'm not sure it's clear what the number in parenthesis refers to here.

falexwolf:
I think this is the most elegant way to do this. People will quickly understand what it means.

Another row simply for a dimension seems odd to me.

But if you don't want to merge, happy to learn about another suggestion.

ivirshup:
I was thinking if it split out onto another line, it could include more information. Like:

```
AnnData object with n_obs × n_vars = 2638 × 1838 
    obs: 'n_genes', 'percent_mito', 'n_counts', 'louvain'
    var: 'n_cells', 'ensemblids'
    uns: 'draw_graph', 'louvain', 'louvain_colors', 'neighbors', 'pca', 'rank_genes_groups'
    obsm: 'X_pca', 'X_tsne', 'X_umap', 'X_draw_graph_fr'
    varm: 'PCs'
    obsp: 'distances', 'connectivities'

    Raw with n_vars = 13714
        var: 'n_cells', 'ensemblids'
```

gokceneraslan:
We can also report cells in raw here, no? like `Raw: 2638 × 13714`

ivirshup:
Sure!

falexwolf:
I'm also happy with this more verbose option but would try to keep it as minimal as possible and keep it on a single line.

Having the default output be so bulky would be a nuisance in my notebooks. If someone wants to have details about what's in `.raw`, they can print out the representation of `Raw`. Don't you think?

The problem is that right now, one might even just *forget* about the existence of raw, as the default output doesn't mention it *at all*. 😄

ivirshup:
> The problem is that right now, one might even just forget about the existence of raw, as the default output doesn't mention it at all. 😄

Completely agree.

>  If someone wants to have details about what's in .raw, they can print out the representation of Raw. Don't you think?

This could be fine, but at the moment `Raw.__repr__` looks like:  `<anndata._core.raw.Raw at 0x13a690850>`.

I was thinking it would nice to have the keys of `raw.var` here since they are used in plotting functions, like when `gene_symbols` is specified. I was also thinking that if `raw` got a better `__repr__`, that could just be added to the `AnnData.__repr__`.

The main issue I thought needed changing here was that, while I would know that a number in parenthesis means the anndata has a `Raw`, I don't think it would be obvious to a beginner. If this was just explicit about that number referring to the shape of `.raw`, I'd be happy to merge.

ivirshup:
Ooh, I would like to get this into the next release. What are your current thoughts on this @falexwolf?

PR #350, opened 2020-03-31
Concatenate uns (with recursion)

ivirshup:
@giovp, here's an alternative implementation of merging `.uns` in `concatenate` calls. This would supersede #343.

This uses a recursive approach so it doesn't matter what depth the keys are at. You can see the merging strategies I've implemented here in in the `uns_compat` argument for `concatenate`.

TODO:

- [ ] Decide on argument names, and merge strategy names
- [x] Copies should be made for each underlying element (this should be tested with `DictViews`)
- [x] Narrative docs demonstrating the merge strategies
- [x] Tests for more types in `uns` (probably just do a few real use cases)
- [x] Figure out how to get black to not make the test cases with large literals ugly, or turn off black for that.


giovp:
Hi @ivirshup ,

first off, it looks great, way more elaborate than what I proposed in #343 and much better in every sense.

I think I'm not on the same page with you on the strategies though, I might be missing something. For instance:
```python
assert merge_unique([{"a": {"b": "c"}}, {"a": {"b": "c"}}]) == {"a": {"b": "c"}}
```
should be 
```
{}
```

but I agree with
```python
 assert merge_common([{"a": {"b": "c"}}, {"a": {"b": "c"}}]) == {"a": {"b": "c"}}
```
where `"b":"c"` should be from the first, or second dictionary (doesn't matter)

Also, with respect to the example you made in the other PR:
```python
>>> concatenate_uns([{"a": {"b": {"c": "d"}}}, {"a": {"b":{"c": "e"}}}], "if_clean")
{"a": {"b": {"c": "e", "d": "e"}}}
```
I don't think it should return that, not sure why keys and values are switched.

In sentences, the way I see `unique` and `common` are the following:
* `unique`: dictionaries should be merged at every level if any of the keys in are unique, so for instance
```python
[{"a": {"b":{"z":"d"}}}, {"a": {"b": {"c":"d"}}}] = {"a":{"b":{"z":"d","c":"d"}}}
```
but
```python
[{"a": {"b":{"c":"d"}}}, {"a": {"b": {"c":"d"}}}] = {}
```
The reason why it should be empty is because all the keys are the same.

* `common`: dictionaries should be merged entirely, irrespective of unique keys. This means that if there are unique keys, their value is preserved, otherwise they will be override by the next dictionary. This is very similar to your version of `common`, except yours is cleaner as you make the distinction with `first`.

Curios to hear what you think!

ivirshup:
> first off, it looks great!

Thanks! This was my probably my third attempt at writing this function. It's been a bit since I've done tree algorithms, and this broke my brain the first couple attempts.

I'll explain my merge strategies in a bit more depth.

I see `uns` kind of like a directory on a file system. A mapping would be a subdirectory, while any other element is a file. The key would be the name of the file or directory. This way, we can say that each value has a "path" in uns, like `["spatial"]["library_1"]["images"]["hires"]` would be a path to an image array.

### `common`

This is finding only elements that are the same between all of passed dictionaries.

My use case here was for something like split-apply-combine. If you take an anndata object, split it into groups, then recombine it, the entries in uns from the first step should all still be valid. Using this mode keeps those entries, but shouldn't keep subset specific values.

### `unique`

This means that there should be one possible value for each element in the merged dictionary. This means that values which don't conflict are kept.

This is like the `common` strategy, but will include more values. For example, if there are multiple visium datasets, this will keep the images, but also keep the `params` field from PCA if they were kept the same.

----------------------------------

In this example:

```python
merge_unique([{"a": {"b": "c"}}, {"a": {"b": "c"}}]) == {"a": {"b": "c"}}
```

There's only one real "value", which is "c". It has the path `["a"]["b"]`. Since there is only one unique value stored at `uns["a"]["b"]`, that value is used in the merged dictionary.

For the result you were thinking of, `{}`, I think I would call this behaviour something like `"only"`. This would mean that a value is used if only one of the passed mappings had a element at this path.

<details>
<summary> possible implementation of `merge_only` </summary>

```python
def only(vals: Collection[T]) -> Union[T, MissingVal]:
    """Return the only value in the collection, otherwise MissingVal."""
    if len(vals) == 1:
        return vals[0]
    else:
        return MissingVal

def merge_only(ds: Collection[Mapping]) -> Mapping:
    return merge_nested(ds, union_keys, only)
```

</details>

For your example for `merge_common`:

```python
merge_common([{"a": {"b": "c"}, "d": "e"}, {"a": {"b": "c"}, "d": 2}]) == {"a": {"b": "c"}}
# Which you thought should give:
{"a": {"b": "c"},"d":2}
```

I don't think this should have a value at path `"d"`, since there are conflicting values between the two mappings. The `common` method will only keep elements where the values at each path are the same across all the mappings.

------------------------

> In general, I would not care about the values, and only assess if something is unique or common at the key level.

To me, the goal in having the different modes is to allow meaningful elements to be preserved. I'm not sure how that can be done without comparing the values. For example, I would probably only want to keep `.uns["pca"]["variance"]` if it had the same value for each anndata being concatenated. How would the user know which ended up in the merged dict?

--------------------------------

> with respect to the example you made in the other PR:

Ah, I see I made a typo there, I've updated it. I meant to put:

```python
>>> concatenate_uns([{"a": {"b": {"c": "e"}}}, {"a": {"b":{"d": "e"}}}], "if_clean") 
{}
```

which I expected to return:

```python
{"a": {"b": {"c": "e", "d": "e"}}}
```

ivirshup:
Some updates:

* I've added a new strategy `"only"`, which I think is similar to how @giovp was describing `"unique"`
* I've updated the strategy name from `"common"` to `"same"`.
* I've added a bunch more tests

ivirshup:
I've added some narrative docs demonstrating this. They could probably use some more work. It would also be good to make sure we were running doc tests, so that we know the examples are correct.

ivirshup:
@falexwolf Could I get you to take a look at the API, in particular the names of the merge strategies? I'm considering removing `"first"` and `"only"` for now.

I've gone with `uns_merge` as the argument name, since I think we could have more arguments controlling merging. Something like `uns_exclude` for specific exclusion of some fields.

As another note, I think we can use all of these merge strategies for merging `var`, `varm`, and `varp` while concatenating as well.

ivirshup:
Thanks! Looks like I forgot what I was doing there.

-------------------------

About obsm concatenation:

So, the way I see it, when you are combining anndata objects there are two ways the data can be merged. The first is along the axis the object is being joined. In this, the axis is being extended. Right now, only the `obs` dimension really has support for this.

When joining arrays that aligned to `obs` you need to choose how the other dimensions of these arrays are handled. We do that with the `join` argument. If it's `inner` we take the intersection of the other axis. If it's `outer` we take the union. I think this can just extend to keys of `obsm` for now.

For all elements of the anndata object which are not aligned to the `obs` axis, they are not growing, so it's really more selecting which elements are going to be used in the final object. That's what the `merge` strategies in this PR address.

---------------------------

@falexwolf I'm going to merge this for now, since it's blocking stuff. I would like a review of the docs for this. I'm currently thinking about removing the `only` and `first` strategies, since I'm not sure how useful they are, and it'd probably be better to start with fewer options.

PR #352, opened 2020-04-07
Concatenate restructure & outer join for layers

### LINKED ISSUE #351: concatenate layers with outer join
Outer join concatenation is not yet available [here](https://github.com/theislab/anndata/blob/e4581abc6297752b8954645b0fe08e0aee843b7f/anndata/_core/anndata.py#L1744). I fear, I'm currently out of capacity to make a PR. Anyone who could take care of that?

ivirshup:
This is a refactor of the `concatenate` method to make it easier to follow and extend, as well as to add some new features.

* Fixes #351 by allowing for concatenation of layers
* Gets us closer to closing #295
* Dataframes in `obsm` keep being dataframes after concatenation
* Categorical columns can keep being categorical after concatenation (basically, pandas rules get followed)

This moves pretty much all of the concatenation logic to the `anndata._core.merge` module, in particular the `concat` function. At some point I would like to deprecate `AnnData.concatenate` and replace it with a `ad.concat`/ `sc.concat` function. There is more work to do before that (most importantly, hammering out the API), but this is a good step forward.

## Breaking change

This was reverted.

~~Currently, there is only one, but I think it's more of a fix really. Previously, an outer join would fill missing values in numpy arrays with `nan`, while sparse arrays would use zeros. This kinda breaks most of what we've been trying to with pretending numpy arrays and scipy sparse arrays are interchangeable.~~

## Performance

This makes outer concatenation significantly faster.

Using this code:

```python
import numpy as np

def permuted_var_adata(nobs, nvar, nvar_kept): 
    adata = gen_adata((nobs, nvar)) 
    return adata[:, np.random.choice(adata.var_names, nvar_kept, replace=False)].copy() 

adatas = [permuted_var_adata(1000, 1000, 900) for i in range(10)]


%timeit adatas[0].concatenate(adatas[1:])
%timeit adatas[0].concatenate(adatas[1:], join="outer")
```

|                          |   inner join |   outer join |
|:-------------------------|-------------:|-------------:|
| master (9e0767b761)      |          240 ms ± 20.5 ms |          1.6 s ± 16 ms |
| current HEAD (0b14e1610) |          237 ms ± 12.4 ms |          304 ms ± 11.2 ms |

## TODO

- [ ] This could probably use a more thorough set of tests. 
- [x] I would also like to deal with the `obsm` issue by making the `join` keyword apply to those entries.
- [x] Black really doesn't like fluent style when it's only two lines. I should make that pretty again.
- [ ] Doc string for concatenate needs to be updated

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -135,7 +135,7 @@ class ImplicitModificationWarning(UserWarning):
     Examples
     ========
     >>> import pandas as pd
-    >>> adata = AnnData(obs=pd.DataFrame(index=[0, 1, 2]))
+    >>> adata = AnnData(obs=pd.DataFrame(index=[0, 1, 2]))  # doctest: +SKIP
     ImplicitModificationWarning: Transforming to str index.
     """
 
@@ -546,10 +546,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     def _gen_repr(self, n_obs, n_vars) -> str:
         if self.isbacked:
-            backed_at = f"backed at {str(self.filename)!r}"
+            backed_at = f" backed at {str(self.filename)!r}"
         else:
             backed_at = ""
-        descr = f"AnnData object with n_obs × n_vars = {n_obs} × {n_vars} {backed_at}"
+        descr = f"AnnData object with n_obs × n_vars = {n_obs} × {n_vars}{backed_at}"
         for attr in [
             "obs",
             "var",
@@ -1533,8 +1533,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         >>> adata = adata1.concatenate(adata2, adata3)
         >>> adata
         AnnData object with n_obs × n_vars = 6 × 2
-            obs_keys = ['anno1', 'anno2', 'batch']
-            var_keys = ['annoA-0', 'annoA-1', 'annoB-2', 'annoA-2']
+            obs: 'anno1', 'anno2', 'batch'
+            var: 'annoA-0', 'annoA-1', 'annoA-2', 'annoB-2'
         >>> adata.X
         array([[2., 3.],
                [5., 6.],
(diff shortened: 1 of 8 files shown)


giovp:
Hi @ivirshup , just saw this, really cool! From my understanding, this should also close #338 , as `join='outer'` would return `obsm` with missing values for obs entries that did not have that `obsm` in the original data. Does it make sense?

gokceneraslan:
That looks really great Isaac, thanks 😃

ivirshup:
@giovp, yes. I wasn't sure it made sense to add another argument for this. To me concatenating the elements of `obsm` seems a lot like concatenating `X` or elements of `layers`, so I figured this could keep the solution simple.

falexwolf:
This is great, Isaac! Only remaining Q the fill value discussion on Slack. Otherwise happy if you merge!

ivirshup:
I believe this is pretty much ready to go.

Notes on other things that should get done in future

* The code for `fill_values` is a bit of a mess, and should be cleaned up at some point. It's just passed around too much.
* In the recent commits I also realized I wasn't performing a full outer join on the layers ( layers that were only present in one object wouldn't show up in the final object), so that's been fixed, but should get a test. I can add this separately so @giovp can get going.

anndata/_core/anndata.py
```diff
@@ -1463,6 +1462,7 @@ def concatenate(
         batch_categories: Sequence[Any] = None,
         uns_merge: Optional[str] = None,
         index_unique: Optional[str] = "-",
+        fill_value=None,
```
falexwolf, line 1465: White space.

anndata/_core/anndata.py
```diff
@@ -1555,52 +1562,55 @@ def concatenate(
...
         annoA-1  2  1
-        annoB-2  2  1
         annoA-2  3  2
+        annoB-2  2  1
 
         Joining on the union of variables.
 
-        >>> adata = adata1.concatenate(adata2, adata3, join='outer')
-        >>> adata
+        >>> outer = adata1.concatenate(adata2, adata3, join='outer')
```
falexwolf, line 1570: Great idea to call it `outer`, here!

giovp:
Really amazing @ivirshup !!!

gheimberg:
Thanks for this awesome improvement @ivirshup ! Trying this out now and I sometimes get a 

```pytb
<timed eval> in <module>

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/anndata.py in concatenate(self, join, batch_key, batch_categories, uns_merge, index_unique, fill_value, *adatas)
   1691             uns_merge=uns_merge,
   1692             fill_value=fill_value,
-> 1693             index_unique=index_unique,
   1694         )
   1695 

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/merge.py in concat(adatas, join, batch_key, batch_categories, uns_merge, index_unique, fill_value)
    472     elif join == "outer":
    473         layers = outer_concat_aligned_mapping(
--> 474             [a.layers for a in adatas], reindexers, fill_value=fill_value
    475         )
    476         obsm = outer_concat_aligned_mapping(

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/merge.py in outer_concat_aligned_mapping(mappings, reindexers, index, fill_value)
    389             cur_reindexers,
    390             index=index,
--> 391             fill_value=fill_value,
    392         )
    393     return result

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/merge.py in concat_arrays(arrays, reindexers, index, fill_value)
    309             [
    310                 f(as_sparse(a), fill_value=fill_value)
--> 311                 for f, a in zip(reindexers, arrays)
    312             ],
    313             format="csr",

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/merge.py in <listcomp>(.0)
    309             [
    310                 f(as_sparse(a), fill_value=fill_value)
--> 311                 for f, a in zip(reindexers, arrays)
    312             ],
    313             format="csr",

~/.anaconda3/envs/dev/lib/python3.7/site-packages/anndata/_core/merge.py in reindexer(X, fill_value)
    274             dtype=out_dtype,
    275         )
--> 276         out = X @ idxmtx
    277 
    278         if fill_value != 0:

~/.anaconda3/envs/dev/lib/python3.7/site-packages/scipy/sparse/base.py in __matmul__(self, other)
    562             raise ValueError("Scalar operands are not allowed, "
    563                              "use '*' instead")
--> 564         return self.__mul__(other)
    565 
    566     def __rmatmul__(self, other):

~/.anaconda3/envs/dev/lib/python3.7/site-packages/scipy/sparse/base.py in __mul__(self, other)
    481         if issparse(other):
    482             if self.shape[1] != other.shape[0]:
--> 483                 raise ValueError('dimension mismatch')
    484             return self._mul_sparse_matrix(other)
    485 

ValueError: dimension mismatch
```

@gokceneraslan suggested I look at what `.layers` and saw that one anndata has a counts layer while the other did not. Would be nice to be able to concatenate even when layers don't match up perfectly.

PR #362, opened 2020-05-01
Improve chunk_X

Koncopd:
Rename `.chunk_X` to `.chunk_obs`.
Now it can return a selection from annotations in `.obs`.
Also now it can return a selection from a key in `.obsm` or `.layers`.

```
adata.chunk_obs(100, obs='louvain') 
#returns (adata.X[random_chunk], adata.obs['louvain'][random_chunk])

adata.chunk_obs(100, obs='louvain', obsm='X_pca') 
#returns (adata.obsm['X_pca'][random_chunk], adata.obs['louvain'][random_chunk])

adata.chunk_obs([0, 1, 1, 2, 3], obs='louvain')
#returns (adata.X[0, 1, 1, 2, 3], adata.obs['louvain'][0, 1, 1, 2, 3])
```

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1899,28 +1899,41 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         if start < n:
             yield (self.X[start:n], start, n)
 
-    def chunk_X(
+    def chunk_obs(
         self,
         select: Union[int, Sequence[int], np.ndarray] = 1000,
+        obs: Optional[str] = None,
+        obsm: Optional[str] = None,
+        layer: Optional[str] = None,
         replace: bool = True,
     ):
         """\
-        Return a chunk of the data matrix :attr:`X` with random or specified indices.
+        Return a chunk of the data matrix :attr:`X`, a key from :attr:`obsm` or
+        a key from :attr:`layers` with random or specified indices.
+        If both obsm and layer are not specified, returns the selection
+        from the data matrix :attr:`X`.
 
         Parameters
         ----------
         select
             Depending on the type:
-
             :class:`int`
                 A random chunk with `select` rows will be returned.
             :term:`sequence` (e.g. a list, tuple or numpy array) of :class:`int`
                 A chunk with these indices will be returned.
-
+        obs
+            If specified, returns a chunk of the annotation from :attr:`obs`.
+        obsm
+            If specified, returns a chunk of the specified key from :attr:`obsm`.
+        layer
+            If specified, returns a chunk of the specified key from :attr:`layers`.
         replace
             If `select` is an integer then `True` means random sampling of
             indices with replacement, `False` without replacement.
         """
+        if obsm is not None and layer is not None:
+            raise ValueError("You can't specify both obsm and layer")
+
         if isinstance(select, int):
             select = select if select < self.n_obs else self.n_obs
             choice = np.random.choice(self.n_obs, select, replace)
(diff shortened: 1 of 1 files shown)


anndata/_core/anndata.py
```diff
@@ -1899,28 +1899,41 @@ def chunked_X(self, chunk_size: Optional[int] = None):
         if start < n:
             yield (self.X[start:n], start, n)
 
-    def chunk_X(
```
ivirshup, line 1902: Let's keep backwards compatibility. Could `chunk_X` remain as a deprecated method?

anndata/_core/anndata.py
```diff
@@ -1929,18 +1942,26 @@ def chunk_X(
...
+                selection = self.X[indices.tolist()]
+            else:
+                selection = self.X[choice]
+            selection = selection.toarray() if issparse(selection) else selection
+            selection = selection if reverse is None else selection[reverse]
 
-        selection = selection.toarray() if issparse(selection) else selection
-        return selection if reverse is None else selection[reverse]
+        if obs is not None:
+            selection = (selection, self.obs[obs])
```
ivirshup, line 1963: Would it be like this?

```suggestion
            selection = (selection, self.obs[obs][choice])
```

ivirshup (changes requested):
* It looks like the doc-string is making the docs fail to build. Could you fix that?
* Tests plz

I'm not sure I get the use case for this method. Why not just just: `adata.X[np.random.choice(adata.n_obs, n)]`? Or something like `sc.pp.subsample`?

anndata/_core/anndata.py
Koncopd: Yep, thank you for catching this.

Koncopd:
`adata.X[np.random.choice(adata.n_obs, n)]` doesn't work in backed mode - indexing elements must be in increasing order. Actually `.chunk_X` is used mainly for drawing a random subsample from backed `X`.
`sc.pp.subsample` draws fraction, not a fixed number.

Koncopd:
About docs - i had some problems but now i get only this error locally

> Warning, treated as error:
> c:\Projects\anndata\anndata\_core\anndata.py:docstring of anndata.AnnData:44:py:
> class reference target not found: anndata.compat.Literal_

No idea why this happens.

anndata/_core/anndata.py
Koncopd: Done.

Koncopd:
About docs.
There is a link to `.chunk_X` in release_notes.rst. Adding `@utils.deprecated` seems to break the link.

ivirshup:
> Yep, lets discuss at first if we really need this function.

Sure! Why did you decide to open this PR? What was your use-case?

> adata.X[np.random.choice(adata.n_obs, n)] doesn't work in backed mode - indexing elements must be in increasing order. 

`adata[np.random.choice(adata.n_obs, n)].X` should work for unique indices regardless of order. I think we could make it work for non-unique elements too. This should also allow getting any of `obsm`, `layers`, `etc.`

> `sc.pp.subsample` draws fraction, not a fixed number.

If you pass `n_obs` instead of `fraction` to `subsample` it should work. There is the issue of the whole object being copied, where as you just want to get a subset of one array. 

--------------------------

> There is a link to .chunk_X in release_notes.rst. Adding @utils.deprecated seems to break the link.

Hmm. I think I see what's happening. It looks like `DeprecationMixinMeta` is meant to make it so that no doc page is built for `chunk_X`. But this has the side effect of breaking links. I figure we should avoid breaking docs when we're not doing a breaking release, though it would be good to make sure the docs reflect deprecations.

@flying-sheep, you're the sphinx expert, any ideas of what we do here? I'm thinking adding a note about deprecation and replacement with the decorator. I'm just not sure how to do this.

Koncopd:
I use it to get batches from a backed anndata for stochastic gradient descent. And i think it is useful to get also annotations from obs along with a batch from X.
Getting a copy of the whole object is not an option in this case.

`adata[np.random.choice(adata.n_obs, n)].X` - yep, this seems fine now, if i remember correctly it didn't work for a backed X at the time of implementation of `.chunk_X`. However, the small chance of failing for a repeating index is still a problem. Also the whole machinery of creating a view of the full object is unneeded in my case.

ivirshup:
> However, the small chance of failing for a repeating index is still a problem. 

This seems fixable.

> Also the whole machinery of creating a view of the full object is unneeded in my case.

Views should be very light weight now. They pretty much only store the indices they were subset to, and then lazily get subsets of their elements. Might be worth benchmarking? A quick test on my end:

```python
import numpy as np

subsets = [np.random.choice(range(a.n_obs), 1000, replace=False) for i in range(10)]

%%timeit 
for s in subsets: 
    X = a.X[s] 
    obs = a.obs.iloc[s] 
# 52.2 ms ± 1.05 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit 
for s in subsets: 
    v = a[s] 
    # `.tocsr()` and `.copy()` are to make these not the view subclasses
    X = v.X.tocsr()
    obs = v.obs.copy()
# 75.8 ms ± 4.14 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

It's only a 1.5x increase in time spent. Is this a bottleneck?

-------------------------------

* I think we can just remove the references from the release notes to make the docs build. It's not super important for old release notes to have working links to deprecated attributes.
* I'm not sure I like the `chunk_obs` name here. I think I had assumed the `chunk_X` method would return an iterator over chunks of X. Would `sample` be a better name for this?

Koncopd:
This can become a bottleneck with thousands of draws. But it should be checked also.
I would say i'm fine with views if they work with `replace=True` in the backed mode. Now your benchmark fails in this case.

The iterator over chunks of X is `.chunked_X`. Yes, the name is similar and this can be confusing.

PR #364, opened 2020-05-05
Create parent directory if needed in write_h5ad

LustigePerson:
As opposed to older anndata versions a missing parent directory was not created anymore when write_h5ad was used.

### DIFF
--- a/anndata/_io/h5ad.py
+++ b/anndata/_io/h5ad.py
@@ -84,6 +84,10 @@ def write_h5ad(
     mode = "a" if adata.isbacked else "w"
     if adata.isbacked:  # close so that we can reopen below
         adata.file.close()
+    # create directory if it doesn't exist
+    dirname = filepath.parent
+    if not dirname.is_dir():
+        dirname.mkdir(parents=True, exist_ok=True)
     with h5py.File(filepath, mode) as f:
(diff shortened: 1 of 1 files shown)


ivirshup:
I'm inclined to think this is likely to introduce bugs in peoples code. For example, misspelling the intended path should probably throw an error. Additionally this is not how most python libraries work, so I think it would be unexpected.

Can you provide some more justification for why you want this?

LustigePerson:
Actually, this code is coming from anndata 0.6.22 and your 3k PBMC Tutorial uses this function, for example. So, the 3k PBMC tutorial throws an error without it right now. For the scanyp plots, the tutorial also stores to a subfolder and there it is creating the required subfolders.
However, I agree that this is not the pythonic way and that it might be better without auto-creation, but I guess it should then be 1) consistent with scanpys plotting functions and 2) changed in your tutorial.

ivirshup:
That all seems fair. I would say the tutorial is out of date.

I agree on the plotting. I would say it's a little different since there isn't quite the same user control over what directory the plots go in. However, I think it would make more sense if the user just passed a relative path to `save`.

LustigePerson:
That's also fine for me. Explicit is better than implicit ;)

ivirshup:
That tutorial should be fixed now.

PR #367, opened 2020-05-12
add layers_names attribute to read_loom

VolkerBergen:
Adding layers_names, so that the user can specify which data layers to read. 

This is particularly useful for large files where not all layers are needed.

### DIFF
--- a/anndata/_io/read.py
+++ b/anndata/_io/read.py
@@ -149,6 +149,7 @@ def read_loom(
     obsm_names: Optional[Mapping[str, Iterable[str]]] = None,
     var_names: str = "Gene",
     varm_names: Optional[Mapping[str, Iterable[str]]] = None,
+    layers_names: Optional[Mapping[str, Iterable[str]]] = None,
     dtype: str = "float32",
     **kwargs,
 ) -> AnnData:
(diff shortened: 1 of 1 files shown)


ivirshup:
Thanks for the PR!

Can you add a test making sure it works?

VolkerBergen:
I'll extend this PR further for those users with a preprocessed `loom` coming e.g. from Seurat:
- recognizing np.ndarray along var/obs to be read into `obsm`/`varm`.
- detect which layer should be read sparse (raw counts) or dense (transformed data), by a sparsity threshold of e.g. 0.5 (i.e. containing at least 50% zero-entries).

ivirshup:
> I'll extend this PR further for those users with a preprocessed loom coming e.g. from Seurat:

Could this be multiple small PRs? It's easier to review that way, and individual features can get in faster. That is, unless you're doing a major restructuring of the function.

> detect which layer should be read sparse (raw counts) or dense (transformed data), by a sparsity threshold of e.g. 0.5 (i.e. containing at least 50% zero-entries).

My preference is for user control over heuristics. Plus we already kind of have an API for specifying which dense matrices to read as sparse in `read_h5ad` that it'd be good to be consistent with.

PR #370, opened 2020-05-13
account for obsp/varp during transpose

falexwolf:


anndata/_core/anndata.py
```diff
@@ -1277,6 +1277,8 @@ def t_csr(m: sparse.spmatrix) -> sparse.csr_matrix:
             self._uns,
             self._varm.flipped(),
             self._obsm.flipped(),
+            obsp=self.varp,
+            varp=self.obsp,
```
ivirshup, line 1281: Should copies get made here?

PR #375, opened 2020-05-14
deprecate concatenate, introduce concat

falexwolf:


ivirshup:
I'm planning on going a little bit further with this. I'd like go even closer to a `pandas` or `xarray` style concatenation function. I'm planning on eventually just exporting `anndata._core.merge.concat` as a top-level function and deprecating `AnnData.concatenate` in favor of that. This would match more with `pandas`, `xarray`, and `numpy` since they all have concatenation as function.

What I'm waiting on for that is adding a few more features, and making sure the API is right. The biggest feature I want to add is an `axis` argument (which I have a branch for). One of my main questions for API is how much flexibility to give users in terms of selecting which elements should be merged.

falexwolf:
OK - this sounds great and much better! 😄

PR #378, opened 2020-05-18
Concat var

ivirshup:
Allow's concatenation along variables. I'm thinking this will be useful for merging multimodal datasets after preprocessing has been done with separate objects. Also makes split-apply-combine over the variables more feasible.

Here's a summary by test-case:

```python
    a = concat([lhs, rhs], axis=axis, join=join_type, merge=merge_strategy)
    b = concat(
        [lhs.T, rhs.T], axis=abs(axis - 1), join=join_type, merge=merge_strategy
    ).T

    assert_equal(a, b)
```

Where `merge` controls what elements are kept from the axis not being concatenated. I.e., in the call `concat([a, b, c], axis=0, merge=merge_strategy)`, it's controlling what elements will be in `varm` and (in future) `varp`.

## TODO:

- [x] Move all backwards compatibility code to `AnnData.concatenate` wrapper
  - [x] Figure out what (if any) new features should be added here
- [x] Finalize defaults for `concat`
  - [x] Don't add batch key or modify index names by default – these should require keyword arguments
  - [x] Make alternative axis dataframe (e.g. `obs` or `var`) have `merge`-like behavior
  - [x] Add `pairwise` boolean argument for concatenating pairwise observations
- [x] Export `concat` (possibly mark as experimental until next version)
- [x] Docs
  - [x] Narrative (pending review)
  - [x] API
  - [x] Run examples in docs as part of test suite
- [x] `obsp`, `varp`
  - *Note to self:*: `sparse.block_diag`
- [ ] Do some benchmarks (initial checks make it look pretty close to optimal – i.e. same speed as using underlying libraries)

* Bugs to fix:
  - [x] When arrays contain `nan`, they don't compare as equal. This breaks merging. Either figure out how to do this unambiguously, or wait for numpy 1.19 https://github.com/numpy/numpy/pull/16128
    - Turns out this is bugged at the moment https://github.com/numpy/numpy/issues/16377. If fixed, we could also just port this forwards for backwards compatibility. I would really like to not roll our own array equivalency function.
  - [ ] Generally make sure we are testing cases with null values


ivirshup:
I've updated the `concat` function to make it do less by default.

For example:

```python
In [14]: lhs = gen_adata((10, 10))                                                                                          

In [15]: rhs = gen_adata((10, 10))                                                                                          

In [16]: lhs.concatenate(rhs)                                                                                               
Out[16]: 
AnnData object with n_obs × n_vars = 20 × 10
    obs: 'obs_cat', 'cat_ordered', 'int64', 'float64', 'uint8', 'batch'
    var: 'var_cat-0', 'cat_ordered-0', 'int64-0', 'float64-0', 'uint8-0', 'var_cat-1', 'cat_ordered-1', 'int64-1', 'float64-1', 'uint8-1'
    obsm: 'array', 'sparse', 'df'
    layers: 'array', 'sparse'

In [17]: concat([lhs, rhs])                                                                                                 
Observation names are not unique. To make them unique, call `.obs_names_make_unique`.
Out[17]: 
AnnData object with n_obs × n_vars = 20 × 10
    obs: 'obs_cat', 'cat_ordered', 'int64', 'float64', 'uint8'
    obsm: 'array', 'sparse', 'df'
    layers: 'array', 'sparse'
```

Differences are:

* `.var` does not get filled with `colname + batch_key` when values aren't identical. Instead this is handled by a new argument `merge` which takes the same values as `uns_merge` but is used for `var` and `varm`. For example, here the `"gene_symbol"` column is kept because it is the same between the objects:

```python
In [21]: lhs.var["gene_symbol"] = list(ascii_letters[:10])

In [22]: rhs.var["gene_symbol"] = list(ascii_letters[:10])

In [23]: concat([lhs, rhs], merge="same")
Observation names are not unique. To make them unique, call `.obs_names_make_unique`.
Out[23]: 
AnnData object with n_obs × n_vars = 20 × 10
    obs: 'obs_cat', 'cat_ordered', 'int64', 'float64', 'uint8'
    var: 'gene_symbol'
    obsm: 'array', 'sparse', 'df'
    layers: 'array', 'sparse'
```

* `obs` does not get `"batch"` column unless `"batch_col"` keyword argument is used
* Only visible through the warning, but `obs_names` also do not get modified by default. To append the batch keys, now the `index_unique` argument must be passed. This was because:
    * This operation is actually quite slow
    * Apart from toy-examples (like this), I've never wanted the names to be modified

docs/concatenation.rst
```diff
@@ -3,20 +3,231 @@ Concatenation
 
 .. note::
 
-    This section is currently under construction. There will be more here in the future.
+    The :func:`~anndata.concat` function is marked as experimental for the `0.7` release series, and will supercede the :meth:`AnnData.concatenate() <anndata.AnnData.concatenate>` method in future releases. While the current API is not likely to change much, this gives us a bit of freedom to make sure we've got the arguments and feature set right.
 
-Merging `.uns`
---------------
+With function :func:`~anndata.concat`, `AnnData` objects can be combined via a composition of two operations: concatenation and merging.
```
falexwolf, line 8: I'd either delete the word "function" or say "With the function". But you're the native speaker.

As this is the first occurrence, here, I'd also do :class:`~anndata.AnnData` instead of just `AnnData`.

PR #381, opened 2020-05-20
Workaround for large memory usage from using np.ix_ with scipy.sparse

### LINKED ISSUE #380: adata.obsp.copy() and adata.obsp access causes extremely rapid memory consumption
This is a relatively new bug, because this used to work fine until I updated scanpy and anndata. 

```
scanpy==1.5.0 anndata==0.7.2 umap==0.3.10 numpy==1.18.1 scipy==1.4.1 pandas==1.0.3 scikit-learn==0.22.1 statsmodels==0.11.1 python-igraph==0.8.0 louvain==0.6.1
``` 

On my ~250k cell dataset (csr sparse matrix):
```
adata.X

<233963x38615 sparse matrix of type '<class 'numpy.float32'>'
	with 437654478 stored elements in Compressed Sparse Row format>
```

This works great:
```
cell_subset, num_per_cell = sc.pp.filter_cells(adata.X, min_genes=1)
adata.obs['n_genes'] = num_per_cell
adata = adata[cell_subset]
```

However this rapidly explodes to > 128G of memory (the amount on my server).  
```
sc.pp.filter_cells(adata, min_genes=1)
```

I've been digging to see why, and get the same memory explosion doing
```
cell_subset, num_per_cell = sc.pp.filter_cells(adata.X, min_genes=1)
adata.obs['n_genes'] = num_per_cell
adata._inplace_subset_obs(cell_subset)
```

Further digging produces that this also explodes (this is what is done inside of _inplace_subset_cell):
```
adata = adata[cell_subset].copy()
```

I walked through the copy() function and found the following immediately consumes all memory on my machine
```
adata = adata[cell_subset]
adata.obsp.copy()
```

Obsp only contains: 
```
PairwiseArraysView with keys: distances, connectivities
```

Hell even accessing the 'distances' element of obsp causes an immediate memory explosion:
```
type(adata.obsp['distances'])
```

I'm going to keep digging but there are lots of layers of abstraction down own this far, and you might be more familiar than I.  I'll also try to get a simple reconstruction.  Do you happen to know off the top of your head what computes the 'distances' obsp entry?

ivirshup:
Fixes #380

Basically, `scipy.sparse` has weird memory usage behavior around orthogonal indexing:

```python
from scipy import sparse
import numpy as np

s = sparse.random(10000, 10000, format="csr")
idx1 = np.random.choice(10000, 1000, replace=False)
idx2 = np.random.choice(10000, 1000, replace=False)

# This is fine
%memit s[idx1.reshape(-1, 1), idx2]
# peak memory: 102.25 MiB, increment: 1.36 MiB

# This is not
%memit s[idx1.reshape(-1, 1), idx2.reshape(1, -1)]
# peak memory: 126.00 MiB, increment: 23.38 MiB  # Increases with size of subset
```

Though the latter is what you're supposed to pass to `np.ndarrays`.

### DIFF
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -118,11 +118,19 @@ def unpack_index(index: Index) -> Tuple[Index1D, Index1D]:
 
 
 @singledispatch
-def _subset(a: Union[np.ndarray, spmatrix, pd.DataFrame], subset_idx: Index):
+def _subset(a: Union[np.ndarray, pd.DataFrame], subset_idx: Index):
     # Select as combination of indexes, not coordinates
     # Correcting for indexing behaviour of np.ndarray
     if all(isinstance(x, cabc.Iterable) for x in subset_idx):
         subset_idx = np.ix_(*subset_idx)
+    return a[subset_idx]
+
+
+@_subset.register(spmatrix)
+def _subset_spmatrix(a: spmatrix, subset_idx: Index):
+    # Correcting for indexing behaviour of sparse.spmatrix
+    if len(subset_idx) > 1 and all(isinstance(x, cabc.Iterable) for x in subset_idx):
+        subset_idx = (subset_idx[0].reshape(-1, 1), *subset_idx[1:])
     return a[subset_idx]
 
 


ivirshup:
To demonstrate the effect on anndata:

### Creating demo data:

```python
import scanpy as sc
a = sc.datasets.blobs(n_observations=10000)
# This is actually faster than sparse.random for equivalent density, wtf
sc.pp.pca(a)
sc.pp.neighbors(a)

a.write_h5ad("test_with_obsp.h5ad", compression="lzf")
```

### Test code

```python
%load_ext memory_profiler
import anndata as ad

a = ad.read_h5ad("test_with_obsp.h5ad")
%memit a[range(a.shape[0])].copy()
```

### Results

On master:

```
peak memory: 2410.82 MiB, increment: 2303.93 MiB
```

On d0839e6144fdce8ff0 

```
peak memory: 128.78 MiB, increment: 21.05 MiB
```

falexwolf:
Wow! Super impressive! 😄

PR #398, opened 2020-07-08
Fix documentation of argument types for anndata.concat

ivirshup:
Types were previously not showing up.

This was due to sphinx_autodoc_typehints failing to process `Optional` and `collections.Mapping[str, AnnData]` as valid typehints (since they are not). However, instead of creating any warning or error, this just silently passed and did nothing.

### DIFF
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -6,7 +6,8 @@ from collections.abc import Mapping, MutableSet
 from functools import reduce, singledispatch
 from itertools import repeat
 from operator import and_, or_, sub
-from typing import Callable, Collection, Iterable, Optional, Tuple, TypeVar, Union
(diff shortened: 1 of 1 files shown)


flying-sheep:
The warning happens here: https://github.com/agronholm/sphinx-autodoc-typehints/blob/e6b611955baf00ef3206326e7b6a942ba39ab604/sphinx_autodoc_typehints.py#L212-L213

the default for all loggers is to be set to show warnings and errors, so we need to figure out why there’s some suppression of logs happening somewhere.

ivirshup:
I believe I was getting a `TypeError` first, which wasn't leading to a log.

Two of the errors I remember were `"Mapping[str, AnnData]"` where `Mapping` was imported from `collections.abc`, and a bare `Optional` which is apparently invalid.

 I figured out what was happening by using a modified version of this code with a bunch of print statements.

PR #402, opened 2020-07-21
[WIP] Make obsm/varm creation lazy

### LINKED ISSUE #360: Anndata not properly garbage collected

fhausmann:
Makes `AxisArrays` creation for `obsm` and `varm` lazy, meaning only create these objects if needed and no reference inside the `AnnData` object is needed.

This should fixes the increased memory consumption when creating multiple `AnnData` objects.

Should fix #360 and was proposed here https://github.com/theislab/anndata/pull/363#issuecomment-624367298

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -232,8 +232,8 @@ class AxisArrays(AlignedActualMixin, AxisArraysBase):
             self.update(vals)
 
     def __setitem__(self, key: str, value: V):
-        if self._axis == 0:
-            self.parent._obsm[key] = value#.copy()
+        if self._axis == 0 and isinstance(self.parent, anndata.AnnData):
+            self.parent._obsm[key] = value
         super(AxisArrays, self).__setitem__(key, value)
 
 
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -243,6 +243,17 @@ class AxisArrays(AlignedActualMixin, AxisArraysBase):
         if parent_needs_update:
             self.parent._obsm[key] = value
 
+    def __delitem__(self, key: str):
+        parent_needs_update = (
+            self._axis == 0
+            and isinstance(self.parent, anndata.AnnData)
+            and self._parent_needs_update
+        )
+        super(AxisArrays, self).__delitem__(key)
+        if parent_needs_update:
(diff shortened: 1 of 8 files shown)


fhausmann:
Status Update:

- [x] obsm lazy
- [ ] varm lazy
- [ ] Add test for memory usage
- [ ] Fix circular reference in AnnDataFileManager

kaizhang:
When will this be merged? I would like to use `obsm` to store a large matrix and do not want it to be loaded to memory.

ivirshup:
@kaizhang I think it's fair to say this PR has stalled and a new one will need to be created, will close for now but will start tracking issue. Shouldn't be awful just needs to get done.

While this does cause temporary increases in memory usage (until a gc cycle collects these) it should not cause an on-disk array to be loaded into memory.

kaizhang:
@ivirshup Are you suggesting that matrices stored in `obsm` or `varm` will not be loaded into memory when opening an anndata file? I know `X` will not be loaded into memory by default. But it seems to me that all other annotations will be kept in memory. Is this documented somewhere?

ivirshup:
> Are you suggesting that matrices stored in obsm or varm will not be loaded into memory when opening an anndata file?

No, I'm saying that the issue this PR is addressing is orthogonal to loading things into memory. This is more of a memory-leak/ poor garbage collection issue.

>  I know X will not be loaded into memory by default.

It only won't be loaded if you pass `ad.read_h5ad(..., backed=True)`

kaizhang:
Thanks for the clarification. How do I tell `anndata` not to load my matrix then? It seems to me that the only way is to use the upcoming `filespec` feature to define custom file format. Will `filespec` also support subsetting, e.g., get access to current view, in future?

PR #404, opened 2020-07-29
Add obsp to adata.raw.to_adata()

gokceneraslan:


ivirshup:
Could this get a test? Does this method even have a test yet?

Don't worry about travis, just rebase on master. I've pinned pandas below `1.1` so those tests won't fail.

PR #416, opened 2020-08-09
Multi-file/ AnnData access and torch data loader: AnnCollection and AnnLoader

Koncopd:
This PR brings two new data objects to the main API.

* `anndata.dev.AnnLoader`: An interface to pytorch DataLoader.
* `anndata.dev.AnnCollection`: An interface to collections of AnnData objects and files, to iterate over and index into.  

The PR started out with open-sourced code written by @falexwolf & @yugeji at @Cellarity: https://github.com/theislab/anndata/pull/412.


adamgayoso:
Hi @Koncopd, this is very intriguing and looks good! There are a few early comments I have. 

1. For our implementation, we opted not to copy any data from the anndata object. I see here that if it's sparse, you can make it dense. We instead make a minibatch dense in the getitem of the torch dataset. This is a bit slower but if your data is massive it could be the only way to do it?
2. We tried to make it flexible enough such that you can get many tensors from the anndata (e.g., X, and obsm df or ndarray, obs vectors, var vectors etc.) To do this, we have the getitem return a dictionary (btw I think your getitem just needs to return  numpy as torch should make them tensors for you, though we handle cuda a bit differently than you). Generally, this flexibility to grab tensors will help us when we implement the option to condition on many covariates
3. We don't split the anndata into train/test, instead we just specify train/test/val indices in the dataloader to reduce any additional data copying.

I'm not sure you should/need to implement any of my comments. I just think that it's probable we will need a custom implementation anyway. We also need custom code to check that categorical things like labels or RNA-seq batches are encoded consistently. I see here that you use `LabelEncoder` -- we decided to add `.obs` fields with the cats and codes for the labels, as well as store the mapping in `.uns` so that you can "transfer" the codes to a new anndata that potentially has a subset of labels of the source anndata. Honestly not sure what the best way to do this is -- we just wanted to make things user friendly.

Koncopd:
Thank you for the comments.
As for 2, i think about implementing batch as an anndata view which can convert arrays to tensors inside `getitem`.

adamgayoso:
If I recall, we found that slicing the anndata itself is much slower than each of it's arrays/dfs individually.

falexwolf:
Just seeing this now, @adamgayoso! Thank you very much! Let's see how much we can adopt and through that, potentially, converge to a canonical solution for interfacing anndata with pytorch. NVIDIA themselves said they're interested in contributing, too.

---

> For our implementation, we opted not to copy any data from the anndata object. I see here that if it's sparse, you can make it dense. We instead make a minibatch dense in the getitem of the torch dataset. This is a bit slower but if your data is massive it could be the only way to do it?

@Koncopd, I thought that this is how we also did it at least in the early chunking iterators?

> We don't split the anndata into train/test, instead we just specify train/test/val indices in the dataloader to reduce any additional data copying.

Agreed!

>  I see here that you use LabelEncoder -- we decided to add .obs fields with the cats and codes for the labels, as well as store the mapping in .uns so that you can "transfer" the codes to a new anndata that potentially has a subset of labels of the source anndata. Honestly not sure what the best way to do this is -- we just wanted to make things user friendly.

I've done the same in the beginning, but we then switched to LabelEncoder as mapping across datasets with potentially highly divergent labels turned out much easier with less custom code than with categories. We're then storing the "trained" LabelEncoder just with the trained model, and not inside the AnnData. Will discuss this.

falexwolf:
@ivirshup just suggested to cache all batches on disk to save the reindexing time. That's a really, really cool idea!

falexwolf:
Indexing logic for a batch of data, given a backed `adata`:
```
if not cached:
     batch = adata[batch_idcs].copy()  # load observations batch into memory
     batch = batch.to_dense()  # densify
     batch = batch[:, target_variables]  # subset to target variable names defined for AnnDataSet/ AdCollection
     batch.write(cache_path)
else:
     batch = sc.read(cache_path)

# ... apply model to batch
```

Need to consider whether one should cache each batch individually or the whole concatenated batches together.

Koncopd:
This should be optional. If we use it with pytorch, then we have random indices which don't repeat, so this caching just slows the process.

ivirshup:
Sergei, my worry was less with the concatenation along obs (which should be fast-ish, assuming csr matrices) but with reindexing along the variables (which will be slow). If objects have been filtered differently or genes are ordered differently then you'd could save a fair amount of time on each batch if it was just cached.

Koncopd:
Yes, now i see your point. I will add obs index caching.

Koncopd:
O, this is ready now in terms of functionality.
A small example is [here ](https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataloader_example.ipynb).

It still needs documentation though.

falexwolf:
Consider
```
autodoc_mock_imports = [
    "torch",
]
```

Koncopd:
Here is a simple [AnnDataSet example](https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataset_simple_example.ipynb) with some simple benchmarks.

gokceneraslan:
Looks cool! Are you guys planning to implement a .write() method too? 🤓

Koncopd:
Maybe.
I there is no need of X or layers, it can be done with `a.to_adata().write(...)`.

ivirshup (commented):
Re @gokceneraslan's comment: as I've discussed with Sergei before, I think this would be useful for combining large datasets without having to do it all in memory. It could also have some performance benefits for iterating over the data so you don't have to keep indexing along the slow axis.

# Comments on the tutorial

## Consistency with concat

```python
dataset = AnnDataSet([covid, pbmc], join_vars='inner', join_obs='outer', join_obsm='inner')
```

It would be nice if this followed the same conventions as concat for specifying joins. I do see how it can be useful to specify different joins for the different collections, but maybe this could be made more general? Maybe we could allow nested input for `join`, like `join={"vars": "inner", "obs": "outer"}`?

## `label` seems broken:

This errors:

```python
dataset = AnnDataSet([covid, pbmc], join_vars='inner', label="dataset")
```

<details>
<summary> traceback </summary>

```pytb
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-24-f0c875e4c055> in <module>
----> 1 dataset = AnnDataSet([covid, pbmc], join_vars='inner', label="dataset")

~/github/anndata/anndata/dev/multi_files/_anndataset.py in __init__(self, adatas, join_obs, join_obsm, join_vars, label, keys, index_unique, convert, harmonize_dtypes)
    483             self.obs = pd.DataFrame(index=self.obs_names)
    484         if label is not None:
--> 485             self.obs[label] = label_col
    486 
    487         # process obsm inner join

UnboundLocalError: local variable 'label_col' referenced before assignment
```

</details>

## AnnDataSet indexing

### benchmarks

I was surprised to see the benchmarks for indexing. Looking into it a bit, I think a large amount of the timing difference comes from the default anndata object call removing unused categories from the generated dataframe. We could reconsider this behavior, though arguably the dataframes should be lazily generated anyways.

<details>
<summary> Example (easier to see with snakeviz) </summary>

```python
%prun -s cumtime covid[np.random.choice(ds_covid.shape[0], 200, replace=False)]
```

```
         18873 function calls (18614 primitive calls) in 0.022 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.022    0.022 {built-in method builtins.exec}
        1    0.000    0.000    0.022    0.022 <string>:1(<module>)
        1    0.000    0.000    0.020    0.020 anndata.py:1106(__getitem__)
        1    0.000    0.000    0.020    0.020 anndata.py:282(__init__)
        1    0.000    0.000    0.020    0.020 anndata.py:325(_init_as_view)
        2    0.000    0.000    0.016    0.008 anndata.py:1111(_remove_unused_categories)
       19    0.000    0.000    0.006    0.000 frame.py:3147(__setitem__)
       19    0.000    0.000    0.006    0.000 frame.py:3231(_set_item)
```

</details>

### Semantics

It's a little unclear to me what the semantics of the `AnnDataSetView` object are supposed to be. Could you clarify this?

For example, `X` is just a matrix, but `obs` and `obsm` are some kind of view? What are these views meant to do?

On the implementation side, these views look like they are creating circular references which can be bad for garbage collection.

### Backed datasets giving incorrect results:

It looks like indexing the `AnnDataSet` made of backed objects is returning incorrect data for `X`

<details>
<summary> Example </summary>

Setup:

```python
both_concat = sc.concat({"covid": covid, "pbmc": pbmc}, label="dataset")

assert both_concat.obs_names.equals(dataset.obs_names)
assert both_concat.var_names.equals(dataset.var_names)

# Backed AnnDataSet
pbmc_b = sc.read('pbmc_seurat_v4.h5ad', backed='r')
covid_b = sc.read('covid_cite.h5ad', backed='r')

dataset_b = AnnDataSet([covid_b , pbmc_b], join_vars='inner', join_obs=None, join_obsm='inner')

assert dataset_b.var_names.equals(both_concat.var_names)
assert dataset_b.obs_names.equals(both_concat.obs_names)
```

Checking indexing:

```python
for _ in range(10):
    idx = np.random.choice(both_concat.shape[0], 200, replace=False)
    # This should be zero if they are equivalent
    print((dataset_b[idx].X != both_concat.X[idx, :]).sum())
```

```
838873
811316
813318
841387
790324
802966
777457
787657
797884
777062
```

</details>

### Regular datasets giving incorrect results as well

Going back and checking the in memory example, the same problem occurs. It looks like it has to do with the order of the indices. If `idx` is sorted, values seem to be correct. If it's not sorted, I think the values start being wrong at the first row where the datasets aren't in their original order. E.g. in this example, the rows have the wrong values for all rows following the first integer in `idx` that maps to the pbmc dataset.

Probably not worth reading too much into the combined benchmarking results until this is fixed.

## Naming

I'm starting to agree more with @falexwolf about the name being confusing. In particular, this line is just weird:

```python
dataset.obs['dataset'] = pd.Series(dtype='O')
```

I think there's to much confusion about what a "dataset" is here, since each `AnnData` is a dataset, but they're combined into an `AnnDataSet`.

Koncopd:
@ivirshup thank you for the comments, i have fixed the bugs and updated the example with benchmarks.

Koncopd:
Here is an [example](https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataloader_example_two_adatas.ipynb) of `AnnDataLoader` usage with 2 adatas.

ivirshup:
>  i have fixed the bugs and updated the example with benchmarks.

Could you add tests for these before I look into this more? I'd like to know that these stay fixed through further updates. Plus, tests are a good way to find bugs.

Related: I think there might be an issue with access to the `.obs` of an `AnnDataSet`, as they appear to be returning numpy arrays instead of pandas `Series`.

Koncopd:
Yes, i will add the tests. 

Yes, it returns numpy objects arrays instead of series in several situations, it is intended. We can discuss if it is a good approach.

anndata/dev/multi_files/_anndataset.py
```diff
@@ -0,0 +1,670 @@
...
+            keys = np.arange(len(adatas)).astype(str)
+        label_col = pd.Categorical.from_codes(
+            np.repeat(np.arange(len(adatas)), [a.shape[0] for a in adatas]),
+            categories=keys,
+        )
+        if index_unique is not None:
+            concat_indices = concat_indices.str.cat(
+                label_col.map(str), sep=index_unique
+            )
+        self.obs_names = pd.Index(concat_indices)
```
falexwolf, line 493: Shouldn't this set the index of a `self._obs` attribute that stores the joint obs annotations? And `self.obs` is a property?

A similar comment applies below.

anndata/dev/multi_files/_anndataset.py
```diff
@@ -0,0 +1,670 @@
...
+            X = self._gather_X()
+            shape = None
+
+        adata = AnnData(X, obs=obs, obsm=obsm, layers=layers, shape=shape)
+        adata.obs_names = self.obs_names
+        adata.var_names = self.var_names
+        return adata
+
+
+class AnnDataSet(_ConcatViewMixin):
```
falexwolf, line 434: Can we get a full spec of the docstring that clarifies all subtleties of the params?

falexwolf: Most importantly, describe that `AnnDataSet` does not give access to most fields of `AnnData`, but `AnnDataSetView` does.

docs/api.rst
```diff
@@ -65,6 +65,16 @@ Writing to other formats.
    AnnData.write_zarr
 
 
+Experimental development API
+----------------------------
+
+.. autosummary::
+   :toctree: .
+
+   dev.multi_files.AnnDataSet
```
falexwolf, line 74: I'd suggest to have both objects exported at the root level of `anndata.dev`. The submodules are confusing.

Koncopd:
I have this when i try to build the docs

`anndata\anndata\_core\anndata.py:docstring of anndata._core.anndata.AnnData:68:py:meth reference target not found: pandas.DataFrame.iloc`

ivirshup:
Yeah, I'm seeing that too. Not sure what to do about it yet, since it looks like that's how pandas itself refers to `iloc`.

Koncopd:
gathering the latest tutorials:

- `AnnDataLoader`
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataloader_example.ipynb
- `AnnDataLoader` + `AnnDataSet`
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataloader_example_two_adatas.ipynb
- `AnnDataSet`
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/multi-files/anndataset_simple_example.ipynb

anndata/dev/multi_files/_anndataset.py
Koncopd: Done.

docs/api.rst
Koncopd: Done.

anndata/dev/pytorch/_anndataloader.py
```diff
@@ -0,0 +1,156 @@
...
+
+# AnnDataLoader has the same arguments as DataLoader, but uses BatchIndexSampler by default
+class AnnDataLoader(DataLoader):
+    def __init__(
+        self,
+        adatas,
+        batch_size=1,
+        shuffle=False,
+        use_default_converter=True,
+        use_cuda=False,
```
azkalot1, line 95: pl lightning and other frameworks handle moving to\out of cuda\specific gpu automatically + if using multiple gpus tensors need to be sent separately.

Koncopd: Hi, yes, i know, this is why it is set to `False` by default.

falexwolf:
This is all progressing well!

I realize that the following attribute does not seem to be documented?
```
convert = {'obs':
           {'orig.ident': lambda s: enc.transform(s.to_numpy()[: , None]),
            labels: lambda c: enc_ct.transform(c.to_numpy()[: , None])[:, :-1]
           }
          }

dataset.convert = convert
```

How do we think about that? https://icb-anndata--416.com.readthedocs.build/en/416/anndata.dev.AnnDataSet.html

Koncopd:
It is documented [here](https://icb-anndata--416.com.readthedocs.build/en/416/anndata.dev.multi_files._anndataset.AnnDataSetView.convert.html#anndata.dev.multi_files._anndataset.AnnDataSetView.convert).

falexwolf:
Thanks! I've just hidden `AnnDataSetView` from the main API; users should not have to know it.

Can you add a link to this feature from `AnnDataSet` itself, even if it's just a property of the view?

I would suggest to make this a property of AnnDataSet that errors as long as it's not yet a view, and becomes functional once it's a view.

And then only have the AnnDataSet.convert property documented.

Koncopd:
Actually i will make it a property (it isn't now), it even should not give any errors on access or setting.

Koncopd:
@falexwolf 
It seems that `hidden` doesn't work, it breaks the docs. You also deleted `read_hdf` from the docs, was it deliberate?

falexwolf:
> You also deleted read_hdf from the docs, was it deliberate?

Yes, was deliberate! It's legacy that clutters the page. Isaac wanted to deprecate it in the code soon, anyway.

Strange that hidden doesn't work, I've been using it a lot for hiding non-top level objects from the main APIs, but still generating a page to reference from sub-pages deeper in the hierarchy. We'll converge this.

anndata/dev/pytorch/_anndataloader.py
```diff
@@ -0,0 +1,207 @@
...
+                else:
+                    as_ks = attrs_keys[attr]
+                new_convert[attr] = _convert_on_top(convert[attr], top_convert, as_ks)
+    return new_convert
+
+
+# AnnDataLoader has the same arguments as DataLoader, but uses BatchIndexSampler by default
+class AnnDataLoader(DataLoader):
+    """\
+    PyTorch DataLoader for AnndData objects.
```
falexwolf, line 99: Typo!

anndata/dev/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .multi_files import AnnDataSet
```
falexwolf, line 1: Both of the submodules should be private so that users do not see the lookups: `._multi_files` and `._pytorch`.

ivirshup (commented):
## harmonize_dtypes

> harmonize_dtypes
> If True, all retrieved arrays from subset objects will have the same dtype.

When would you want this to be false? What does that look like? Object arrays? Depends on the subset retrieved?

## join_obs, join_obsm, vs convert={"obs": ..., "obsm": })

Why not have either special keyword arguments for `convert_obs`, or allow `join={"obs": ...., "obsm": ...}`

Koncopd:
`convert` is kind of uniform across attributes and\or keys (you provide an attribute name, a key and a converter function), but `join_obs` and `join_obsm` have different sets of possible values (`None, 'inner'` for the latter, `'outer', 'inner', None` for the former). `convert` is not only an initialization option but also a property of both objects (`AnnDataSet` and view).

Koncopd:
For `harmonize_dtypes` what is meant for example when you have 2 adatas and `.X` for the first has dtype float32 and for the second float64. When you create `AnnDataSet` with `harmonize_dtypes=True` every selection will give you `.X` with dtype float64, even when the subset is only from the first adata. This obviously has to copy the .X during dtype conversion.
You can have `harmonize_dtypes=False` to avoid this conversion of dtypes when you don't care, for example when you need only to have a sum along axis as in the example with size factors, or any iterative preprocessing for which dtypes ae not important.

Koncopd:
As discussed with @falexwolf and @ivirshup `AnnDataLoader` was renamed to `AnnLoader` and `AnnDataSet` to `AnnCollection`. In the tutorials also. This is the final naming.

Koncopd:
@falexwolf `:hidden:` goes through but still doesn't hide the `AnnCollectionView` link.

ivirshup:
So, after discussion with @falexwolf, it sounds like it's more useful to have this merged since other people are going to be working with it. Planning on doing that once the CI checks pass.

I've made one major change, which is renaming the `.dev` module to `.experimental` so it fits with what we're doing for scanpy (https://github.com/theislab/scanpy/pull/1715).

PR #418, opened 2020-08-14
add version references

flying-sheep:
We need to solve this first: https://stackoverflow.com/q/63411072/247482

maybe we should define a `.. version::` directive in scanpydoc that takes care of this, as well as nicely formatting version and date. We could even make it automatically link to the PyPI release or GitHub or so!


flying-sheep:
we now use a custom sphinx directive to render our version notes, which works well!

PR #423, opened 2020-08-19
Consolidate setup information

flying-sheep:
See theislab/scanpy#1381


pyproject.toml
```diff
@@ -1,7 +1,19 @@
 [build-system]
-requires = ['setuptools', 'setuptools_scm', 'wheel']
+requires = ['setuptools', 'setuptools_scm', 'wheel', 'pytoml']
 build-backend = 'setuptools.build_meta'
 
+# we’ll move to flit, but for now this only holds package data
+[tool.flit.metadata]
+author = 'Philipp Angerer, Alex Wolf, Isaac Virshup, Sergei Rybakov'
+# We don’t need all emails, the main authors are sufficient.
+author-email = 'phil.angerer@gmail.com, f.alex.wolf@gmx.de'
```
ivirshup, line 9: Can this be stored in a more tool independent way? Like how for `setup.cfg` there's the `metadata ` and `options.extras_require` tables.

setup.py
```diff
@@ -18,19 +21,37 @@
...
+            "codacy-coverage",
+            "docutils",  # for rst2html.py
+            "zarr",
+            "matplotlib",
+            "sklearn",
+            "xlrd",
+            "joblib",
+            "boltons",
+            "scanpy",
+        ],
```
ivirshup, line 54: Could these all be in a config file instead of `setup.py`? I'm thinking something like [this](https://github.com/jaraco/skeleton/blob/master/setup.cfg), but as entries in `pyproject.toml` and not tool specific.

ivirshup:
Oh, there's a pep for that: https://www.python.org/dev/peps/pep-0621. That seems like a good idea. Seems weird that it's independent of 517 and 518.

flying-sheep:
Thank you for digging up PEP 621. I fear that until it exists, no standard for that exists other than setup.cfg (which is on its way out due to pyproject.toml). So we can store email and author in one of the following:

- use setup.cfg (another file that *will* be replaced)
- use the WIP PEP 621 format (“specifying dependencies” is its one [open issue](https://www.python.org/dev/peps/pep-0621/#id119))
- invent our own standard and store them under `tool.anndata[.*]` (we can only use a section under `tool` for which we control the PyPI entry)
- use the format of the tool we use

personally I see no need to not use `[tool.flit.metadata]` if we use a compatible format anyway, but sure.

PR #434, opened 2020-09-22
Remove dtype argument from AnnData constructor

### LINKED ISSUE #129: Should the anndata constructor convert datatypes by default?
I was just converting some data from R to python and was trying to debug why I was getting a different amount of total counts. In particular why `sum(adata.X.data)` was off by over 1000 from `np.sum(adata.X.data)`. Ended up turning the difference was floating point error, and `sum` was more accurate (also gave a more similar answer to R).

I'd generally expect calculations like "how many counts did this cell/ dataset have" to be precise.

Following the decision that `float32` isn't intended to be an automatic conversion anymore for most functions in scanpy, could we change the by-default conversion to `float32` in the `AnnData` constructor? I'm thinking we can keep the argument, just have the default be `None`.

ivirshup:
Fixes #129 

Brought up again by https://github.com/theislab/scanpy/issues/1415

This PR makes the `dtype` argument of the `AnnData` constructor default to `None`, and throws a warning if it is passed. The idea is that we should just accept whatever array was passed as `X`.

This code change breaks some tests in scanpy. AFAICT, this can all be fixed on the scanpy side.

@falexwolf @flying-sheep, how do we want to handle this breaking change? A few thoughts:

* One option: this goes in 0.8, and we throw a warning if `dtype` is passed
* Should we warn about this before it's in a breaking release? Might be hard to do, since it's a changing default.

# TODO

- [ ] Fix tests on scanpy side
- [ ] Figure out how we want to handle this change
- [ ] Check for/ remove old tests for dtype argument

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -288,7 +288,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         varm: Optional[Union[np.ndarray, Mapping[str, Sequence[Any]]]] = None,
         layers: Optional[Mapping[str, Union[np.ndarray, sparse.spmatrix]]] = None,
         raw: Optional[Mapping[str, Any]] = None,
-        dtype: Union[np.dtype, str] = "float32",
+        dtype: Optional[Union[np.dtype, str]] = None,
         shape: Optional[Tuple[int, int]] = None,
         filename: Optional[PathLike] = None,
         filemode: Optional[Literal["r", "r+"]] = None,
@@ -387,7 +387,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         obsp=None,
         raw=None,
         layers=None,
-        dtype="float32",
+        dtype=None,
         shape=None,
         filename=None,
         filemode=None,
@@ -397,6 +397,13 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         self._adata_ref = None
         self._oidx = None
         self._vidx = None
+
+        if dtype is not None:
+            warnings.warn(
+                FutureWarning(
+                    "The dtype argument is deprecated and will be removed in anndata 0.8."
+                )
+            )
 
         # ----------------------------------------------------------------------
         # various ways of initializing the data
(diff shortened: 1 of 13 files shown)


flying-sheep:
Totally for it. Each layer can have its own dtype.

So you want to leave the read functions defaulting to float32? For formats that have their own intrinsic precisison, I think that makes no sense, for text formats it does:

- read_excel: I guess it makes sense. Even if Excel has a way to internally use f32, few people probably use it
- text-based (umi-tools, csv, …): No intrinsic precision
- hdf: Intrinsic precision is already derived from the file
- **loom: Has intrinsic precision, but we convert the data anyway. I think we should change that.**
- matrix market: No intrinsic precision

ivirshup:
> So you want to leave the read functions defaulting to float32? For formats that have their own intrinsic precisison, I think that makes no sense, for text formats it does:

This was pretty much my thoughts. If you pass in something with precision, we'll use that. If you don't we can probably assume float32.

ivirshup:
@flying-sheep, this has a companion PR in scanpy that shows everything this breaks https://github.com/theislab/scanpy/pull/1430. This needs a plan for release. I'm trying to figure out what the right way to do this is, so code will still work with both versions, without too many warnings being thrown. How does this sound:

* This changes in `0.8`
* In 0.7 (the next bugfix release), throw a `FutureWarning("In a future version this will not be converted")` if `dtype` was not passed and `X` would be cast to float32.
* In `0.8.0`, `dtype` is None by default, but passing `dtype` will still convert. A `FutureWarning("Conversion via argument dtype is deprecated")` will be thrown if a conversion would happen. Passing the argument at all throws a `DeprecationWarning` which will be visible to package developers, but not users.
* Some time into `0.8` (giving people a chance to upgrade) passing any value for `dtype` would throw a user visible `FutureWarning`
* The argument is completely removed for the following breaking release.

I feel like this will give people the easiest upgrade path. If they get warnings now, specify the `dtype`. Next version, developers get warned that `dtype` is deprecated and get time to update before users get warnings. Finally users see deprecation warnings, and soon after it's gone.

An alternative would be to rip the bandaid off faster, but I don't want to be saying "if you don't want your data to convert pass dtype, though this will break soon".

flying-sheep:
> This was pretty much my thoughts. If you pass in something with precision, we'll use that. If you don't we can probably assume float32.

Okay, so we should change it for loom. For Matrix Market, I guess we could get fancy and derive it from the amount of digits in the first few numbers, but there’s probably not many people using that one anyway.

> If they get warnings now, specify the `dtype`

Hmm, this requires two actions for users going forward. Would be nice if there was a way to do The Right Thing immediately, but I can’t come up with one. If we allow setting some kind of `._noconvert` field on the object being passed, we risk that being cargo-culted to everywhere.

So yeah! Sounds very solid! :smile:

ivirshup:
Yeah, I'd like this to be quicker, but I can't think of another way to have code that will work with `0.7-0.8` or `0.8-0.9`

flying-sheep:
Superseded by #854

PR #461, opened 2020-11-22
Make AxisArrays and AxisArraysView dim_names a property

### LINKED ISSUE #459: Allowing obsm/varm to have their own dim_names leads to weird issues
Imagine a scenario where the user reads the mtx, genes and barcode files to load a dataset (e.g. [this](https://singlecell.broadinstitute.org/single_cell/study/SCP1038/the-human-and-mouse-enteric-nervous-system-at-single-cell-resolution#study-download)):

```python
adata = sc.read_mtx('gene_sorted-hli.matrix.mtx')
adata.obs_names = pd.read_table('hli.barcodes.tsv', header=None).values.ravel()

# or
# obs = pd.read_table('hli.barcodes.tsv', header=None, index_col=0)
# obs.index.name = None
# adata.obs = obs

adata.var_names = pd.read_table('hli.genes.tsv', header=None).values.ravel()
```

and further imagine that the user wants to store something in varm:

```python
ad = adata[:10]
fake_varm = pd.DataFrame(ad.X.A, index=ad.obs_names, columns=ad.var_names).T
adata.varm['something'] = fake_varm
```

Unfortunately this leads to the following exception:

```pytb
<ipython-input-120-2aa021228d99> in <module>
----> 1 adata.varm['something'] = fake_varm

~/.miniconda3/lib/python3.8/site-packages/anndata/_core/aligned_mapping.py in __setitem__(self, key, value)
    149 
    150     def __setitem__(self, key: str, value: V):
--> 151         value = self._validate_value(value, key)
    152         self._data[key] = value
    153 

~/.miniconda3/lib/python3.8/site-packages/anndata/_core/aligned_mapping.py in _validate_value(self, val, key)
    210         ):
    211             # Could probably also re-order index if it’s contained
--> 212             raise ValueError(
    213                 f"value.index does not match parent’s axis {self.axes[0]} names"
    214             )

ValueError: value.index does not match parent’s axis 1 names
> /home/gokcen/.miniconda3/lib/python3.8/site-packages/anndata/_core/aligned_mapping.py(212)_validate_value()
    210         ):
    211             # Could probably also re-order index if it’s contained
--> 212             raise ValueError(
    213                 f"value.index does not match parent’s axis {self.axes[0]} names"
    214             )

```

because obs/var or obsm/varm assignments do not update obsm and varm `dim_names` attribute:

image

I know that this exception is designed as a protection against copying random data frames to obsm/varm (see https://github.com/theislab/anndata/pull/228#issuecomment-545771620 for more detailed explanation and discussion), but I don't understand why there are two things like `adata.obs_names` and `adata.obsm.dim_names` that are independently updatable? Is the right solution here to define setters for obs/var and obs_names/var_names which updates obsm/varm.dim_names?

I think this is a rather likely scenario, since many people distribute data in this form and I believe that the user should not do anything extra here to avoid the `ValueError` (e.g. manually syncing indices). WDYT?

gokceneraslan:
Fixes #459 .

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -226,10 +226,13 @@ class AxisArrays(AlignedActualMixin, AxisArraysBase):
         if axis not in (0, 1):
             raise ValueError()
         self._axis = axis
-        self.dim_names = (parent.obs_names, parent.var_names)[self._axis]
         self._data = dict()
         if vals is not None:
             self.update(vals)
+
+    @property
+    def dim_names(self):
+        return (self._parent.obs_names, self._parent.var_names)[self._axis]
 
 
 class AxisArraysView(AlignedViewMixin, AxisArraysBase):
@@ -243,7 +246,10 @@ class AxisArraysView(AlignedViewMixin, AxisArraysBase):
         self._parent = parent_view
         self.subset_idx = subset_idx
         self._axis = parent_mapping._axis
-        self.dim_names = parent_mapping.dim_names[subset_idx]
+
+    @property
+    def dim_names(self):
+        return self.parent_mapping.dim_names[self.subset_idx]
 
 
 AxisArraysBase._view_class = AxisArraysView
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -214,6 +214,10 @@ class AxisArraysBase(AlignedMapping):
             )
         return super()._validate_value(val, key)
 
+    @property
+    def dim_names(self) -> pd.Index:
+        return (self.parent.obs_names, self.parent.var_names)[self._axis]
+
 
 class AxisArrays(AlignedActualMixin, AxisArraysBase):
     def __init__(
@@ -230,10 +234,6 @@ class AxisArrays(AlignedActualMixin, AxisArraysBase):
         if vals is not None:
             self.update(vals)
 
-    @property
-    def dim_names(self) -> pd.Index:
-        return (self._parent.obs_names, self._parent.var_names)[self._axis]
-
 
 class AxisArraysView(AlignedViewMixin, AxisArraysBase):
     def __init__(
@@ -249,7 +249,7 @@ class AxisArraysView(AlignedViewMixin, AxisArraysBase):
 
     @property
     def dim_names(self) -> pd.Index:
-        return self.parent_mapping.dim_names[self.subset_idx]
+        return super().dim_names[self.subset_idx]
 
 
 AxisArraysBase._view_class = AxisArraysView
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -247,10 +247,6 @@ class AxisArraysView(AlignedViewMixin, AxisArraysBase):
         self.subset_idx = subset_idx
         self._axis = parent_mapping._axis
 
-    @property
-    def dim_names(self) -> pd.Index:
-        return super().dim_names[self.subset_idx]
-
 
 AxisArraysBase._view_class = AxisArraysView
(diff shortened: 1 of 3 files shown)


gokceneraslan:
image

ivirshup:
Can you add a test that would catch the case you showed in #459?

gokceneraslan:
Done I guess.

anndata/_core/aligned_mapping.py
```diff
@@ -243,7 +246,10 @@ def __init__(
         self._parent = parent_view
         self.subset_idx = subset_idx
         self._axis = parent_mapping._axis
-        self.dim_names = parent_mapping.dim_names[subset_idx]
+
+    @property
+    def dim_names(self) -> pd.Index:
+        return self.parent_mapping.dim_names[self.subset_idx]
```
ivirshup, line 252: Shouldn't this just be the same as for the `Actual` class?

ivirshup: So this could just be a property on the parent class

ivirshup (changes requested):
Thanks for catching this:

Two things left to do:

- [ ] Move `dim_names` property to `AxisArraysBase` class
- [ ] Add bug fix release note

anndata/_core/aligned_mapping.py
gokceneraslan: I am a bit confused, is there already a dim_names in AxisArraysBase: 

https://github.com/theislab/anndata/pull/461/files/bac91c25d502ed72daa499b07b08d198302acb3e#diff-af8b8ed3c983f1ed77b81030eb9d80df8136e0640324f180354e6e0d3b49e7e7R198

gokceneraslan: Here I mean:

https://github.com/theislab/anndata/blob/master/anndata/_core/aligned_mapping.py#L198

If yes, where does it come from?

ivirshup: Are you asking how ` AxisArraysBase` can refer to the `dim_names` attribute? If so:

The idea is that there will never actually be an instance of `AxisArraysBase`. The class is only meant to be inherited from so `AxisArrays` and `AxisArraysView` can share code.

gokceneraslan: Hmm I see. But then things like `_axis` and `dim_names` must be a "requirement" in the AlignedMapping class, no?

gokceneraslan: What I mean is: if I were to write another python package where I define classes inheriting from `AxisArraysBase`, I wouldn't know that I have to have `_axis` and `dim_names` in my class.

anndata/_core/aligned_mapping.py
```diff
@@ -249,7 +249,7 @@ def __init__(
 
     @property
     def dim_names(self) -> pd.Index:
-        return self.parent_mapping.dim_names[self.subset_idx]
+        return super().dim_names[self.subset_idx]
```
ivirshup, line 252: I don't think you need to redefine this property, since the indices of the `parent` should already be subsetted (the parent `AnnData` of a `AlignedView` should always be a view)

anndata/_core/aligned_mapping.py
ivirshup: True, but we don't export `AxisArraysBase`, so it's a bit use-at-your-own-risk. For now, I'd call these classes strictly internal.

gokceneraslan:
> * [ ]  Add bug fix release note

Is there are place to write release notes for the next release?

ivirshup:
`docs/release-latest.rst`. You might need to rebase on master, since I just committed a modification to that an hour or so ago.

ivirshup (approved):

Issue #331, opened 2020-02-27
TypeError: No conversion path for dtype: dtype('<U10')  Above error raised while writing key 'obs' of <class 'h5py._hl.files.File'> from /.


lila167:
Hi, I have problem with writing anndata into a file using the newer version of Anndata and Scanpy. 
In the following code, I read .h5ad and try to write it and it works. However when I set the .obs, I can't write my anndata anymore. With the older versions, it used to work. Has something changed that I am missing? Thanks.

```python
adata = sc.read(results_file+"adata_raw.h5ad")
adata_proc = adata.copy()
adata_proc.write(results_file+"./adata_proc.h5ad")

for s in meta_old.columns.values:
    adata_proc.obs[s] = meta_old[s].tolist()
for s in meta_raw.columns.values:
    adata_proc.obs[s] = meta_raw[s].tolist()
adata_proc.write(results_file+"./adata_proc.h5ad")
```

```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/.local/lib/python3.6/site-packages/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)

~/.local/lib/python3.6/site-packages/anndata/_io/h5ad.py in write_dataframe(f, key, df, dataset_kwargs)

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

/usr/lib64/python3.6/site-packages/h5py/_hl/attrs.py in __setitem__(self, name, value)
     94         """
---> 95         self.create(name, data=value, dtype=base.guess_dtype(value))
     96 

/usr/lib64/python3.6/site-packages/h5py/_hl/attrs.py in create(self, name, data, shape, dtype)
    170             if use_htype is None:
--> 171                 htype = h5t.py_create(original_dtype, logical=True)
    172                 htype2 = h5t.py_create(original_dtype)  # Must be bit-for-bit representation rather than logical

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

TypeError: No conversion path for dtype: dtype('<U10')

The above exception was the direct cause of the following exception:

TypeError                                 Traceback (most recent call last)
<ipython-input-266-204bf022b1f7> in <module>
----> 1 adata_proc.write(results_file+"./adata_proc.h5ad")

[...]

Issue #330, opened 2020-02-27
adata.X = None raises “incompatible shape” error


flying-sheep:
If we only want to support `del adata.X`, we need to change the pbmc tutorial.

ivirshup:
Do you want to allow `del adata.X`?

flying-sheep:
If `adata.X = None` should be legal, `del adata.X` should be too, right?

But yeah, right now there’s no way to do what the tutorial wants to …

ivirshup:
I forget why this came up. I don't think we should allow `del adata.X` or `adata.X = None`.

Can we close this?

LustigePerson:
I stumbled on the same thing while running your 3k PBMC tutorial. Are you planning to allow `del adata.X` or `adata.X = None`? Or is this functionality gone?

ivirshup:
I would say the tutorial is out of date. I don't think we currently support having no value for `X`, since it's pretty central to how AnnData objects work.

PR #463, opened 2020-11-23
Allow `del adata.X`, `adata.X = None`

rcannood:
Feature proposal for #464.

Need to test:
- [x] Getter (`test_base.py > test_x_is_none()`)
- [x] Setter (`test_base.py > test_x_is_none()`)
- [x] Deleter (`test_base.py > test_x_is_none()`)
- [x] Initializer (`test_base.py > test_x_is_none()`)
- [ ] Views
    * [ ] Check that .X = None doesn't change other view behavior
    * [ ] a[idx].X doesn't do anything weird
    * [ ] copying views still works
- [ ] _subset_inplace
- [ ] Copying
- [ ] Backed mode
- [ ] Concatenation
- [x] Transpose
- [ ] Methods which default to values from X, e.g. obs_vector, to_df.
- [ ] IO (seems to work, needs testing)
    * [ ] Maybe specialized writers like to_csvs will need some attention here?
- [ ] Raw (pretty similar set of things to be checked)
    * [ ] Initializer
    * [ ] Backed
    * [ ] Copies
    * [ ] Concatenation
    * [ ] IO

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -627,7 +627,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
         if value is None:
             if self.isbacked:
-                raise ValueError("Cannot currently remove data matrix from backed object.")
+                raise ValueError(
+                    "Cannot currently remove data matrix from backed object."
+                )
             if self.is_view:
                 self._init_as_actual(self.copy())
             self._X = None
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1280,6 +1280,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             else:
                 return m.T
 
+        if X is not None:
+            dtype = X.dtype
+        else:
+            dtype = "float32"
+
         return AnnData(
             t_csr(X),
             obs=self.var,
@@ -1292,7 +1297,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             varp=self.obsp.copy(),
             filename=self.filename,
             layers={k: t_csr(v) for k, v in self.layers.items()},
-            dtype=self.X.dtype.name # problem: X might be none so dtype might not be available
+            dtype=dtype,
         )
 
     T = property(transpose)
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -627,7 +627,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
         if value is None:
             if self.isbacked:
-                raise ValueError(
+                raise NotImplementedError(
                     "Cannot currently remove data matrix from backed object."
                 )
             if self.is_view:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -625,11 +625,6 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     @X.setter
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
-        if not isinstance(value, StorageType.classes()) and not np.isscalar(value):
-            if hasattr(value, "to_numpy") and hasattr(value, "dtypes"):
-                value = ensure_df_homogeneous(value, "X")
-            else:  # TODO: asarray? asanyarray?
-                value = np.array(value)
         if value is None:
             if self.is_view:
                 raise ValueError(
@@ -639,6 +634,12 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 raise ValueError("Not implemented.")
             self._X = None
             return
+        if not isinstance(value, StorageType.classes()) and not np.isscalar(value):
+            if hasattr(value, "to_numpy") and hasattr(value, "dtypes"):
+                value = ensure_df_homogeneous(value, "X")
+            else:  # TODO: asarray? asanyarray?
+                value = np.array(value)
+
         # If indices are both arrays, we need to modify them
         # so we don’t set values like coordinates
         # This can occur if there are succesive views
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -626,10 +626,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     @X.setter
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
         if value is None:
-            if self.is_view:
-                self._init_as_actual(self.copy())
             if self.isbacked:
                 raise ValueError("Not implemented.")
+            if self.is_view:
+                self._init_as_actual(self.copy())
             self._X = None
             return
         if not isinstance(value, StorageType.classes()) and not np.isscalar(value):
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1273,20 +1273,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             )
 
         def t_csr(m: sparse.spmatrix) -> sparse.csr_matrix:
-            if m is None:
-                return None
-            elif sparse.isspmatrix_csr(m):
-                return m.T.tocsr()
-            else:
-                return m.T
-
-        if X is not None:
-            dtype = X.dtype
-        else:
-            dtype = "float32"
+            return m.T.tocsr() if sparse.isspmatrix_csr(m) else m.T
 
         return AnnData(
-            t_csr(X),
+            X=t_csr(X) if X is not None else None,
             obs=self.var,
             var=self.obs,
             # we're taking a private attributes here to be able to modify uns of the original object
@@ -1297,7 +1287,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             varp=self.obsp.copy(),
             filename=self.filename,
             layers={k: t_csr(v) for k, v in self.layers.items()},
-            dtype=dtype,
+            dtype=self.X.dtype.name if X is not None else "float32",
         )
 
     T = property(transpose)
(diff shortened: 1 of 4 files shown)


ivirshup:
I don't think we currently allow `X` to be `None` (#331).

I have come across a case recently where this would be useful, though I think it would take some doing to make the entire codebase work again if we allowed this. `X` being an array with `shape` is pretty baked into the data model at this point.

Maybe this could be opened as a feature request issue, and you could pitch your use case?

rcannood:
Thanks for your input.

`X` being `None` is already part of the unit tests: [test_base.py#L46-L47](https://github.com/theislab/anndata/blob/cf915f51f6d53ee621e9c200cd0aa56fc82d8661/anndata/tests/test_base.py#L46-L47). Also, if `X` would not be allowed to be `None`, why is there a check in the code I changed? ;)

A possible use case I have is when working with read-only h5ad files: if your input is immutable, it may be interesting to exclude X from the output file to reduce the file size.

ivirshup:
At one point it was allowed, but I can say for certain that a lot of code has been written since then that does not expect `X` to be `None`.

For example:

```python
a = ad.AnnData(X=None, shape=(100, 200))

a.copy()
```

<details>
<summary> Traceback </summary>


```pytb
<ipython-input-12-02bb7baf10a6> in <module>
----> 1 a.copy()

~/github/anndata/anndata/_core/anndata.py in copy(self, filename)
   1428                 X = _subset(self._adata_ref.X, (self._oidx, self._vidx)).copy()
   1429             else:
-> 1430                 X = self.X.copy()
   1431             # TODO: Figure out what case this is:
   1432             if X is not None:

AttributeError: 'NoneType' object has no attribute 'copy'
```

</details>

At one point I think I saw `_X` being `None` as an indicator for if the object was in backed mode.

If this is something we were to support more, it would need a fair amount of testing.

rcannood:
Alright :) I hereby turn this PR into a feature request. I'd like `X` to be optional. I mean, it [already is](https://github.com/theislab/anndata/blob/cf915f51f6d53ee621e9c200cd0aa56fc82d8661/anndata/_core/anndata.py#L283), but apparently more thorough testing is needed to verify that it works properly. 

What is the opinion of the major anndata developers? :+1: or :-1:?

If we agree `X` should be able to be `None`, we can use this PR to add tests and fix the code where necessary.

ivirshup:
Oops, I linked the wrong issue before. #330 is where this has come up before. Also I'm going to open this discussion as a new issue so it's more visible.

anndata/tests/test_base.py
```diff
@@ -581,3 +581,28 @@ def assert_eq_not_id(a, b):
...
+    shape = (3, 5)
+    adata = AnnData(None, uns=dict(test=np.array((3, 3))), shape=shape)
+    assert adata.X is None
+    assert adata.shape == shape
+
+    # test transpose
+    adataT = adata.transpose()
+    assert_equal(adataT.shape, (5, 3))
+    assert_equal(adataT.obsp.keys(), adata.varp.keys())
+    assert_equal(adataT.T, adata)
```
ivirshup, line 608: I think this might actually be a good time to expand the tests for transpose. Mind if we do this in a separate PR?

ivirshup: #468 should set the basis for this

rcannood:
> Plus it's much easier to review and merge small PRs. Let's keep that PR's scope to just fixing up the .X property.

@ivirshup Alright, shall we merge this PR in order to resolve the first box in #467?

anndata/_core/anndata.py
```diff
@@ -1268,7 +1273,17 @@ def transpose(self) -> "AnnData":
...
+            else:
+                return m.T
+
+        if X is not None:
+            dtype = X.dtype
+        else:
+            dtype = "float32"
 
         return AnnData(
             t_csr(X),
```
ivirshup, line 1289: Instead adding all that stuff to `t_csr`, could this be: `t_csr(X) if X is not None else None`

My thinking here is that I'm not sure if we should keep the behavior of `t_csr`, so I don't want to put more logic in it.

rcannood: Fair enough!

anndata/_core/anndata.py
```diff
@@ -683,6 +684,10 @@ def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
                 f"need to be {self.shape}."
             )
 
+    @X.deleter
+    def X(self):
+        self.X = None
```
ivirshup, line 689: With other attributes, deleting them when an object is a view causes it to become an "actual" object. This happens for any modification. For example:

```python
In [1]: import scanpy as sc
pbmc
In [2]: pbmc = sc.datasets.pbmc3k_processed()

In [3]: v = pbmc[:, :50]

In [4]: v
Out[4]: 
View of AnnData object with n_obs × n_vars = 2638 × 50
    obs: 'n_genes', 'percent_mito', 'n_counts', 'louvain'
    var: 'n_cells'
    uns: 'draw_graph', 'louvain', 'louvain_colors', 'neighbors', 'pca', 'rank_genes_groups'
    obsm: 'X_pca', 'X_tsne', 'X_umap', 'X_draw_graph_fr'
    varm: 'PCs'
    obsp: 'distances', 'connectivities'

In [5]: del v.obsm

In [6]: v
Out[6]: 
AnnData object with n_obs × n_vars = 2638 × 50
    obs: 'n_genes', 'percent_mito', 'n_counts', 'louvain'
    var: 'n_cells'
    uns: 'draw_graph', 'louvain', 'louvain_colors', 'neighbors', 'pca', 'rank_genes_groups'
    varm: 'PCs'
    obsp: 'distances', 'connectivities'
```

That should happen here too. Could you add a test similar to `test_attr_deletion` which checks for this?

I think it's fair to say that `adata.X = None` should be equivalent to `del adata.X`, so you could also make sure that works when the `AnnData` is a view? It looks like it's set to throw an error right now.

ivirshup: Oh, the lines where I should have commented this are 628-631, where:

```python
        if value is None:
             if self.is_view:
                 raise ValueError(
                    "Copy the view before setting the data matrix to `None`."
                )
```

rcannood: See 72618c3 -- is this what you meant?

anndata/_core/anndata.py
```diff
@@ -627,9 +627,7 @@ def X(self) -> Optional[Union[np.ndarray, sparse.spmatrix, ArrayView]]:
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
         if value is None:
             if self.is_view:
-                raise ValueError(
-                    "Copy the view before setting the data matrix to `None`."
-                )
+                self._init_as_actual(self.copy())
```
ivirshup, line 630: I think this should error if the view is of a backed object, since I'm not sure we have the logic to deal with that at the moment.

anndata/tests/test_base.py
```diff
@@ -118,6 +118,15 @@ def test_attr_deletion():
     assert_equal(full, empty, exact=True)
 
 
+def test_not_view_after_deletion():
+    full = gen_adata((30, 30))
+    for attr in ["X", "obs", "var", "obsm", "varm", "obsp", "varp", "layers", "uns"]:
+        view = full[:, :20]
+        assert view.is_view
+        delattr(view, attr)
+        assert not view.is_view
```
ivirshup, line 127: Could this instead be a modification of `test_view_delattr` in `test_views.py`?

ivirshup: In general, I think `test_base.py` covers too much API surface, and has too much in it. When possible, try to add tests in more specific files. Should make it easier to find tests in the future.

anndata/_core/anndata.py
ivirshup: I think these checks need to be flipped, so the lower error get's thrown, not an error from trying to copy.

At the moment:

```python
import anndata as ad

gen_adata((10, 20)).write_h5ad("tmp.h5ad")
b = ad.read_h5ad("tmp.h5ad", backed="r")
v = b[::2, ::2]
del v.X
```

```pytb
<ipython-input-11-6b92c05b346f> in <module>
----> 1 del v.X

~/github/anndata/anndata/_core/anndata.py in X(self)
    685     @X.deleter
    686     def X(self):
--> 687         self.X = None
    688 
    689     @property

~/github/anndata/anndata/_core/anndata.py in X(self, value)
    628         if value is None:
    629             if self.is_view:
--> 630                 self._init_as_actual(self.copy())
    631             if self.isbacked:
    632                 raise ValueError("Not implemented.")

~/github/anndata/anndata/_core/anndata.py in copy(self, filename)
   1460 
   1461             if filename is None:
-> 1462                 raise ValueError(
   1463                     "To copy an AnnData object in backed mode, "
   1464                     "pass a filename: `.copy(filename='myfilename.h5ad')`."

ValueError: To copy an AnnData object in backed mode, pass a filename: `.copy(filename='myfilename.h5ad')`.
```

Also the error should be more descriptive, and probably be an actual `NotImplementedError`.

Maybe:

```python
raise NotImplementedError("Cannot currently remove data matrix from backed object.")
```

rcannood: Done! :)

anndata/_core/anndata.py
```diff
@@ -625,20 +625,21 @@ def X(self) -> Optional[Union[np.ndarray, sparse.spmatrix, ArrayView]]:
...
     def X(self, value: Optional[Union[np.ndarray, sparse.spmatrix]]):
-        if not isinstance(value, StorageType.classes()) and not np.isscalar(value):
-            if hasattr(value, "to_numpy") and hasattr(value, "dtypes"):
-                value = ensure_df_homogeneous(value, "X")
-            else:  # TODO: asarray? asanyarray?
-                value = np.array(value)
         if value is None:
-            if self.is_view:
+            if self.isbacked:
                 raise ValueError(
```
ivirshup, line 630:
```suggestion
                raise NotImplementedError(
```

ivirshup:
Great, thanks for implementing this and making the suggestion!

PR #468, opened 2020-11-27
Minor refactor of transpose tests

ivirshup:
Move transpose test to it's own file, added some property based tests. They will still need to be expanded, but this makes that easier.

### DIFF
--- a/anndata/tests/test_transpose.py
+++ b/anndata/tests/test_transpose.py
@@ -1,0 +1,74 @@
+from scipy import sparse
+
+import pytest
+
+from anndata.tests.helpers import gen_adata, assert_equal
+
+
+def test_transpose_orig():
+    """
+    Original test for transpose, should be covered by more thorough tests below, but
+    keeping around just in case.
+    """
+    adata = gen_adata((5, 3))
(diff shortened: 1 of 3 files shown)


anndata/tests/test_transpose.py
```diff
@@ -25,15 +25,18 @@ def _add_raw(adata, *, var_subset=slice(None)):
...
+# * X is None should have the xfail marker removed
 # * Backed
 @pytest.fixture(
     params=[
         pytest.param(gen_adata((50, 20)), id="csr_X"),
         pytest.param(gen_adata((50, 20), sparse.csc_matrix), id="csc_X"),
         pytest.param(_add_raw(gen_adata((50, 20))), id="with_raw"),
+        pytest.param(
+            gen_adata((20, 10), X_type=None), id="None_X", marks=pytest.mark.xfail
+        ),
```
ivirshup, line 39: @rcannood, this should suffice for a transpose test. For getting transpose to work with `X=None`, I'd remove the `marks` from this case, and try to get this to pass.

rcannood: Done in #463.

PR #470, opened 2020-11-27
Load backed data into memory more conveniently

VolkerBergen:
Enables loading a backed anndata into memory via `AnnData.copy(filename=None)`, which previously required `filename` to be not `None` in backed mode.

Minimal example:
```
mem_adata = sc.datasets.blobs()
mem_adata.write(tmp_path)

backed_adata = ad.read(tmp_path, backed='r')
backed_to_mem = backed_adata[:, :5].copy()  # assert equal mem_adata[:, :5] 
```

Particularly useful to efficiently read large files only on a subset of genes.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1419,7 +1419,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     def copy(self, filename: Optional[PathLike] = None) -> "AnnData":
         """Full copy, optionally on disk."""
-        if not self.isbacked:
+        if not self.isbacked or filename is None:
             if self.is_view:
                 # TODO: How do I unambiguously check if this is a copy?
                 # Subsetting this way means we don’t have to have a view type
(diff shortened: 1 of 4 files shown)


falexwolf:
As discussed in person, great Volker!

We can also use this PR to talk about the following; given that backed may become something much more popular essentially as of now:
```
with ad.read(backed='r') as adata:
    # do stuff to adata while it's in backed    
```

Let's discuss with Isaac!

ivirshup:
Just talked with @falexwolf about this. Yeah, the current API is super not convenient, and I hadn't realized we didn't have a more ergonomic way to do this yet! I definitely agree that we need something like this, but think a new method (maybe `.to_memory`? This is what's used for `SparseDataset`) might be the right way to do it. Some arguments for this

`.copy` is called pretty frequently in code, and I don't think there are frequently guardrails for backed mode. This can mean code using a ton of memory when maybe it should have errored.

Is it weird that `.copy` now sometimes creates another backed object, and sometimes loads the whole thing into memory? It could make more sense for `.copy()` to make an "anonymous" backed object using a temporary file.

In the grander scheme of our API, I don't think I like how backed mode works at the moment. If "backed" describes the `AnnData` object and not just the state of its elements, I think it should probably be its own class. In this case, I think a method like `.to_memory` would be the right choice for that class.

For prior art: `xarray` also allows [lazy loading of data through `dask`](http://xarray.pydata.org/en/stable/dask.html#using-dask-with-xarray). Their method for loading an object to memory is [`.load`]( http://xarray.pydata.org/en/stable/generated/xarray.Dataset.load.html#xarray.Dataset.load).

falexwolf:
After calling among all 3 of us: `.to_memory()` it is!

(And I'll just remain sad forever about the impossibility of the exact same code running in memory and in backed. 😢 😆 )

PR #471, opened 2020-11-27
anndata.__sizeof__()

VolkerBergen:
determines the correct size of an anndata object:
- iterates over elements in obsm/varm etc.
- and account for correct size of sparse matrices

Minimal example
```
adata = sc.datasets.pbmc3k()
X = adata.X
print('size (current):', X.__sizeof__())
print('size (after):', X.data.nbytes + X.indptr.nbytes + X.indices.nbytes)
adata.__sizeof__(show_stratified=True)
```

Output
```
size (current): 32
size (after): 18305876

Size of .X     : 17.46 MB
Size of .obs   : 0.27 MB
Size of .var   : 5.53 MB

24379414
```

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -555,7 +555,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             else:
                 s = get_size(getattr(self, attr))
             if s > 0 and show_stratified:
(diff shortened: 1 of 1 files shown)


falexwolf:
Cool! 😄

mxposed:
I think this still misses the `raw` attribute's size, not sure if intended or not

PR #473, opened 2020-12-01
Write obsm keys to zattrs

keller-mark:
Hi, we are interested in using the zarr-based output of `.write_zarr`. Currently the `obsm` keys are not written to any metadata file. It would help if these keys were written to `obsm/.zattrs`

### DIFF
--- a/anndata/_io/zarr.py
+++ b/anndata/_io/zarr.py
@@ -72,6 +72,8 @@ def write_attribute_zarr(f, key, value, dataset_kwargs=MappingProxyType({})):
 
 
 def write_mapping(f, key, value: Mapping, dataset_kwargs=MappingProxyType({})):
+    group = f.create_group(key)
+    group.attrs["column-order"] = list(value.keys())
     for sub_k, sub_v in value.items():
         if not isinstance(key, str):
             warn(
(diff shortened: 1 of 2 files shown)


ivirshup:
Could you explain the problem, and how this solves it?

keller-mark:
Sure, the problem arises from the fact that we are using the Zarr outputs of `.write_zarr` in the web browser with zarr.js (to render interactive visualizations, for example with scatterplots after running UMAP and t-SNE with ScanPy)

When loading the zarr store, we want to obtain a list of the available embeddings. For instance, that there are two `obsm` keys called "X_umap" and "X_tsne" in the zarr store. Normally you could simply list the names of the `data.zarr/obsm/` subdirectory, for example to determine that there are subdirectories `data.zarr/obsm/X_umap` and `data.zarr/obsm/X_tsne`. However, most web servers do not list directory contents over HTTP for security reasons, so instead we need to list these in a known file, such as `data.zarr/obsm/.zattrs`

ilan-gold:
Just to chime in here, and this is something I can help with as well @keller-mark, but are there other places in the zarr store that it would make sense to do this @ivirshup where people may add things that should be "searchable"?

ivirshup:
@ilan-gold, I may not be entirely following here, but I suppose it should all be "searchable"? I think that it already is though, on the basis that the elements all have keys.

@keller-mark Right now I'm not sure I like adding this information to what anndata writes out. Now what happens if I've removed some elements or added some? Previously this was all kept up to date naturally by whether keys were present in the group. How bad is it if this isn't completely up to date?

Is this better solved by a function like [`consolidate_metadata`](https://zarr.readthedocs.io/en/stable/tutorial.html#consolidating-metadata) from `zarr`, or some other way of sending your app a schema of the object?

ivirshup:
Closing this, as I think you've managed to make this work without changes over here. LMK if that's not the case

PR #475, opened 2020-12-03
revert switch to flit

Koncopd:


flying-sheep (changes requested):
why?

Koncopd:
There is a discussion on slack. Just doing what already done for scanpy to have a uniform setup.

ivirshup (approved):

flying-sheep:
@ivirshup https://scanpy.slack.com/archives/CK717EH08/p1607015213001100

ivirshup:
I think the flit changes will be easier to evaluate in a PR than while they are on master. I also think that because these changes haven't been evaluated by anyone else on the team there is a good chance this will lead to more churn of bugfix PRs.

ivirshup:
Interesting point on this, the `pip install git+https://github.com/theislab/anndata` was failing ([build history](https://dev.azure.com/theislab/scanpy/_build/results?buildId=47&view=logs&j=5d8ea360-f0ee-555b-5164-7dd0f0ec5e6c)) for https://github.com/theislab/scanpy/pull/1529 until I merged this.

It was just on a hunch that I tried the build again (since this was the only change to anndata since that step worked a day ago, and I can't actually see the error logs), but it seems to have fixed this. Not sure how this could have caused that, but something we should check for in a new PR.

flying-sheep:
The build you linked to passed.

PR #478, opened 2020-12-04
Switch to flit

flying-sheep:
See theislab/scanpy#1527


ivirshup:
To check sure this won't cause CI problems after merging, can you start a pr on scanpy that makes the `anndata-dev` build point to this PR? We'll just delete that pr once this is ready.

flying-sheep:
Done, too bad such a nice issue number is wasted on it lol theislab/scanpy#1700

flying-sheep:
d339903 can probably be removed, see theislab/scanpy#1702

ivirshup (approved):
Just to be sure, did you not have to update the `azure` setup?

flying-sheep:
No need it’s just standard pip usage and simply works. If we want, we can add a build check later like scanpy has

PR #496, opened 2021-01-25
Enable codecov

ivirshup:
Testing enabling codecov (will be useful for some file format refactoring)


ivirshup:
It looks like this might need to be on master so there are base reports to enable most of the functionality.

PR #508, opened 2021-02-15
Correct variable names for write_loom output

ACastanza:
Currently write_loom outputs incorrect attribute names that are incompatible with downstream analysis. This change corrects the variable names to be the expected ones.


ivirshup:
Can you provide examples of where this causes issues?

@VolkerBergen, any thoughts on this? I think you're a bit more familiar with the `loom` format. Is this part of the `loom` spec?

PR #517, opened 2021-03-05
Fix typing of `read_h5ad`

flying-sheep:
Small thing I noticed. Maybe other functions need to be updated in a similar way.


ivirshup (commented):
Is this something we want to say we support? My understanding was this didn't work is many cases ([see note here](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects)). Also, do we have any tests for this case?

PR #521, opened 2021-03-17
WIP: add _repr_html_() method to AnnData for nicer rendering in Jupyter

VolkerH:
## About
This is a draft PR that adds a `_repr_html_()` method to the AnnData class for nicer rendering in Jupyter.
This is heavily inspired from similar functionality in dask and re-uses some of dasks code for generating SVG/HTML.
The new functionality is best described by this screenshot:

image

## WIP status

Currently this is still very much in draft status, there are no unit tests, the formatting is a bit wonky and I have not looked at visualizing all aspects of an AnnData object as I am a newcomer to anndata and have only used a very limited subset.

Before I invest more time into this I just wanted to see, whether something like this is of interest at all.
If you think it is worthwhile, there's more work ahead, such as shortening very long lists of variable or observation keys with `...` entries (similar to pandas).

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -588,19 +588,18 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         else:
             return self._gen_repr(self.n_obs, self.n_vars)
 
-
-    def _repr_html_(self) -> str:    
+    def _repr_html_(self) -> str:
         overview = []
-        
+
         if self.is_view:
             overview.append("is a view")
 
         overview.append(f"n_obs: {self.n_obs}")
         overview.append(f"n_vars: {self._n_vars}")
-        overview.append(svg_anndata(self.n_obs,self.n_vars))
+        overview.append(svg_anndata(self.n_obs, self.n_vars))
 
         cols = [make_single_column_table("AnnData object", overview)]
-        
+
         for attr in [
             "obs",
             "var",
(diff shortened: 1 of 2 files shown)


VolkerH:
Arrggh ... i just saw #346 by @ivirshup 
I had searched the code base and the issues for `repr_html` and didn't find anything before coding up the above, but only after submitting this PR did it occur to me to look into existing PRs. Well, maybe there is something useful in here anyway.

ivirshup:
Sorry about the late response on this! As you can tell from the other PR, I'd definitely be interested in seeing something like this for AnnData. But that's a bit stalled.

If this is something you'd like to work on I'd be happy for the help on it! I'd also like to hear what you think are the most useful aspects for this kind of feature.

> there's more work ahead, such as shortening very long lists of variable or observation keys

I rather liked the way xarray handles this with dropdown ([example](http://xarray.pydata.org/en/stable/examples/weather-data.html)), so you can expand the features you're interested in looking at (and getting more information about them).

VolkerH:
Hi,
thanks for taking the time to reply.
I saw that your other PR uses the drop-down approach. 

TBH, I have not worked with AnnData/scanpy much so far, so I am not sure what the most useful aspects for most users would be. For me, it was important to see the shape of the data and the keys that are present in one glance, that's why I went with what I implemented. I can see that this will run into limitations if there are many keys and that the pull down-type information can hold a lot more additional information, but will always require additional interaction. I am not sure I can devote enough time to work on the drop-down approach.

PR #527, opened 2021-03-24
Sum by method (to create pseudobulk)

vals:
I implemented a helper method for AnnData that allow a user to sum gene expression in `.X` based on groups in columns of `.obs`.

The primary use case for this is 'pseudo-bulk' style analysis. Cells can be grouped by e.g. donor, experimental condition, and cell type. Then the UMI counts can be studied as if there were from a population of cells rather than individual cells.

Here is an example:
```
> import anndata
> adata = anndata.read('/mnt/efs/h5ad/External/Hrvatin et al 2018/GSE102827_merged_all_raw.h5ad')
> adata
AnnData object with n_obs × n_vars = 65539 × 25187
    obs: 'stim', 'sample', 'maintype', 'celltype', 'subtype', 'sample_idx'
    var: 'required', 'highly_variable', 'means', 'dispersions', 'dispersions_norm', 'highly_variable_nbatches', 'highly_variable_intersection'
> sum_adata = adata.sum_by(['stim', 'sample', 'celltype'])
> sum_adata
AnnData object with n_obs × n_vars = 909 × 25187
    obs: 'stim', 'sample', 'celltype'
    var: 'required', 'highly_variable', 'means', 'dispersions', 'dispersions_norm', 'highly_variable_nbatches', 'highly_variable_intersection'
> print(sum_adata.obs.head())
                   stim   sample celltype
0h-B1_1_0h-ExcL4     0h  B1_1_0h    ExcL4
0h-B1_1_0h-ExcL5_3   0h  B1_1_0h  ExcL5_3
0h-B1_1_0h-ExcL5_2   0h  B1_1_0h  ExcL5_2
0h-B1_1_0h-ExcL6     0h  B1_1_0h    ExcL6
0h-B1_1_0h-ExcL23    0h  B1_1_0h   ExcL23
> sum_adata.X
<909x25187 sparse matrix of type '<class 'numpy.float32'>'
	with 7374777 stored elements in Compressed Sparse Row format>
```

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1270,6 +1270,37 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def __len__(self) -> int:
         return self.shape[0]
 
+    def sum_by(self, group_list: List) -> "AnnData":
+        """
+        Group observations by the columns names in `group_list` and sum expression in `X`.
+
+        Params
+        -------
+        group_list: List of columns names in `.obs` to group observations by.
+
+        Returns
+        -------
+        A new `AnnData` where `obs` are individual groups and `X` is summed by group.
+        """
+        groupy_object = self.obs.groupby(group_list, observed=True)
+        N_obs = groupy_object.ngroups
+        N_var = self.X.shape[1]
+        X = sparse.lil_matrix((N_obs, N_var))
+        
+        group_names = []
+        index_names = []
+        row = 0
+        for group_columns, idx_ in groupy_object.indices.items():
+            X[row] = self.X[idx_].sum(0)
+            row += 1
+            group_names.append(group_columns)
+            index_names.append('-'.join(map(str, group_columns)))
+
+        X = X.tocsr()
+        obs = pd.DataFrame(group_names, columns=group_list, index=index_names)
+
+        return AnnData(X=X, obs=obs, var=self.var)
+    
     def transpose(self) -> "AnnData":
         """\
         Transpose whole object.
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1272,7 +1272,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     def sum_by(self, group_list: List) -> "AnnData":
         """
-        Group observations by the columns names in `group_list` and sum expression in `X`.
+        Group observations by the columns names in `group_list` and sum expression in `.X`.
 
         Params
         -------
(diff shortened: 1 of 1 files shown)


vals:
Any feedback on this?

ivirshup:
Thanks for the PR! This is a useful operation I would like to be easier, but I have some about implementing it this way. `anndata` should have a more general split-apply-combine system, which this operation should fit into very well. However, I think putting all of the logic into a single call isn't a manageable way to implement this. It would mean a new top level method for every aggregation, and over all a less flexible model than the established paradigm.

I would love to see more progress towards having a split-apply-combine framework here. I've prototyped this a bit myself [in a separate repo](https://github.com/ivirshup/mantis#group-by), and in a [separate language](https://github.com/rafaqz/DimensionalData.jl/issues/3#issuecomment-590061903). Something I find particularly interesting about your implementation is that you've combined into an `AnnData` object, as I wasn't sure what the right thing is to do here. Some topics I had thought about:

* What about reductions over elements other than X? E.g., per group medoids in PC space.
* What if the result is one dimensional – e.g. `ad.groupby("cluster").sum()`? E.g. the reduction also occurs over the variables.
* What elements should be preserved? Does this fit with grouping over both axes? E.g. some correlation between pathways per cluster (groupby clusters, groupby pathways).

I'll open an issue for discussion of this, since it's come up multiples times and should be tracked.

-------------

### Pseudobulk

I've done a little bit with creating pseudo bulk, and ran into some issue with this type of approach. It was a while ago, so the details are a bit fuzzy, but IIRC the issues mostly had to do with clusters having very different numbers of cells in them. This would basically result in "library complexity" issues, where clusters with large numbers of cells would have far more genes expressed. I'd worked around this taking multiple sample of a fixed number of cells from each cluster to make the pseudo bulk samples.

<details>
<summary> Here's the code I used for that: </summary>

```python
def simulate_bulk(adata, by, skip_groups=(), n_samples=1, n_draws=None):
    """
    Simulate bulk data by summing single cells
    """
    sampled_from = []
    if n_draws is None:
        n_draws = np.inf
    new = pd.DataFrame(index=adata.var_names)
    for key, subdf in adata.obs.groupby(by):
        if key in skip_groups:
            continue
        sampled_from.append(key)
        subset = adata[subdf.index]
        for i in range(n_samples):
            sampled = subset.X[
                np.random.choice(
                    np.arange(len(subset.obs_names)),
                    size=min(n_draws, len(subset.obs_names)), 
                    replace=True)
            ].sum(axis=0)
            new[f"{key}_{i}"] = np.ravel(sampled)
    new_obs = pd.DataFrame(
        {
            by: list(chain.from_iterable(
                repeat(k, n_samples) for k in sampled_from
            )),
        },
        index=new.columns
    )
    if pd.api.types.is_categorical_dtype(adata.obs[by]):
        new_obs[by] = pd.Categorical(new_obs[by], categories=adata.obs[by].cat.categories)
    colorkey = f"{by}_colors"
    if colorkey in adata.uns:
        uns = {colorkey: adata.uns[colorkey]}
    else:
        uns = {}
    return sc.AnnData(new.T, obs=new_obs, uns=uns)
```

</details>

vals:
Thanks for the feedback!

I definitely agree that a general split-apply-combine framework would be best. (For some reason I thought there was one, but there wasn't so I patched in this method).

I also thought of having the sum_by method work along both obs and var. In particular thinking of e.g. gene families or so. But I didn't have an immediate use for it so I figured that could be left for another time.

Regarding what elements should be preserved, I also didn't think of a good solution. So I just throw everything not explicitly specified away. For a bit I was thinking you could do something similar to the `aggregate` function in Pandas and specify what to with which columns and matrices and where. This would be an interesting use case for a general split-apply-combine framework.

This is a pretty small and simple method. I thought it was kind of handy so figured it could be useful for people in lieu of a general split-apply-combine framework. For many questions pseudobulks + GLMs is a quick and easy way to get answers.

---

With respect to pseudo-bulk issues in general, it really only makes sense for UMI counts as the unit. Some randomization sounds like an interesting strategy too investigate heterogeneity. In my implementation I have done similar things by making a column of random categories for replicates per group in the group-by specification. When using with a GLM, as long as log library size is included as an offset, discrepancies in cell numbers between experimental conditions are reflected in the bulked library size, so it works out in the end.

PR #528, opened 2021-03-31
Add requirements file

ivirshup:
Testing to see if we can restore the `Packages` info for the project by adding the requirements.txt back.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,9 @@
+# auto-generated from pyproject.toml, DO NOT EDIT
+pandas>=1.1.1
+numpy>=1.16.5
+scipy~=1.0
+h5py
+natsort
+packaging>=20
+xlrd<2.0
+importlib_metadata>=0.7; python_version < "3.8"


flying-sheep:
You forgot to add a test that makes sure the requirements aren’t out of sync as we decided

ivirshup:
I want to see if it actually fixes the problem before adding a task to generate it/ make sure it's up to date

flying-sheep:
too bad one has to commit to master for this, but I understand

ivirshup:
Dependencies of our package seem to be back (if a little strange), but dependents are not. I wonder if making a release will trigger that again. It's not obvious to me if Dependabot is even what's being used here. Sadly, got no feedback on the support forum: https://github.community/t/projects-not-listing-dependencies-anymore/168530

flying-sheep:
Yeah, I’m not even sure if GitHub’s aquiring of depandabot means that anything fixed there ends up in GitHub, or if there’s still an independent version …

But I guess that’s what the OSS world gets for deciding for closed source infrastructure …

flying-sheep:
How is it autogenerated? Did you add the test checking if the file is synced?

PR #533, opened 2021-04-01
fix copying of backed files (closes #519)

### LINKED ISSUE #519: copying backed AnnData overwrites backing file of the original object with new file
When copying a backed AnnData object, this is what happens:
```
In [3]: test = ad.read_h5ad("rna.h5ad", backed=True)

In [4]: test2 = test.copy("rna2.h5ad")

In [5]: test2.filename                                                                                                                                                                                    
Out[5]: PosixPath('rna2.h5ad')

In [6]: test.filename                                                                                                                                                                                     
Out[6]: PosixPath('rna2.h5ad')
```
I don't know if this is intended, but it's certainly counterintuitive. I'd be happy to take a stab at a PR to fix this if it is indeed a bug.

ilia-kats:
don't overwrite the original file's filename with the new one

### DIFF
--- a/anndata/_io/h5ad.py
+++ b/anndata/_io/h5ad.py
@@ -116,8 +116,6 @@ def write_h5ad(
         write_attribute(f, "varp", adata.varp, dataset_kwargs=dataset_kwargs)
         write_attribute(f, "layers", adata.layers, dataset_kwargs=dataset_kwargs)
         write_attribute(f, "uns", adata.uns, dataset_kwargs=dataset_kwargs)
-    if adata.isbacked:
-        adata.file.open(filepath, "r+")
 
 
 def _write_method(cls: Type[T]) -> Callable[[H5Group, str, T], None]:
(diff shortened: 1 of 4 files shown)


anndata/_io/h5ad.py
```diff
@@ -87,8 +87,14 @@ def write_h5ad(
...
-    mode = "a" if adata.isbacked else "w"
-    if adata.isbacked:  # close so that we can reopen below
+    copy = (
+        not adata.isbacked
+        or not filepath.exists()
+        or not filepath.samefile(adata.filename)
+    )
+
+    mode = "a" if not copy else "w"
+    if not copy:  # close so that we can reopen below
```
ivirshup, line 97: I'm not sure this is necessary. What if we just didn't call `adata.file.open(filepath, "r+")` at the bottom?

ivirshup (changes requested):
Thanks for opening a PR for this!

* I think there might be an easier solution, basically, just don't reassign the file in this function. If that behaviour is needed it can be implemented in whatever function calls this one instead.
* Could this get some tests?

ivirshup (approved):
LGTM. Thanks!

PR #535, opened 2021-04-03
Update CI to be on par with scanpy

ivirshup:
- [x] Azure pipelines should have equivalent steps
  - [x] Tests
  - [x] Build
  - [x] Remove static analysis
- [x] pre-commit
- [x] Coverage

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,41 @@
+# Can't yet be moved to the pyproject.toml due to https://github.com/PyCQA/flake8/issues/234
+[flake8]
+max-line-length = 88
+ignore = # module imported but unused -> required for Scanpys API
+         F401,
+         # line break before a binary operator -> black does not adhere to PEP8
+         W503,
+         # line break occured after a binary operator -> black does not adhere to PEP8
+         W504,
+         # line too long -> we accept long comment lines; black gets rid of long code lines
+         E501,
+         # whitespace before : -> black does not adhere to PEP8
+         E203,
+         # missing whitespace after ,', ';', or ':' -> black does not adhere to PEP8
+         E231,
+         # module level import not at top of file -> required to circumvent circular imports for Scanpys API
+         E402,
+         # continuation line over-indented for hanging indent -> black does not adhere to PEP8
+         E126,
+         # E266 too many leading '#' for block comment -> Scanpy allows them for comments into sections
+         E262,
+         # inline comment should start with '# ' -> Scanpy allows them for specific explanations
+         E266,
+         # Do not assign a lambda expression, use a def -> Scanpy allows lambda expression assignments,
+         E731,
+         # allow I, O, l as variable names -> I is the identity matrix, i, j, k, l is reasonable indexing notation
+         E741
+ per-file-ignores =
+    # F811 Redefinition of unused name from line, does not play nice with pytest fixtures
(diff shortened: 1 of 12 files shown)


ivirshup:
The github action isn't triggering, but I think that's because CI didn't run on the first commit. I rebased to put the action yaml in the first commit, but that seems insufficient (https://github.com/theislab/scanpy/pull/1684#issuecomment-784837291).

.flake8
```diff
@@ -0,0 +1,41 @@
+# Can't yet be moved to the pyproject.toml due to https://gitlab.com/pycqa/flake8/-/issues/428#note_251982786
```
flying-sheep, line 1: @asottile moved the project to GitHub, losing all the metadata and breaking all the links, so idk which comment this linked to. The migrated issue is this one: https://github.com/PyCQA/flake8/issues/234

anndata/tests/test_base.py
```diff
@@ -467,8 +467,13 @@ def test_pickle():
...
 def test_to_df_dense():
-    df = adata_dense.to_df()
-    df = adata_dense.to_df(layer="test")
+    X_df = adata_dense.to_df()
+    layer_df = adata_dense.to_df(layer="test")
+
+    np.testing.assert_array_equal(adata_dense.layers["test"], layer_df.values)
+    np.testing.assert_array_equal(adata_dense.X, X_df.values)
+    pd.testing.assert_index_equal(X_df.columns, layer_df.columns)
+    pd.testing.assert_index_equal(X_df.index, layer_df.index)
```
flying-sheep, line 476: did this make it in here by accident?

flying-sheep (approved):
Looks great! I’m a bit concerned about all this config desyncing by making a change here or there, but idk how to fix it: `eslint` allows creating shared configs, I don’t think flake8 does.

flying-sheep:
Regarding the build step’s pip:

`python -m build` uses build environment isolation, i.e. it creates a whole new venv just for the build. That venv’s pip is the one that comes with the version of pip that comes with the `python`.

If that ever makes problems, we can use `--no-isolation`, but I think it’s generally a good idea to make sure it builds in a minimal environment containing only what we specify.

anndata/tests/test_base.py
ivirshup: Nah, just leaving the code better than I found it. This was throwing a flake8 error (assigning to variable that's never used) and instead of renaming the variables and leaving a bad test I made the test more useful.

ivirshup:
> it creates a whole new venv just for the build ... I think it’s generally a good idea to make sure it builds in a minimal environment containing only what we specify.

Is using a `venv` meaningfully contributing here if we're already making the build in a container 🤔?

--------------------

I'm going to squash this whole thing to see if I can trigger the github action, and will merge if there aren't any issues there.

.flake8
ivirshup: Updated link

ivirshup:
Hmm. I'm going to assume it's github being weird about recognizing the config

flying-sheep:
> Is using a venv meaningfully contributing here if we're already making the build in a container thinking?

If the build happens in a completely clean Python environment in that container, it’s superfluous.

If Azure preinstalls some useful stuff like Travis does, the isolation actually improves how thorough our tests are.

PR #537, opened 2021-04-06
Fix error when concatenating anndata with an axis of size 0

### LINKED ISSUE #526: Anndata.concatenate() fails when .X has a dimension of size 0
```
>>> from anndata import AnnData
>>> import numpy as np
>>> adata = AnnData(X=np.empty([1,0]))
>>> adata
AnnData object with n_obs × n_vars = 1 × 0
>>> AnnData.concatenate(adata, adata)
```

Error:
`IndexError: index 0 is out of bounds for axis 0 with size 0`

Running anndata '0.7.5'

The reason this is an issue for me is that the scirpy TCR package builds adata like that https://github.com/icbi-lab/scirpy/blob/507c943e50a2d2ba5b0011575030db85dd5b316b/scirpy/io/_convert_anndata.py#L81

A fix or workaround would be great, thanks

ivirshup:
Fixes #526

Tests are a bit weak

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -9,6 +9,7 @@ from enum import Enum
 from functools import partial, singledispatch
 from pathlib import Path
 from os import PathLike
+from textwrap import dedent
 from typing import Any, Union, Optional  # Meta
 from typing import Iterable, Sequence, Mapping, MutableMapping  # Generic ABCs
 from typing import Tuple, List  # Generic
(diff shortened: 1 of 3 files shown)


gokceneraslan:
Not exactly related to this but any idea why version looks old when I install anndata via pip git+https:

image

ivirshup:
@gokceneraslan, not sure. My guess would be that the master branch on my fork is pretty old, so somehow the version is being grabbed from there. It's also very possible there are no newer tags there, but I would've thought those would be in this branches history.

@flying-sheep has mentioned pip can sometimes cache out of date version numbers.

PR #538, opened 2021-04-06
Allow passing dimension names to and from loom

### LINKED ISSUE #509: Incorrect attribute names in AnnData write_loom()
Loom file created by anndata names the gene and sample attributes incorrectly. In the AnnData loom they are called "var_names" and "obs_names" containing the Gene IDs and Cell IDs respectively, but the expected name for the attribute "var_names" is "Gene" and for "obs_names", "CellID". This is what the attributes are named in the Loompy sample files, and are what is expected by, for example, seurat's function to load loom files.

If you create an attribute named "Gene" or "CellID", AnnData.write_loom seems to ignore those attributes.

I believe this change https://github.com/theislab/anndata/pull/508 will allow AnnData to write out files with the expected attribute names for downstream analysis.

ivirshup:
Fixes #509
Supersedes #508

Also did a little bit of consolidation in `read_loom`. 

## TODO:

- [ ] ~~Figure out what's going on with singleton values in `read_loom` (e.g. `cleanup=True` code)~~
  - These are uniform arrays stored in loom. Not sure why we have an argument to remove these, but it's opt-in so leaving it for now.
- [x] Release note

### DIFF
--- a/anndata/_io/read.py
+++ b/anndata/_io/read.py
@@ -1,16 +1,19 @@
 from pathlib import Path
 from os import PathLike, fspath
+from types import MappingProxyType
 from typing import Union, Optional, Mapping, Tuple
 from typing import Iterable, Iterator, Generator
 from collections import OrderedDict
 import gzip
 import bz2
+from warnings import warn
 
 import h5py
 import numpy as np
 import pandas as pd
 
 from .. import AnnData
+from ..compat import _deprecate_positional_args
 from .utils import is_float
 from .h5ad import read_h5ad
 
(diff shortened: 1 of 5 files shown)


VolkerBergen:
I would propose to change `obsm_names` to `obsm_mapping` or `obsm`, and add a minimal example in the docstring such as `obsm = {'X_umap': ['umap_1', 'umap_2']}` (same for `varm`). Otherwise this looks great!

VolkerBergen (commented):
I would propose to change `obsm_names` to `obsm_mapping` or `obsm`, and add a minimal example in the docstring such as `obsm = {'X_umap': ['umap_1', 'umap_2']}` (same for `varm`). Otherwise this looks great!

ivirshup:
Fixed! Thanks for the review.

PR #542, opened 2021-04-07
To memory review pass

ivirshup:
A bit of a belated review on the `.to_memory` method. Main changes:

* Renamed and simplified internal function a bit (`_create_anndata` -> `_mutated_copy`)
* Made `to_memory` work with `raw`
* Made `to_memory` work with dense backed arrays
* More tests
* Centralized subsetting of backed arrays (added `h5py.Dataset` and `SparseDataset` methods to `_subset`)
* Fixed bug in `assert_equals` where if the first argument didn't have raw, the second arguments `raw` wasn't checked

Could probably use a benchmark run.

TODO

- [x] Probably make `.to_memory` error if the object isn't backed.
- [x] Change log

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -609,7 +609,6 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             X = self.file["X"]
             if isinstance(X, h5py.Group):
                 X = SparseDataset(X)
-            # TODO: This should get replaced/ handled elsewhere
             # This is so that we can index into a backed dense dataset with
             # indices that aren’t strictly increasing
             if self.is_view:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -611,18 +611,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             # TODO: This should get replaced/ handled elsewhere
             # This is so that we can index into a backed dense dataset with
             # indices that aren’t strictly increasing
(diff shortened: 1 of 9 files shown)


ivirshup:
@VolkerBergen two quick questions about this:

### `raw`

Originally this didn't bring `.raw` from a backed object into memory. Was this a design decision or a technical limitation?

If it was about not wanting to read too much data into memory, `raw` can be `del`'d from the backed object before bringing it into memory.

### Warning vs error

Originally this warned if `to_memory` was called on a in memory object. I took a bit more conservative approach and made it error. My philosophy here is that if you plan on handling backed objects, and they need special handling, it should be done explicitly.

Does this sound reasonable?

VolkerBergen (approved):
`.raw` was more of a technical limitation (quite some effort to implement). Loading `raw` in memory would be useful when subsetting along `.obs`, e.g., when reading a small celltype from a huge atlas. In our case, `.raw` was not relevant, thus didn't follow that up. 

The error makes sense, of course. 

I have no objections with the implementation. Thanks, looks great!

anndata/_core/anndata.py
```diff
@@ -612,18 +612,8 @@ def X(self) -> Optional[Union[np.ndarray, sparse.spmatrix, ArrayView]]:
             # TODO: This should get replaced/ handled elsewhere
```
ivirshup, line 612:
```suggestion
```

ivirshup:
Great, thanks for the review!

PR #543, opened 2021-04-07
fix #540

### LINKED ISSUE #540: flake8 config error
There is a space in front of the `per-file-ignore` entry in the `flake8` config (added in https://github.com/theislab/scanpy/pull/1689). This is the same in scanpy, as the config here was just copied from there. This causes a parse failure which leads to the the `per-file-ignore`s becoming global instead. To demonstrate:

<details>
<summary> Parse failure </summary>

From the root of the repo

```python
>>> from flake8.api.legacy import get_style_guide
>>> get_style_guide().options.ignore
['#',
 'module',
 'imported',
 'but',
 'unused',
 '->',
 'required',
 'for',
 'Scanpys',
 'API',
 'F401',
 'W503',
 'W504',
 'E501',
 'E203',
 'E231',
 'E402',
 'E126',
 'E262',
 'E266',
 'E731',
 'E741',
 'per-file-ignores',
 '=',
 'tests/test*.py:',
 'F811',
 'anndata/compat/_overloaded_dict.py:',
 'F821',
 'anndata/tests/test_readwrite.py:',
 'E721']
```

</details>

Fixing the parse error reveals a number of flake8 errors which were accidentally being ignored. I came across this while adding back `F401`, since it should only be ignored for files which define a namespace.

It is kind of ironic that a tool for enforcement of code style is permissive to a fault about config parsing.

(ping @Zethson)

Zethson:
Signed-off-by: zethson <lukas.heumos@posteo.net>

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -25,9 +25,9 @@ ignore = # module imported but unused -> required for Scanpys API
          E731,
          # allow I, O, l as variable names -> I is the identity matrix, i, j, k, l is reasonable indexing notation
          E741
- per-file-ignores =
+per-file-ignores =
     # F811 Redefinition of unused name from line, does not play nice with pytest fixtures
-    tests/test*.py: F811
+    anndata/tests/test*.py: F811
     # F821 Undefined name, can't import AnnData or it'd be a circular import
     anndata/compat/_overloaded_dict.py: F821
     # E721 comparing types, but we specifically are checking that we aren't getting subtypes (views)
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -8,6 +8,7 @@ import pandas as pd
 from pandas.api.types import is_bool_dtype
 from scipy import sparse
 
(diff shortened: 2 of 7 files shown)


Zethson:
Didn't expect this to throw errors.

Issue: `AttributeError: partially initialized module 'anndata' has no attribute 'AnnData' (most likely due to a circular import)`
I'll have a look.

ivirshup:
There are going to be a lot of strings used for type annotation because they would otherwise be circular imports. Basically, functions which operate on `AnnData` objects, which are used in `AnnData` methods. This may be solved once we can bump minimum python to 3.7 and use `from __future__ import annotations`, which should be reasonably soon (May-ish?, next major release)

Zethson:
Ahh, I see. That's why they were strings before. I was already wondering.

flying-sheep:
This is obsolete, we use Ruff now

PR #547, opened 2021-04-09
Concat empty dim 1

### LINKED ISSUE #546: concat([a, b], axis=1, join="outer") errors when b.shape[1] == 0
@ivirshup the following code (from my poor man's reindex) still errors out, is this expected? (axis=0 version still works though)

```python
import scanpy as sc
import pandas as pd
import anndata as ad
import numpy as np

adata = sc.datasets.paul15()
labels = ['A', 'B', '0', '1']

ph = ad.AnnData(np.ones([len(labels), 0]), obs=pd.DataFrame(index=labels))   
ret = ad.concat([adata, ph], join='outer', axis=1)
```

```pytb
WARNING: In Scanpy 0.*, this returned logarithmized data. Now it returns non-logarithmized data.
... storing 'paul15_clusters' as categorical
Trying to set attribute `.uns` of view, copying.

<ipython-input-2-70a65516d436> in <module>
      8 
      9 ph = ad.AnnData(np.ones([len(labels), 0]), obs=pd.DataFrame(index=labels))
---> 10 ret = ad.concat([adata, ph], join='outer', axis=1)

~/Code/anndata/anndata/_core/merge.py in concat(adatas, axis, join, merge, uns_merge, label, keys, index_unique, fill_value, pairwise)
    820     )
    821 
--> 822     X = concat_arrays(
    823         [a.X for a in adatas], reindexers, axis=axis, fill_value=fill_value
    824     )

~/Code/anndata/anndata/_core/merge.py in concat_arrays(arrays, reindexers, axis, index, fill_value)
    425         )
    426     else:
--> 427         return np.concatenate(
    428             [
    429                 f(x, fill_value=fill_value, axis=1 - axis)

<__array_function__ internals> in concatenate(*args, **kwargs)

ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 2732 and the array at index 1 has size 4
```

_Originally posted by @gokceneraslan in https://github.com/theislab/anndata/issues/526#issuecomment-815870234_

ivirshup:
Fixes #546

The underlying issue was me making assumptions about which axis was going to be reindexed when one of the dimensions of the arrays was 0. This is because I used these "empty" arrays as a sentinel value when an element was missing, but it's indices should be filled.

## TODO:

- [x] add checks for other fill values being added correctly
  - [x] X, layers
  - [x] obsm, varm (these weren't really special cases, but there are issues around dataframe dtype promotion)
  - [ ] ~~obsp, varp~~ These weren't actually special cases.

I may need to draw this out. Additionally noticed a "bug" where pairwise entries are always converted to sparse matrices. I remember discussing this, and we had some reason for doing this (probably because we always expect these entires to be graphs), but we may want to revisit this.


anndata/_core/anndata.py
```diff
@@ -1124,7 +1124,8 @@ def _remove_unused_categories(self, df_full, df_sub, uns):
             if not is_categorical_dtype(df_full[k]):
                 continue
             all_categories = df_full[k].cat.categories
-            df_sub[k] = df_sub[k].cat.remove_unused_categories()
+            with pd.option_context("mode.chained_assignment", None):
+                df_sub[k] = df_sub[k].cat.remove_unused_categories()
```
ivirshup, line 1128: why isn't this different from what's on master?

ivirshup: Because my local clone got messed up...... Still not sure how, but my local `master` had started tracking my forks `master`

ivirshup:
Docs are failing due to sphinx 4 being out. I believe we can just remove `work_around_issue_6785` and pin sphinx above 4.0 to resolve this. Waiting on @flying-sheep to hear if this seems fine.

@gokceneraslan, could you try this out? I think I've solved this, but kept getting confused while writing the tests (concatenation where the result can be smaller is a weird case). Is there anything I forgot?

PR #554, opened 2021-04-15
Specs for all elements

ivirshup:
This PR adds specs for each element of an anndata object. That is, on disk each element will be have attributes `"encoding-type"` and `"encoding-version"`. For more info see this issue #555

## TODO:

- [x] Remove old code
- [x] Handling of dynamic types
  - For elements like `list`s, we convert them to arrays. But we handle arrays differently based on the type of their values. All Most of this system is build around `element-type -> on-disk-format`, but these types break this a little. Currently handling this by not setting `encoding-type`/ `encoding-version` if they're already set, so we can set them for the converted value.
  - Not 100% sure what I meant by this. Writing lists works, but they are converted to arrays. Should they come back out as lists, or is that getting too python specific?
- [x] AnnData level element
- [x] Zarr backend
  - [x] Zarr backend for new implementation
  - [x] Zarr backwards compatibility backend
- [x] Warnings for ancient anndata objects
- [x] More tests
  - ~[ ] Tests for SparseDataset indexing (sometimes it wasn't returning a `csr_matrix`~ out of scope
  - [x] Test encoding metadata for all types
- [x] Example cases for current formats
- [x] Docs
- ~[ ] Don't allow unlabelled elements in more recently written objects~
    - This means, all backwards compat is turned off if we know the file was written using a more current spec
    - Future task, figure out what to do with unrecognized elements.

## Maybe this PR, maybe not:

*I have decided: not*

~- [ ] Consolidate "modified" reading methods~
~- [ ] Figure out what the scope is for an individual resolver~
~- [ ] Partial IO? (probably not until during the 0.8.x release series)~

This might also be the right time to cut a 0.7 maintenance branch, and make master 0.8 specific.


ivirshup:
We now have warnings for when an element is found that doesn't have a spec. These are not user visible, but we should silence them further for the tests.

ivirshup:
Working on zarr implementation. Seems mostly good, but running into the json only being able to represent a limited set of values. 

Current problem: `bool`. Maybe I just encode these attributes as `0` and `1` for now?

ivirshup:
Huh, pretty sure I meant to call c0112fc something different. Like "start zarr support".

Still needs some work. Like most backwards compat and figuring out what I can delete.

ivirshup:
I'm having a bit of an existential crisis about how the encodings work for AnnData elements.

So we specify how some group or dataset in our store should be read into memory with the `encoding-type` and `encoding-version` fields. The question I'm facing is how specific should the `encoding-type` field be. Up until now it's basically been the amount of extra information needed to read the element. For example, `encoding-type` would be `"string-array"` for arrays of strings, but just `"array"` for `float`s, `int`s, and `bool`s – since those have one-to-one representations between hdf5 and python.

So, should we be more specific, and more like a real type system? E.g. `encoding-type: "array", "element-type": string`? `encoding-type: "array", "element-type": int32`? This leads to many more specifications needing to be written, and would be a greater change from how things are currently done. Do we need to have the whole type system encoded in the specification? If so, is this something that needs to be done now – or can it wait for a later PR?

flying-sheep:
I think with #662, there’s no real way around `encoding-type: "array", "element-type": "string"`, right?

ivirshup:
Long term yes, but I think this can go through without it. I think designing that system is worth a bit more care, and we may be able to base that design (or at least get help with it) from a "single cell data interchange" schema (will hopefully hear more on that in January).

The way around this is for now is basically just long encoding names. So `"string-array"`, `"datetime-array"`, `"nullable-datetime-array"`, etc.

PR #569, opened 2021-05-28
fix reading of fixed-size strings in data set attributes – hdf5.jl compat

ilia-kats:
This is needed for compatibility with other languages. E.g. HDF5.jl in Julia as well as hdf5r in R write string attributes as fixed-size strings, which are read as bytes by h5py. As far as I can tell, h5py does not offer an equivalent to `Dataset.asstr` for attributes, so we have to convert the bytes objects ourselves.


ivirshup (commented):
Thanks for the PR!

Two quick informational questions:

* Can you make little examples that create files which are being read wrong? I think I've figured it out for hdf5.jl, but haven't for the r library.
* Is there an upstream issue in h5py about this?

## Some changes I'd like to see

### Checking whether we should read it as a string

I think I would like to at least check if the attribute is meant to be a string, and if so, read it as a string. I want to leave this open because I think writing arbitrary binary blobs is useful here, and I don't want to read them in as strings by accident.

We can check if something is "string-like" with:

```python
attr_id = el.attrs.get_id(key)
```

We can get the encoding via:

```
attr_id.dtype.metadata
```

```
mappingproxy({'h5py_encoding': 'utf-8'})
```

But we should read it as a python string even if it says it's ascii, we could maybe choose to explicitly encode it as ascii. I'm not sure I see a reason to other than being pedantic.

### attribute access

Instead of using `_from_bytes` at specific points I think I'd rather have a `read_attribute` function that handles any necessary conversions. API could be either: `value = read_attribute(element.attrs, key)` or `attrs: dict = read_attributes(element)`. Just some wrapper function to encapsulate special handling of attributes that's used everywhere we need to access them.

### Tests

This needs some tests. Could be files you've written, but then I'd want to review the code that writes them. Alternatively, could create specific kinds of attributes using the lower level h5py api and ensure those types are read in correctly.

ilia-kats:
So for HDF5.jl this works:
```julia
using HDF5

file = h5open("test.h5", "cw")
dset = create_dataset(file, "hdf5.jl", collect(1:10))[1]
attrs = attributes(dset)
attrs["single_string"] = "teststring"
attrs["array_of_strings"] = ["string1", "string2"]
close(file)
```
The `single_string` attribute will then be a bytestring. As for hdf5r, I think you're right, it is being read as proper strings by h5py. I was using ViTables to debug, and ViTables showed the corresponding attributes as bytestrings. So now there are these possibilities:
- variable-length string with UTR-u encoding (h5py): works everywhere
- variable-length string with ASCII encoding (hdf5r): read as string by h5py, read as bytestring by ViTables
- fixed-length string (HDF5.jl): read as bytestring by h5py and ViTables

I am not aware of any h5py issue about this, and a quick search of their tracker didn't reveal anything related.

I'll try to implement the changes you suggested. Regarding tests, I think it would be sufficient to just write a bytestring as attribute using the high-level h5py API, this should have the same effect.

ivirshup:
> Regarding tests, I think it would be sufficient to just write a bytestring as attribute using the high-level h5py API, this should have the same effect.

I would eventually like the ability to write a `bytes` object to an attribute and have it come back as `bytes`. `hdf5` has the ability to represent this, so I'd like to keep the option open for now. Basically I would like to keep the distinction between "text" and "data". HDF5 makes this distinction in it's type system, so we can just lean on that.

> As for hdf5r, I think you're right, it is being read as proper strings by h5py.

Question about this, what encoding does hdf5r say it uses? I've seen issues with unicode strings which were annotated as ascii, but it wasn't clear to me where this was being set.

> I was using ViTables to debug

I've been finding the `HDF5.datatype` julia function quite helpful for seeing how hdf5 represents the data.

ilia-kats:
> Question about this, what encoding does hdf5r say it uses? I've seen issues with unicode strings which were annotated as ascii, but it wasn't clear to me where this was being set.

I haven't looked at the code, but it seems to use ASCII:
```
HDF5 "test.h5" {
ATTRIBUTE "single_string" {
   DATATYPE  H5T_STRING {
      STRSIZE H5T_VARIABLE;
      STRPAD H5T_STR_NULLTERM;
      CSET H5T_CSET_ASCII;
      CTYPE H5T_C_S1;
   }
   DATASPACE  SCALAR
   DATA {
   (0): "teststring"
   }
}
}
```
(created with
```
library(hdf5r)
file <- H5File$new("test.h5", mode="a")
dset <- file$create_dataset("hdf5r", 1:10)
dset$create_attr("single_string", "teststring", space=H5S$new("scalar"))
h5attr(dset, "array_of_strings") <- c("test1", "test2")
file$close_all()
```
)

ilia-kats:
OK, I tried to generalize the conversion function and I added tests. Note that the conversion function cannot be in `h5ad.py` due to circular imports: `_io/h5ad.py` imports `_core/anndata.py` imports `_core/sparse_dataset.py` imports `_io/utils.py` which pulls in `_io/h5ad.py` through `_io/__init__.py`

anndata/_io/h5ad.py
```diff
@@ -522,19 +523,20 @@ def read_group(group: h5py.Group) -> Union[dict, pd.DataFrame, sparse.spmatrix]:
     if "h5sparse_format" in group.attrs:  # Backwards compat
         return SparseDataset(group).to_memory()
 
-    encoding_type = group.attrs.get("encoding-type")
-    if encoding_type:
+    try:
```
ivirshup, line 526: I would rather check that that the key existed than be permissive about any KeyError. Could this be done with `if "encoding-type" in group.attrs` instead?

anndata/tests/test_readwrite.py
```diff
@@ -619,6 +620,24 @@ def test_write_string_types(tmp_path, diskfmt):
         write(adata_pth)
 
 
+@pytest.mark.parametrize(
+    "teststring", ["teststring", np.asarray(["test1", "test2", "test3"])]
```
ivirshup, line 624: What `dtype` are you expecting `np.asarray(["test1", "test2", "test3"])` to be read back in as?

anndata/compat/__init__.py
```diff
@@ -62,12 +62,37 @@ class Literal(metaclass=LiteralMeta):
...
+
+
 def _from_fixed_length_strings(value):
     """\
     Convert from fixed length strings to unicode.
 
     For backwards compatability with older h5ad and zarr files.
     """
+    if not hasattr(value, "dtype"):
+        return value
```
ivirshup, line 95: What case is this for? My understanding is that Julia will write `Array{String, N}` as variable length strings.

My general concern here is that I would rather not touch the backwards compat stuff unless we need to.

anndata/compat/__init__.py
```diff
@@ -62,12 +62,37 @@ class Literal(metaclass=LiteralMeta):
...
+    dtype = h5py.check_string_dtype(attr_id.dtype)
+    if dtype is None:
+        return attr
+    else:
+        if dtype.length is None:  # variable-length string, no problem
+            return attr
+        elif len(attr_id.shape) == 0:  # Python bytestring
+            return attr.decode("utf-8")
+        else:  # NumPy array
+            return attr.astype("U")
```
ivirshup, line 85: We generally don't use the numpy unicode type, instead opting for just arrays of python strings. The numpy unicode type is really weird (utf-32, not supported by h5py), and we haven't really had a use for fixed length text data. Also numpy just doesn't have good tools for working with text data, generally we hand this to pandas which uses object arrays.

anndata/compat/__init__.py
```diff
@@ -62,12 +62,37 @@ class Literal(metaclass=LiteralMeta):
...
+    attr = attrs[name]
+    attr_id = attrs.get_id(name)
+    dtype = h5py.check_string_dtype(attr_id.dtype)
+    if dtype is None:
+        return attr
+    else:
+        if dtype.length is None:  # variable-length string, no problem
+            return attr
+        elif len(attr_id.shape) == 0:  # Python bytestring
+            return attr.decode("utf-8")
```
ivirshup, line 83: I'm not sure I see how this case happens.

```python
import h5py

f = h5py.File("h5py.h5", "w")
f.attrs["python_bytes"] = b"some bytes"
attr_id = f.attrs.get_id("python_bytes")
h5py.check_string_dtype(attr_id.dtype)
```

```
string_info(encoding='ascii', length=None)
```

This would just hit the previous case, right?

ivirshup (changes requested):
Thanks for the updates. Here's my review, looks good overall, mostly had some questions about different cases.

anndata/compat/__init__.py
ilia-kats:
```
import h5py
f = h5py.File("test.h5", "w")
f.attrs.create("test", "some bytes", dtype=h5py.h5t.string_dtype(length=15))
h5py.check_string_dtype(f.attrs.get_id("test").dtype)
```
gives
```
 string_info(encoding='utf-8', length=15)
```
and
```
f.attrs["test"]
```
gives 
```
b'some bytes'
```

anndata/compat/__init__.py
ilia-kats: The problem is that you can get fixed-length strings from HDF5 files:
```
f.attrs.create("test", np.asarray(["test1", "test2"]), dtype=h5py.h5t.string_dtype(length=15))
f.attrs["test"]
```
gives 
```
array([b'test1', b'test2'], dtype='|S15')
```
and now it needs to be converted to something that we can work with. `np.astype("O")` doesn't work here, and you do use `np.astype("U")` in a couple of places, e.g. [here](https://github.com/theislab/anndata/blob/master/anndata/utils.py#L151) [here](https://github.com/theislab/anndata/blob/master/anndata/compat/__init__.py#L82) [here](https://github.com/theislab/anndata/blob/master/anndata/_core/merge.py#L783) (the last of which, `astype(str)` also results in a `U` dtype)

anndata/compat/__init__.py
ilia-kats: Ah, apologies, I think I was trying to use that function at some point and forgot to remove my changes. I'll remove this.

anndata/tests/test_readwrite.py
ilia-kats: depending on how it was written, it can be either `dtype=object` (for variable-length UTF-8 and ASCII strings, in the latter case the array will contain bytestring objects) or `dtype=S` for fixed-length strings

anndata/compat/__init__.py
ivirshup: > The problem is that you can get fixed-length strings from HDF5 files

We can get fixed length strings from hdf5 files, but we won't keep them as fixed length strings in memory. I'd say our goal is not to have a one-to-one mapping with hdf5, but to be able to represent useful data types. There isn't really a useful fixed length text type in our dependencies or a use case for it, so we don't have a representation for it.

> and now it needs to be converted to something that we can work with. np.astype("O") doesn't work here

This should work in this case:

```python
import codecs, h5py

string_type = h5py.check_string_type(attrs.get_id(key).dtype)
[codecs.decode(x, string_type.encoding) for x in attrs[key]]
```

Passing the encoding isn't even necessary, since hdf5 only uses ascii and utf-8 so utf-8 should always work. So just

```python
[codecs.decode(x) for x in attrs[key]]
```

> in a couple of places, e.g. here here here

As for those examples:

* `convert_dictionary_to_structured_array` is never called as far as I can tell. We've also moved significantly away from structured arrays since then. I also think this code predates me working on AnnData.
* `_from_fixed_length_strings` is specifically for backwards compatibility. The main goal of that function is to convert fixed length strings to python `str`s anyways. We definitely would not recommend putting anything in structured arrays these days, as they are such a massive pain to work with.
* `np.arange(len(adatas)).astype(str)` I hadn't realized this was fixed length, but I'm not sure it ultimately matters since the only way it's getting into the `AnnData` object is through

```python
    label_col = pd.Categorical.from_codes(
        np.repeat(np.arange(len(adatas)), [a.shape[axis] for a in adatas]),
        categories=keys,
    )
```

Where:

```python
label_col.categories.dtype
# dtype('O')
```

Could totally be replaced with `keys = [str(k) for k in range(len(adatas))]`.

anndata/compat/__init__.py
ivirshup: Ahh, yeah, this makes sense, not sure why I didn't get this at the time.

anndata/tests/test_readwrite.py
ivirshup: Even if it's ascii, I would like it to be read in as a string. It could be a string encoded as ascii, but it would be equivalent to read it encoded as utf-8.

My main concern with this exact test is that `np.asarray(["test1", "test2", "test3"])` creates a numpy's unicode strings, which h5py (and hdf5) doesn't support. I think it would make sense for this to be more like:

```suggestion
    "teststring", ["teststring", np.asarray(["test1", "test2", "test3"], dtype=object)]
```

ivirshup (approved):
LGTM

Thanks for reminding me about this! I would like to get some h5ad files written from R and Julia into a cloud directory somewhere so we can test against them.

Issue #578, opened 2021-06-17
Import error from `write_zarr()` is not propagated correctly


romanhaa:
**Situation**

Try to write an anndata object to ZARR format with `adata.write_zarr('test.zarr')` when `write_zarr()` function cannot be imported (not all dependencies are installed).

**Observed behavior**

When the `write_zarr()` function cannot be imported, e.g. because the `numcodecs` module hasn't been installed, I receive the following error:

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    adata.write_zarr('test.zarr')
  File "/Users/roman/hyve/python_envs/scanpy_fresh/lib/python3.7/site-packages/anndata/_core/anndata.py", line 1969, in
write_zarr
    write_zarr(store, self, chunks=chunks)
  File "/Users/roman/hyve/python_envs/scanpy_fresh/lib/python3.7/site-packages/anndata/_io/write.py", line 22, in write_
zarr
    raise e
NameError: name 'e' is not defined
```

**Expected behavior**

Receive an error such as the following:

```
ModuleNotFoundError: No module named 'numcodecs'
```

Without this import error, debugging (understanding what modules are missing) can be a little tricky.

PR #579, opened 2021-06-17
Fix propagation of ImportError for `write_zarr`

romanhaa:
See [issue #578](https://github.com/theislab/anndata/issues/578).

This fixes the propagation of the import error for the `write_zarr` function when it cannot be imported, e.g. because dependencies are not available.

### DIFF
--- a/anndata/utils.py
+++ b/anndata/utils.py
@@ -225,25 +225,25 @@ class DeprecationMixinMeta(type):
         ]
 
 
-def import_function(module, name):
+def import_function(module: str, name: str):
     """\
     Try to import function from module. If the module is not installed or
     function is not part of the module, it returns a dummy function that raises
     the respective import error once the function is called. This could be a
     ModuleNotFoundError if the module is missing or an AttributeError if the
     module is installed but the function is not exported by it.
+
+    Params
+    -------
+    module
+        Module to import from. Can be nested, e.g. "sklearn.utils".
+    name
+        Name of function to import from module.
     """
     try:
         module = __import__(module, fromlist=[name])
-        try:
-            func = getattr(module, name)
-        except AttributeError as e:
-            error = e
-
-            def func(*_, **__):
-                raise error
-             
-    except ImportError as e:
+        func = getattr(module, name)
+    except (ImportError, AttributeError) as e:
         error = e
 
         def func(*_, **__):
--- a/anndata/utils.py
+++ b/anndata/utils.py
@@ -223,3 +223,30 @@ class DeprecationMixinMeta(type):
             for item in type.__dir__(cls)
             if not is_deprecated(getattr(cls, item, None))
         ]
+
+
+def import_function(module, name):
+    """\
+    Try to import function from module. If the module is not installed or
+    function is not part of the module, it returns a dummy function that raises
+    the respective import error once the function is called. This could be a
+    ModuleNotFoundError if the module is missing or an AttributeError if the
+    module is installed but the function is not exported by it.
+    """
+    try:
+        module = __import__(module, fromlist=[name])
+        try:
+            func = getattr(module, name)
+        except AttributeError as e:
+            error = e
+
+            def func(*_, **__):
+                raise error
+             
+    except ImportError as e:
+        error = e
+
+        def func(*_, **__):
+            raise error
+
+    return func
--- a/anndata/utils.py
+++ b/anndata/utils.py
@@ -1,6 +1,6 @@
 import warnings
 from functools import wraps, singledispatch
-from typing import Mapping, Any, Sequence, Union
+from typing import Mapping, Any, Sequence, Union, Callable
 
 import h5py
 import pandas as pd
@@ -225,7 +225,7 @@ class DeprecationMixinMeta(type):
         ]
 
 
-def import_function(module: str, name: str):
+def import_function(module: str, name: str) -> Callable:
     """\
     Try to import function from module. If the module is not installed or
     function is not part of the module, it returns a dummy function that raises
@@ -240,8 +240,9 @@ def import_function(module: str, name: str):
     name
         Name of function to import from module.
     """
+    from importlib import import_module
     try:
-        module = __import__(module, fromlist=[name])
+        module = import_module(module)
         func = getattr(module, name)
     except (ImportError, AttributeError) as e:
         error = e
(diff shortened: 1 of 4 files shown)


ivirshup:
Thanks for the PR! Could you add a test making sure this throws the right error? It'd be good to make sure the error stays helpful.

romanhaa:
Sure, will look into that.

romanhaa:
Alright, I implemented a new function called `import_function()` (found in `utils.py`) that, when provided with a module name and a function name, attempts to import the function and, when that's not possible because either the module doesn't exist or the function isn't exported by the module, will return a dummy function with the provided name, which itself returns the import error when being called. In the context of the `write_zarr()` function, it works just like the solution I had before, but this can be used in other cases if needed. I added a few tests that should make sure the correct errors are returned. Let me know what you think.

anndata/utils.py
```diff
@@ -223,3 +223,30 @@ def is_deprecated(attr):
...
+
+
+def import_function(module, name):
+    """\
+    Try to import function from module. If the module is not installed or
+    function is not part of the module, it returns a dummy function that raises
+    the respective import error once the function is called. This could be a
+    ModuleNotFoundError if the module is missing or an AttributeError if the
+    module is installed but the function is not exported by it.
+    """
```
ivirshup, line 235:
```suggestion
def import_function(module: str, name: str) -> Callable:
    """\
    Try to import function from module. If the module is not installed or
    function is not part of the module, it returns a dummy function that raises
    the respective import error once the function is called. This could be a
    ModuleNotFoundError if the module is missing or an AttributeError if the
    module is installed but the function is not exported by it.
    
    Params
    -------
    module
        Module to import from. Can be nested, e.g. "sklearn.utils".
    name
        Name of function to import from module.
    """
```

anndata/utils.py
```diff
@@ -223,3 +223,30 @@ def is_deprecated(attr):
...
+            def func(*_, **__):
+                raise error
+             
+    except ImportError as e:
+        error = e
+
+        def func(*_, **__):
+            raise error
+
+    return func
```
ivirshup, line 252: Why not:

```suggestion
    try:
        module = __import__(module, fromlist=[name])
        func = getattr(module, name)
    except (ImportError, AttributeError) as e:
        error = e

        def func(*_, **__):
            raise error

    return func
```

ivirshup (commented):
This looks fantastic, thanks!

Apart from the pre-commit checks:

* Could you document the arguments to `import_function`? I've added some suggestions.
* Does the suggested simplification to the code change anything meaningful about the function?

anndata/utils.py
romanhaa: Great suggestion! Looks a lot cleaner like this.

romanhaa:
Thank you for the good suggestions, they definitely improve the code. Hopefully the new commit passes all the checks.

romanhaa:
Note: I removed the `-> Callable` from the function definition as that resulted in build issues (`name 'Callable' is not defined`).

anndata/utils.py
```diff
@@ -223,3 +223,30 @@ def is_deprecated(attr):
...
+
+    Params
+    -------
+    module
+        Module to import from. Can be nested, e.g. "sklearn.utils".
+    name
+        Name of function to import from module.
+    """
+    try:
+        module = __import__(module, fromlist=[name])
```
ivirshup, line 244:
```suggestion
    """
    from importlib import import_module 
    try:
        module = import_module(module)
```

From the docs for `__import__`, I think this might be the way to go.

```
__import__(name, globals=None, locals=None, fromlist=(), level=0) -> module

Import a module. Because this function is meant for use by the Python
interpreter and not for general use, it is better to use
importlib.import_module() to programmatically import a module.
```

ivirshup (commented):
I've added another possible change to the `__import__` call after looking at it's documentation. Seems like `importlib.import_module` may be the preferred way to do that.

Could you also add a release note for this? Should be good to go then.

> Note: I removed the -> Callable from the function definition as that resulted in build issues (name 'Callable' is not defined).

You would have to add `Callable` to what's imported from `typing` for this.

romanhaa:
Thank you for the explanations. Regarding the release notes, do you mean I should document the change in a new release, i.e. `v0.7.7`, in the `docs/release-latest.rst` file? I would've expected that this will only be part of the next release, but I couldn't find a place where the changes that will be included in the next release are collected.

anndata/utils.py
```diff
@@ -223,3 +223,31 @@ def is_deprecated(attr):
...
+
+    Params
+    -------
+    module
+        Module to import from. Can be nested, e.g. "sklearn.utils".
+    name
+        Name of function to import from module.
+    """
+    from importlib import import_module
+    try:
```
ivirshup, line 244:
```suggestion
    from importlib import import_module

    try:
```

docs/release-latest.rst
```diff
@@ -1,6 +1,13 @@
 .. role:: small
 .. role:: smaller
 
+0.7.7 :small:`21 June, 2021`
```
ivirshup, line 4:
```suggestion
On `master` :small:`the future`
```

docs/release-latest.rst
```diff
@@ -1,6 +1,13 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. rubric:: Bug fixes
+
+- Fixed propagation of import error when importing `write_zarr` but not all dependencies are installed :pr:`579` :smaller:`R Hillje` :smaller:`I Virshup`
```
ivirshup, line 9:
```suggestion
- Fixed propagation of import error when importing `write_zarr` but not all dependencies are installed :pr:`579` :smaller:`R Hillje`
```

ivirshup (approved):
We collect the notes in the same file, though it looks like I forgot to create a section for new notes after the last release. Thanks for adding that, and thanks for this PR!

docs/release-latest.rst
```diff
@@ -1,6 +1,13 @@
 .. role:: small
 .. role:: smaller
 
+On `master` :small:`the future`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ivirshup, line 5:
```suggestion
On `master` :small:`the future`
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #583, opened 2021-06-30
add pytest github actions

rcannood:
This PR adds a GitHub Actions CI for testing anndata with pytest.

Uncomment the last section if so desired :)

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,39 @@
+name: pytest
+
+on: [push]
+
+jobs:
+  pytest:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: |
+          python -m pip install --upgrade pip
+          pip --version
+      - name: Install dependencies
+        run: |
+          pip install ".[dev,test]"
+      - name: Test with pytest
+        run: |
+          pytest
+      - name: Lint with flake8
+        run: |
+          # pip install --upgrade flake8
+          # stop the build if there are Python syntax errors or undefined names
+          # flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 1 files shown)


rcannood:
Ugh. Only after creating the PR do I see that anndata is being tested by azure devops.

@ivirshup What do you think about this PR? The benefit of running it in this way is that people who make forks for the theislab/anndata repository (e.g. me) can automatically run the pytest CI.

ivirshup:
Ha, I was wondering what was up when you sent that link before. Thanks for going through the effort of setting this up! CI is definitely pretty key to the development process.

I'm not sure I want to have multiple CI systems, just because debugging CI is never fun and I don't want to do it twice as often. I do appreciate that there's value to distributing the CI load onto forks; however, I don't think we're having CI load issues at the moment. There are some features we like in Azure like being able to see plots from tests for scanpy, being able to use custom instances for CI in the future. I am also a bit concerned with frequently [github actions has issues](https://twitter.com/githubstatus).

How useful is this to you? Would you be fine with just using the CI on this repo?

rcannood:
I understand that the Azure CI has clean benefits for theislab.

I am definitely sufficiently happy only using this CI for my own purposes. It'll be a bit annoying to merge my any potential contributions to theislab/anndata, because I need to remove the CI before creating a PR :P

I think there may be some middle ground to this conversation. It would be possible to merge this PR; but then [disable the pytest workflow from running in this repository](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow). This way, theislab/anndata can rely on Azure CI but potential contributors can use GitHub Actions CI. 

If any issues arise that seem to be due to github actions rather than anndata itself, feel free to ping me.

What do you think? :)

ivirshup:
Main thing I think I'm not understanding here: Why not just open a draft PR and use our CI instead of having the CI run on a fork?

rcannood:
I understand your concern regarding having multiple CI systems running in parallel. Disabling the GitHub CI in the theislab/anndata repository would solve this issue.

I think there are benefits to using CI for non-theislab developers working on separate forks:

* CI runs as soon as they start creating commits. If they eventually decide to create a PR, the proposed changes will (/should) have passed the CI checks.
* If all potential contributors would start creating PRs to see if their code passes CI, the repository would become bloated with PRs of code that the potential contributor does not want to merge just yet.
* As a non-theislab developer, I can't actually see the output of Azure CI in a PR. If the CI fails, fixing bugs or debugging the CI would have to be done by you as opposed to the potential contributor, since they can't see the error message.

What is your opinion on this? Would you like to propose me to change something in the PR?

rcannood:
Closing this PR

flying-sheep:
we switched to GH actions in #1906 because it’s faster, more convenient to access for us and others, and we use it everywhere else (making fixes easier)

PR #585, opened 2021-07-06
Test Old CI

ilan-gold:


ivirshup:
I think this is going to an incompatibility with the latest pandas release that needs fixing. Don't worry about that test.

ivirshup:
Should now be fixed on master

PR #588, opened 2021-07-12
Replace xlrd with openpyxl

ivirshup:
Attempt to fix broken tests


ivirshup:
Docs failing due to https://github.com/theislab/scanpydoc/issues/35

Will check if that's good to go, otherwise will pin sphinx

ivirshup:
Pinning sphinx for now

Issue #515, opened 2021-02-26
Multimodal Data


atarashansky:
Hi there. I’m working on extending SAM to allow joint integration of multimodal datasets. For SAM, I refactored the code to work with a dictionary of ‘assays’, where each assay is a separate AnnData. This works well-enough, but it’s kind of a headache to output the associated h5ad files since I need to make one for each assay.

It would be really great if an AnnData could encapsulate multiple AnnData’s, via something similar to the `.raw` attribute. 

AnnData could have a default assay: this is default AnnData (`adata`), the contents of which will be displayed if we output it in a Jupyter cell, for example. Other assays (including an inplace reference to the default AnnData) will be stored as attributes (e.g. `adata.rna`, `adata.atac`, `adata.cite`, etc), with assay names specified by the user as they load the data. The user should also be able to change which assay is the default assay.

I’m not too familiar with the inner-workings of AnnData, is something like this possible?

atarashansky:
I just saw the discussion in #237.

I think this solution could work very well (assuming there are no obvious implementation obstacles that I’m unaware of) as none of the functions in scanpy/anndata would have to be modified — they would just automatically operate on whatever assay is specified as default. If the user wants to change what AnnData the functions operate on, they would just switch the identity of the default assay.

flying-sheep:
Hi, thanks for that PR and this discussion. I’m going to close this in favor of #237

If we reconsider multimodal support in AnnData, it will be talked about there.

PR #602, opened 2021-08-03
multimodal update

atarashansky:
I was toying around with a implementation of the idea I proposed in #515. This is highly experimental and really only meant to serve as a proof-of-concept. I'm doing some really ugly things by modifying the `__dict__` attribute in place, but it gets the job done for a prototype.

In general, the idea is to be able to combine multiple modalities (each with their own dimensionality and metadata) into one AnnData object with a shared `obs` index. 

I added the following to the AnnData class in `_core/anndata.py`:

Functions:
- `add_multimodal_layer`: Adds the input AnnData as an attribute to `self`
- `del_multimodal_layer`: Deletes the AnnData attribute associated with the input name
- `set_default_multimodal_layer`: Replaces `self` with the specified AnnData attribute, and adds `self` as an attribute with the name `new_name`. The default multimodal layer is used by all of scanpy's functions by default. I would also update all of scanpy's functions to accept a keyword argument like `mlayer=None` to signify which AnnData attribute the functions should operate on.

Attributes:
- `_multimodal_layers`: Keeps track of the attribute names for the other AnnData objects corresponding to the various loaded modalities

I also modified `__getitem__` to propagate `obs` indexing from `self` to all of the AnnData attributes tracked by `_multimodal_layers`.

One design question I had was if all multimodal AnnData's should all reference the same `obs` DataFrame (i.e. writing to one writes to all others).

So, for example:
```python
adata1 = anndata.read_h5ad(filename1) # RNA
adata2 = anndata.read_h5ad(filename2) # ADT

adata1.add_multimodal_layer(adata2,'ADT')

print(adata1)
print(adata1.ADT)
print(adata1[:100]) #indexing first 100 rows
print(adata1[:100].ADT) # this index is propagated to the ADT layer

adata1.set_default_multimodal_layer('ADT','RNA') #ADT is set as default, and the previous default is set to attribute RNA
print(adata1) # This is now ADT
print(adata1.RNA) # This is the RNA layer
```

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -300,6 +300,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         oidx: Index1D = None,
         vidx: Index1D = None,
     ):
+        self._multimodal_layers = []
+
         if asview:
             if not isinstance(X, AnnData):
                 raise ValueError("`X` has to be an AnnData object.")
@@ -858,6 +860,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     @obs.deleter
     def obs(self):
         self.obs = pd.DataFrame(index=self.obs_names)
+        for mlayer in self._multimodal_layers:
+            self.__dict__[mlayer].obs = pd.DataFrame(index=self.obs_names)
 
     @property
     def obs_names(self) -> pd.Index:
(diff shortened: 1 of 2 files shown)


ivirshup:
Hey! Thanks for starting to work on this.

At the moment, I don't think we're going to be adding multimodal support to the core anndata object. Instead, I think we're going to recommend people use the [`MuData`](https://muon.readthedocs.io/en/latest/io/mudata.html) object developed by @gtca which uses a collection of `AnnData` instances over a semi-overlapping set of observations.

That said, I don't think anyone on the team is against an alternative implementation being tried out, we just think this would have to be developed outside of the core anndata package.

PR #604, opened 2021-08-11
Fix vertical scrolling in docs

### LINKED ISSUE #601: Can't scroll

michalk8:
closes #601


flying-sheep:
Thank you, but this solution looks ugly:

image

Also, fixing this should happen in scanpydoc, not here.

michalk8:
Nope, completely, would this be better?
anndata

PR #611, opened 2021-09-03
references to the same object should be equal

adamgayoso:
This fixes an issue with #272 where two references of the same object are not equal. Strangely we need this in scvi-tools for pytorch lightning related reasons.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -591,10 +591,13 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     def __eq__(self, other):
         """Equality testing"""
-        raise NotImplementedError(
-            "Equality comparisons are not supported for AnnData objects, "
-            "instead compare the desired attributes."
-        )
+        if self is other:
+            return True
+        else:
+            raise NotImplementedError(
+                "Equality comparisons are not supported for AnnData objects, "
+                "instead compare the desired attributes."
+            )
 
     @property
     def shape(self) -> Tuple[int, int]:
--- a/anndata/tests/test_base.py
+++ b/anndata/tests/test_base.py
@@ -428,6 +428,7 @@ def test_n_obs():
 def test_equality_comparisons():
     adata1 = AnnData(np.array([[1, 2], [3, 4], [5, 6]]))
     adata2 = AnnData(np.array([[1, 2], [3, 4], [5, 6]]))
+    assert adata1 == adata1
     with pytest.raises(NotImplementedError):
         adata1 == adata1
     with pytest.raises(NotImplementedError):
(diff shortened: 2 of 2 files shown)


adamgayoso:
@ivirshup any chance we can get this in?

flying-sheep:
In which cases would this behavior be useful?

flying-sheep:
I think we should simply move the existing equality check from https://github.com/theislab/anndata/blob/19ade7ad134a773e44db50f92022ccd632f27368/anndata/tests/helpers.py#L378 into a submodule and then call it from both the test helpers and `__eq__`

ivirshup:
@adamgayoso, did you end up looking into whether your check could be `is` instead of `==`? And could you link to the issue you're running into?

> I think we should simply move the existing equality check from assert_adata_equal

@flying-sheep, that check is fairly inefficient in many cases, and I wouldn't recommend people use it for equality checking. The whole assertion thing doesn't exactly make sense.

Also, I'm wondering if `__eq__` should work more like it does for `xarray.Dataset`?

flying-sheep:
> The whole assertion thing doesn't exactly make sense.

how so?

> Also, I'm wondering if __eq__ should work more like it does for xarray.Dataset?

how is that?

ivirshup:
@flying-sheep could you open an issue for starting a test utility module? I think it makes more sense for additions to be at least tracked there.

> > The whole assertion thing doesn't exactly make sense.
> how so?

With the assertions we are failing with what should be a useful error message for figuring out what's different. Comparison requires less information at each step and shouldn't error in most cases.

> how is that? returning a comparison object showing where differences are like a numpy array?

Yeah, it's broadcast.

ivirshup:
After going over this a bit with Adam, the check that's causing problems is [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/8d810d614472b8c68caaac6cd413210569c5df63/pytorch_lightning/trainer/data_loading.py#L247):

```python
{name for name, p in params.items() if name in attrs and p.default != attrs[name]}
```

Which actually requires we do `!=` which is more complicated cause we can't just do:

```python
if a is b:
    return True
else:
    raise NotImplementedError()
```

Which was the conservative solution I was thinking of. Also, this would appear to mean you can't use any array like as an attribute of a DataLoader with pytorch lightning?

flying-sheep:
> @flying-sheep could you open an issue for starting a test utility module? I think it makes more sense for additions to be at least tracked there.

Done:  #646

ivirshup:
I'm going to close this, since we're (a) not sure what equality means here, (b) the underlying library doesn't seem to want this kind of thing to work/ should be handled higher up.

Sound fair @adamgayoso?

adamgayoso:
> I'm going to close this, since we're (a) not sure what equality means here, (b) the underlying library doesn't seem to want this kind of thing to work/ should be handled higher up.

I agree. Thanks!

PR #613, opened 2021-09-07
Split function

Koncopd:
Moved the PR https://github.com/theislab/scanpy/pull/1939 here.


Koncopd:
Hm, docs fail with 
`py:docstring of anndata._core.split.split:9:py:class reference target not found: pandas.core.series.Series`

Koncopd:
@ivirshup merging it?

ivirshup:
Sorry! I haven't had a chance to take a look yet, still busy with meetings getting familiar with the lab. Hopefully will have time to look soon!

ivirshup (commented):
Just took a look again. I'm still don't love the `groups` and `others_key` arguments. To me, these are more operations on the categorical vector which are generally useful, rather than operations specific to a groupby type operation.

Would you be alright with splitting these out of the function and instead providing these transformations as a utility function? I've outlined on in my review comments.

Also, I would like to name this function `splitby` or `split_by` just to keep the similarities with `groupby`. Other functions in python and the pydata ecosystem called `split` deal with splitting an object into contiguous regions at cut points, while this splits the objects by values.

I forget where we landed on this, but did you still want this to be a method?

Koncopd:
`groups` and `others_key` were the main motivation to write this function. Also note that they are optional.

ivirshup:
I agree that they are useful, but I think they're quite generally useful.

The most useful thing these arguments do is allow renaming of categories, without having to provide a 1:1 mapping like pandas wants. I wish pandas let us do this efficiently, but it does not. I think this is useful enough to be broken out to be stand alone.

<details>
<summary> Quick union_cats implementation </summary>

```python
def reverse_map(d: "dict[Any, list[Any]]") -> "dict[Any, Any]":
    out = {}
    for v, l in d.items():
        for k in l:
            if k in out:
                raise KeyError(f"Specified mapping for {k} more than once.")
            out[k] = v
    return out


def union_cats(
    c: pd.Series,
    to_union: "dict[str, list]",
    *,
    name: Optional[str]=None,
    union_rest: Optional[str] = None
) -> pd.Series:
    """
    Params
    ------
    c
        Categorical array to operate on
    to_union
        Mapping of {new_categories: [categories, to, combine]}
    name
        Name of returned Series
    union_rest
        If passed, categories not enumerated in to_union are combined with this
        name.
    """
    # Inefficient implementation due to intermediate string arrays, but should work
    mapping = reverse_map(to_union)
    if union_rest is not None:
        out = c.map(lambda x: mapping.get(x, union_rest))
    else:
        out = c.map(lambda x: mapping.get(x, x))
    if name is not None:
        return pd.Series(out, name=name, dtype="category")
    else:
        return pd.Series(out, dtype="category")
```

</details>

We could write this:

```python
ad.split(
    pbmc,
    "bulk_labels",
    groups={
        "T": ['CD4+/CD25 T Reg', 'CD4+/CD45RA+/CD25- Naive T', 'CD8+ Cytotoxic T'],
        "Mono": ['CD14+ Monocyte', "CD34+", "Dendritic"]
    },
    others_key="Others"
)
```

Which gives us a useful grouping of observations. If we instead wrote:

```python
pbmc.obs["highlevel"] = union_cats(
    pbmc.obs["bulk_labels"],
    {
        "T": ['CD4+/CD25 T Reg', 'CD4+/CD45RA+/CD25- Naive T', 'CD8+ Cytotoxic T'], 
        "Mono": ['CD14+ Monocyte', "CD34+", "Dendritic"]
    }, 
    union_rest="Others"
)
ad.split(pbmc, "highlevel")
```

We'd get the same thing, but we'd also be able to use this grouping for plotting and DE. If we didn't want to store the new series, it'd be:

```python
ad.split(
    pbmc,
    union_cats(
        pbmc.obs["bulk_labels"],
        {
            "T": ['CD4+/CD25 T Reg', 'CD4+/CD45RA+/CD25- Naive T', 'CD8+ Cytotoxic T'], 
            "Mono": ['CD14+ Monocyte', "CD34+", "Dendritic"]
        }, 
        union_rest="Others",
        name="highlevel",
    )
)
```

It's very similar code, but if we split the functionality of these arguments out I think we gain functionality elsewhere.

------------------------

The other functionality of `groups` is that it lets you filter which groups get used, but I'm not sure this is adding much, since subsetting should be lightweight and you can just iterate over the keys you want to use.

PR #633, opened 2021-10-29
Suggest a name convention for zarr directories

falexwolf:


falexwolf:
Close this until it is clear whether there will be a frequent usage patterns for sharing zipped zarr directories.

PR #653, opened 2021-11-30
maintain sparsity in to_df

joshua-gould:


ivirshup:
Thanks for the PR

Are these useful now? Last I tried using pandas SparseArrays, they were painfully slow.

Some use cases with benchmarks would be great. Also, would this make sense as a flag?

joshua-gould:
Here's an example. Performance seems to be comparable to dense data frame, except it can handle much bigger matrices do to maintaining sparsity.

```
import pandas as pd
import scipy.sparse

d = scipy.sparse.random(100000, 100, format="csr")
df = pd.DataFrame.sparse.from_spmatrix(d)
result = df @ df.T
df = pd.DataFrame(d.toarray())
result = df @ df.T
```

ivirshup:
> Here's an example.

Oh, I had meant to ask for an example of `to_df` returning a sparse dataframe. I'm wondering what kind of operations you want to do with it afterwards.

-------------------

Just playing around with it, I'm finding some performance problems that are a bit worrying. For example, this densifies the whole dataframe:

```python
(s_df * np.ones(s_df.shape[0]))
```

<details>
<summary> Indexing by row is also much slower. </summary>


```python
s_array = sparse.random(10_000, 10_000, format="csr", dtype=np.float32, density=0.05)
s_df = pd.DataFrame.sparse.from_spmatrix(s_array)

def array_idx(max_idx, n):
    return np.random.choice(max_idx, n, replace=False)

idx1 = array_idx(10_000, 1000)

%timeit s_array[idx1]
```

```
841 µs ± 39.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

```python
%timeit s_df.iloc[idx1]
```

```
1.88 s ± 68.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

</details>

I'm also not sure how to reason about performance, since pandas is quite opaque about it's backend (which might be in the middle of a replacement?)

joshua-gould:
Performance for multiplying a sparse data frame by a dense data frame is comparable to dense times dense. Other operations are much slower for sparse, but are still be useful to avoid requiring lots of memory.

```
m = scipy.sparse.random(100000, 100, format="csr")
df_s = pd.DataFrame.sparse.from_spmatrix(m)
df = pd.DataFrame(m.toarray())
d = np.random.random((100, 100))
%timeit result = df @ d
60.2 ms ± 3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit result = df_s @ d
81.8 ms ± 1.44 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

ivirshup:
What I'm mostly wondering about here is what kinds of operations you are doing with the dataframe you get from `to_df`. What does it allow you to do that you couldn't accomplish with an `AnnData` containing a sparse `X` or with a dense dataframe? And is it the best way to accomplish those tasks?

PR #664, opened 2021-12-14
New schema figure

ivirshup:
Updating schema figure to be one from the paper. Currently having trouble building anndata docs locally due to something with the pytorch stuff.


docs/conf.py
```diff
@@ -30,6 +30,7 @@
 
 # default settings
 templates_path = ["_templates"]
+html_static_path = ["_static"]
```
ivirshup, line 33: I'm not sure if this actually does anything

ivirshup:
I would like the schema to sit on the right side of the page, an allow text to be on the same level.

README.rst
```diff
@@ -17,7 +17,7 @@
 .. |Coverage| image:: https://api.codacy.com/project/badge/Coverage/b92ae35b691141ceb5f2ee74beaf39d3
    :target: https://www.codacy.com/manual/theislab/anndata
 
-anndata - Annotated Data
+anndata - Annotated data
```
falexwolf, line 20: See: https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings

docs/index.rst
```diff
@@ -3,7 +3,7 @@
 
 .. image:: _static/img/anndata_schema.svg
    :align: right
-   :width: 350px
+   :width: 260px
```
falexwolf, line 6: This should not have broken the docs.

anndata/_core/anndata.py
```diff
@@ -138,18 +138,15 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     variables :attr:`var` (:attr:`varm`, :attr:`varp`),
     and unstructured annotations :attr:`uns`.
 
-    .. figure:: https://falexwolf.de/img/scanpy/anndata.svg
-       :width: 350px
+    .. figure:: _static/img/anndata_schema.svg
```
falexwolf: This might have broken the docs.

docs/conf.py
flying-sheep: I think it needs to be set. We didn’t have files, so didn’t need it

anndata/_core/anndata.py
```diff
@@ -138,18 +138,15 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     variables :attr:`var` (:attr:`varm`, :attr:`varp`),
     and unstructured annotations :attr:`uns`.
 
-    .. figure:: https://falexwolf.de/img/scanpy/anndata.svg
-       :width: 350px
+    .. figure:: ../_static/img/anndata_schema.svg
+       :width: 260px
```
flying-sheep, line 142: the font looks a bit wonky for me:

grafik

ivirshup: The PDF looks fine though, could probably go from pdf -> svg

ivirshup:
I think 40% looks consistently good, especially since the central container has it's max width capped

falexwolf:
Sounds good!

falexwolf (approved):

ivirshup:
Oh, I still need to update the svg with one where the font isn't weird

ivirshup:
It turns out the issue was too few decimal points for the coordinates in the SVG – so the new version will be 1000x more precise

Not so scalable when you encode numbers as text I guess

flying-sheep:
> Not so scalable when you encode numbers as text I guess

hilariously, some JSON files are actually smaller than some binary variations of the same data, since each ASCII character is only 8 bit, so `3.2` is 24 bit and `1` is 8 bit, but both are 32 bit in the smallest common fixed-sized integer representation.

PR #676, opened 2022-01-11
Attempt to handle windows numpy dtypes

### LINKED ISSUE #672: np.float128 and np.complex256 are not supported on windows
https://github.com/theislab/anndata/blob/664e32b0aa6625fe593370d37174384c05abfd4e/anndata/_io/specs/methods.py#L666

Now the master branch fails on import on Windows because at least the standard numpy library versions on conda (tried 1.19 and 1.20) do not support `np.float128` and `np.complex256`. So trying to access them throws an error
`module 'numpy' has no attribute 'float128'`
`module 'numpy' has no attribute 'complex256'`

ivirshup:
@Koncopd, could you check this locally?

Should be the thing that fixes #672

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -709,8 +709,8 @@ for numeric_scalar_type in [
     bool, np.bool_,
     np.uint8, np.uint16, np.uint32, np.uint64,
     int, np.int8, np.int16, np.int32, np.int64,
-    float, np.float16, np.float32, np.float64, np.float128,
-    np.complex64, np.complex128, np.complex256,
+    float, *np.floating.__subclasses__(),
+    *np.complexfloating.__subclasses__(),
 ]:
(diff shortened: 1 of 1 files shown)


Koncopd:
Hi, i will check. But one immediate question is how necessary these subclasses are? On 64 bit machines these 128 bit datatypes are useless and their usage would make adata files with these datatypes unreadable on Windows, wouldn't it?

ivirshup:
These would have worked in the past, so it would be removing features. That said, I don't think anyone is using these types. It does allow round-tripping with other languages which do support those types on windows.

That said, this does seem more complicated than I hoped (`float128` [may not actually be 128 bit](https://numpy.org/doc/stable/user/basics.types.html#extended-precision) on systems where it's available, `complex256` is [more complicated](https://github.com/h5py/h5py/pull/749)).

Maybe we could remove support, but consider adding it back as a bug fix if anyone complains?

Koncopd:
@ivirshup this works fine on Windows.

Issue #455, opened 2020-11-16
Writing to H5AD fails if AnnData contains datetime64 objects


lazappi:
Writing an `AnnData` object to disk fails if it contains **numpy** `datetime64` objects. Example:

```python
import anndata
import numpy as np

ad = anndata.AnnData()

ad.uns['now'] = np.datetime64('now')

ad.write("test_h5py.h5ad")
```

<details>
<summary>ERROR MESSAGE</summary>
Traceback (most recent call last):
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_io/utils.py", line 188, in func_wrapper
    return func(elem, key, val, *args, **kwargs)
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_io/h5ad.py", line 144, in write_not_implemented
    raise NotImplementedError(
NotImplementedError: Failed to write value for uns/now, since a writer for type <class 'numpy.datetime64'> has not been implemented yet.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/luke.zappia/Desktop/test_h5py.py", line 8, in <module>
    ad.write("test_h5py.h5ad")
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_core/anndata.py", line 1846, in write_h5ad
    _write_h5ad(
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_io/h5ad.py", line 112, in write_h5ad
    write_attribute(f, "uns", adata.uns, dataset_kwargs=dataset_kwargs)
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/functools.py", line 877, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_io/h5ad.py", line 126, in write_attribute_h5ad
    _write_method(type(value))(f, key, value, *args, **kwargs)
  File "/Users/luke.zappia/miniconda3/envs/test-h5py/lib/python3.9/site-packages/anndata/_io/h5ad.py", line 286, in write_mapping
    write_attribute(f, f"{key}/{sub_key}", sub_value, dataset_kwargs=dataset_kwargs)
[...]

PR #684, opened 2022-01-14
Datetimes

ivirshup:
Basic datetime IO support.

Currently this converts everything to numpy datetime arrays at write time. I'm not preserving pandas array types since there are multiple seemingly overlapping ways to deal with datetimes in pandas. This implementation also does not support time zones but that would be easy to add.

It would be good to get someone working with time series data to try this out and see if it meets their needs.

(I thought this would solve #455, but now see that was for datetime scalars which this does not currently support)

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -681,6 +681,59 @@ def read_nullable_boolean(elem):
         return pd.array(read_elem(elem["values"]))
 
 
+#############
+# datetimes #
+#############
+
+
+@_REGISTRY.register_write(
+    ZarrGroup, (views.ArrayView, "M"), IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(
+    H5Group, (views.ArrayView, "M"), IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(
+    ZarrGroup, (np.ndarray, "M"), IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(H5Group, (np.ndarray, "M"), IOSpec("datetime-array", "0.1.0"))
+def write_datetime_array(f, k, elem, dataset_kwargs=MappingProxyType({})):
+    unit, step = np.datetime_data(elem.dtype)
+
+    if step != 1:
+        raise ValueError(
+            f"Datetime had non-unit step '{step}', which is currently unsupported."
+        )
+
+    dset = f.create_dataset(k, data=elem.view("uint64"), **dataset_kwargs)
+    dset.attrs["unit"] = unit
+
+
+@_REGISTRY.register_write(
+    ZarrGroup, pd.arrays.DatetimeArray, IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(
+    H5Group, pd.arrays.DatetimeArray, IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(
+    ZarrGroup, pd.DatetimeIndex, IOSpec("datetime-array", "0.1.0")
+)
+@_REGISTRY.register_write(H5Group, pd.DatetimeIndex, IOSpec("datetime-array", "0.1.0"))
+def write_datetime_array_from_pandas(f, k, elem, dataset_kwargs=MappingProxyType({})):
+    if elem.tz is not None:
+        raise NotImplementedError(
+            "Datetime had timezone specified, which is currently not supported."
+        )
+
+    write_elem(f, k, elem.to_numpy(), dataset_kwargs=dataset_kwargs)
+
+
+@_REGISTRY.register_read(ZarrArray, IOSpec("datetime-array", "0.1.0"))
+@_REGISTRY.register_read(H5Array, IOSpec("datetime-array", "0.1.0"))
+def read_np_datetime_array(elem):
+    unit = _read_attr(elem.attrs, "unit")
+    return elem[()].view(f"datetime64[{unit}]")
+
+
 ###########
 # Scalars #
 ###########
--- a/anndata/_io/specs/registry.py
+++ b/anndata/_io/specs/registry.py
@@ -53,19 +53,27 @@ class IORegistry(object):
 
         return _register
 
-    def get_writer(self, dest_type, typ, modifiers=frozenset()):
+    def get_writer(self, dest, elem, modifiers=frozenset()):
         import h5py
+
+        typ = type(elem)
+        dest_type = type(dest)
 
         if dest_type is h5py.File:
             dest_type = h5py.Group
         modifiers = frozenset(modifiers)
 
-        if (dest_type, typ, modifiers) not in self.write:
+        if (
+            hasattr(elem, "dtype")
+            and (dest_type, (typ, elem.dtype.kind), modifiers) in _REGISTRY.write
+        ):
+            return self.write[(dest_type, (typ, elem.dtype.kind), modifiers)]
+        elif (dest_type, typ, modifiers) in _REGISTRY.write:
+            return self.write[(dest_type, typ, modifiers)]
+        else:
             raise TypeError(
                 f"No method has been defined for writing {typ} elements to {dest_type}"
             )
-
-        return self.write[(dest_type, typ, modifiers)]
 
     def has_writer(self, dest_type, typ, modifiers):
         modifiers = frozenset(modifiers)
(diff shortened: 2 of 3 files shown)


Zethson:
> It would be good to get someone working with time series data to try this out and see if it meets their needs.

we do technically. If we detect datetime we always just copy it to obs directly.

ivirshup:
An example (or you giving this branch a shot) would be great.

Do you have a way of saving these `AnnData`'s at the moment?

Zethson:
@imipenem can you help here?

Imipenem:
> Do you have a way of saving these AnnData's at the moment?

At ehrapy, it just worked out of the box when writing these `AnnData`s to `.h5ad` files. But this might be due to the fact, that we save columns with `datetime` values in `obs` only (and `pandas` treats these `datetimes` kind of different in comparison to `numpy` from what I've read), neither in `uns` or `X`. So we do not have any `np.datetime` values stored in the `AnnData` object at any time, which (IMO) fits our needs here (for now). So this would not affect us currently or do I miss something @Zethson?

Zethson:
Thought so as well. We didn't run into any issues.

ivirshup:
I'm a little confused here. If I put any sorts of dates into `obs`, that anndata will fail to write in `0.7.8`.

Can you make an example of this? For me:

<details>
<summary> Failing example </summary>

```python
import anndata as ad, pandas as pd, numpy as np
from vega_datasets import data
print(ad.__version__)
```

```
0.7.8
```

```python
cars = data.cars()

dt_array = cars["Year"]
np_dt_array = dt_array.to_numpy()

N = np_dt_array.shape[0]
adata = ad.AnnData(X=np.ones((N, N)), obs=pd.DataFrame({"dt": dt_array}))

adata.write_h5ad("test_dt.h5ad")
```

```pytb
~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    208         try:
--> 209             return func(elem, key, val, *args, **kwargs)
    210         except Exception as e:

~/github/anndata/anndata/_io/h5ad.py in write_array(f, key, value, dataset_kwargs)
    184         value = _to_hdf5_vlen_strings(value)
--> 185     f.create_dataset(key, data=value, **dataset_kwargs)
    186 

/usr/local/lib/python3.9/site-packages/h5py/_hl/group.py in create_dataset(self, name, shape, dtype, data, **kwds)
    148 
--> 149             dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
    150             dset = dataset.Dataset(dsid)

/usr/local/lib/python3.9/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, name, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times, external, track_order, dcpl, allow_unknown_filter)
     90             dtype = numpy.dtype(dtype)
---> 91         tid = h5t.py_create(dtype, logical=1)
     92 

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

h5py/h5t.pyx in h5py.h5t.py_create()

TypeError: No conversion path for dtype: dtype('<M8[ns]')

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    208         try:
--> 209             return func(elem, key, val, *args, **kwargs)
    210         except Exception as e:

~/github/anndata/anndata/_io/h5ad.py in write_series(group, key, series, dataset_kwargs)
    288     else:
--> 289         write_array(group, key, series.values, dataset_kwargs=dataset_kwargs)
    290 

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    211             parent = _get_parent(elem)
--> 212             raise type(e)(
    213                 f"{e}\n\n"

TypeError: No conversion path for dtype: dtype('<M8[ns]')

Above error raised while writing key 'dt' of <class 'h5py._hl.group.Group'> from /.

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    208         try:
--> 209             return func(elem, key, val, *args, **kwargs)
    210         except Exception as e:

~/github/anndata/anndata/_io/h5ad.py in write_dataframe(f, key, df, dataset_kwargs)
    262     for col_name, (_, series) in zip(col_names, df.items()):
--> 263         write_series(group, col_name, series, dataset_kwargs=dataset_kwargs)
    264 

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    211             parent = _get_parent(elem)
--> 212             raise type(e)(
    213                 f"{e}\n\n"

TypeError: No conversion path for dtype: dtype('<M8[ns]')

Above error raised while writing key 'dt' of <class 'h5py._hl.group.Group'> from /.

Above error raised while writing key 'dt' of <class 'h5py._hl.group.Group'> from /.

/var/folders/bd/43q20k0n6z15tdfzxvd22r7c0000gn/T/ipykernel_4792/2332825967.py in <module>
----> 1 adata.write_h5ad("test_dt.h5ad")

~/github/anndata/anndata/_core/anndata.py in write_h5ad(self, filename, compression, compression_opts, force_dense, as_dense)
   1910             filename = self.filename
   1911 
-> 1912         _write_h5ad(
   1913             Path(filename),
   1914             self,

~/github/anndata/anndata/_io/h5ad.py in write_h5ad(filepath, adata, force_dense, as_dense, dataset_kwargs, **kwargs)
    109         else:
    110             write_attribute(f, "raw", adata.raw, dataset_kwargs=dataset_kwargs)
--> 111         write_attribute(f, "obs", adata.obs, dataset_kwargs=dataset_kwargs)
    112         write_attribute(f, "var", adata.var, dataset_kwargs=dataset_kwargs)
    113         write_attribute(f, "obsm", adata.obsm, dataset_kwargs=dataset_kwargs)

/usr/local/Cellar/python@3.9/3.9.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/functools.py in wrapper(*args, **kw)
    875                             '1 positional argument')
    876 
--> 877         return dispatch(args[0].__class__)(*args, **kw)
    878 
    879     funcname = getattr(func, '__name__', 'singledispatch function')

~/github/anndata/anndata/_io/h5ad.py in write_attribute_h5ad(f, key, value, *args, **kwargs)
    128     if key in f:
    129         del f[key]
--> 130     _write_method(type(value))(f, key, value, *args, **kwargs)
    131 
    132 

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    210         except Exception as e:
    211             parent = _get_parent(elem)
--> 212             raise type(e)(
    213                 f"{e}\n\n"
    214                 f"Above error raised while writing key {key!r} of {type(elem)}"

TypeError: No conversion path for dtype: dtype('<M8[ns]')

Above error raised while writing key 'dt' of <class 'h5py._hl.group.Group'> from /.

Above error raised while writing key 'dt' of <class 'h5py._hl.group.Group'> from /.

Above error raised while writing key 'obs' of <class 'h5py._hl.files.File'> from /.
```

</details>

Zethson:
Sure:

import ehrapy.api as ep

```
adatas = ep.dt.mimic_3_demo(encoded=False, mudata=False)
print(adatas["INPUTEVENTS_CV"].obs)
adata = adatas["INPUTEVENTS_CV"]
# This may take 5-20 minutes
ep.pp.knn_impute(adata)
adata_encoded = ep.pp.encode(adata, autodetect=True)
ep.io.write("test.h5ad", adata_encoded)
```

I would not be surprised if we store things differently than you somewhere, but feel free to play around with it. Feedback is always appreciated!

ivirshup:
> I have the suspicion that the datetimes are somewhere just read as strings and then mapped to categoricals.

That seems to be the case. 

```python
adata_encoded.obs["charttime"].cat.categories.dtype
```

```
dtype('O')
```

Would it be useful if these were actual datetimes? The you could do things like ask how far apart the times were.

Zethson:
> > I have the suspicion that the datetimes are somewhere just read as strings and then mapped to categoricals.
> 
> That seems to be the case.
> 
> ```python
> adata_encoded.obs["charttime"].cat.categories.dtype
> ```
> 
> ```
> dtype('O')
> ```
> 
> Would it be useful if these were actual datetimes? The you could do things like ask how far apart the times were.

Not surprised. Our primary motivation was the coloring of plots and things like that.

Yeah, your suggested use-case is a good one. Although, in general I am trying to reduce the dependency on real time as much as possible with ehrapy and to work more with pseudotime :)

Zethson:
@ivirshup is this PR still one approach that you'd follow or did it change since Pandas 2.0 got released? Datetime support would still be great for ehrapy - especially for stuff like comparing them and more

Issue #675, opened 2022-01-11
HTML Repr


ivirshup:
We should have a nicer repr for the anndata object. Ideally a rich one with more info and collapsibility. [`mudata` already has this implemented](https://muon.readthedocs.io/en/latest/notebooks/quickstart_mudata.html#Rich-representation), and we've been given permission to steal it here.

Previous discussion #346, #521, #671

Zethson:
@ivirshup scvi tools also had an option. See: https://github.com/theislab/ehrapy/issues/233

ivirshup:
Could you point me at the deprecated thing in scvi-tools?

To me, collapsibility is a must, since the "tons of columns" makes it long quick!

katosh:
I really like this suggestion! Especially, for large objects it often becomes hard to keep an overview.

I made an implementation for this and will submit my PR tomorrow.

PR #694, opened 2022-01-31
Initial draft of AnnData HTML repr

gtca:
This is a draft to address #675. As with [mudata](https://github.com/scverse/mudata), it is collapsable — [check out e.g. this notebook](https://mudata.readthedocs.io/en/latest/notebooks/quickstart_mudata.html) with `mudata`.

Maybe we can iterate on an implementation of it, and then we'll refactor the code in `mudata` to re-use AnnData representation inside the MuData representation.

Currently both [mudata](https://github.com/scverse/mudata/blob/master/mudata/_core/config.py) and [muon](https://github.com/scverse/muon/blob/master/muon/_core/config.py) have a config manager, which is used in `mudata` for collapse/expand settings. Not sure if we want to introduce it in `anndata` as the setting is still somewhat reachable:

```py
from anndata._core.repr import repr_html
from IPython.core.display import HTML
display_html = lambda adata, expand: HTML(repr_html(adata, expand=0b11))

HTML(repr_html(adata, 0b11))
```

---

I have also just noticed there was a [`html-repr` branch](https://github.com/theislab/anndata/commits/html-repr) a while ago, maybe we can drag over some ideas with aligned mappings from there, — or vice versa.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -57,6 +57,7 @@ from ..compat import (
     _overloaded_uns,
     OverloadedDict,
 )
+from .repr import repr_html
 
 
 class StorageType(Enum):
@@ -589,6 +590,13 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             return "View of " + self._gen_repr(self.n_obs, self.n_vars)
         else:
             return self._gen_repr(self.n_obs, self.n_vars)
+
+    def _repr_html_(self) -> str:
+        """
+        HTML formatter for the AnnData object
+        for rich display in notebooks.
+        """
+        return repr_html(self, expand=0b10)
 
     def __eq__(self, other):
         """Equality testing"""
--- a/anndata/_core/repr.py
+++ b/anndata/_core/repr.py
@@ -1,0 +1,311 @@
+#
+# Utility functions for AnnData._repr_html_()
+#
+
+from typing import Tuple, Iterable
+from numbers import Integral, Real, Complex
+from warnings import warn
+import numpy as np
+import pandas as pd
+
+def repr_html(adata: "AnnData", expand=None) -> str:
+    """
+    HTML formatter for the AnnData object.
+
+    This formatter has an optional argument `expand`
+    which is a 2-bit flag:
+    10 - expand AnnData slots
+    01 - expand individul slots
+    """
+
+    if expand is None:
+        expand = 0b10
+
+    data = ""
+
+    data += "<div class='block-anndata'><div>"
+    data += "<details{}>".format(" open" if (expand & 0b10) >> 1 else "")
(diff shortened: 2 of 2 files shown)


ivirshup (commented):
Thanks for opening this! Really excited to get this added.

# Examples

This is more a point for tracking of this PR. Could you share examples of what this looks like embedded in the issue? I'd like to be able to easily track changes to what this looks like, and maybe we could just agree now including an image or stable jsfiddle or something?

Also, do you have an example for how dataframes are displayed? I don't think there was one in the linked docs.

# Dispatch

Could we use `functools.singledispatch` for generating the div for each element? That could make registering new types possible. E.g.

```python
@ad.repr_html(MuData)
def repr_html_mudata(mudata, name: Optional[str] = None, expand=None) -> str:
    data = ""
    data += ... # MuData header
    data += ad.repr_html(mudata.obs, "obs")
    data += ad.repr_html(mudata.var, "var")
    ...

    data += ad.repr_html(adata.mod, "modalities") # which calls repr_html on each AnnData object, but indented.
    ...
```

# "nice" names

My preference would be to remove the more "semantic" names, like "Distances" or "Embeddings". Reasons being:

* These fields can contain other things. You could put QC metrics in obsm, or alternative annotations. `connectivities` or `spatial_connectivities` aren't distances.
* Brevity. In a diagram I might add more saying what these fields are used for, but I think identification is enough for the representation that's shown every time.

# Config

On the config, I was wondering if we could maybe normalize on a single config for `scverse` tools? Like a `scverse.toml`?

This would let us have shared locations for downloading data and annotations. Subtables within the table could be package specific.

Added this to the meeting agenda to gauge thoughts.

gtca:
> Thanks for opening this! Really excited to get this added.
> 
> # Examples
> This is more a point for tracking of this PR. Could you share examples of what this looks like embedded in the issue? I'd like to be able to easily track changes to what this looks like, and maybe we could just agree now including an image or stable jsfiddle or something?
> 
> Also, do you have an example for how dataframes are displayed? I don't think there was one in the linked docs.

Let's start one here: https://jsfiddle.net/325vxk6u/.

The rest makes sense as well, will get back to you with the code refactored.

---

> On the config, I was wondering if we could maybe normalize on a single config for scverse tools? Like a scverse.toml?

As it's a setting adjustable by the user, we would still need to add [a config manager](https://github.com/theislab/scanpy/blob/master/scanpy/_settings.py) to `anndata` – unless I'm missing the point.

ivirshup:
> As it's a setting adjustable by the user, we would still need to add a config manager to anndata – unless I'm missing the point.

It was a bit of a tangent, but I was thinking we could have either a common config manager or at least a base class. Would be a bit like how bioconductor does it

Plus should be easy enough with [`pydantic`](https://pydantic-docs.helpmanual.io/usage/settings/#environment-variable-names)

Issue #693, opened 2022-01-26
Dask support


ivirshup:
It would be nice if we could seamlessly hold `dask` arrays and dataframes inside of an AnnData object.

I would consider this issue closed once most operations on an AnnData would work when it contained `dask` objects. Reading in an `AnnData` with all elements backed by `dask` would be separate and downstream task.

See also:

* Possible API for reading an AnnData object with some dask backed arrays #659
* Out of core support #690
* Performance advantages of dask dataframes may be strongly limited here until https://github.com/dask/dask/issues/7933 is resolved (currently cannot infer only one column should be read from disk from `ddf.read_*(...)["col"]`)

ivirshup:
I think this issue will have may subparts, so I think individual issues for that level of specificity I think

Issue #659, opened 2021-12-09
AnnData IO modifiers


ivirshup:
# AnnData selectors

*Note: I will be updating this to be easier to read, just wanted it up to share*

## Use case: Modified IO

I would like to able to specify read time transforms on specific elements of an AnnData object. These transforms would include:

* Read in as an alternate class, examples:
    * dataframe: `polars.DataFrame`, `arrow.Table`, `dask.dataframe.DataFrame`
    * sparse array: pydata `sparse`, `jax.sparse`
    * array: `dask`
* Include/ exclude
* Sparsify
* Densify
* Lazily read -> `delayed` elements
* Partial read
    * E.g. read only the first hundred obs

How do we specify these transforms?

### Specification

I think we would want to select elements by their type and location. These would be two separate fields. 

* *location* can be a one or multiples paths to locations in anndata
* *type* will refer to the type of the object. This could be either with our `IOSpec` mapping of types.

### Query by location

```python 
{
    "select": {"location": "obsm/*"},
    "modifier": Exclude()
}
```

### Query by type

```python
{
    "select": {"type": pd.DataFrame},
    "modifier": AsPolars(),
}
```

### Query by both

```python
{
    "select": {"type": "array", "location": ["layers", "obsm"]}
    "modifier": Exclude()
}
```

### Read objects in lazily to dask arrays

```python
{
    "select": {"location": "*"},
    "modifer": AsDask()
}
```

### Usage:

```python
adata = ad.read_h5ad(
    "path/to/file.h5ad",
    modifiers = [
        {
            "select": {"loc": ["X", "obs", "obsm/X_umap"]},
            "modifier": Include(),
        },
        {"select": {"loc": "X"}, "modifier": AsCSC()},
    ]
)
```

### Only delayed

Only delayed ops, but this would have to include a way to access data in it's backed format (which does not exist for dataframes, may not exist for future types).

#### Exclude

```python
adata = ad.read_as_dask("path/to/file.h5ad")
adata = adata.select_elements(["X", "obs", "obsm/X_umap"])
adata.to_memory()
```

#### Read time modifiers

```python
adata = ad.read_backed("path/to/file.h5ad")
adata.X = ad.io.utils.read_dense_as_csr(adata.X)
adata.to_memory()
```

(will be adding to this)

ivirshup:
Related: zarr PR for reading in data to a specific array type. The particular use case being GPU arrays https://github.com/zarr-developers/zarr-python/pull/934.

[...]

PR #695, opened 2022-01-31
Add support for DeviceArray from JAX

gtca:
Allow AnnData objects have `.X` as a `DeviceArray`. Layer support should also be there but seems to currently fail just as it does e.g. for `DaskArray`, as mentioned in #693.

```py
from jax import random
from anndata import AnnData

x = random.normal(random.PRNGKey(1), (100,10))
adata = AnnData(x)

adata.write("demojax.h5ad")
```

As mentioned in #659, it would be nice to also get `DeviceArray` on read in future if possible, currently it is a NumPy Array:

```py
adata = anndata.read("demojax.h5ad")
type(adata.X)
# => numpy.ndarray
```

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -51,6 +51,7 @@ from ..compat import (
     ZarrArray,
     ZappyArray,
     DaskArray,
+    DeviceArray,
     Literal,
     _slice_uns_sparse_matrices,
     _move_adj_mtx,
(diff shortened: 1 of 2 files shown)


ivirshup:
I would support JAX as another backend.

Would probably be worth opening another issue for tracking JAX as a backend. For IO, I was thinking that could be done with #659

PR #701, opened 2022-02-03
Fix X dtype when reading in from zarr

### LINKED ISSUE #700: read_zarr() ignores array data type
Hello! Thanks for the great library!

I was playing around with reading/writing AnnData from/to zarr and I noticed that when reading from zarr, the dtype of the arrays are not taken into account. For example, if I create an AnnData object with dtype float64, the resulting zarr file has arrays with dtype float64, but when I read it back in, it now has dtype float 32. I have pasted the code to reproduce below.

I'm not 100% sure, I think it might be because dtype isn't added to the AnnData kwargs [here](https://github.com/theislab/anndata/blob/555f5469382168e9d8fc6d845511038fd5ff7cf4/anndata/_io/zarr.py#L77-L90).

<details>
  <summary>example code</summary>

```python
from anndata import AnnData, read_zarr
import numpy as np
import pandas as pd

n_rows = 3
n_cols = 3
rng = np.random.default_rng(42)
X = rng.random((n_rows, n_cols), dtype=float)

print(f'X dtype: {X.dtype}')

# create points data
n_points = 10
n_dim = 2
obs = pd.DataFrame(
    {
        'x': rng.random((n_cols,), dtype=float),
        'y': rng.random((n_cols,), dtype=float),
        'z': rng.random((n_cols,), dtype=float)
    }
)

original_anndata = AnnData(X=X, obs=obs, dtype=float)
print(f'anndata X dtype: {original_anndata.X.dtype}')

# write to zarr
fname = 'test_anndata.zarr'
original_anndata.write_zarr(fname)

# load the anndata
loaded_anndata = read_zarr(fname)
print(f'loaded X dtype: {loaded_anndata.X.dtype}')
```

</details>

ivirshup:
Fixes #700

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -246,6 +246,8 @@ def read_anndata(elem):
     ]:
         if k in elem:
             d[k] = read_elem(elem[k])
+        if "X" in d:
+            d["dtype"] = d["X"].dtype
     return AnnData(**d)
 
 
--- a/anndata/_io/zarr.py
+++ b/anndata/_io/zarr.py
@@ -85,6 +85,9 @@ def read_zarr(store: Union[str, Path, MutableMapping, zarr.Group]) -> AnnData:
 
     d["raw"] = _read_legacy_raw(f, d.get("raw"), read_dataframe, read_elem)
 
+    if "X" in d:
+        d["dtype"] = d["X"].dtype
+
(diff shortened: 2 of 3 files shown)


kevinyamauchi:
I tested this against the my use case I reported in #700 and it fixes my issue. Thanks!

PR #734, opened 2022-03-16
Improve interpretability of IOSpec errors

flying-sheep:
This adds error messages based on what wasn’t found. There’s 3-4 levels of detail (3 for writing, 4 for reading)

1. Nothing registered for source type (write) / destination type (read)

   > No such write function registered: Unknown destination type SomeType.

   > No such read function registered: Unknown source type SomeType.

2. Destination type (write) / IOSpec (read) not found

    > No such write function registered: Destination type Group not found for source type \<class SomeType>.

    1. > No such read function registered: Unknown encoding type “floob” for source type Group.
    2. > No such read function registered: Unknown encoding version “10000.0” for source type Group’s encoding “rec-array”.

3. Modifier set not found

    > No such write function registered: Unknown modifier set {...} for source type SomeType and destination type Group

    > No such read function registered: Unknown modifier set {...} for source type Group’s encoding IOSpec(...)


anndata/_io/specs/registry.py
```diff
@@ -33,16 +87,18 @@ def wrapper(g, k, *args, **kwargs):
...
     def register_write(
         self,
-        dest_type,
-        typ: Union[type, tuple[type, str]],
-        spec,
-        modifiers: frozenset(str) = frozenset(),
+        dest_type: type,
+        typ: type | tuple[type, str],
+        spec: IOSpec | Mapping[str, str],
+        modifiers: Iterable[str] = frozenset(),
```
ivirshup, line 101: frozenset is actually critical here because it's an unordered hashable collection.

flying-sheep: https://github.com/theislab/anndata/blob/iospec-errors/anndata/_io/specs/registry.py#L104

anndata/_io/specs/registry.py
```diff
@@ -17,6 +16,61 @@ class IOSpec(NamedTuple):
     encoding_version: str
 
 
+class NoSuchIO(KeyError):
```
ivirshup, line 19: Could this live with the other errors?

ivirshup (changes requested):
Thanks for the PR! Looks broadly good.

Things I'd like to see:

* More tests (I don't think the write error actually gets thrown)
* Working docs
* Cleaned up error formatting. I'm struggling to follow the logic in there. I think splitting out the read and write cases would simplify this a lot.
* The error type should be exported so users can catch it.

anndata/_io/specs/registry.py
```diff
@@ -65,13 +126,23 @@ def get_writer(self, dest_type, typ, modifiers=frozenset()):
                 f"No method has been defined for writing {typ} elements to {dest_type}"
             )
 
-        return self.write[(dest_type, typ, modifiers)]
+        try:
+            return self.write[(dest_type, typ, modifiers)]
+        except KeyError:
```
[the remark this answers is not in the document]
flying-sheep: IDK if `__contains__` is implemented that way, but since `NoSuchIO` is a subclass of `KeyError`, that’s irrelevant.

anndata/_io/specs/registry.py
flying-sheep: That’s not a conflict, that’s just two separate types.

anndata/_io/specs/registry.py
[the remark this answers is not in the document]
ivirshup: Oh, I meant the block of code immediately above this, with:

```python

        if (dest_type, typ, modifiers) not in self.write:
            raise TypeError(
                 f"No method has been defined for writing {typ} elements to {dest_type}"
             )
```

anndata/tests/test_io_elementwise.py
```diff
@@ -110,3 +107,20 @@ def test_write_to_root(store):
...
+    ],
+)
+def test_read_version_error(store, attr, val, pattern):
+    adata = gen_adata((3, 2))
+
+    write_elem(store, "/", adata)
+    store["obs"].attrs[f"encoding-{attr}"] = val
+    full_pattern = rf"No such read function registered: {pattern}"
+    with pytest.raises(NoSuchIO, match=full_pattern):
+        read_elem(store)
```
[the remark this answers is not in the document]
flying-sheep: Sure!

anndata/_io/specs/registry.py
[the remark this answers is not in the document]
flying-sheep: oh, I missed that! yeah, I’ll remove that part.

anndata/_io/specs/registry.py
```diff
@@ -17,6 +16,61 @@ class IOSpec(NamedTuple):
...
+        self.key = key
+        super().__init__(str(self))
+
+    def __str__(self) -> str:
+        return f"No such {self.registry} function registered: {self._get_msg()}."
+
+    def _get_msg(self) -> str:
+        registry: Mapping[
+            tuple[type, IOSpec | type | tuple[type, str], frozenset[str]], Callable
+        ] = getattr(_REGISTRY, self.registry)
```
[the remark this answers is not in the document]
flying-sheep: maybe, I’ll think about it

anndata/_io/specs/registry.py
flying-sheep: For sure!

anndata/_io/specs/registry.py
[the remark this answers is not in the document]
ivirshup: Most errors can be thrown like `raise FooError(msg)`, and I would like to keep that API here. I would suggest that the method to format the message is split out into functions. E.g.: `raise NoSuchIO(_read_error_msg(...))`.

This would also make the write case super easy, since you don't even need a function at the moment. You could probably just replace `TypeError` with `NoSuchIO`.

anndata/_io/specs/registry.py
```diff
@@ -17,6 +16,61 @@ class IOSpec(NamedTuple):
...
+        is_read = self.registry != "write"
+        assert isinstance(spec_or_src_type, IOSpec) == is_read
+
+        dir_ = "destination" if is_read else "source"
+        desc_type = f"{dir_} type {typ.__name__}"
+        types = {typ for typ, _, _ in registry}
+        if typ not in types:
+            return f"Unknown {desc_type}"
+
+        sosts = {sost for typ, sost, _ in registry if typ in types}
```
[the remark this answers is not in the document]
flying-sheep: see below, `spec_or_src_type`, but point taken regarding readability.

anndata/_io/specs/registry.py
flying-sheep: wait, we don’t have a place for other errors:

grafik

anndata/_io/specs/registry.py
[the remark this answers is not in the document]
flying-sheep: I don’t like that API at all. It makes so much more sense to have structured data in the error in most cases.

CPython has very slowly expanded that, but is still not there. An `AttributeError` should have a field that tells you the name of the Attribute that doesn’t exist!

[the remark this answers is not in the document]
flying-sheep: I split it into different code paths, but IDK if this is much clearer. certainly more verbose.

flying-sheep:
OK, this gets more involved now. I noticed that

> Most errors can be thrown like raise `FooError(msg)`, and I would like to keep that API here.

Likely refers to the current `report_write_key_on_error`. Doing `type(e)("some msg")` isn’t guaranteed to work, Errors can and do have different constructors. I think the `raise AnnData*Error(...) from e` approach used in `report_read_key_on_error` is much more elegant.

I therefore made things symmetric:  `report_read_key_on_error` raises `AnnDataReadError` and `report_write_key_on_error` raises `AnnDataWriteError`. The original error is always available as `e.__cause__`

anndata/_io/specs/registry.py
```diff
@@ -17,6 +21,108 @@ class IOSpec(NamedTuple):
...
+            )
+        return (
+            f"Unknown encoding version {self.spec.encoding_version} "
+            f"for {self.type_desc}’s encoding “{self.spec.encoding_type}”"
+        )
+
+    def _get_modifiers_desc(self) -> str:
+        return f"{self.type_desc}’s encoding {self.spec}"
+
+
```
ivirshup, line 125: I agree this is too verbose. I think you could do something more like this:

```suggestion
class NoSuchIO(KeyError):
    pass


class NoSuchWrite(NoSuchIO):

    @staticmethod
    def _from_parts(cls, dest_type, typ, modifiers) -> NoSuchWrite:
        msg = f"No method registered for writing {typ} into {dest_type}"
        if modifiers:
            msg += f" with {modifiers}"
        return cls(msg)


class NoSuchRead(NoSuchIO):

    @staticmethod
    def _from_parts(cls, method: str, registry, src_typ, spec) -> NoSuchRead:
        versions = [s.encoding_version for s in self.registry[method] if spec.encoding_type == s.encoding_type]
        msg = f"No {method} method registered for {spec} from {src_typ}"
        if versions:
            msg += f", readers only found for versions: {versions}"
        return cls(msg)
```

BTW, the way of doing look ups for reading methods needs to change (for cases where forward compatible reading is possible) so I wouldn't hook into that logic much.

anndata/_io/specs/registry.py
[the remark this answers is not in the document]
ivirshup: I would prefer the `if` check. Easier to swap out when the logic around method look up changes. Maybe even:

```python
if not self.has_writer(dest_type, typ, modifiers):
    raise NoSuchWrite._from_parts((dest_type, typ, modifiers))
```

anndata/_io/specs/registry.py
```diff
@@ -101,9 +235,17 @@ def _register(func):
...
+        src_type: type,
+        spec: IOSpec | Mapping[str, str],
+        modifiers: Iterable[str] = frozenset(),
+    ):
         modifiers = frozenset(modifiers)
-        return self.read_partial[(src_type, spec, modifiers)]
+        try:
+            return self.read_partial[(src_type, spec, modifiers)]
+        except KeyError:
+            raise NoSuchRead("read_partial", (src_type, spec, modifiers)) from None
```
ivirshup, line 248: Similar to above, would prefer `if` over `try`

anndata/_io/specs/registry.py
ivirshup: Also IDK if we need three classes here.

ivirshup (changes requested):
Mostly good!

Two main changes:

* Simplify formatting of error messages. It's overkill, and if they're ever thrown in another context we'd need to redo the classes.
* I would keep the current error structuring. I am up for a conversation on a better way to do this, but I think it's out of scope for this PR. More on this below:

> I therefore made things symmetric: report_read_key_on_error raises AnnDataReadError and report_write_key_on_error raises AnnDataWriteError. The original error is always available as e.__cause__.
> ...
> Doing type(e)("some msg") isn’t guaranteed to work, Errors can and do have different constructors. 

They can, but in practice pretty much all do. It's not ideal, but the case where they don't hasn't come up. At one point I actually checked which errors from all the libraries installed on my machine didn't work with this, and I think it was ~5.

> I think the raise AnnData*Error(...) from e approach used in report_read_key_on_error is much more elegant and will always work.

I specifically haven't done this because then `try: ...; catch NoSuchIO` doesn't work.

Maybe this is the right way to go ultimately, but changing the type of any error encountered within `read_*` is a breaking change. I would suggest this move to a different PR.

ivirshup:
One other point:

Is `NoSuchIO` a `KeyError`? Maybe it's more of a `TypeError` since it's basically dispatch?

anndata/_io/specs/registry.py
flying-sheep: that’s the pythonic way:

- https://docs.python.org/3.7/glossary.html#term-eafp
- https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

anndata/_io/specs/registry.py
flying-sheep: The tests are now complete. Feel free to shorten my code if you think it’s worth the less expressive error messages.

ivirshup:
Some changes I'm making:

* I'm going to revert the addition of AnnDataWriteError for now in favor of using [PEP-678](https://peps.python.org/pep-0678/). Details in: #868.
* I've collapsed the registry errors to one class and simplified it

PR #752, opened 2022-04-04
Remove pre-commit workflow

flying-sheep:
We already use pre-commit.ci


flying-sheep:
Since it serves no purpose and nobody replied, I’m just removing it, we can always reinstate it if we need it.

PR #765, opened 2022-04-29
(feat): Backed zarr and hdf5 sparse matrices (separate from backed mode)

ivirshup:
Initial draft of backed sparse array support for zarr.

- [x] I intend to export `sparse_dataset`, `CSRDataset`, and `CSCDataset` class from `experimental`.
- [x] Get tests passing
- [ ] Figure out if I'm going to expose any arguments through `read_zarr`


ilan-gold:
> Initial draft of backed sparse array support for zarr.
> 
>     * [ ]  I intend to export `sparse_dataset`, `CSRDataset`, and `CSCDataset` class from `experimental`.
> 
>     * [x]  Get tests passing
> 
>     * [ ]  Figure out if I'm going to expose any arguments through `read_zarr`

At the moment, the first and last are still open ended as is the question of setting the data.  Otherwise, tests pass.

ilan-gold:
Decisions:
1. don't expose arguments to `read_zarr`
2. do export the classes from `experimental`.  
3. `DeprecationWarning` for sparse setting - eventually remove probably, but leave space for a new implementation.

ilan-gold:
@ivirshup `__setitem__` does work in so far as the tests pass.  So I think it would be worth adding the `DeprecationWarning` (or perhaps some sort of `InstabilityWarning` which probably better describes this) to the `__setitem__` method.  Does that work?

anndata/tests/test_backed_sparse.py
```diff
@@ -49,17 +87,21 @@ def test_backed_indexing(ondisk_equivalent_adata, subset_func, subset_func2):
         pytest.param(sparse.csc_matrix, sparse.hstack),
     ],
 )
-def test_dataset_append_memory(tmp_path, sparse_format, append_method):
-    h5_path = tmp_path / "test.h5"
+def test_dataset_append_memory(tmp_path, sparse_format, append_method, diskfmt):
+    path = (
+        tmp_path / f"test.{diskfmt.replace('ad', '')}"
+    )  # diskfmt is either h5ad or zarr
```
ilan-gold, line 93: Added this and the one below.  These tests are "positive" in that they tests if appending works but I didn't add `zarr` to the negative tests (wrong shape/format).  I can though, but like I said, we already have slow tests so not sure it's super necessary.

anndata/_core/sparse_dataset.py
```diff
@@ -304,10 +300,17 @@ def _normalize_index(
             row, col = np.ix_(row, col)
         return row, col
 
+    # def __setitem__(self, index: Union[Index, Tuple[()]], value):
+
+    #     row, col = self._normalize_index(index)
+    #     mock_matrix = self._to_backed()
+    #     mock_matrix[row, col] = value
+
```
ilan-gold, line 308: @ivirshup `__setitem__` does work in so far as the tests pass.  So I think it would be worth adding the `DeprecationWarning` (or perhaps some sort of `InstabilityWarning` which probably better describes this) to the `__setitem__` method.  Does that work?

ivirshup (commented):
@ilan-gold looking very good!

I've added a few comments, but otherwise can we add a changelog entry?

anndata/_core/sparse_dataset.py
ilan-gold: I didn't have anything in mind specifically, but I also thought we were not planning on actually deprecating this since I imagine eventually we do want to bring it back somehow.  So the instability would be referring to behavior surrounding performance and/or write location (i.e., if we ever move to `kerchunk`).

anndata/_core/sparse_dataset.py
```diff
@@ -382,6 +385,23 @@ def to_memory(self) -> ss.spmatrix:
...
-@_subset.register(SparseDataset)
+class CSRDataset(BaseCompressedSparseDataset):
+    format_str = "csr"
+
+
+class CSCDataset(BaseCompressedSparseDataset):
+    format_str = "csc"
+
+
+def sparse_dataset(group) -> BaseCompressedSparseDataset:
```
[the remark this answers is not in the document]
ilan-gold: Good point, will do.

ilan-gold:
@ivirshup Where is the changelog? Is it the `release-notes`?  And if so, which release?

ivirshup:
> Where is the changelog? 

`release-notes/0.10.0.md`

anndata/_core/sparse_dataset.py
ivirshup: I think deprecation would be fine.

Even if there were atomic updates via `kerchunk` (what I think you're talking about?) this would require changes in user code to setup.

ilan-gold:
@ivirshup have  look at the deprecation warning! thanks!

anndata/_core/sparse_dataset.py
```diff
@@ -304,10 +305,20 @@ def _normalize_index(
             row, col = np.ix_(row, col)
         return row, col
 
+    def __setitem__(self, index: Union[Index, Tuple[()]], value):
+        warnings.warn(
+            "__setitem__ will likely be removed in the near future. We do not recommend relying on its stability.",
+            PendingDeprecationWarning,
+        )
+        row, col = self._normalize_index(index)
+        mock_matrix = self._to_backed()
```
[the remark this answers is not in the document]
ivirshup: This is currently throwing an error, but I do like the idea of making it private.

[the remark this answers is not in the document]
ilan-gold: It's used publicly.....

ilan-gold:
I'm not sure where I got the idea in my head that this branch was passing CI but in any case, should we make `zarr` a dep of `anndata` now?  It's only used for type checking right now in `sparse_dataset.py`

ivirshup:
In `anndata.compat` there are dummy classes defined for this. Basically replace `zarr.Array` and `zarr.Group` with `anndata.compat.ZarrArray` and `anndata.compat.ZarrGroup`

anndata/_core/file_backing.py
```diff
@@ -7,8 +7,8 @@
 import h5py
 
 from . import anndata
-from .sparse_dataset import SparseDataset
-from ..compat import ZarrArray, DaskArray, AwkArray
+from .sparse_dataset import BaseCompressedSparseDataset
+from ..compat import ZarrArray, ZarrGroup, DaskArray, AwkArray
```
ivirshup, line 11: @ilan-gold like this

Issue #781, opened 2022-06-07
Simple comparative benchmarks


ivirshup:
We should have a simple comparative benchmark (with prose) comparing using AnnData and our on disk format to alternative structures.

Initial measurements could look at:

* IO time
* IO memory
* On disk size
* In memory size

Alternatives to compare against include:

* Serialized R object (SeuratObject, maybe also SingleCellExperiment)
* Loom

Variables to consider here include file format (hdf5 vs zarr), chunking, and compression.

ivirshup:
Good resources to use here include:

* [Benchmarking chapter from Jake VanderPlas' Python Data Science HandBook](https://jakevdp.github.io/PythonDataScienceHandbook/01.07-timing-and-profiling.html)
* [loompy docs](http://linnarssonlab.org/loompy/index.html)
* [Seurat docs](https://satijalab.org/seurat/index.html)

PR #786, opened 2022-06-15
Seurat Benchmark R Markdown

ganesh25:
Added the Seurat Benchmark R markdown file that I implemented in R studio cloud (unfortunately I am not able to knit to pdf/html due to memory constraints) 

Draft PR for issue #781


ivirshup (changes requested):
Thanks for the initial PR

We're going to want this running locally with results. Have you been able to set up a conda environment with Seurat installed?

I would also specifically want to be benchmarking the same things as we are benchmarking for AnnData, so that we can compare them.

PR #800, opened 2022-08-08
Support buffer IO for .h5ad

flying-sheep:
So far we only support reading/writing from paths, this PR adds tests / type hints for binary buffers / file-like objects.

- [x] TODO: tests


anndata/_io/h5ad.py
```diff
@@ -68,19 +71,23 @@ def write_h5ad(
...
         f.attrs.setdefault("encoding-type", "anndata")
         f.attrs.setdefault("encoding-version", "0.1.0")
 
         if "X" in as_dense and isinstance(adata.X, (sparse.spmatrix, SparseDataset)):
             write_sparse_as_dense(f, "X", adata.X, dataset_kwargs=dataset_kwargs)
-        elif not (adata.isbacked and Path(adata.filename) == Path(filepath)):
+        elif not (
+            adata.isbacked
+            and (
+                not isinstance(file, (str, Path)) or Path(adata.filename) == Path(file)
```
flying-sheep, line 88: If `file` is a buffer, we can’t reliably get a name from it.

anndata/_io/h5ad.py
```diff
@@ -128,10 +135,13 @@ def write_sparse_as_dense(f, key, value, dataset_kwargs=MappingProxyType({})):
         del f[key]
 
 
-def read_h5ad_backed(filename: Union[str, Path], mode: Literal["r", "r+"]) -> AnnData:
-    d = dict(filename=filename, filemode=mode)
+def read_h5ad_backed(
+    file: Union[str, Path, BinaryIO], mode: Literal["r", "r+"]
+) -> AnnData:
+    file_name = Path(file) if isinstance(file, (str, Path)) else getattr(file, "name")
+    d = dict(filename=file_name, filemode=mode)
```
flying-sheep, line 142: attempt to get the name here. until now we’d just add the file object as `filename`

anndata/_io/h5ad.py
```diff
@@ -68,19 +71,23 @@ def write_h5ad(
...
 
         if "X" in as_dense and isinstance(adata.X, (sparse.spmatrix, SparseDataset)):
             write_sparse_as_dense(f, "X", adata.X, dataset_kwargs=dataset_kwargs)
-        elif not (adata.isbacked and Path(adata.filename) == Path(filepath)):
+        elif not (
+            adata.isbacked
+            and (
+                not isinstance(file, (str, Path)) or Path(adata.filename) == Path(file)
+            )
+        ):
```
flying-sheep, line 90: does this make sense?

anndata/_io/h5ad.py
```diff
@@ -128,10 +135,20 @@ def write_sparse_as_dense(f, key, value, dataset_kwargs=MappingProxyType({})):
...
 
 
-def read_h5ad_backed(filename: Union[str, Path], mode: Literal["r", "r+"]) -> AnnData:
-    d = dict(filename=filename, filemode=mode)
+def _maybe_get_file_name(file) -> Path | None:
+    if isinstance(file, (str, Path)):
+        return Path(file)
+    if hasattr(file, "name"):
+        return Path(file.name)
+    return None
```
flying-sheep, line 143: best effort to set `filename`. If we can’t, we can’t.

jeffjc97:
Hi! I just started playing around with `anndata` but I think this would solve a problem I'm working on. Any chance you were thinking of merging this soon @flying-sheep? Thank you!

ivirshup:
General question about the PR, why buffers specifically and not `h5py.File`/ `h5py.Group` or `zarr` stores directly?

flying-sheep:
That’s also a good idea, but this PR is about the most generic serialization format: a byte stream.

We can add the others at another date.

ivirshup:
Currently, this works:

```python
import anndata as ad
from anndata.experimental import read_elem, write_elem

import h5py

from io import BytesIO

adata = ad.read_h5ad("/Users/isaac/data/pbmc3k_raw.h5ad")

bio = BytesIO()
with h5py.File(bio, "w") as f:
    write_elem(f, "/", adata)

with h5py.File(bio, "r") as f:
    from_bytes = read_elem(f["/"])
```

I don't think I want to directly support file-like objects as an input type to `read_h5ad`. It's more to support on our side, and the docs in h5py for this feature are full of "use at your own risk" warnings. See also h5py/h5py#1698

ivirshup:
What is the use case here? I'm not too familiar with Quilt.

flying-sheep:
Check out their promo material: https://quiltdata.com/

Basically it’s a catalog to browse and manage in-house or public data. The data is stored in versioned “packages” (file trees on S3) together with searchable/queryable metadata. Metadata can be validated using schemas.

But if `[read|write]_elem` works, it’s true, they can just use that.

ivirshup:
For quilt:

I saw that, but the specifics were a little unclear. Looks like it's only s3?

I think there are some similarities to things I'd like to do with zarr and OME, but not quite sure yet.

-------------

For specific use case, are you expecting the data to always be delivered as streams? If so, I think pickling would have less overhead from chunking the arrays. If the data is on the cloud, `zarr` may be a nicer choice for storage format (plus defaults to more modern compression libraries).

flying-sheep:
I agree about zarr, hdf5 is a legacy thing. I also don’t know if their (de)serialization supports S3 file hierarchies as opposed to individual files.

Regarding pickling: I avoid that format for anything but caches because of its lack of stability.

ivirshup:
One can have backwards compatible pickled objects. I believe both numpy and pandas do this.

I don't think it should be too hard for us, as anndata is basically builtin python types and those. I'd be up for a PR implementing this.

Of course you could have then have references to items which don't have compatibility guarantees.

flying-sheep:
yeah, that’s the problem, right? it’s easy that something sneaks in and everything breaks. Recovery is near impossible.

PR #647, opened 2021-11-14
first attempt to support awkward arrays

giovp:
First draft at supporting [awkward arrays](https://awkward-array.org/quickstart.html#).
As discussed with @ivirshup this would be useful for Squidpy @hspitzer and potentially @Zethson EHR project. 

Here's a walkthrough showcasing some basic functionality we could use for sub observation annotations (e.g. spatial coordinates of rna-molecules/segmentations).

<details>
<summary>Details</summary>

```python
import numpy as np
import scanpy as sc
import squidpy as sq
import matplotlib.pyplot as plt
from numpy.random import default_rng
from sklearn.datasets import make_blobs
import awkward as ak

adata = sq.datasets.visium_hne_adata()
adata = adata[:10, :].copy()
adata.obsm["spatial"] = (
    adata.obsm["spatial"] - np.std(adata.obsm["spatial"], 0)
) / np.mean(adata.obsm["spatial"], 0)

# generate data
obs_list = []
rng = default_rng(42)
for idx in adata.obs_names.values:
    coord, _ = make_blobs(
        n_samples=rng.integers(5, 15),
        cluster_std=0.02,
        centers=adata[idx].obsm["spatial"],
        random_state=42,
    )
    obs_list.append(coord)

sub_obs = ak.Array(obs_list)

fig, ax = plt.subplots(1, 2, figsize=(10, 5))
ax[0].set_aspect("equal")
ax[0].scatter(x=adata.obsm["spatial"][:, 0], y=adata.obsm["spatial"][:, 1])
for i in range(adata.shape[0]):
    ax[1].scatter(
        x=sub_obs[i, :, 0],
        y=sub_obs[i, :, 1],
    )
    ax[1].axis("equal")
```

image

```python
adata.obsm["sub_obs"] = sub_obs  # sub_obs is an awkward array
adata_subset = adata[:5]  # let's subset, it also works for `.copy()`

fig, ax = plt.subplots(1, 2, figsize=(10, 5))
ax[0].set_aspect("equal")
ax[0].scatter(
    x=adata_subset.obsm["spatial"][:, 0], y=adata_subset.obsm["spatial"][:, 1]
)
for i in range(adata_subset.shape[0]):
    ax[1].scatter(
        x=adata_subset.obsm["sub_obs"][i, :, 0],
        y=adata_subset.obsm["sub_obs"][i, :, 1],
    )
    ax[1].set_aspect("equal")
```
image
</details>

[...]

PR #813, opened 2022-09-05
Dask array support

syelman:
A PR to get started on #693.
_Going to update this as we move forward. Also, please correct me if I got something wrong._

We should start looking on having a dask.Array in obsm, varm, layers, etc. First objective is to write the suitable test functions and maybe discuss solutions for them to pass the tests.

## Current state
So one should understand the AnnData repository overall (especially the ``_core`` and ``_io`` folders) to see the expected behavior and write tests according to it (i.e., find the methods that should raise a NotImplementedError in this state).

After the tests or meanwhile writing the tests, we can discuss possible solutions.

Update: I just added `dask.Array` to the default  of `tests.helpers.gen_adata` for obsm and varm like this
```python
obsm = dict(
    array=np.random.random((M, 50)),
    ...
    da=da.random.random((M, 50), chunks=(min(M,50)//5+1,)*2), 
)
```
Now it gives the all the errors we want to see I think when we call test_readwrite or something. Example output
```
FAILED anndata/tests/test_readwrite.py::test_changed_obs_var_names[h5ad] - TypeError: No method has been defined for writing <class 'dask.array.core.Array'> elements to <class 'h5py._hl.group.Group'> 
ERROR anndata/tests/test_readwrite.py::test_maintain_layers - TypeError: No method has been defined for writing <class 'dask.array.core.Array'> elements to <class 'h5py._hl.group.Group'>
```
I think I will do some reading on this h5 and zarr group concepts, then try to solve these test_readwrite errors.

ping: @ivirshup @rahulbshrestha

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1486,8 +1486,6 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         """
         new = {}
 
-        self.uns = to_memory(self.uns)
-
         if self.obsm is not None:
             new["obsm"] = AxisArrays(
                 self, 0, vals={k: to_memory(v) for k, v in self.obsm.items()}
@@ -1504,8 +1502,12 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["varp"] = PairwiseArrays(
                 self, 1, vals={k: to_memory(v) for k, v in self.varp.items()}
             )
+        if self.uns is not None:
+            new["uns"] = {k: to_memory(v) for k, v in self.uns.items()}
+        if self.layers is not None:
+            new["layers"] = {k: to_memory(v) for k, v in self.layers.items()}
 
-        for key in ["X", "obs", "var", "layers", "uns", "raw"]:
+        for key in ["X", "obs", "var", "raw"]:
             elem = getattr(self, key)
             if elem is not None:
                 elem = to_memory(elem)
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1547,6 +1547,14 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def to_memory(self, copy=False) -> "AnnData":
         """Load backed AnnData object into memory.
 
+        Args:
+            copy (bool, optional):
+                If file is not backed and if set it will return a copy of
+                anndata with the lazy objects computed. Original anndata will
+                still contain the lazy objects.
+                If false the lazy objects will be replaced by their computed
+                 versions. Defaults to False.
+
         Example
         -------
 
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1480,7 +1480,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["raw"] = self.raw.copy()
         return AnnData(**new)
 
-    def _to_memory_compute_lazy(self):
+    def _to_memory_replace_lazy(self):
         """
         Instantiates lazy objects stored.
         For example calls compute() on dask.array
@@ -1518,9 +1518,33 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             if isinstance(self.X, DaskArray):
                 self.X = self.X.compute()
 
-        return self
+    def _to_memory_copy(self):
+        """Creates a new adata that has the computed results of the previous adata.
+        This is for the unbacked case
+        """
+        new = {}
+        for attr_name in ["obsm", "varm", "obsp", "varp", "layers"]:
+            attr = getattr(self, attr_name, None)
+            if attr is not None:
+                new[attr_name] = {k: to_memory(v) for k, v in attr.items()}
+        if self.uns is not None:
+            new["uns"] = {k: to_memory(v) for k, v in self.uns.items()}
+        if self.raw is not None:
+            new["raw"] = {
+                "X": to_memory(self.raw.X),
+                "var": self.raw.var,
+                "varm": {k: to_memory(v) for k, v in self.raw.varm.items()},
+            }
+        for key in ["X", "obs", "var"]:
+            elem = getattr(self, key)
+            if elem is not None:
+                elem = to_memory(elem)
+                new[key] = elem
+        if self._has_X():
+            new["dtype"] = new["X"].dtype
+        return AnnData(**new)
 
-    def to_memory(self) -> "AnnData":
+    def to_memory(self, copy=False) -> "AnnData":
         """Load backed AnnData object into memory.
 
         Example
@@ -1533,7 +1557,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         if not self.isbacked:
-            adata = self._to_memory_compute_lazy()
+            if copy:
+                adata = self._to_memory_copy()
+            else:
+                self._to_memory_replace_lazy()
+                adata = self
         else:
             elems = {"X": to_memory(self.X)}
             if self.raw is not None:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -26,7 +26,7 @@ from scipy.sparse import issparse, csr_matrix
 from anndata._warnings import ImplicitModificationWarning
 from .raw import Raw
 from .index import _normalize_indices, _subset, Index, Index1D, get_vector
-from .file_backing import AnnDataFileManager, to_memory
+from .file_backing import AnnDataFileManager, to_memory, _to_memory_helper
 from .access import ElementRef
 from .aligned_mapping import (
     AxisArrays,
@@ -1480,64 +1480,33 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["raw"] = self.raw.copy()
         return AnnData(**new)
 
-    def _to_memory_replace_lazy(self):
-        """
-        Instantiates lazy objects stored.
-        For example calls compute() on dask.array
-        """
-
-        for attr_name in ["obsm", "varm", "obsp", "varp", "layers"]:
-            attr = getattr(self, attr_name, None)
-            if attr is not None:
-                for k, v in attr.items():
-                    if isinstance(v, DaskArray):
-                        attr[k] = v.compute()
-
-        if self.uns is not None:
-
-            maps = []
-            if isinstance(self.uns, Mapping):
-                maps.append(self.uns)
-            while maps:
-                uns = maps.pop()
-                for k, v in uns.items():
-                    if isinstance(v, DaskArray):
-                        uns[k] = v.compute()
-                    elif isinstance(v, Mapping):
-                        maps.append(v)
-
-        if self.raw is not None:
-            if isinstance(self.raw.X, DaskArray):
-                self.raw._X = self.raw.X.compute()
-            if self.raw._varm is not None:
-                for k, v in self.raw._varm.items():
-                    if isinstance(v, DaskArray):
-                        self.raw._varm[k] = v.compute()
-
-        if self.X is not None:
-            if isinstance(self.X, DaskArray):
-                self.X = self.X.compute()
-
-    def _to_memory_copy(self):
-        """Creates a new adata that has the computed results of the previous adata.
-        This is for the unbacked case
+    def _to_memory_unbacked(self, copy=False):
+        """Creates a new adata that has the computed results of the previous
+        adata. This is for the unbacked case.
         """
         new = {}
-        for attr_name in ["obsm", "varm", "obsp", "varp", "layers", "uns"]:
+        for attr_name in [
+            "X",
+            "obs",
+            "var",
+            "obsm",
+            "varm",
+            "obsp",
+            "varp",
+            "layers",
+            "uns",
+        ]:
             attr = getattr(self, attr_name, None)
             if attr is not None:
-                new[attr_name] = {k: to_memory(v) for k, v in attr.items()}
+                new[attr_name] = _to_memory_helper(attr, copy)
+
         if self.raw is not None:
             new["raw"] = {
-                "X": to_memory(self.raw.X),
-                "var": self.raw.var,
-                "varm": {k: to_memory(v) for k, v in self.raw.varm.items()},
+                "X": _to_memory_helper(self.raw.X, copy),
+                "var": _to_memory_helper(self.raw.var, copy),
+                "varm": _to_memory_helper(self.raw.varm, copy),
             }
-        for key in ["X", "obs", "var"]:
-            elem = getattr(self, key)
-            if elem is not None:
-                elem = to_memory(elem)
-                new[key] = elem
+
         if self._has_X():
             new["dtype"] = new["X"].dtype
         return AnnData(**new)
@@ -1549,9 +1518,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             copy (bool, optional):
                 If file is not backed and if set it will return a copy of
                 anndata with the lazy objects computed. Original anndata will
-                still contain the lazy objects.
-                If false the lazy objects will be replaced by their computed
-                versions. Defaults to False.
+                still contain the lazy objects. If false the lazy objects
+                will be replaced by their computed versions in the returned
+                object but, their other contents reference to the objects from
+                original adata. Defaults to False.
 
         Example
         -------
@@ -1563,11 +1533,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         if not self.isbacked:
-            if copy:
-                adata = self._to_memory_copy()
-            else:
-                self._to_memory_replace_lazy()
-                adata = self
+            adata = self._to_memory_unbacked(copy=copy)
         else:
             elems = {"X": to_memory(self.X)}
             if self.raw is not None:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1480,6 +1480,53 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["raw"] = self.raw.copy()
         return AnnData(**new)
 
+    def _to_memory_copy(self):
+        """Creates a new adata that has the computed results of the previous adata.
+        This is for the unbacked case
+        """
+        already_in_mem = False
+        new = {}
+
+        # TODO: How to to_mem uns?
+        # TODO: If there is a case that the object
+        # can both include lazy attributes and be in backed mode
+        # support that case
+
+        # TODO: uns
+        # unstructured annotations
+        # self.uns = uns or OrderedDict()
+
+        # TODO: check if its in memory already
+        if self.obsm is not None:
+            new["obsm"] = AxisArrays(
+                self, 0, vals={k: to_memory(v) for k, v in self.obsm.items()}
+            )
+        if self.varm is not None:
+            new["varm"] = AxisArrays(
+                self, 1, vals={k: to_memory(v) for k, v in self.varm.items()}
+            )
+        if self.obsp is not None:
+            new["obsp"] = PairwiseArrays(
+                self, 0, vals={k: to_memory(v) for k, v in self.obsp.items()}
+            )
+        if self.varp is not None:
+            new["varp"] = PairwiseArrays(
+                self, 1, vals={k: to_memory(v) for k, v in self.varp.items()}
+            )
+
+        for key in ["X", "obs", "var", "layers", "uns", "raw"]:
+            elem = getattr(self, key)
+            if elem is not None:
+                prev_type = type(elem)
+                elem = to_memory(elem)
+                if not isinstance(elem, prev_type):
+                    # TODO Delete this if already_in_mem is not used
+                    already_in_mem = False
+                new[key] = elem
+        if self._has_X():
+            new["dtype"] = new["X"].dtype
+        return AnnData(**new), already_in_mem
+
     def to_memory(self) -> "AnnData":
         """Load backed AnnData object into memory.
 
@@ -1493,7 +1540,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         if not self.isbacked:
-            raise ValueError("Object is already in memory.")
+            adata, already_in_mem = self._to_memory_copy()
+            if already_in_mem:
+                raise ValueError("Object is already in memory.")
         else:
             elems = {"X": to_memory(self.X)}
             if self.raw is not None:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -26,7 +26,7 @@ from scipy.sparse import issparse, csr_matrix
 from anndata._warnings import ImplicitModificationWarning
 from .raw import Raw
 from .index import _normalize_indices, _subset, Index, Index1D, get_vector
-from .file_backing import AnnDataFileManager, to_memory, _to_memory_helper
+from .file_backing import AnnDataFileManager, to_memory
 from .access import ElementRef
 from .aligned_mapping import (
     AxisArrays,
@@ -1480,9 +1480,22 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["raw"] = self.raw.copy()
         return AnnData(**new)
 
-    def _to_memory_unbacked(self, copy=False):
-        """Creates a new adata that has the computed results of the previous
-        adata. This is for the unbacked case.
+    def to_memory(self, copy=True) -> "AnnData":
+        """Return a new AnnData object with all backed arrays loaded into memory.
+
+        Params
+        ------
+            copy:
+                Whether the arrays that are already in-memory should be copied.
+
+        Example
+        -------
+
+        .. code:: python
+
+            import anndata
+            backed = anndata.read_h5ad("file.h5ad", backed="r")
+            mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         new = {}
         for attr_name in [
@@ -1498,53 +1511,22 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         ]:
             attr = getattr(self, attr_name, None)
             if attr is not None:
-                new[attr_name] = _to_memory_helper(attr, copy)
+                new[attr_name] = to_memory(attr, copy)
 
         if self.raw is not None:
             new["raw"] = {
-                "X": _to_memory_helper(self.raw.X, copy),
-                "var": _to_memory_helper(self.raw.var, copy),
-                "varm": _to_memory_helper(self.raw.varm, copy),
+                "X": to_memory(self.raw.X, copy),
+                "var": to_memory(self.raw.var, copy),
+                "varm": to_memory(self.raw.varm, copy),
             }
 
         if self._has_X():
             new["dtype"] = new["X"].dtype
-        return AnnData(**new)
 
-    def to_memory(self, copy=False) -> "AnnData":
-        """Load backed AnnData object into memory.
-
-        Args:
-            copy (bool, optional):
-                If file is not backed and if set it will return a copy of
-                anndata with the lazy objects computed. Original anndata will
-                still contain the lazy objects. If false the lazy objects
-                will be replaced by their computed versions in the returned
-                object but, their other contents reference to the objects from
-                original adata. Defaults to False.
-
-        Example
-        -------
-
-        .. code:: python
-
-            import anndata
-            backed = anndata.read_h5ad("file.h5ad", backed="r")
-            mem = backed[backed.obs["cluster"] == "a", :].to_memory()
-        """
-        if not self.isbacked:
-            adata = self._to_memory_unbacked(copy=copy)
-        else:
-            elems = {"X": to_memory(self.X)}
-            if self.raw is not None:
-                elems["raw"] = {
-                    "X": to_memory(self.raw.X),
-                    "var": self.raw.var,
-                    "varm": self.raw.varm,
-                }
-            adata = self._mutated_copy(**elems)
+        if self.isbacked:
             self.file.close()
-        return adata
+
+        return AnnData(**new)
 
     def copy(self, filename: Optional[PathLike] = None) -> "AnnData":
         """Full copy, optionally on disk."""
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -339,7 +339,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         self._layers = adata_ref.layers._view(self, (oidx, vidx))
         self._obsp = adata_ref.obsp._view(self, oidx)
         self._varp = adata_ref.varp._view(self, vidx)
-        # Speical case for old neighbors, backwards compat. Remove in anndata 0.8.
+        # Special case for old neighbors, backwards compat. Remove in anndata 0.8.
         uns_new = _slice_uns_sparse_matrices(
             copy(adata_ref._uns), self._oidx, adata_ref.n_obs
         )
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1506,8 +1506,14 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             new["uns"] = {k: to_memory(v) for k, v in self.uns.items()}
         if self.layers is not None:
             new["layers"] = {k: to_memory(v) for k, v in self.layers.items()}
+        if self.raw is not None:
+            new["raw"] = {
+                "X": to_memory(self.raw.X),
+                "var": self.raw.var,
+                "varm": {k: to_memory(v) for k, v in self.raw.varm.items()},
+            }
 
-        for key in ["X", "obs", "var", "raw"]:
+        for key in ["X", "obs", "var"]:
             elem = getattr(self, key)
             if elem is not None:
                 elem = to_memory(elem)
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1484,7 +1484,6 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         """Creates a new adata that has the computed results of the previous adata.
         This is for the unbacked case
         """
-        already_in_mem = False
         new = {}
 
         # TODO: How to to_mem uns?
@@ -1517,15 +1516,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         for key in ["X", "obs", "var", "layers", "uns", "raw"]:
             elem = getattr(self, key)
             if elem is not None:
-                prev_type = type(elem)
                 elem = to_memory(elem)
-                if not isinstance(elem, prev_type):
-                    # TODO Delete this if already_in_mem is not used
-                    already_in_mem = False
                 new[key] = elem
         if self._has_X():
             new["dtype"] = new["X"].dtype
-        return AnnData(**new), already_in_mem
+        return AnnData(**new)
 
     def to_memory(self) -> "AnnData":
         """Load backed AnnData object into memory.
@@ -1540,9 +1535,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         if not self.isbacked:
-            adata, already_in_mem = self._to_memory_copy()
-            if already_in_mem:
-                raise ValueError("Object is already in memory.")
+            adata = self._to_memory_copy()
         else:
             elems = {"X": to_memory(self.X)}
             if self.raw is not None:
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -468,7 +468,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 #       as in readwrite.read_10x_h5
                 if X.dtype != np.dtype(dtype):
                     X = X.astype(dtype)
-            elif isinstance(X, ZarrArray):
+            elif isinstance(X, (ZarrArray, DaskArray)):
                 X = X.astype(dtype)
             else:  # is np.ndarray or a subclass, convert to true np.ndarray
                 X = np.array(X, dtype, copy=False)
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1523,12 +1523,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         This is for the unbacked case
         """
         new = {}
-        for attr_name in ["obsm", "varm", "obsp", "varp", "layers"]:
+        for attr_name in ["obsm", "varm", "obsp", "varp", "layers", "uns"]:
             attr = getattr(self, attr_name, None)
             if attr is not None:
                 new[attr_name] = {k: to_memory(v) for k, v in attr.items()}
-        if self.uns is not None:
-            new["uns"] = {k: to_memory(v) for k, v in self.uns.items()}
         if self.raw is not None:
             new["raw"] = {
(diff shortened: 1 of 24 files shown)


anndata/tests/test_dask.py
```diff
@@ -20,3 +21,87 @@ def test_dask_X_view():
     adata.X = da.ones((M, N))
     view = adata[:30]
     view.copy()
+
+
+# TODO: Appropriate path to write in testing.
+# TODO: Check written object
```
ivirshup, line 27:
```suggestion
# TODO: Appropriate path to write in testing.
# TODO: Check written object
# TODO: Check that this works with `zarr`
```

ivirshup (commented):
Thanks for getting this started!

ivirshup:
I think you should also start looking at having a `dask.Array` in `obsm`, `varm`, `layers`, etc. In many cases I think those will be more straightforward.

ivirshup (commented):
## Tests

Glad you're getting started with tests! I think the tests in test_dask look like a good starting place.

Is the goal to get these passing first?

### `gen_adata`

It can be useful to have the helper function change, but I would also consider looking at fixtures which define array types.

If a test that uses `gen_adata` errors around a specific type, you don't really know that the other types wouldn't have thrown an error. If a test is parameterized by types then you can see which types are working and which are failing. This is especially useful if you start touching the generic code. For example, see fixtures with names like `array_type` or `matrix_type`.

### Comparing dask arrays

Right now it looks like there are some errors coming from trying to compare dask arrays to in memory values. This can probably be handled a little generically in `assert_equal` by calling `compute` when comparing a dask object to an in memory object. I expect the dask and xarray test suites may have some other useful strategies to look at here.

anndata/tests/test_dask.py
```diff
@@ -20,3 +21,88 @@ def test_dask_X_view():
...
+
+    M, N = 50, 30
+    adata = ad.AnnData(
+        obs=pd.DataFrame(index=[f"cell{i:02}" for i in range(M)]),
+        var=pd.DataFrame(index=[f"gene{i:02}" for i in range(N)]),
+    )
+    adata.X = da.ones((M, N))
+    view = adata[:30]
+    view.copy()
+    view.write("test.h5ad")
```
ivirshup, line 60: I would suggest something a little more like this.

```suggestion
@pytest.fixture(params=["h5ad", "zarr"])
def diskfmt(request):
    return request.param


def test_dask_write(tmp_path, diskfmt):
    import dask.array as da

    pth = tmp_path / f"test_write.{diskfmt}"
    write = lambda x, y: getattr(x, f"write_{diskfmt}")(y)
    read = lambda x: getattr(ad, f"read_{diskfmt}")(x)

    M, N = 50, 30
    adata = ad.AnnData(
        obs=pd.DataFrame(index=[f"cell{i:02}" for i in range(M)]),
        var=pd.DataFrame(index=[f"gene{i:02}" for i in range(N)]),
    )
    adata.X = da.random.randint((M, N))
    write(adata, pth)
    result = read(pth)

    assert_equal(adata, result)
```

* The diskfmt fixture lets us test both the hdf5 and zarr backends
* We're checking that our result is equal to what we expect
* I would have one test for views, one which doesn't use views.

syelman: After applying your suggestions, I realized making these changes makes the tests wrong indeed. But I modified the write tests, and I was surprised that they passed. Turns out, AnnData converts X to ndarrays before storing.
First I will focus on the `test_dask_write` cases which gives not supported write error when obsm/varm gets assigned.

And in the future for X, we could first focus on this case:
``` python
assert type(adata.X) == da.Array # doesn't pass
```

syelman:
@ivirshup @rahulbshrestha, I did some changes that enables the Zarr support. I still don't know what encoding-type and encoding-version refer to, so there is probably a cleaner way to do this.

anndata/_io/specs/methods.py
```diff
@@ -303,10 +306,30 @@ def write_list(f, k, elem, dataset_kwargs=MappingProxyType({})):
 @_REGISTRY.register_write(ZarrGroup, h5py.Dataset, IOSpec("array", "0.2.0"))
 @_REGISTRY.register_write(ZarrGroup, np.ma.MaskedArray, IOSpec("array", "0.2.0"))
 def write_basic(f, k, elem, dataset_kwargs=MappingProxyType({})):
-    """Write methods which underlying library handles nativley."""
+    """Write methods which underlying library handles natively."""
     f.create_dataset(k, data=elem, **dataset_kwargs)
 
 
+@_REGISTRY.register_write(ZarrGroup, da.Array, IOSpec("dask-array", "0.2.0"))
```
ivirshup, line 313: This should just write as a normal array, since their on disk representation is the same. So `IOSpec("array", "0.2.0")`

anndata/_io/specs/methods.py
```diff
@@ -303,10 +306,30 @@ def write_list(f, k, elem, dataset_kwargs=MappingProxyType({})):
...
+def write_zarr(f, k, elem, dataset_kwargs=MappingProxyType({})):
+    g: ZarrGroup = f.create_group(k)
```
ivirshup, line 315: You shouldn't need to create a group here. As for  the other dense arrays, this can be saved directly to the key.

anndata/_io/specs/methods.py
```diff
@@ -303,10 +306,30 @@ def write_list(f, k, elem, dataset_kwargs=MappingProxyType({})):
...
+    # TODO: Understand IOSpec stuff
+    da.Array.to_zarr(
```
ivirshup, line 317: I don't think this is a static method. I believe it should be used like:

```python
elem.to_zarr(...)
# or
da.to_zarr(elem, ...)
```

ivirshup: Actually, looking into `da.to_zarr` and `da.to_hdf5` I think you may not want to use them. You might want to go a bit lower level and use the `dset = f.require_dataset(k, ..., **dataset_kwargs)` and `da.store(elem, dset)`

ivirshup (commented):
Added some comments around IO. Good start!

syelman:
Added Hdf5 and Zarr support for read and write operations as array. Optional dependency issue is also solved. I will now be looking to the index operations and `test_view.py`.

anndata/tests/test_dask.py
```diff
@@ -93,3 +93,19 @@ def test_assign_X(adata):
 
 
 # TODO: Views
+
+def test_assert_eq(adata):
```
syelman, line 98: I think this is just for demonstration, right? Because we don't use the adata argument. Anyways, it seems like this function should be enough, so I will add it to the dispatch flow.

anndata/tests/test_dask.py
```diff
@@ -93,3 +93,19 @@ def test_assign_X(adata):
...
+
+    import dask.array as da
+    import numpy as np
+    from dask.array.utils import assert_eq
```
syelman, line 102: I think this is just for demonstration, right? Because we don't use the adata argument. Anyway, it seems like this function should be enough, so I will add it to the dispatch flow.

rahulbshrestha: yup, only for demo

syelman:
@ivirshup @rahulbshrestha, I added DaskArray to the view dispatch and added the `matrix_type`. But the thing is, this doesn't cover all the tests.  

Should I create more tests or just refactor the old test functions so that they take `matrix_type` as an argument when they can? For example this function https://github.com/syelman/anndata/blob/687df0686cb6a3210ea550f2f4c5e826f2a62116/anndata/tests/test_views.py#L86. These functions use the `gen_adata` function often, should I add a dask.array support somehow in it?

anndata/tests/helpers.py
```diff
@@ -17,6 +17,8 @@
 from anndata._core.aligned_mapping import AlignedMapping
 from anndata.utils import asarray
 
+import dask.array as da
```
syelman, line 20: do we need this import?

anndata/tests/helpers.py
```diff
@@ -335,6 +337,13 @@ def assert_equal_h5py_dataset(a, b, exact=False, elem_name=None):
     assert_equal(b, a, exact, elem_name=elem_name)
 
 
+@assert_equal.register(da.Array)
+def assert_equal_dask_array(a, b, exact=False, elem_name=None):
+    from dask.array.utils import assert_eq
+
+    assert_eq(a, b, check_dtype=False)
```
syelman, line 344:
```suggestion
assert da.equal(a,b).all()
```
I would just go with this so that the azure tests pass

anndata/tests/helpers.py
ivirshup: I would import it inside the function

ivirshup:
> Should I create more tests or just refactor the old test functions so that they take matrix_type as an argument when they can? For example [this](https://github.com/syelman/anndata/blob/687df0686cb6a3210ea550f2f4c5e826f2a62116/anndata/tests/test_views.py#L86) function:

This would be nice, but I think I'd wait on this until you're more familiar with the test suite.

That specific test is fairly old, and I would like to keep it as is. If there are multiple tests which the same logic but different datatypes it would be great to combine these. Sometimes the logic here can be tricky.

> These functions use the gen_adata function often, should I add a dask.array support somehow in it?

I think this is a broadly a good idea. However, this is used all over the place and many tests would start failing. There are also places like IO tests where the logic is different for dask arrays.

If you want to do this, I would probably make objects returned by `gen_adata` not contain dask arrays by default. So  tests would have to opt in to getting a dask array.

anndata/tests/helpers.py
ivirshup: Is this a python version thing? No need to worry about that, since we'll drop 3.7 in the release this is going into.

ivirshup: Implementing this in #820

syelman: Ok then we are good I think

syelman:
> I would probably make objects returned by `gen_adata` not contain dask arrays by default. So tests would have to opt in to getting a dask array.

I was planning to do it that way. Then I will move on with this.

syelman:
Added new cases

having errors like these:
```
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-X] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-obs] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-var] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-obsm] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-varm] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-obsp] - NotImplementedError: Don't yet support nd fancy indexing
FAILED anndata/tests/test_views.py::test_view_delattr[array_subset-varp] - NotImplementedError: Don't yet support nd fancy indexing
```
The error comes from Dask itself. I will check the subset functions. Dask seams to be supporting masking etc so far. So I might have to adjust the test cases.

ivirshup:
@syelman, for `arr[[1,2],[1,4]]`, we handle this through the `subset` single dispatch method. The dask array could probably use pretty similar code to numpy.

We actually do this because if `arr` was a numpy array, `arr[[1,2],[1,4]]` would return a 1d array with two values. E.g. you are grabbing values at (1, 1) and (2, 4). But I don't think anyone likes that numpy does this, so most other types don't do this.

ivirshup:
`_subset` is defined in `anndata/_core/index.py`

anndata/compat/__init__.py
```diff
@@ -53,6 +53,18 @@ def __repr__():
...
+    # If this was on _core/index.py there would be a circular import.
+    # So registering the _subset function with the import here.
+
+    @_subset.register(DaskArray)
+    def _subset_dask(a: DaskArray, subset_idx):
+        if all(isinstance(x, cabc.Iterable) for x in subset_idx):
+            subset_idx = np.ix_(*subset_idx)
+            return a.vindex[subset_idx]
+        return a[subset_idx]
+
```
syelman, line 67: See [dask.array.Array.vindex](https://docs.dask.org/en/stable/generated/dask.array.Array.vindex.html)

I had to register the _subset here. I am not sure if this is the best practice.

ivirshup: Could you instead do:

```python
from anndata.compat import DaskArray

...

@_subset.register(DaskArray)
def _subset_dask(a: DaskArray, subset_idx) -> DaskArray:
    ...
```

In `anndata._core.index`?

ivirshup: Ah seeing now, what's the circular import?

ivirshup: This has been handled in the awkward array PR, I think you can just copy what was done there: https://github.com/scverse/anndata/pull/647/files#diff-14bdd0423d47482c6db457fa70c6118cff28bbc3675a8e309c962b6c209f157b

syelman: Fixed it on my recent commit. I will now focus on concatenation.

anndata/_io/specs/methods.py
```diff
@@ -301,9 +303,22 @@ def write_list(f, k, elem, dataset_kwargs=MappingProxyType({})):
...
+        DaskArray.store(elem, g)
+    else:
+        f.create_dataset(k, data=elem, **dataset_kwargs)
+
+
+@_REGISTRY.register_read(H5Array, IOSpec("array", "0.2.0"))
+@_REGISTRY.register_read(ZarrArray, IOSpec("array", "0.2.0"))
+@_REGISTRY.register_read(ZarrArray, IOSpec("string-array", "0.2.0"))
+def read_array(elem):
+    return elem[()]
```
ivirshup, line 321:
```suggestion
```

Looks like this has been duplicated

anndata/_io/specs/methods.py
```diff
@@ -301,9 +303,22 @@ def write_list(f, k, elem, dataset_kwargs=MappingProxyType({})):
...
 @_REGISTRY.register_write(ZarrGroup, np.ma.MaskedArray, IOSpec("array", "0.2.0"))
+@_REGISTRY.register_write(ZarrGroup, DaskArray, IOSpec("array", "0.2.0"))
+@_REGISTRY.register_write(H5Group, DaskArray, IOSpec("array", "0.2.0"))
 def write_basic(f, k, elem, dataset_kwargs=MappingProxyType({})):
-    """Write methods which underlying library handles nativley."""
-    f.create_dataset(k, data=elem, **dataset_kwargs)
+    """Write methods which underlying library handles natively."""
+    if isinstance(elem, DaskArray):
+        g = f.require_dataset(k, shape=elem.shape, dtype=elem.dtype, **dataset_kwargs)
+        DaskArray.store(elem, g)
```
ivirshup, line 312: Could this be in a separate function? Like:

```python
@_REGISTRY.register_write(ZarrGroup, DaskArray, IOSpec("array", "0.2.0"))
@_REGISTRY.register_write(H5Group, DaskArray, IOSpec("array", "0.2.0"))
def write_dask_array(f, k, elem, dataset_kwargs=MappingProxyType({})):
     g = f.require_dataset(k, shape=elem.shape, dtype=elem.dtype, **dataset_kwargs)
     DaskArray.store(elem, g)
```

The `write_elem` logic already handles dispatch based on type, so I'd just rely on that rather than putting dispatch flow control within a write method.

anndata/_core/views.py
```diff
@@ -126,6 +126,7 @@ def as_view(obj, view_args):
 
 
 @as_view.register(np.ndarray)
+@as_view.register(DaskArray)
```
ivirshup, line 129: I think this might need it's own view class. 

* It would be nice to know that it's a view of a dask array at a type level.
* Views are being loaded into memory:

```python
import anndata as ad
import dask.array as da
import numpy as np


a = ad.AnnData(
    np.ones((5, 10), dtype=np.float32),
    obsm={"dask": da.random.random((5, 20))}
)
display(a.obsm["dask"])
```

image

```python
view = a[:2]
display(view.obsm["dask"])
```

```
ArrayView([[0.51141181, 0.80505702, 0.72229628, 0.8530678 , 0.84690909,
            0.1601411 , 0.97500309, 0.4163151 , 0.79551743, 0.2428621 ,
            0.50578874, 0.47671809, 0.02821872, 0.99338238, 0.81772692,
            0.84456787, 0.13378153, 0.58699808, 0.17240616, 0.87100689],
           [0.98989405, 0.17072794, 0.94920147, 0.93408043, 0.23794826,
            0.82444377, 0.0869244 , 0.55049418, 0.94464469, 0.79332537,
            0.67213581, 0.41819875, 0.0634449 , 0.0353851 , 0.78107282,
            0.94189   , 0.710782  , 0.35208005, 0.90897056, 0.67785694]])
```

```python
b = view.copy()
display(b.obsm["dask"])
```

```
array([[0.51141181, 0.80505702, 0.72229628, 0.8530678 , 0.84690909,
        0.1601411 , 0.97500309, 0.4163151 , 0.79551743, 0.2428621 ,
        0.50578874, 0.47671809, 0.02821872, 0.99338238, 0.81772692,
        0.84456787, 0.13378153, 0.58699808, 0.17240616, 0.87100689],
       [0.98989405, 0.17072794, 0.94920147, 0.93408043, 0.23794826,
        0.82444377, 0.0869244 , 0.55049418, 0.94464469, 0.79332537,
        0.67213581, 0.41819875, 0.0634449 , 0.0353851 , 0.78107282,
        0.94189   , 0.710782  , 0.35208005, 0.90897056, 0.67785694]])
```

anndata/tests/test_dask.py
```diff
@@ -20,3 +55,69 @@ def test_dask_X_view():
...
+
+    M, N = adata.X.shape
+    adata.obsm["a"] = da.ones((M, 10))
+    adata.varm["a"] = da.ones((N, 10))
+
+    orig = adata
+    write(orig, pth)
+    curr = read(pth)
+    assert da.equal(curr.varm["a"], orig.varm["a"]).all()
+    assert da.equal(curr.obsm["a"], orig.obsm["a"]).all()
```
ivirshup, line 75: Could you use `assert_equal` here? Also `exact=True` should work.

Ideally this would also check `obsp`, `varp` and `layers`.

This would help address the ndarray io test we discussed at the meeting.

anndata/tests/test_dask.py
```diff
@@ -20,3 +55,69 @@ def test_dask_X_view():
...
+def test_dask_write(adata, tmp_path, diskfmt):
+    import dask.array as da
+
+    pth = tmp_path / f"test_write.{diskfmt}"
+    write = lambda x, y: getattr(x, f"write_{diskfmt}")(y)
+    read = lambda x: getattr(ad, f"read_{diskfmt}")(x)
+
+    M, N = adata.X.shape
+    adata.obsm["a"] = da.ones((M, 10))
+    adata.varm["a"] = da.ones((N, 10))
```
ivirshup, line 69: Could you make these randomly generated? This makes checking that the values are the same more strict.

anndata/tests/test_dask.py
```diff
@@ -20,3 +55,69 @@ def test_dask_X_view():
...
+
+    a = np.array([[1, 2, 3], [4, 5, 6]])
+    b = da.from_array([[1, 2, 3], [4, 5, 6]])
+    assert_eq(a, b)
+
+    c = np.ones(10, dtype="int32")
+    d = np.ones(10, dtype="int64")
+    assert_eq(c, d, check_dtype=False)
+
+
```
ivirshup, line 111: As discussed, we don't need this here.

```suggestion
```

ivirshup (commented):
Looking good! I've left some comments on things to fix, but also have a few other requested tests:

- [ ] `adata.to_memory` should load dask arrays into memory
- [ ] `adata.copy()` should not load dask arrays into memory
	- This is currently true, but should be tested

And from our meeting:

- [ ] Tests for assert_equal
  - [x] Test dask array vs sparse array
  - [x] Test that exact works, e.g. that type and dtype are checked

And then onto concatenation!

It would also be good to start thinking about how this should be documented. At the very least, a release note!

anndata/_io/specs/methods.py
syelman: right! fixed it

anndata/_core/views.py
syelman: Ok, in the recent commit the behavior is like this:
First output looks the same.

```python
view = a[:2]
display(view.obsm["dask"])
```
```
DaskArrayView([[0.67451575, 0.6255451 , 0.29382805, 0.32638486,
                0.94882371, 0.96557258, 0.32915445, 0.68920697,
                0.82376327, 0.32344818, 0.65802747, 0.90954457,
                0.89987801, 0.44695522, 0.68305018, 0.97012088,
                0.45402617, 0.8436169 , 0.30581718, 0.66890336],
               [0.00803059, 0.12620636, 0.48831987, 0.91531449,
                0.23068138, 0.33108069, 0.88164289, 0.19952604,
                0.35021352, 0.90056961, 0.49632834, 0.72886952,
                0.69350378, 0.98114544, 0.35685473, 0.52749217,
                0.55908432, 0.5112443 , 0.07758866, 0.92408913]])
```
```python
b = view.copy()
display(b.obsm["dask"])
```
```
array([[0.67451575, 0.6255451 , 0.29382805, 0.32638486, 0.94882371,
        0.96557258, 0.32915445, 0.68920697, 0.82376327, 0.32344818,
        0.65802747, 0.90954457, 0.89987801, 0.44695522, 0.68305018,
        0.97012088, 0.45402617, 0.8436169 , 0.30581718, 0.66890336],
       [0.00803059, 0.12620636, 0.48831987, 0.91531449, 0.23068138,
        0.33108069, 0.88164289, 0.19952604, 0.35021352, 0.90056961,
        0.49632834, 0.72886952, 0.69350378, 0.98114544, 0.35685473,
        0.52749217, 0.55908432, 0.5112443 , 0.07758866, 0.92408913]])
```
Did I get this right or should the outputs look different somehow

syelman:
Hi @ivirshup,
While, waiting for clarifications on my questions above, I added dask array to `test_concatenate.py` functions. There are errors coming from this function in `anndata/__core/merge.py`

https://github.com/scverse/anndata/blob/919d34cfe08faf54daa9f9caa76090739fd66262/anndata/_core/merge.py#L367-L381

This is probably because dask doesn't support `pd.api.extensions` thing. I am wondering if I should go with what was done in **AwkardArrays** or do some reading to make this call work on `_apply_to_array`.

syelman:
Apart from this, I fixed another compatibility issue on concatenation. Traced the merge functions and found out there was a bug because we didn't have equal implemented for **DaskArray**. So I implemented it like this
```python
@equal.register(DaskArray)
def equal_dask_array(a, b) -> bool:

    import dask.array as da

    if isinstance(b, DaskArray):
        return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all().compute()
    return equal(b, asarray(a))
```

The problem with dask equal was it was returning false when np.nan == np.nan.

Now I added plenty of test functions for concatenation. There are two tests don't pass right now, but it's a bug from ``array_equal`` probably.

rahulbshrestha:
> Apart from this, I fixed another compatibility issue on concatenation. Traced the merge functions and found out there was a bug because we didn't have equal implemented for **DaskArray**. So I implemented it like this
> 
> ```python
> @equal.register(DaskArray)
> def equal_dask_array(a, b) -> bool:
> 
>     import dask.array as da
> 
>     if isinstance(b, DaskArray):
>         return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all().compute()
>     return equal(b, asarray(a))
> ```
> 
> The problem with dask equal was it was returning false when np.nan == np.nan.
> 
> Now I added plenty of test functions for concatenation. There are two tests don't pass right now, but it's a bug from `array_equal` probably.



> The problem with dask equal was it was returning false when np.nan == np.nan.

I think we can leave np.nan to be not equal to np.nan as it is part of the definition.

Relevant discussion: https://stackoverflow.com/questions/10034149/why-is-nan-not-equal-to-nan

syelman:
> I think we can leave np.nan to be not equal to np.nan as it is part of the definition.

But in anndata we don't want this behavior. Please see
https://github.com/scverse/anndata/blob/919d34cfe08faf54daa9f9caa76090739fd66262/anndata/_core/merge.py#L93-L102

anndata/tests/helpers.py
```diff
@@ -476,3 +493,18 @@ def fmt_name(x):
...
+
+    return da.from_array(arr, chunks="auto")
+
+
+def darr_from_arr_dense(arr):
+    import dask.array as da
+    import scipy
+
+    if scipy.sparse.issparse(arr):
+        return da.from_array(asarray(arr.toarray()), chunks="auto")
```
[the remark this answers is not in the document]
syelman: Ah, I shouldn't have committed this version. So we were trying to see where it fails on a test with Rahul and I wanted to see if it worked or not. 

The new commit has what I wanted to do. Why do I think it is useful? Because it helped me and Rahul catch some bugs.

anndata/_core/merge.py
```diff
@@ -107,6 +108,16 @@ def equal_dataframe(a, b) -> bool:
     return a.equals(b)
 
 
+@equal.register(DaskArray)
+def equal_dask_array(a, b) -> bool:
+
+    import dask.array as da
+
+    if isinstance(b, DaskArray):
+        return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all().compute()
```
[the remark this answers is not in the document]
syelman: Removing the compute call worked. But the main problem is probably the indexing stuff, I will do some reading.

anndata/tests/test_dask.py
```diff
@@ -20,3 +55,106 @@ def test_dask_X_view():
...
+    adata.varm["a"] = da.random.random((N, 10))
+
+    orig = adata
+    write(orig, pth)
+    curr = read(pth)
+
+    # orig.to_memory() Both give an error
+    # In what case should we assume it shouldn't?
+    # If we assign X should we get into backed mode?
+    # curr.to_memory() # TODO: why does this give an error
```
syelman, line 80: @ivirshup,
One thing I didn't understand with the `to_memory` task, is how it relates to the backed mode. `to_memory` directly checks if the adata is backed or not. But I thought those two concepts were different things.

ivirshup:
@syelman, I have an example for outer indexing to share with you:

Here we are adding columns 5, 6, and 7 to the array with our indexer:

```python
import numpy as np, pandas as pd, dask.array as da
from anndata._core import merge

np_array = np.broadcast_to(np.arange(5), (5, 5))
dask_array = da.array(np_array)

r = merge.Reindexer(pd.Index(range(5)), pd.Index(range(8)))
r.apply(np_array, axis=1)
```

```
array([[ 0.,  1.,  2.,  3.,  4., nan, nan, nan],
       [ 0.,  1.,  2.,  3.,  4., nan, nan, nan],
       [ 0.,  1.,  2.,  3.,  4., nan, nan, nan],
       [ 0.,  1.,  2.,  3.,  4., nan, nan, nan],
       [ 0.,  1.,  2.,  3.,  4., nan, nan, nan]])
```

```python
r.apply(dask_array, axis=1).compute()
```

```
array([[0, 1, 2, 3, 4, 4, 4, 4],
       [0, 1, 2, 3, 4, 4, 4, 4],
       [0, 1, 2, 3, 4, 4, 4, 4],
       [0, 1, 2, 3, 4, 4, 4, 4],
       [0, 1, 2, 3, 4, 4, 4, 4]])
```

It looks like it's being padded with the wrong values, as the failing test would suggest.

anndata/_core/merge.py
```diff
@@ -364,6 +380,24 @@ def _apply_to_df(self, el: pd.DataFrame, *, axis, fill_value=None):
...
+            shape[axis] = len(self.new_idx)
+            return da.broadcast_to(fill_value, tuple(shape))
+
+        indexer = self.old_idx.get_indexer(self.new_idx)
+
+        # Indexes real fast, and does outer indexing
+        # TODO this
+
+        return _subset(el, make_slice(indexer, axis, len(shape)))
+
```
ivirshup, line 400:
```suggestion
        sub_el = _subset(el, make_slice(indexer, axis, len(shape)))
        sub_el[make_slice(indexer == -1, axis, len(shape))] = fill_value
        return sub_el

```

I think this will fix it.

The indexer being generated by `self.old_idx.get_indexer(self.new_idx)` used `-1` to represent missing values. Pandas expects this, but numpy does not. By numpy indexing conventions, this just meant to grab the last value.

I bet the solution I've proposed isn't particularly efficient, but it should work.

ivirshup:
Xarray achieves similar behavior, though I'm not sure how yet:

```python
import numpy as np, pandas as pd, xarray as xr, dask.array as da

dask_dataarray = xr.DataArray(
    da.array(np.arange(6).reshape(2, 3)), [("x", ["a", "b"]), ("y", [10, 20, 30])]
)

xr.concat([dask_dataarray[:, :1], dask_dataarray], dim="x").compute()
```

```
array([[ 0., nan, nan],
       [ 3., nan, nan],
       [ 0.,  1.,  2.],
       [ 3.,  4.,  5.]])
```

But it has a fairly complicated dask graph to show for it:

```python
xr.concat([dask_dataarray[:, :1], dask_dataarray], dim="x").variable.data.visualize()
```

<details>
<summary> Task graph for outer index </summary>

image

</details>

anndata/_core/merge.py
syelman: Great it works!

anndata/_core/merge.py
```diff
@@ -107,6 +109,19 @@ def equal_dataframe(a, b) -> bool:
...
+def equal_dask_array(a, b) -> bool:
+    import dask.array as da
+
+    # TODO: Add potential performance warning
+    # TODO: In which cases do we test the else case?
+    # Why don't we use these dispatches in assert_array_equal anyways?
+    if not isinstance(b, DaskArray):
+        raise NotImplementedError()
+        # TODO: is it better to convert b to DaskArray here?
+    return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all()
```
syelman, line 122: @ivirshup, should I focus on doing this the way xarray did? or should I give an error or warning?

anndata/_core/anndata.py
```diff
@@ -1493,7 +1540,9 @@ def to_memory(self) -> "AnnData":
             mem = backed[backed.obs["cluster"] == "a", :].to_memory()
         """
         if not self.isbacked:
-            raise ValueError("Object is already in memory.")
+            adata, already_in_mem = self._to_memory_copy()
+            if already_in_mem:
+                raise ValueError("Object is already in memory.")
```
syelman, line 1545: The tests fail because, in some cases, we expect `to_memory` to fail when called. So I am not sure if I should check every element and to see if they are dask array or not, just to give an error.

Should the `to_memory` behavior change, i.e., not give errors anymore 

Or should I focus on developing a way to first traverse through the adata to see if it's already on memory, then give an error. This way it would be faster if the call is meant to give an error.

anndata/_core/merge.py
```diff
@@ -107,6 +109,23 @@ def equal_dataframe(a, b) -> bool:
...
+    if a is b:
+        return True
+    if a.shape != b.shape:
+        return False
+    if isinstance(b, DaskArray):
+        if tokenize(a) == tokenize(b):
+            return True
+    # TODO: If we are going to give a performance warning what should it
+    # we say or recommend in the message?
+    return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all()
```
syelman, line 126: I did it similarly to xarray. If I need to write a warning, what should the content be like, I am not sure. I am also not sure in which cases.

Also, one more question, I don't think we ever had a case where a and b were different types. Should I write tests for that, or do we not expect it?

anndata/_core/anndata.py
syelman, line 1545: The tests fail because, in some cases, we expect to_memory to fail when called. So I am not sure if I should check every element and to see if they are dask array or not, just to give an error.

Should the to_memory behavior change, i.e., not give errors anymore

Or should I focus on developing a way to first traverse through the adata to see if it's already on memory, then give an error. This way it would be faster if the call is meant to give an error.

syelman (commented):
hi @ivirshup,

I have only the questions on the reviews remaining, after I add tests for `to_memory()` on `obsp`, `varp`, `layers`, `uns`, and `raw`, all tasks we set would be done.

anndata/_core/anndata.py
ivirshup: I think I'd have it not error anymore. This is what `xarray` does with `load` and `compute`.

anndata/_core/merge.py
ivirshup: Found the code in xarray that does this: https://github.com/pydata/xarray/blob/main/xarray/core/merge.py#L149

I believe this would work for in memory and dask backed arrays.

Also it looks like we are comparing sparse arrays as equal to dense arrays, so type shouldn't matter.

```python
from scipy import sparse
import numpy as np, anndata as ad

varm_array = sparse.random(5, 200, density=0.5, format="csr")

ad1 = ad.AnnData(X=np.ones((5, 5)), varm={"a": varm_array}, dtype=np.float64)
ad2 = ad.AnnData(X=np.zeros((5, 5)), varm={"a": varm_array.A}, dtype=np.float64)

ad.concat([ad1, ad2], merge="unique")
# AnnData object with n_obs × n_vars = 10 × 5
#     varm: 'a'
```

syelman: I wrote the tests, and they work, but I didn't understand why you send that line in the code. Should I implement the same error behavior in our case? If that is the case, it might take more time than I thought, since I would have to read deeper about the merge logic in our code.

syelman:
hi @ivirshup ,

Besides a conversation, we didn't resolve about the warning/error on concatenation, I added all the necessary features and their tests. Here is a summary of what is done last week:

## Last week

- Added tests for concatenation for different types to test the equal dispatch with different arguments.
- Added `dask-array` option to `gen_adata` helper functions (most of them)

About `to_memory`:

- `to_memory` now doesn't give an error when not backed
- When `to_memory` is called each object gets copied, if they are lazy objects they are instantiated by a `compute()` call
- Added tests for `X`, `raw`, `uns`, `obsm`, `varm`, `layers` to see if they are the expected type (dask-array or just array) and if their contents match.

## In general

So I need a general review. Except the behavior about warning on concatenation. These are completed/tested:
- `copy` method
- `to_memory` method
- read/write
- concatenation

anndata/tests/helpers.py
```diff
@@ -452,3 +491,29 @@ def fmt_name(x):
...
+# we use in AnnData this function can be helpful.
+# We wouldn't want all inputs to be called asarray
+# since we already assume they should have the same semantic.
+def darr_from_arr_dense(arr):
+    # For cases when we take scipy.sparse as input
+    # and don't care if it gets turned into a dense array.
+    # (e.g., comparing only the contents and not the type)
+    import dask.array as da
+
+    return da.from_array(asarray(arr), chunks="auto")
```
ivirshup, line 519:
```suggestion
@singledispatch
def as_dense_dask_array(a):
    import dask.array as da

    return da.asarray(a)


@as_dense_dask_array.register(sparse.spmatrix)
def _(a):
    return as_dense_dask_array(a.toarray())
```

And then a find and replace for all `darr_from_arr(_dense)?` to `as_dense_dask_array`.

anndata/tests/test_dask.py
```diff
@@ -3,12 +3,53 @@
...
+@pytest.fixture
+def adata(sizes):
+    import dask.array as da
+
+    (M, N), chunks = sizes
+    X = da.random.random((M, N), chunks=chunks)
+    obs = pd.DataFrame(
+        dict(batch=da.array(["a", "b"])[da.random.randint(0, 2, M)]),
+        index=[f"cell{i:03d}" for i in range(M)],
+    )
```
ivirshup, line 47:
```suggestion
    obs = pd.DataFrame(
        {"batch": np.random.choice(["a", "b"], M)},
        index=[f"cell{i:03d}" for i in range(M)],
    )
```

The dask array becomes an in-memory array here

anndata/tests/test_dask.py
```diff
@@ -20,3 +61,182 @@ def test_dask_X_view():
...
+    adata.varm["a"] = da.random.random((N, 10))
+
+    orig = adata
+    write(orig, pth)
+    curr = read(pth)
+
+    # orig.to_memory() Both give an error
+    # In what case should we assume it shouldn't?
+    # If we assign X should we get into backed mode?
+    # curr.to_memory() # TODO: why does this give an error
```
ivirshup, line 86: Is this resolved?

anndata/tests/test_dask.py
```diff
@@ -20,3 +61,182 @@ def test_dask_X_view():
...
+def test_assert_equal_dask_sparse_arrays():
+
+    import dask.array as da
+    from scipy import sparse
+
+    x = sparse.random(10, 10, format="csr", density=0.1)
+    y = da.from_array(asarray(x))
+
+    assert_equal(x, y)
+    assert_equal(y, x)
```
ivirshup, line 204: Can these live in `test_helpers`?

anndata/_core/anndata.py
```diff
@@ -1480,6 +1480,48 @@ def _mutated_copy(self, **kwargs):
...
+    def _to_memory_copy(self):
+        """Creates a new adata that has the computed results of the previous adata.
+        This is for the unbacked case
+        """
+        new = {}
+
+        if self.obsm is not None:
+            new["obsm"] = AxisArrays(
+                self, 0, vals={k: to_memory(v) for k, v in self.obsm.items()}
+            )
```
ivirshup, line 1492: Why are these `AxisArrays` and not just `dict`s?

ivirshup (commented):
Looking very good! A lot of small changes, but broadly close to done.

## `as_dense_dask_array`

Biggest code change I would like to see:

I want `darr_from_arr` and `darr_from_arr_dense` merged into a single function `as_dense_dask_array` (suggestion in review). After this, there's a big find and replace to do. Then I'd like to see some tests merged to one case. These include:

- [ ] `test_concatenate_layers_misaligned` & `test_concatenate_layers_misaligned_dense`
- [ ] `test_nan_merge` & test_nan_merge_dense`
- [ ] Other special cases made in `test_concatenate.py`

## `to_memory`

> -   When `to_memory` is called each object gets copied, if they are lazy objects they are instantiated by a `compute()` call

I'd like to be able to do this without copying. Can we get a `copy` argument to this?

Also I would be happy to postpone this to a separate PR if this isn't very easy.

## Documentation

It would be great to get a little documentation of features added here. I'd be fine to leave it until another PR, but please start thinking about it.

anndata/tests/test_dask.py
syelman: yes

anndata/_core/anndata.py
syelman: fixed

anndata/tests/helpers.py
```diff
@@ -322,6 +349,18 @@ def assert_equal_h5py_dataset(a, b, exact=False, elem_name=None):
...
 
+@assert_equal.register(DaskArray)
+def assert_equal_dask_array(a, b, exact=False, elem_name=None):
+
+    from dask.array.utils import assert_eq
+
+    if exact:
+        assert_eq(a, b, check_dtype=True, check_type=True, check_graph=False)
+    else:
+        # TODO: Why does it fail when check_graph=True
```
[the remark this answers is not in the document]
syelman: After reading the assert_equal function I think that options checks if the computation graphs are overlapping or not because it calls the following method inside
```python
def _check_dsk(dsk):
    """Check that graph is well named and non-overlapping"""
```
Then I realized .copy() does nothing for dask arrays. I could copy the computation graph (see https://github.com/dask/dask/issues/6674 and https://docs.dask.org/en/latest/graph_manipulation.html#api) but I don't know if I should.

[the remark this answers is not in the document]
ivirshup: I don't think we need to do that here

ivirshup (approved):

PR #814, opened 2022-09-05
Add stale bot

grst:
Add an action to label stale issues as discussed in https://github.com/scverse/governance/issues/41

There are two separate actions for "mark stale" and "close" for the following reason: 
 * According to the specifications in https://github.com/scverse/governance/issues/41, we want to mark any issue that is not `pinned` or an `enhancement` to be marked as stale
 * We only want to *close* issues if they are marked as stale and have the `needs info` label. 

afaict, having separate conditions for closing and labeling is not possible with a single stale action.

### DIFF
--- a/.github/actions/close-stale.yml
+++ b/.github/actions/close-stale.yml
@@ -9,12 +9,9 @@ jobs:
     steps:
       - uses: actions/stale@v5
         with:
-          days-before-issue-stale: 65535 # We don't want to mark issues as stale in this action
-          days-before-close: 14
+          days-before-issue-stale: -1 # We don't want to mark issues as stale in this action
+          days-before-issue-close: 14
+          days-before-pr-close: -1 # don't close PRs
           stale-issue-label: stale
           any-of-labels: "needs info"
-          stale-issue-message: |
-            This issue has been automatically marked as stale because it has not had recent activity.
-            It will be closed if no further activity occurs. Thank you for your contributions!
-          stale-pr-message: null # if not provided, will not mark PRs as stale
           debug-only: true # enable dry-run, remove when we know from the logs it's working.
--- a/.github/actions/close-stale.yml
+++ b/.github/actions/close-stale.yml
@@ -2,6 +2,7 @@ name: "Close stale issues"
 on:
   schedule:
     - cron: "0 2 * * *"
+  workflow_dispatch:
 
 jobs:
   stale:
--- a/.github/actions/close-stale.yml
+++ b/.github/actions/close-stale.yml
@@ -1,0 +1,20 @@
+name: "Close stale issues"
+on:
+  schedule:
+    - cron: "0 2 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 65535 # We don't want to mark issues as stale in this action
+          days-before-close: 14
+          stale-issue-label: stale
+          any-of-labels: "needs info"
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs. Thank you for your contributions!
+          stale-pr-message: null # if not provided, will not mark PRs as stale
+          debug-only: true # enable dry-run, remove when we know from the logs it's working.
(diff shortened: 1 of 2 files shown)


.github/actions/close-stale.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 65535 # We don't want to mark issues as stale in this action
+          days-before-close: 14
+          stale-issue-label: stale
+          any-of-labels: "needs info"
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs. Thank you for your contributions!
```
Zethson, line 18: Guess we can get rid of the state issue message? It should not mark things as stale anyways. Cleans the workflow a bit.

.github/actions/label-stale.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 60
+          days-before-close: 65535 # We don't want to close issues in this action
+          stale-issue-label: stale
+          exempt-issue-labels: pinned,bug,enhancement
```
Zethson, line 15: Does this work with pinned? Do all pinned issues get a pinned label?

.github/actions/label-stale.yml
```diff
@@ -0,0 +1,20 @@
...
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs. Thank you for your contributions!
```
Zethson, line 18: Maybe an explicit "Please comment to keep this issue open"?

.github/actions/label-stale.yml
Zethson: Also, should we add another issue label such as "not stale" or whatever to ensure that we don't have to free important long term issues that are not bugs or enhancements from the bot every 2 months? Might never need it, but you never know.

Zethson (commented):
Thank you @grst ! Couple of comments...

.github/actions/label-stale.yml
grst: actually, don't think the `pinned` label is related to "pinning" an issue. I thought it was just a conventional name for "not stale". 
E.g. nextflow: https://github.com/nextflow-io/nextflow/issues/2702, but I think I also saw the "pinned" label in other places.

.github/actions/close-stale.yml
grst: Hmm, does it still close issues then... but worth a try!

Zethson: If not, at least shorten it to the bare minimum.

ivirshup:
Are you sure you want to exclude `enhancement` and `bug`? I'm thinking something can be a bug report, but not have enough info to reproduce or figure out what's wrong.

ivirshup:
> I was just following the specification

I'd be fine with sticking with the specification for now. But I don't see it mentioning `"bug"`. Really I had more issue with that than `enhancement` 😅

grst:
oh well :sweat_smile: 
So what do we do? Get rid of it altogether, or keep "enhancement"?

ivirshup:
I'd be fine either way.

As an argument for keeping it: I think if it's obvious it's an enhancement, and not a "question" or "won't implement" maybe it's useful to keep. Maybe it's a feature request that needs fleshing out, but is broadly defined.

grst:
kept `enhancement` for now, can still make the bot more "aggressive" later if required.

.github/actions/label-stale.yml
flying-sheep: I always interpreted it as “don’t mark as stale please”, not as having anything to do with pinned issues. But maybe that could be clearer?

Zethson (approved):
LGTM. I can live with "pinned" although it seems to be a weird community standard.

grst:
Added `on workflow_dispatch`, as this might be useful for testing the dry-runs.

grst:
@ivirshup, let's merge this? 
With the dry-run activated nothing bad can happen (famous last words), and we'll likely anyway need to follow up on this after inspecting the dry run results.

ivirshup:
Sure. Are you expecting "dry run" results soon?

grst:
As soon as I hit the button? 
And oh well - this: https://github.com/scverse/anndata/pull/817

Issue #797, opened 2022-07-12
Integration with Loompy


pranzatelli:
@edg1983 mentioned [here](https://github.com/theislab/scvelo/issues/864) that anndata fails to read in the loom files generated by loompy (necessary to link loompy and scvelo). I've fixed this locally by changing some code in `io_/read.py`'s `_fmt_loom_axis_attrs()` from
```for k, v in input.items():
        if v.ndim > 1 and v.shape[1] > 1:
            axis_mapping[k] = v
        else:
            axis_df[k] = v
```
to
```for k, v in input.items():
        try:
            if v.ndim > 1 and v.shape[1] > 1:
                axis_mapping[k] = v
            else:
                axis_df[k] = v
        except:
            warn('Failed to import item '+k+' from loom file.')
```
However, I'm sure there's a more elegant (and professional) solution.

ivirshup:
Hi @pranzatelli, thanks for reporting this.

From the linked issue:

> The problem is that the loom file produced by loompy using their reference file for hg38 has a row attribute Aliases which contains None object and thus crash the code.

AFAICT, this is a bug in loompy, since [the spec says](http://linnarssonlab.org/loompy/format/index.html#row-and-column-attributes):

> There MUST be a group /row_attrs
> There can OPTIONALLY be one or more datasets at /row_attrs/{name} whose first dimension has length N

I've opened an issue on loompy about this: https://github.com/linnarsson-lab/loompy/issues/175

flying-sheep:
We deprecated loom in #2143 since it doesn’t support numpy 2

Issue #788, opened 2022-06-21
Indexing error with views


joshua-gould:
```
import anndata
import numpy as np
import pandas as pd

adata = anndata.AnnData(
    X=np.array([[-1.0, 0, 2.0, 10], [3, 5, 0, 1], [0, 0, 0, 0]]),
    var=pd.DataFrame(index=['A', 'B', 'C', 'E']),
)

weights = anndata.AnnData(X=np.array([[1., 1., 0., 0., 0.],
                                      [0., 0., 1., 1., 1.]]), var=pd.DataFrame(
    index=['B', 'A', 'D', 'C', 'X']))
common_genes = adata.var.index.intersection(weights.var.index)
# common_genes is ['A', 'B', 'C']
adata = adata[:, common_genes]
weights = weights[:, common_genes]

X_transformed = adata.X @ weights.X.T

# making a copy works fine
X_transformed_copy = X_transformed.copy()
X_transformed_copy[X_transformed_copy > 0] = X_transformed_copy[
    X_transformed_copy > 0]

# this fails with IndexError: boolean index did not match indexed array along dimension 1; dimension is 3 but corresponding boolean dimension is 2
X_transformed[X_transformed > 0] = X_transformed[X_transformed > 0]
```

ivirshup:
I think the main issue here is `X_transformed` is an `ArrayView`, when it definitely shouldn't be.

The view classes are not great, and need to be redone. I think this could be bandaged by defining `__array_ufunc__` and `__array_function__` classes for the views.

PR #818, opened 2022-09-06
Follow up PR: enable stale bot

grst:
Follow-up PR on #814, to discuss its behavior and eventually disable the dry-run mode.


grst:
Here's the output of 'label stale issue': 
https://github.com/scverse/anndata/runs/8204846798?check_suite_focus=true

Here's the output of 'close stale issues':
https://github.com/scverse/anndata/runs/8205105299?check_suite_focus=true

I think the "label" process looks already like what we'd expect. I'm not so sure about the "close" workflow. We can only tell once the first issues are labelled as stale and 14 days have passed. I would therefore suggest to only enable the "label" process first, and activate the "close" workflow later. 

P.S.
Since the workflow has now been executed once from the main branch, it is now possible to execute it from a feature branch. So no need to merge anything until we have a final solution: 
image

ivirshup:
@grst, sorry about the late response here. Yes, happy to enable stale.

That can be done in this PR? Or does it need to be merged?

grst:
This needs to be merged to enable the cron-job. 

(It would be possible to manually dispatch the workflow from this PR. But this needs to be done multiple times as not all actions are performed at once due to API limits)

ivirshup:
Alright, I'll merge in the morning tomorrow – so we're more available in case anything blows up

grst:
It will anyway blow up at night because of the cron job 🙃

Ok, let's first run a manual dispatch tomorrow, then merge

grst:
@ivirshup, this is what the action produced by running it once: 
https://github.com/scverse/anndata/issues?q=label%3Astale

(more will be labelled when it is run again, e.g. by the cronjob)

ivirshup:
* Should we mark upstream issues as stale? https://github.com/scverse/anndata/issues/797
* We shouldn't mark bugs: https://github.com/scverse/anndata/issues/788

grst:
> Should we mark upstream issues as stale? 

maybe it's a good reminder to check if the upstream issue got fixed? 

> We shouldn't mark bugs:

Actually, you asked me to do so, see https://github.com/scverse/anndata/pull/814#issuecomment-1236954096 ff. :wink:

ivirshup:
Ahh, missed the notification for this – sorry

> maybe it's a good reminder to check if the upstream issue got fixed?

Maybe? Is the logic here just we check, comment saying it's still a problem, then the bot pings again in however many weeks?

grst:
>  Is the logic here just we check, comment saying it's still a problem, then the bot pings again in however many weeks?

Yes, it would ping again after 60 days

flying-sheep:
is this ready? would be great to be able to cut down on old issues

grst:
Yes, I think so. I inspected the dry run some time ago. Maybe take another look at that one if it does what you would expect.

grst:
It's currently configured to be very lenient. It will only label issues as stale, but not close them. 
The second part that will close them is still in "dry-run" mode and would only close issues that are stale and have the label "needs-info".

PR #821, opened 2022-09-15
Lazy compute obs/var

michalk8:
Whenever `adata_view.{obs,var,uns}` is accessed, it's not computed lazily. This now allows for the below code to work:
```python
from anndata import AnnData
import numpy as np

adata = AnnData(np.random.normal(size=(1000, 30)), dtype=float)
adata_view = adata[:30]
adata.obs['foo'] = 'bar'
assert 'foo' in adata_view.obs
```

This is more of a proof-of-concept: am worried about performance penalty/user expectation when handling large `AnnDatas`.

### DIFF
--- a/anndata/_core/descriptors.py
+++ b/anndata/_core/descriptors.py
@@ -1,0 +1,91 @@
+from typing import TYPE_CHECKING, MutableMapping, Optional
+from collections import OrderedDict
+from copy import copy
+
+from anndata.compat import _overloaded_uns, OverloadedDict, _slice_uns_sparse_matrices
+from typing_extensions import Literal
+import pandas as pd
+from anndata._core.views import DataFrameView, DictView
+
+if TYPE_CHECKING:
+    from anndata._core.anndata import AnnData
+
+
+class DataFrameDescriptor:
+    """\
+    One-dimensional annotation of observations or features.
+
+    Parameters
+    ----------
+    attr:
+        Named of the attribute.
+    """
+
+    def __init__(self, attr: Literal["obs", "var"]):
+        self._attr = attr
+
+    def __get__(self, adata: "AnnData", objtype: Optional[type] = None) -> pd.DataFrame:
+        print(objtype)
+        if not adata.is_view:
+            return adata._obs if self.attr == "obs" else adata._var
+
+        adata_ref = adata._adata_ref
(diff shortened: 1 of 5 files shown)


anndata/_core/descriptors.py
```diff
@@ -0,0 +1,90 @@
...
+        if self.attr == "obs":
+            container_full = adata_ref.obs
+            container = container_full.iloc[adata._oidx]
+        elif self.attr == "var":
+            container_full = adata_ref.var
+            container = adata_ref.var.iloc[adata._vidx]
+        else:
+            raise NotImplementedError(self.attr)
+
+        # we don't care about uns at this point
```
ivirshup, line 41: Why don't we care about `uns` here?

anndata/_core/descriptors.py
```diff
@@ -0,0 +1,90 @@
...
+
+        adata_ref = adata._adata_ref
+        # special case for old neighbors, backwards compat. Remove in anndata 0.8.
+        # fix categories
+        uns = _slice_uns_sparse_matrices(
+            copy(adata_ref._uns), adata._oidx, adata_ref.n_obs
+        )
+
+        adata._remove_unused_categories(adata_ref.obs, adata.obs, uns, update_sub=False)
+        adata._remove_unused_categories(adata_ref.var, adata.var, uns, update_sub=False)
```
ivirshup, line 73: Doesn't this mean every time you `uns` on a view, a full copy of `obs` and `var` are made and full passes over their categorical values are done?

ivirshup (commented):
Right now I'm inclined to agree that the performance penalties for this would be quite large. I think a better path forward would involve removing all the special casing for categoricals (e.g. removing unused ones). I think they have improved significantly since we first used them. However, this will break downstream packages so will take some planning and mitigation.

Ideally we would also be able to compute the columns of the dataframe or the entries of uns lazily rather than have to copy the whole thing. Probably do-able for uns, not so much for `obs` and `var`.

anndata/_core/descriptors.py
ivirshup: Ah, because you're being lazy in copying. I think in practice, many times that someone is accessing a categorical on a view they also want to get it's colors from uns because they are plotting.

PR #822, opened 2022-09-19
create much smaller files by writing fixed-length strings to HDF5 in some circumstances

ilia-kats:
Apparently, variable-length strings incur quite some overhead in terms of file size, so write arrays of fixed-length strings when the length differences between the elements of the array are small.

For example, [the 10x Genomics CITE-seq data set](https://support.10xgenomics.com/single-cell-gene-expression/datasets/3.0.2/5k_pbmc_protein_v3) is 146 MB (for the unfiltered matrix). Reading this into an AnnData object and writing it out results in a 300 MB h5ad file (gzip-compressed). With this patch, the h5ad file is only 50 MB.

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -340,9 +340,21 @@ def read_array_partial(d, items=None, indices=slice(None)):
 @_REGISTRY.register_write(H5Group, (np.ndarray, "U"), IOSpec("string-array", "0.2.0"))
 @_REGISTRY.register_write(H5Group, (np.ndarray, "O"), IOSpec("string-array", "0.2.0"))
 def write_vlen_string_array(f, k, elem, dataset_kwargs=MappingProxyType({})):
-    """Write methods which underlying library handles nativley."""
-    str_dtype = h5py.special_dtype(vlen=str)
-    f.create_dataset(k, data=elem.astype(str_dtype), dtype=str_dtype, **dataset_kwargs)
+    """Write methods which underlying library handles natively. Variable-length strings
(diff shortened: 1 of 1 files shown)


ivirshup:
@ilia-kats, do you know why the variable length strings are taking up more space? E.g. are you sure it's a fundamental problem with the type or could this be a bug?

ilia-kats:
According to [the HDF5 docs](https://docs.hdfgroup.org/archive/support/HDF5/doc/TechNotes/VLTypes.html) it's a fundamental HDF5 property.

ivirshup:
But those docs say it should be 32 bytes per element, but you're saying this would save 250 Mb?

I don't think we can switch storage to a fixed length system, since that would break out-of-core concatenation.

ilia-kats:
Yes, it's 32 bytes per string, so for a large array such as `obs_names` it adds up.

ivirshup:
Found some more details in this RFC: https://docs.hdfgroup.org/hdf5/rfc/var_len_data_sketch_design_190715.pdf

Basically, that describes the problem and proposes a solution. However, I can't find any discussion of the RFC. 

The page you've linked doesn't seem to be included [in recent documentation](https://docs.hdfgroup.org/hdf5/develop/_t_n.html). I'm not sure what to make of that.

ivirshup:
@ilia-kats, I don't think we can just change the binary format of strings. It would have to be a larger decision. From the hdf5.jl package docs, it's seems like it wouldn't even be able to read these fixed length unicode strings.

I've opened an issue for further discussion.

PR #823, opened 2022-09-28
Dask dataframe support 

rahulbshrestha:
This PR introduces support for Dask dataframes in anndata. 

TODOs:
- [ ] Indexing
- [ ] Writing / Reading
- [ ] Concatenation
- [x] assert_equal for tests
- [ ] adata.to_memory() / adata.copy()

Related PR (Dask array support): https://github.com/scverse/anndata/pull/813
Contributors: @rahulbshrestha @syelman

### DIFF
--- a/anndata/compat/__init__.py
+++ b/anndata/compat/__init__.py
@@ -58,7 +58,16 @@ except ImportError:
     class DaskArray:
         @staticmethod
         def __repr__():
-            return "mock dask.array.core.Array"
+            return "mock dask.array.core.Array" 
+
+try:
+    from dask.dataframe import DataFrame as DaskDataFrame 
+except ImportError:
+
+    class DaskDataFrame:
+        @staticmethod
+        def __repr__():
+            return "mock dask.dataframe.core.DataFrame"
 
 
 @singledispatch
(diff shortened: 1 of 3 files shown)


ivirshup:
So, I've looked into the length thing a bit. It looks like there is still no way to include info on number of rows for a dask dataframe. This is tracked multiple places in the dask repo, but this issue looks most recent: https://github.com/dask/dask/issues/5633

It's possible we can do something clever to work around this, like persisting the index of the data frame and doing length checks there. We could also not do length checks on dask dataframes until we try to compute, and error then.

@ryan-williams, any chance you have thoughts here? Is it best to just wait on dask some more?

ivirshup:
Here is [a gist](https://gist.github.com/ivirshup/884df705592186a55b56c2c863e1f884) with some code for reading a dataframe saved in AnnData to a dask DataFrame

ilan-gold:
@ivirshup I've got a branch with your gist - I can start an issue for this but so far what I see is that:
1. calling `len(df)` when `df` is a dask dataframe loads the whole dataframe into memory
2. the index has no `is_unique` attribute
Both seems manageable as PR's into dask (if they're actually issues) but just figured I'd document this somewhere.

PR #827, opened 2022-10-09
Fix zarr partial reads for sparse and dense arrays

Koncopd:
For sparse array `read_partial` wasn't registered. For dense arrays it actually didn't work.


ivirshup:
Could you add some tests for this?

PR #828, opened 2022-10-10
Add experimental support for `zstd` compression via hdf5plugin library

nigeil:
These changes implement experimental support for the `zstd` compression filter as an alternative to `gzip` and `lzf` compressors in `anndata.write_h5ad()`. The `zstd` compressor is available via the `hdf5plugin` library, which is pulled in as a new dependency.

Testing on a real scRNA-seq anndata object with n_obs × n_vars = 38860 × 22205 and 4 sparse integer layers, the `zstd` compressor provides read/write speeds comparable to `lzf` while providing a compressed data size comparable to `gzip`. See benchmark results below:

zstd write time = 18.27s
zstd read time = 5.28s
zstd write size = 1261.0MB

lzf write time = 10.33s
lzf read time = 7.04s
lzf write size = 2588.0MB

gzip write time = 72.63s
gzip read time = 12.43s
gzip write size = 1275.0MB

None write time = 1.74s
None read time = 1.71s
None write size = 3723.0MB

A side-effect of using compression filters from the `hdf5plugin` library is that compression encoding does not appear to be written/readable in the resulting h5ad file. Despite this, compression still works as expected. I have wrapped the `test_hdf5_compression_opts()` function to detect when this is happening in the `zstd` testing edge case and ignore it, but further investigation will be required to understand why resulting h5ad files do not store this compression information as a dataset attribute and what we can do to fix that. It may be related to the fact that the `zstd` compressor is passed to the `h5py` `create_dataset()` function as an `int` filter identifier, not by a `str` filter name.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -10,7 +10,7 @@ from functools import partial, singledispatch
 from pathlib import Path
 from os import PathLike
 from textwrap import dedent
-from typing import Any, Union, Optional  # Meta
+from typing import Any, Union, Optional, Literal  # Meta
 from typing import Iterable, Sequence, Mapping, MutableMapping  # Generic ABCs
 from typing import Tuple, List  # Generic
 
@@ -51,7 +51,6 @@ from ..compat import (
     ZarrArray,
     ZappyArray,
     DaskArray,
-    Literal,
     _slice_uns_sparse_matrices,
     _move_adj_mtx,
     _overloaded_uns,
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1902,7 +1902,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             For [`zstd`], see the hdf5plugin :ref:`usage.html <usage>`. Experimental.
         compression_opts
             For [`lzf`, `gzip`], see the h5py :ref:`dataset_compression`.
-            For [`zstd`], see the hdf5plugin :ref:`usage.html <usage>`. Experimental.
+            For [`zstd`], see the hdf5plugin :ref:`hdf5plugin:usage`. Experimental.
             Construct `compression_opts` using the helper functions provided
             by the hdf5plugin package - for instance:
             `hdf5plugin.Zstd(clevel=5).filter_options`
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1931,14 +1931,14 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             Alternative compression filters such as `zstd` can be used by
             passing a compressor from the hdf5plugin library; see
             :ref:hdf5plugin:`usage` for more information. Experimental.
-            Usage example:
-            ```
-            import hdf5plugin
-            adata.write_h5ad(
-                filename,
-                compression=hdf5plugin.FILTERS["zstd"]
-            )
-            ```
+            Usage example::
+
+                import hdf5plugin
+                adata.write_h5ad(
+                    filename,
+                    compression=hdf5plugin.FILTERS["zstd"]
+                )
+            
             .. note::
                 Datasets written with hdf5plugin-provided compressors
                 cannot be opened without first loading the hdf5plugin
@@ -1946,21 +1946,22 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 compression filters such as `zstd`, consider writing to
                 `zarr` format instead of `h5ad`, as the `zarr` library
                 provides a more transparent compression pipeline. 
+
         compression_opts
             For [`lzf`, `gzip`], see the h5py :ref:`dataset_compression`.
             For alternative compressor filters provided by hdf5plugin,
             `compression_opts` can be constructed using the instantiated
              compressor's `filter_options` property; see :ref:hdf5plugin:`usage`
             for more information. Experimental.
-            Usage example (setting `zstd` compression level to 5):
-            ```
-            import hdf5plugin
-            adata.write_h5ad(
-                filename, 
-                compression=hdf5plugin.FILTERS["zstd"],
-                compression_opts=hdf5plugin.Zstd(clevel=5).filter_options
-            )
-            ```
+            Usage example (setting `zstd` compression level to 5)::
+            
+                import hdf5plugin
+                adata.write_h5ad(
+                    filename, 
+                    compression=hdf5plugin.FILTERS["zstd"],
+                    compression_opts=hdf5plugin.Zstd(clevel=5).filter_options
+                )
+            
         as_dense
             Sparse arrays in AnnData object to write as dense. Currently only
             supports `X` and `raw/X`.
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1899,10 +1899,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
             Filename of data file. Defaults to backing file.
         compression
             For [`lzf`, `gzip`], see the h5py :ref:`dataset_compression`.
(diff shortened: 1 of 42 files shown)


ivirshup (commented):
Thanks for opening a PR @nigeil!

I played around with hdf5plugin a bit, and I think this possible without making any changes to the code in `anndata`. This works for me:

```python
import hdf5plugin

adata.write_h5ad("adata.h5ad",  **hdf5plugin.LZ4())
```

If I look at the written file with `h5ls -r -v "adata.h5ad"` it looks like the compression is being used.

It would be good to know if there were cases where this didn't work, and this could get documented.

----------

I think the reason you aren't seeing a compression value on the dataset is because `h5py` [hardcodes the possible values](https://github.com/h5py/h5py/blob/1487a54fb5149603dcc32604df4db418ea4f5236/h5py/_hl/dataset.py#L556-L563). It's probably worth creating an issue over there about this.

anndata/_core/anndata.py
```diff
@@ -1897,9 +1898,14 @@ def write_h5ad(
         filename
             Filename of data file. Defaults to backing file.
         compression
-            See the h5py :ref:`dataset_compression`.
+            For [`lzf`, `gzip`], see the h5py :ref:`dataset_compression`.
+            For [`zstd`], see the hdf5plugin :ref:`usage.html <usage>`. Experimental.
```
flying-sheep, line 1902: Can simply be this I think

```suggestion
            For [`zstd`], see the hdf5plugin :ref:`hdf5plugin:usage`. Experimental.
```

anndata/_core/anndata.py
```diff
@@ -1897,9 +1898,14 @@ def write_h5ad(
...
         compression_opts
-            See the h5py :ref:`dataset_compression`.
+            For [`lzf`, `gzip`], see the h5py :ref:`dataset_compression`.
+            For [`zstd`], see the hdf5plugin :ref:`usage.html <usage>`. Experimental.
```
flying-sheep, line 1905: same

```suggestion
            For [`zstd`], see the hdf5plugin :ref:`hdf5plugin:usage`. Experimental.
```

flying-sheep (changes requested):

nigeil:
@ivirshup thanks for taking a look. I agree, as far as I can tell these compressors can be used without any changes to the anndata library (that is how we've been using it so far). 

However, my understanding of the hdf5plugin library is that reading from hdf5 files compressed using one of their compressors requires that the hdf5plugin library be imported - during the import, compressors are registered with the h5py library. 

I can confirm that reading a `zstd` compressed dataset fails with the following error if I do not first `import hdf5plugin`:
`OSError: Can't read data (can't open directory: /usr/local/hdf5/lib/plugin)`

One use case where I see these changes being useful is when anndata is used in applications like cellxgene, where users can upload their own h5ad files for analysis. For instance, if a containerized cellxgene instance doesn't `import hdf5plugin` somewhere, reading data compressed with that plugin may fail, and it will fall on those software maintainers to make the necessary changes to their code bases.

Perhaps we can consider the compromise of just including the hdf5plugin library as a dependency (and importing it in the anndata class), which will require minimal changes. 

I've updated the docs links with the suggestions from @flying-sheep (looks like they still fail - sorry!), but if you decide to go for a docs-only strategy let me know.

ivirshup:
*tldr*, I would prefer a "just docs" solution.

> One use case where I see these changes being useful is when anndata is used in applications like cellxgene, where users can upload their own h5ad files for analysis. For instance, if a containerized cellxgene instance doesn't import hdf5plugin somewhere, reading data compressed with that plugin may fail, and it will fall on those software maintainers to make the necessary changes to their code bases.

This use case is fair, but I would also like to avoid making the change here. To me, there are two better options for handling this, though they both require upstream work.

## h5py should use entry-points to import hdf5plugin on its own

`h5py` could import `hdf5plugin` on its own. It would need to claim an [entry point](https://docs.python.org/3/library/importlib.metadata.html#entry-points), and `hdf5plugin` would have to register that entry point.

This is a better solution since it solves the problem for everyone. No one needs to be the one considering if they should import `hdf5plugin` before `h5py`.

## `cellxgene` should be able to load `zarr` anndata files

Zarr defaults to more modern compression libraries, has a much simpler way of recognizing them, ships with them, and is overall just like a more modern hdf5.

AFAICT, very little code would need to change in cellxgene to add support for zarr backed anndata's.

At it's core, [this would need to call `read_zarr` if a zarr store was passed](https://github.com/chanzuckerberg/cellxgene/blob/d2b20129f7a3acf89a34128e1594aa7c02adaaa9/server/data_anndata/anndata_adaptor.py).

# Making a change here

The "lightest touch" thing we could do here is try to check if `hdf5plugin` is available at `anndata` import time, then import it. This would mean we're not adding it as a dependency, but if it's available we'd use it.

But this makes any problems that come from that our problem. And it's a dynamic C library plugin problem, not something I know how to deal with. This old discussion from h5py brings up a variety of things that can go wrong plugins, and there are many issues open on h5py about plugins.

Basically: If this led to any bugs, and it seems like it might, we wouldn't be able to do much about it. I would rather someone who is willing to take on dealing with C libraries take responsibility for this.

flying-sheep:
> h5py should use entry-points to import hdf5plugin on its own

definitely, did you file the issue?

ivirshup:
I sorta mentioned it here: https://github.com/h5py/h5py/issues/2161, and that specific solution has been mentioned before (https://github.com/h5py/h5py/issues/928#issuecomment-353715290) but they ended up going in a different direction. Not sure if there was a conflict with making an entry point.

I would feel a little bad about opening three issues in a row without any response 😅.

flying-sheep:
Sure, can do! https://github.com/h5py/h5py/issues/2163

ivirshup:
@nigeil any thoughts on what we should do with this PR?

I would be up for throwing a better error (e.g. suggesting importing hdf5plugin), but I don't want to add a dependency. Docs about this would be good too.

nigeil:
@ivirshup it seems there are a lot of decisions needed upstream to make this integration more seamless, and with that added information I agree that the anndata library should not be responsible for maintaining that interface. I would support adding documentation for the usage pattern so that users know how to pass through `hdf5plugin` compressors if they want. I would also support adding a warning should `ad.read_h5ad` fail, so that users/downstream projects who are given an `hdf5plugin` compressed h5ad file to read know that they may need to import this extra library. I can make a new PR with those changes if you'd like the help, and we can probably close this PR.

ivirshup:
@nigeil, a new PR with docs on usage and a hint for hdf5plugin would be great!

Do you know how you could tell if a compressor would have been available through hdf5plugin? Is the current error specific enough?

nigeil:
@ivirshup I don't believe the currently thrown `OSError` will be informative enough to make the call with 100% accuracy. It currently reads as follows `OSError: Can't read data (can't open directory: /usr/local/hdf5/lib/plugin)`, and that path is probably different across different platforms. Additionally, since `hdf5plugin` compressed datasets don't have their compression type stored in the `h5ad` file properly, we can't intuit that an `hdf5plugin`-compatible compressor was used by reading the file. 

We could catch the `OSError`, check to see if the plugin is referenced at all in the error text, and provide a hint to the user if that is the case. Alternatively, we can leave it as-is (don't catch anything) and just mention this behavior in the documentation. Since these plugins are probably only going to be used by power-users in any case, this may be acceptable. 

We can also add documentation that encourages users wanting to take advantage of these compressors to consider saving to `zarr` as it provides a more accessible compression pipeline, as you previously pointed out.

ivirshup:
@nigeil that all sounds good to me.

I'd be happy with either the doc solution or detection solution. The detection solution would probably take longer to figure out, and could probably even be done separately.

nigeil:
Closing this PR - see https://github.com/scverse/anndata/pull/857 for new docs-only PR.

PR #834, opened 2022-10-13
(feat): `read_dispatched` and `write_dispatched`

ilan-gold:
For #659 - no write method, but this seems like a good standalone feature.  I also figured there was no reason not to "battle test" it by using it for the already existing read methods to which it could apply.

### DIFF
--- a/anndata/_io/h5ad.py
+++ b/anndata/_io/h5ad.py
@@ -3,7 +3,7 @@ from functools import partial
 from warnings import warn
 from pathlib import Path
 from types import MappingProxyType
-from typing import Callable, Type, TypeVar, Union
+from typing import Callable, Type, TypeVar, Union, Literal
 from typing import Collection, Sequence, Mapping
 
 import h5py
@@ -18,7 +18,6 @@ from ..compat import (
     _from_fixed_length_strings,
     _decode_structured_array,
     _clean_uns,
-    Literal,
 )
 from .utils import (
     H5PY_V3,
--- a/anndata/_io/h5ad.py
+++ b/anndata/_io/h5ad.py
@@ -11,6 +11,8 @@ import numpy as np
 import pandas as pd
 from scipy import sparse
 
+from anndata._io.read import read_dispatched
+
 from .._core.sparse_dataset import SparseDataset
 from .._core.file_backing import AnnDataFileManager
 from .._core.anndata import AnnData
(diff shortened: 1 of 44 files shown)


ivirshup (commented):
AFAICT, this is only going to let me override top level elements of the anndata. What I was  thinking was more of a callback point right in front of every `read_elem` call.

I would like to be able to use this function to do things like:

* Read any 2d dense array as a dask array

```python
import dask.array as da
def callback(...):
    ...
    if iospec.encoding_type == "array":
        return da.from_zarr(elem)
    ...
```

* Read any sparse array as a backed sparse array (similar to above example)
* Read only specific columns of a data frame

```python
_keys = ["cell_type", "cell_label"]
def callback(...):
    ...
    if k == "obs":
        return pd.DataFrame(
            {df_col: read_elem(elem[df_col]) for df_col in _keys},
            index=read_elem(elem[elem.attrs["index"]])
        )
    ...
``` 

I think to do this `read_elem` is going to need to be able to reference some state (the callback).

At the moment, I'm thinking this could be done by making `read_elem` an instance method of the `IORegistry` (or some reader instance class), and for that instance to have a callback attribute.

What do you think?

ilan-gold:
@ivirshup I see what you're saying.  Could we export the registry and then allow people to just register their own read functions?  I think the reason I'm having troubling seeing how to implement something so low-level is that there's already a lot of good code that you have written for handling low-level stuff and the `read` methods that exist don't go that low-level by using the existing code.  Also, the current code does take in the IOSpec, so could we match on that based on what is here i.e `get_spec` in `v = dispatch_element(get_reader(group[k]), group, k, get_spec(group[k]))`?

ivirshup:
> Could we export the registry and then allow people to just register their own read functions? 

That's the idea eventually. I would like to keep this stuff private while experimenting with them. Basically, people could use it, but I don't want to promise stability here yet.

### Some things about the future of registries:

In the near future we will likely need the ability to write older versions of the anndata spec. It would be nice to handle this separately. In addition, one could just completley replace the registry, or remove parts of it for a more restrictive spec.

> there's already a lot of good code that you have written for handling low-level stuff and the read methods that exist don't go that low-level by using the existing code.

I'm not totally sure which `read` methods you are referring to here.

> Also, the current code does take in the IOSpec, so could we match on that based on what is here i.e get_spec in v = dispatch_element(get_reader(group[k]), group, k, get_spec(group[k]))?

As opposed to `k` in the example, or overall?

In the example, probably.

Overall, I would like to be able to choose behavior on type and location.

ilan-gold:
> I'm not totally sure which `read` methods you are referring to here.

`read_zarr` and `read_elem` rely on `read_elem` and other dispatched/registered functions that already do what you want here.  In this sense, if people want to write their own reader for, say, all scalar arrays backed by zarr, they can handle the  logic translating that from a key and value in the group if they choose, although we also pass in the `spec` to the callback.

> As opposed to `k` in the example, or overall?

Exactly - it sounds like this already allows for what you are asking.  We have both cases here - reading based purely on key (and this way `read_zarr` and `read_h5ad` can use it)  and also based on the spec i.e reading all `encoding_type`s the same way.  I guess this leaves the `dict` case i.e for `obsm` that needs be handled on its own, but this seems like a small price to pay for users to implement themselves.  We could handle `dict` ourselves in `read_dispatched`.

>  I think to do this `read_elem` is going to need to be able to reference some state (the callback).

Would we export `IORegistry` as an object then?  I'm not clear how we could intercept `read_elem` calls  without exporting this for people to register their  own readers (instead of exporting `_REGISTRY`, and if we did intercept `read_elem` calls, why do we need `read_dispatched`?  Why not just override `IORegistry` and then allow people to use `read_zarr`?   Sorry for not understanding this, but I'm having trouble seeing where

anndata/_io/h5ad.py
```diff
@@ -71,37 +73,36 @@ def write_h5ad(
...
-            write_elem(f, "raw", adata.raw, dataset_kwargs=dataset_kwargs)
-        write_elem(f, "obs", adata.obs, dataset_kwargs=dataset_kwargs)
-        write_elem(f, "var", adata.var, dataset_kwargs=dataset_kwargs)
-        write_elem(f, "obsm", dict(adata.obsm), dataset_kwargs=dataset_kwargs)
-        write_elem(f, "varm", dict(adata.varm), dataset_kwargs=dataset_kwargs)
-        write_elem(f, "obsp", dict(adata.obsp), dataset_kwargs=dataset_kwargs)
-        write_elem(f, "varp", dict(adata.varp), dataset_kwargs=dataset_kwargs)
-        write_elem(f, "layers", dict(adata.layers), dataset_kwargs=dataset_kwargs)
-        write_elem(f, "uns", dict(adata.uns), dataset_kwargs=dataset_kwargs)
+        write_dispatched(f, adata, dispatch_element)
```
ilan-gold, line 105: I am not sure what's happening here, but if I add a `adata.file.close()` after this then it works (outside the `with` block).

ivirshup (commented):
I'm not totally sure I get what the various arguments to the read_dispatched and write_dispatched functions are meant to be used for. Could you explain these?

I also wanted to share #851 with you. I'm not happy with it, but I thought it would help illustrate the direction I was thinking of.

The main differences are:

* It's recursive, as opposed to top level. Dispatch happens on each element. E.g. if I make a rule for how to read dataframes, that can apply to all dataframes regardless of location
* It only asks the user for one callback function

In particular, could you take a look at the [test cases](https://github.com/scverse/anndata/pull/851/files#diff-9d490f09ad3a86b27f14805e04651b44bc296c177430c088451535adada460d5)? I think these illustrate the kinds of things I want to be possible with this function.

ilan-gold:
I will add the test cases to be sure we can do this here, and then respond.

PR #838, opened 2022-10-19
Re-add schema to README

flying-sheep:
As mentioned in https://github.com/github/markup/issues/163#issuecomment-1282676832, this works now!

The docs index page doesn’t change after this, but it includes everything from the readme instead of containg a copy of its text: https://icb-anndata--838.com.readthedocs.build/en/838/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,6 +17,9 @@
 .. |Coverage| image:: https://codecov.io/gh/scverse/anndata/branch/master/graph/badge.svg?token=IN1mJN1Wi8
    :target: https://codecov.io/gh/scverse/anndata
 
+.. image:: docs/_static/img/anndata_schema.svg
+   :align: right
+   :width: 350px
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -17,15 +17,21 @@
...
+   :width: 350px
+
+.. after image
 
 anndata - Annotated data
 ========================
 
 anndata is a Python package for handling annotated data matrices in memory and on disk, positioned between pandas and xarray. anndata offers a broad range of computationally efficient features including, among others, sparse data support, lazy operations, and a PyTorch interface.
 
+* Discuss development on `GitHub <https://github.com/scverse/anndata>`_.
```
flying-sheep, line 31: The only “issue” with this PR is that this list item is quite redundant in the README.

However, only adding it to the docs index and `include`ing the rest of the bullet points would create two bullet lists.

PR #839, opened 2022-10-19
Adapt schema graphic for dark background

flying-sheep:
It has two elements that don’t look too good on a dark background:

- [x] The `uns` text, ellipsis, and braces are plain black. I’ll solve this with a stroke. Since [`stroke-alignment`](https://svgwg.org/specs/strokes/#SpecifyingStrokeAlignment) doesn’t seem to have wide support, I’ll emulate it with `<use>` (which does have wide support)
- [x] the `*_names` texts have white backdrop rectangles to make them readable over their lines. I’ll solve this by using the rectangles as clip paths instead.


Zethson:
Think that Github now supports displaying different images based on the detected theme. We could show the new dark one if the dark mode is detected and the current one elsewise.

flying-sheep:
[indeed](https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/). But I think the image could be improved in any case (so it can e.g. be displayed on non-white backgrounds)

PR #847, opened 2022-11-08
Handle non-csr/csc sparse matrices in anndata.write

scottgigante-immunai:
Attempting to save an AnnData object with non-CSR/CSC sparse matrices throws an error:

```
File "/usr/src/singlecellopenproblems/openproblems/api/utils.py", line 58, in write_h5ad
      adata.write_h5ad(filename)
    File "/usr/local/lib/python3.8/site-packages/anndata/_core/anndata.py", line 1918, in write_h5ad
      _write_h5ad(
    File "/usr/local/lib/python3.8/site-packages/anndata/_io/h5ad.py", line 85, in write_h5ad
      write_elem(f, "X", adata.X, dataset_kwargs=dataset_kwargs)
    File "/usr/local/lib/python3.8/site-packages/anndata/_io/utils.py", line 220, in func_wrapper
      raise type(e)(
  TypeError: No method has been defined for writing <class 'scipy.sparse._coo.coo_matrix'> elements to <class 'h5py._hl.group.Group'>
  
  Above error raised while writing key 'X' of <class 'h5py._hl.group.Group'> to /
```

This PR fixes that by registering an appropriate method which converts all sparse matrices without an explicitly defined conversion to CSR.

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -469,6 +469,21 @@ _REGISTRY.register_write(ZarrGroup, views.SparseCSCView, IOSpec("csc_matrix", "0
 )
 
 
+@_REGISTRY.register_write(H5Group, sparse.coo_matrix, IOSpec("csr_matrix", "0.1.0"))
+@_REGISTRY.register_write(H5Group, sparse.lil_matrix, IOSpec("csr_matrix", "0.1.0"))
+@_REGISTRY.register_write(H5Group, sparse.dia_matrix, IOSpec("csr_matrix", "0.1.0"))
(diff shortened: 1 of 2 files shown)


ivirshup:
Hey. I think I'd be more inclined to say "that anndata is invalid and should have errored when you created it". I think many anndata operations would error with any of the sparse matrix types that don't do indexing right.

I would rather add broad support for them as they gain functionality, rather than later changing conversion behavior.

An alternative would be to do the conversion when the AnnData is constructed, but I think it's largely fine to leave this to the user.

scottgigante-immunai:
I'd be fine with either throwing an error or automatically converting inside `AnnData.X.setter` -- just doing it in construction may not work as it's not that uncommon to set `adata.X = func(adata.X)`.

PR #864, opened 2022-11-29
Centralize pytest options

flying-sheep:
Addding `-v` to pytest shows each test result on a separate line, which plays better with a remote test runner than the default `.` display:

grafik


ivirshup (commented):
> which plays better with a remote test runner than the default . display:

What do you mean by this? Like, there are formatting errors?

----------

I think I like the non-verbose output better. I prefer the high level overview at a glance, which I can then drill down into with the azure test viewer or locally.

flying-sheep:
The view doesn’t show unbuffered output, it will only emit a new line once it’s complete.

Therefore to see when a test hangs, we need to see the verbose output.

flying-sheep:
Hi Isaac, the tests breaking on one of the backport PRs seem to need yet another backport PR: https://github.com/scverse/anndata/pull/1021 and https://github.com/scverse/anndata/pull/1023

With the verbosity changes here, I’d have seen the failing tests in context (i.e. that all parameter combinations fail)

I therefore want to merge this PR. It’s too helpful to see all test names + parametrizations in sequence.

flying-sheep:
@ivirshup ping

ivirshup:
Can you take a look at the test report we're currently get when you click on this?
 
image

You should get a view like this (you may have to unfilter the tests that pass):

image

I think this is basically why we are using the junitxml. Does this cover your use case?

ivirshup:
I believe this used to be publicly available, which is why we started using it, but it's unclear if it is still. Which would be a shame, because it's a nice interface.

flying-sheep:
I can see it, but only after all tests have run. When a test gets stuck, only stdout/stderr get written reliably (if you have to cancel the tests, the xml might not get written, right?)

`-v` also makes the diff output for failed tests visible, saving time (no need to rerun when you can see the issue in the diff).

ivirshup:
> When a test gets stuck, only stdout/stderr get written reliably (if you have to cancel the tests, the xml might not get written, right?)

Yes, but tests can appear to hang on CI for a wide variety of reasons. I agree this can be useful locally.

-----

I agree that these options are useful for debugging a PR, but I think a summarized version is more helpful for reviewing. Getting a sense of which files are seeing errors at a glance is nice. I don't think adding a thousand lines of in every case is a good change. One can always drill down further locally or in the viewer if needed.

flying-sheep:
OK, I’ll revisit if this ends up annoying me

ivirshup (approved):

PR #867, opened 2022-12-05
Fix sklearn package name

Zethson:
```
The 'sklearn' PyPI package is deprecated, use 'scikit-learn'
      rather than 'sklearn' for pip commands.
      
      Here is how to fix this error in the main use cases:
      - use 'pip install scikit-learn' rather than 'pip install sklearn'
      - replace 'sklearn' by 'scikit-learn' in your pip requirements files
        (requirements.txt, setup.py, setup.cfg, Pipfile, etc ...)
```

Signed-off-by: zethson <lukas.heumos@posteo.net>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ test = [
     "pytest-cov>=2.10",
     "zarr",
     "matplotlib",
-    "sklearn",
+    "scikit-learn",
     "openpyxl",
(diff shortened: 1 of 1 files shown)


Zethson:
Don't think that this needs a release note or anything else.

ivirshup:
Thanks for the fix!

Issue #861, opened 2022-11-28
Views on dask arrays copy the array into memory


ivirshup:
Ping @syelman 

Views on dask arrays are currently broken, since it's actually converting the dask array to a numpy array since DaskArrayView is using the code for views of numpy arrays. 

<details>
<summary> Example </summary>

```python
import anndata as ad, dask.array as da, numpy as np, pandas as pd

def adata_with_dask(M, N):
    adata_dict = {}
    adata_dict["X"] = da.random.random((M, N))
    adata_dict["dtype"] = np.float64
    adata_dict["obsm"] = dict(
        a=da.random.random((M, 100)),
    )
    adata_dict["layers"] = dict(
        a=da.random.random((M, N)),
    )
    adata_dict["obs"] = pd.DataFrame(
        {"batch": np.random.choice(["a", "b"], M)},
        index=[f"cell{i:03d}" for i in range(M)],
    )
    adata_dict["var"] = pd.DataFrame(index=[f"gene{i:03d}" for i in range(N)])
    
    
    return ad.AnnData(**adata_dict)


adata = adata_with_dask(10000,10000)
adata
```

```python
%%prun -s cumtime
v = adata[:, :5000]
v.layers["a"]
```

```
         5453 function calls (4977 primitive calls) in 0.309 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.309    0.309 {built-in method builtins.exec}
        1    0.021    0.021    0.309    0.309 <string>:1(<module>)
        1    0.000    0.000    0.288    0.288 aligned_mapping.py:111(__getitem__)
        2    0.000    0.000    0.288    0.144 functools.py:883(wrapper)
        1    0.000    0.000    0.286    0.286 views.py:139(as_view_dask_array)
        1    0.000    0.000    0.286    0.286 views.py:66(__new__)
        1    0.000    0.000    0.286    0.286 {built-in method numpy.asanyarray}
        1    0.000    0.000    0.286    0.286 core.py:1701(__array__)
        1    0.018    0.018    0.286    0.286 base.py:291(compute)
        1    0.000    0.000    0.268    0.268 base.py:539(compute)
        1    0.000    0.000    0.187    0.187 base.py:601(<listcomp>)
        1    0.000    0.000    0.187    0.187 core.py:1278(finalize)
        1    0.186    0.186    0.187    0.187 core.py:5232(concatenate3)
[...]

PR #869, opened 2022-12-06
Views on dask arrays copy the array into memory

syelman:
Hi @ivirshup ,

I started by doing  changing the view class. It seems to create a lazy object for the view but doesn't set the value

```python

import dask
import dask.array as da
import numpy as np
import anndata as ad
import pandas as pd

def adata_with_dask(M, N):
    adata_dict = {}
    adata_dict["X"] = da.random.random((M, N))
    adata_dict["dtype"] = np.float64
    adata_dict["obsm"] = dict(
        a=da.random.random((M, 100)),
    )
    adata_dict["layers"] = dict(
        a=da.random.random((M, N)),
    )
    adata_dict["obs"] = pd.DataFrame(
        {"batch": np.random.choice(["a", "b"], M)},
        index=[f"cell{i:03d}" for i in range(M)],
    )
    adata_dict["var"] = pd.DataFrame(index=[f"gene{i:03d}" for i in range(N)])
    
    
    return ad.AnnData(**adata_dict)


adata = adata_with_dask(8192,8192)
adata_slice_view = adata[:500, :][:, :500]

adata_slice_view.is_view # True

adata_slice_view.X[1,1] = 99 # just calls the getter and on the comp graph it is just a get operation

# while on classic dask 
a = da.random.random((100, 100))
a[1,1] = 1 # works

```

Related issue: https://github.com/scverse/anndata/issues/861


syelman:
Hi @ivirshup ,
The problem was, I wasn't able to cast the dask array to `DaskArrayView` since it just absorbed it lazyly. So I used the parent classes `__new__` constructor. Now it should work as expected.

ivirshup:
Thanks for adding this. Could you also add tests?

I think we could try using [memray](https://bloomberg.github.io/memray/index.html) through [pytest-memray](https://pytest-memray.readthedocs.io/en/latest/)

I'm a little concerned this may be brittle w.r.t. dask internals, but not sure there's anything we can do about that.

syelman:
I will look at the docs of pytest-memray. But what exactly should I be testing for? To check if it copies the dask array when view is modified or not? Or some sorts of memory leaks maybe?

> I'm a little concerned this may be brittle w.r.t. dask internals, but not sure there's anything we can do about that.

I think this concern is valid. Considering I had to access a private attribute. It might break if some changed happen in dask.array.

ivirshup:
> Or some sorts of memory leaks maybe?

This one. Just making sure that memory usage isn't increasing when we don't expect it to.

syelman:
@ivirshup I added some memory leak tests. I tried some values experimentally (since my theoretical estimations for the sizes didn't hold up) then I had to hard-code the memory limits. The tests pass on pipeline because pytest doesn't recognize pytest-memray markers, so it ignores it.

ivirshup (commented):
I would like to have a little bit larger margin of error around indexing overhead. E.g. it's okay if the indexing takes a little more memory. 

Would making the matrix larger give us this margin? Matrix memory usage should be $\propto N^2$, while indexing should be $\propto N$, right? So I'm thinking if we aim for a matrix that would allocate more like ~2-3 mb we'd be pretty safe around the indexing implementation.

> The tests pass on pipeline because pytest doesn't recognize pytest-memray markers, so it ignores it.

I'm not sure what you mean by this, could you explain a bit more?

Shouldn't memray get added to the test requirements in `pyproject.toml`?

anndata/tests/test_dask_view_mem.py
```diff
@@ -0,0 +1,136 @@
...
+    size = ((M, M), (M, M))
+
+    adata = ad.AnnData(
+        da.random.random(*size),
+        **{attr_name: dict(m=da.random.random(M))},
+    )
+    subset = adata[:N, :][:, :N]
+    assert subset.is_view
+    m = getattr(subset, attr_name)
+    m[0, 0] = 100
```
ivirshup, line 136: What is this test meant to be doing?

Right now it's giving `subset.obs` a column `(0, 0)`.

anndata/tests/test_dask_view_mem.py
```diff
@@ -0,0 +1,136 @@
...
+# since the array we allocated would be 4mb this should be ok
+@pytest.mark.usefixtures("alloc_cache")
+@pytest.mark.limit_memory("340 KB")
+def test_modify_view_mapping_component_memory(mapping_name):
+    import dask.array as da
+
+    N = 2**6
+    M = 2**9
+
+    size = ((M, M), (M, M))
```
ivirshup, line 79: Why are you specifying chunk size?

anndata/tests/test_dask_view_mem.py
```diff
@@ -0,0 +1,136 @@
+import pytest
+
+import anndata as ad
+
+# ------------------------------------------------------------------------------
```
ivirshup, line 5:
```suggestion

pytest.importorskip("pytest_memray")

# ------------------------------------------------------------------------------
```

To check that we don't pass these tests if pytest-memray isn't installed.

ivirshup (commented):
Added a few more questions

anndata/tests/test_dask_view_mem.py
syelman: I am not sure how I would estimate the memory allocated if I left it on auto, since it might work differently on different systems. From my experience, the chunk size can change the allocated memory.

anndata/tests/test_dask_view_mem.py
syelman: oh sorry, fixed it

syelman:
> Shouldn't memray get added to the test requirements in `pyproject.toml`?

Ok, added it, it shouldn't skip them now

anndata/tests/test_dask_view_mem.py
syelman: I made an update, it now tests for both cases

anndata/tests/test_dask_view_mem.py
```diff
@@ -0,0 +1,141 @@
...
+# since the array we allocated would be 4mb this should be ok
+@pytest.mark.usefixtures("alloc_cache")
+@pytest.mark.limit_memory("340 KB")
+def test_modify_view_mapping_component_memory(mapping_name):
+    import dask.array as da
+
+    N = 2**6
+    M = 2**9
+
+    size = ((M, M), (M, M)) if give_chunks else ((M, M), "auto")
```
ivirshup, line 86: `give_chunks` should be an argument to the test, right?

Is `if give_chunks` just always evaluating to `True` because it's in the namespace?

ivirshup: Ditto for the other functions.

syelman: Oh sorry again, fixed it

ivirshup:
Looks good, thanks for fixing this!

PR #873, opened 2022-12-16
(feat): `zarr` read/write and `hdf5` read

ilan-gold:
replicates https://github.com/ivirshup/anndata/pull/1


ivirshup (commented):
I think there's a couple issues with legacy reading (legacy zarr reading seems to not work), but otherwise this is looking very close to done!

Sorry about the long wait for a review! End of the year got very busy.

anndata/_io/h5ad.py
```diff
@@ -209,28 +210,25 @@ def read_h5ad(
...
-            else:  # Base case
-                d[k] = read_elem(f[k])
-
-        d["raw"] = _read_raw(f, as_sparse, rdasp)
+                return read_dataframe(elem)
+            return func(elem)
 
-        # Backwards compat to <0.7
-        if isinstance(f["obs"], h5py.Dataset):
-            _clean_uns(d)
```
ivirshup, line 231: I believe this will be what made the wrong entries show up in `uns`

ivirshup:
I've merged master, and am partially done clearing that up. One thing that got merged was better testing for the error reporting, which wasn't working quite right, so I've started fixing that. This is the changes to `@report_{read,write}_key_on_error`. I don't love what I've done there, so am open to other solutions.

ilan-gold:
Sorry @ivirshup I thought there was a more clear to-do here for me - I can work on getting the tests passing but was there something else?

ivirshup:
I think tests passing was the main thing.

Could you (in rough order of priority):

* Write some docs (this could be some examples for dask-io in https://github.com/scverse/anndata-tutorials)
* Take another look at the API. I think we could simplify the signature of the callbacks for `read_dispatched`, `write_dispatched`
* Try implementing `write_h5ad` with write_dispatched? I've done a preliminary version over at: https://github.com/scverse/anndata/pull/851

I'm also not super happy with how `read_h5ad` and `write_h5ad` handle backwards compat. I think the main issue here is that we're moving around elements, which `read_dispatched` isn't great at. I think it could be nice to more quickly branch to a "deals with backwards compat" code branch, and a "will get new features" code branch.

ilan-gold:
Steps: 
1. Simplify the callbacks to infer other needed properties from the `elem` itself
2. Unifying key access however
3. Tutorial -> use this to do (2) properly if needed

ivirshup:
I think docs are the top priority.

For inferring other fields from elem: This was also a question of opinion, do you think we should do that? I think writing examples in the docs would help clarify this. E.g. does removing these fields just add boiler plate to every callback, or does it clean up the API.

ilan-gold:
sounds good! found the repo after a little while to do that in, so will make one.  any preference on what the subject matter be?  maybe something remote-related? something self-contained?

ilan-gold:
A first-impression from working on the tutorial - we should decide how "public" the non-class method `read_elem` needs to be for this to really work well.  Having to import it means we should _probably_ make it more accessible than `._io.specs` from where I import it now (if there is a better way, let me know!).  Otherwise it's probably better to provide it as a default option within the `callback` in addition to `func` for recursion.

ivirshup:
> Having to import it means we should probably make it more accessible than ._io.specs

It's exported through `anndata.experimental`, which is also where `read_dispatched` and `write_dispatched` are exported.

ilan-gold:
> > Having to import it means we should probably make it more accessible than ._io.specs
> 
> It's exported through `anndata.experimental`, which is also where `read_dispatched` and `write_dispatched` are exported.

So is the goal then to make this a part of the "standard" `AnnData` package along with `read_dispatched` and `write_dispatched`?  If not, then I do think it would make sense to have it passed within the callback so the users can access it.

ivirshup:
> So is the goal then to make this a part of the "standard" AnnData package along with read_dispatched and write_dispatched?

Yes

> If not, then I do think it would make sense to have it passed within the callback so the users can access it.

I think we need to pass the function for cases where other registries are used. The function passed in here uses the registry in it's `Reader` or `Writer`, while the `read_elem`, `write_elem` function exported from `anndata.experimental` implicitly uses a registry we've defined in `AnnData`, but one could create a separate registry.

ilan-gold:
> > If not, then I do think it would make sense to have it passed within the callback so the users can access it.
> 
> I think we need to pass the function for cases where other registries are used. The function passed in here uses the registry in it's `Reader` or `Writer`, while the `read_elem`, `write_elem` function exported from `anndata.experimental` implicitly uses a registry we've defined in `AnnData`, but one could create a separate registry.

So you're saying that we should both export `read_elem` and have it passed down as an argument to the callback for use (like the `func` argument now)?  I think this makes it a little less clear what the end-goal of lazy dataframes for me should look like.  For example, one option would be to override the reading behavior within `read_dispatched` to "catch" dataframes as the reading process happens and read them in as a special class.  Another option would be to just change the `Reader` class used to have a custom registry that has the custom class registered.   What did you have in mind here?

ivirshup:
> So you're saying that we should both export read_elem and have it passed down as an argument to the callback for use (like the func argument now)?

Not exactly.

The `read_elem` in `anndata.experimental` is `Reader(anndata._io.specs._REGISTRY, callback=None).read_elem`.

So `read_elem` isn't really passed to `read_dispatched`. Something like it is, but it carries the callback with it. It could also be connected to a different registry, but that's another topic.

anndata 0.8 already exports `read_elem` and `write_elem`, and I don't want to remove them. I would consider them more stable than `read_dispatched` and `write_dispatched.

`read_elem` and `write_elem` were exported just to expose the encoding and decoding of supported types (and their standard in memory representation) to users. One could write a custom function for IO based on these, but it would be a lot of boiler plate.

> For example, one option would be to override the reading behavior within read_dispatched to "catch" dataframes as the reading process happens and read them in as a special class.

This seems like a good use case for `read_dispatched`.

> Another option would be to just change the Reader class used to have a custom registry that has the custom class registered. What did you have in mind here?

One could also do this, though I think you would also have a hard time reaching the flexibility of `read_dispatched` with this approach. It would be difficult to use additional context here, e.g. read only particular entries, do something special based on location, subset while reading, etc.

One could easily generalize arguments like `as_dense` with `read_dispatched`, it would be more difficult with registries.

> In other words, I'm not sure what the "suggested" method of doing customization is going to be 

Neither am I, which is why this is going into experimental. The main idea is to make it more possible to customize IO, figuring out what all the things we want to achieve with customized IO are, and then committing to a stable API.

ilan-gold:
Ok so what needs be updated here is that you want me to (a) remove the superfluous arguments on the callback and (b) unify key handling?

ivirshup:
> (a) remove the superfluous arguments on the callback

Which arguments? Right now, I'm thinking maybe just iospec.

> (b) unify key handling

I think I just did this, but now need to figure out what this other thing that broke was. Pretty sure it's some dependency updating.

---------

Could use docstrings?

ilan-gold:
> Could use docstrings?
For?

> Which arguments? Right now, I'm thinking maybe just iospec.
I think we said that the key can also be inferred?

ivirshup:
This is basically done, minus the doc + finalized API issue. But definitely good enough to merge so we can get a doc build with this + the tutorial together.

PR #896, opened 2023-02-02
Bump memory limit in test

ivirshup:
Started erroring, so bumped the margin by a bit.

cc: @syelman


ivirshup:
It looks kinda like pandas will randomly allocate 1mb, so it would be good if we had at least a ~1.5 mb margin of error.

Issue #897, opened 2023-02-02
Bump margin's of error on dask memory usage tests


ivirshup:
It looks like pandas will sometimes use about 1mb in our memory usage tests. 

<img width="793" alt="image" src="https://user-images.githubusercontent.com/8238804/216399530-aa81ff8d-e017-4582-a808-27d4e56d5155.png">

It looks unpredictable, so I think we should probably just bump the margin of error on these tests by about ~1.5mb. So this would mean just increasing the sizes of arrays we're passing through, and the expected memory usage.

cc: @syelman

PR #899, opened 2023-02-02
Remove AnnDataFileManager._adata

jggatter:
Original post on [discourse](https://discourse.scverse.org/t/avoiding-anndata-file-adata-self-reference/1097)

Slight tweaks to remove the cyclical reference `AnnDataFileManager._adata` which causes stack overflows when the AnnData object is parsed by parsers such as fastapi's [`jsonable_encoder`](https://fastapi.tiangolo.com/tutorial/encoder/)

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -395,9 +395,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
         # init from file
         if filename is not None:
-            self.file = AnnDataFileManager(self, filename, filemode)
+            self.file = AnnDataFileManager(filename, filemode)
         else:
-            self.file = AnnDataFileManager(self, None)
+            self.file = AnnDataFileManager(None)
 
             # init from AnnData
             if isinstance(X, AnnData):
@@ -1051,7 +1051,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         # change from backing-mode back to full loading into memory
         if filename is None:
             if self.filename is not None:
-                self.file._to_memory_mode()
+                self.file._to_memory_mode(self)
             else:
                 # both filename and self.filename are None
                 # do nothing
--- a/anndata/_core/file_backing.py
+++ b/anndata/_core/file_backing.py
@@ -16,11 +16,9 @@ class AnnDataFileManager:
 
     def __init__(
         self,
-        adata: "anndata.AnnData",
         filename: Optional[PathLike] = None,
         filemode: Optional[Literal["r", "r+"]] = None,
     ):
-        self._adata = adata
         self.filename = filename
         self._filemode = filemode
         self._file = None
@@ -76,9 +74,9 @@ class AnnDataFileManager:
         if self._file is not None:
             self._file.close()
 
-    def _to_memory_mode(self):
+    def _to_memory_mode(self, adata: "anndata.AnnData"):
         """Close the backing file, forget filename, *do* change to memory mode."""
-        self._adata.__X = self._adata.X[()]
+        adata.__X = adata.X[()]
         self._file.close()
         self._file = None
         self._filename = None


ivirshup:
Thanks for the PR! 

It looks like the CI failure is due to a flaky test (#897), so not your problem.

Does just removing this reference let you encode the AnnData? There are going to be other reference cycles in the AnnData object besides this (the AlignedMappings for example).

jggatter:
Thank you for the quick response, Isaac!

Ah you're right! It seems that PairwiseArrays (`obsp`/`varp`) and AxisArrays (`obsm`/`varm`) are AlignedMappings present on my AnnData that have the `parent` reference back to the AnnData. 

After some digging I've gotten a bit more context on how the `jsonable_encoder` is being used and why the AnnData is successfully passed to the flow with the changes made in this PR. The function is called [here](https://github.com/PrefectHQ/prefect/blob/26c145b3cbc8559a84e285e3be5fc054f371f3ba/src/prefect/flows.py#L350-L358), where you can see that they are excepting ValueErrors and TypeErrors. If there is an exception, Prefect determines the object is not serializable and does not send it to their backend for storing workflow metadata, etc.. I incorrectly assumed that the `jsonable_encoder` was being used to pass data to workflows themselves.

The `adata.file` field is third or so value checked after the AnnData is converted to a `dict` and iterated upon. Without the `file._adata` field, the cyclical reference traversal no longer happens. The `jsonable_encoder` continues to iterate until it raises an exception (luckily) before encountering any of these AlignedMappings.

For a bit more context, calling `jsonable_encoder(adata)` directly results in two TypeErrors: 
```
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 152, in jsonable_encoder
    data = dict(obj)
TypeError: cannot convert dictionary update sequence element #0 to a sequence

  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 157, in jsonable_encoder
    data = vars(obj)
TypeError: vars() argument must have __dict__ attribute

  File "/home/ubuntu/Projects/cyflows/cyflows/sandbox/adataX-example.py", line 24, in <module>
    main()
  File "/home/ubuntu/Projects/cyflows/cyflows/sandbox/adataX-example.py", line 16, in main
    jsonable_encoder(adata)
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 161, in jsonable_encoder
    return jsonable_encoder(
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 117, in jsonable_encoder
    encoded_value = jsonable_encoder(
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 161, in jsonable_encoder
    return jsonable_encoder(
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 117, in jsonable_encoder
    encoded_value = jsonable_encoder(
  File "/home/ubuntu/venv/anndata310/lib/python3.10/site-packages/fastapi/encoders.py", line 160, in jsonable_encoder
    raise ValueError(errors) from e
ValueError: [TypeError('cannot convert dictionary update sequence element #0 to a sequence'), TypeError('vars() argument must have __dict__ attribute')]
```
I didn't have time to check which field is causing this exception.

Anyways, it seems the fix in this PR only works by luck. The real issue has been that the `jsonable_encoder` is not keeping track of ids when traversing. Also that `jsonable_encoder(adata.obsm)` seems to raise a different ValueError:
```
ValueError: [TypeError("'numpy.float32' object is not iterable"), TypeError('vars() argument must have __dict__ attribute')]
```

I suppose I would appreciate this PR as a quick fix, but I understand if you would rather me file issues with prefect and fastapi.

ivirshup:
So, it looks to me like prefect is trying to bail if it hits something that it can't encode. But it's not bailing when it hits a circular reference, which it can't encode. To me this seems like a bug in prefect – since it should just bail.

Have you tried opening a bug there, seeing if they would make a change?

I'm not totally against making this change in `anndata`, the `FileManager` predates me and I'm not totally sure what it does. But if there's a more direct bug in prefect, I'd prefer that was fixed than making a change here.

jggatter:
Yes, agreed. I would have already filed files with fastapi and prefect, but I haven't been able to reproduce the problem with other objects, e.g.:
```python
from fastapi.encoders import jsonable_encoder

class Foo:
    def __init__(self):
        self._bar = Bar(self)

class Bar:
    def __init__(self, foo: Foo):
        self._foo = foo


foo = Foo()
jsonable_encoder(foo)
```
This simply spits out a `RecursionError: maximum recursion depth exceeded in comparison`.

I will file issues on fastapi and prefect and see what kind of reception I get.

jggatter:
No success with anyone. Anyway this should be Prefect's responsibility.

Thanks Isaac!

PR #900, opened 2023-02-03
Tuned Memory Limits for Tests

syelman:
Hi @ivirshup ,
do you think the limits I put here makes sense? Also, I am not sure about the calculations.
Related: #897


ivirshup (commented):
I checked the tolerances by adding in a `to_memory` call at the end of the functions. A number of them didn't error with that, so I increased the sizes of the subsets, and reduced the tolerance. Now they should error if the array loads into memory.

Right now, `test_size_of_view` is erroring for me locally. What is this test supposed to be doing? No views are taken.

syelman:
@ivirshup I did that as a base case, so if the views we allocate is the same size with the size we allocated we can give a minimum bound on how much we need to allocate. For example, if we use N=2^8 we can also set the N=2^8 on the ```test_size_of_view```. So if that test fails, we shouldn't expect others to pass.

But I can remove it if you find it redundant.

ivirshup (approved):
In theory, this should allocate about 1mb right?

-------

So, I sort of get it. But I think I would prefer a test which we expect to fail (e.g. [`@pytest.mark.xfail(strict=True)`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter)) when it allocates memory. Happy to merge as is, though.

syelman:
> In theory, this should allocate about 1mb right?

Yes.

> But I think I would prefer a test which we expect to fail (e.g. [`@pytest.mark.xfail(strict=True)`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#strict-parameter)) when it allocates memory. Happy to merge as is, though.

By this, do you mean to give a lower bound (e.g., 0.9 mb) so that we know the allocated size is between 0.9 mb and 1.5mb?

ivirshup:
> By this, do you mean to give a lower bound (e.g., 0.9 mb) so that we know the allocated size is between 0.9 mb and 1.5mb?

yep

syelman:
```limit-memory``` returns an object and doesn't raise anything when the limit is exceeded. So technically it doesn't fail.

https://github.com/bloomberg/pytest-memray/blob/986759c170d2f365aaf3a5ce2800d6a82b266fc0/src/pytest_memray/marks.py#L74-L76

I could use a nested test function instead of the decorators, but I don't know how good it would look. If you think it's worth it, I can find a workaround.

Also, I don't have the permissions to merge, just saying in case you are leaving it to me to do it.

**Comment by ivirshup** [comment]
Ah yeah, I see what you mean. It's kind of strange. Since this will get marked as failing:

```python
@pytest.mark.usefixtures("alloc_cache")
@pytest.mark.limit_memory("1 KB")
def test_size_of_view(mapping_name, give_chunks):
    ...
```

But this will fail, with `[XPASS(strict)] `

```python
@pytest.mark.xfail(strict=True)
@pytest.mark.usefixtures("alloc_cache")
@pytest.mark.limit_memory("1 KB")
def test_size_of_view(mapping_name, give_chunks):
    ...
```

If you can figure this out, that'd be great. But I'm going to merge this fix as is.

PR #904, opened 2023-02-06
Fix outer concat when generating fill obsm values

### LINKED ISSUE #901: anndata.concat bug: "Length mismatch"
Reproducible issue posted here: https://github.com/Teichlab/MultiMAP/issues/10

ivirshup:
Fixes #901

The function generating a fill value for missing dataframes in obsm was defining callable in a loop while relying on loop variables. So, the last value in the loop would be used by all closures.

### DIFF
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -593,7 +593,7 @@ def gen_outer_reindexers(els, shapes, new_index: pd.Index, *, axis=0):
         reindexers = [
             (lambda x: x)
             if not_missing(el)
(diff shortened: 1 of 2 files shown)


ivirshup:
I actually [sorta ran into this issue last week](https://scverse.zulipchat.com/#narrow/stream/315570-random/topic/python.20weirdness/near/325177496) with a feature very nearby this one. At the time I couldn't figure out why the new code wasn't working, but the old code was. Answer: the old wasn't working!

Unless: the last object to be concatenated was the one that needed a dataframe generated. Then it worked out. I guess this was surprisingly common?

PR #909, opened 2023-02-10
adding `axis` kwarg to `AnnLoader`

### LINKED ISSUE #908: enable sampling along `vars` axis with `AnnLoader`
Currently it is only possible to sample batches along `obs` axis with `AnnLoader`. Ideally we would be able to do both directions

gszep:
closes #908

### DIFF
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -15,7 +15,7 @@ def _normalize_indices(
     names0: pd.Index,
     names1: pd.Index,
     axis: Literal["obs", "vars"] = "obs",
-) -> tuple[slice, slice]:
+) -> Tuple[slice, slice]:
     # deal with tuples of length 1
     if isinstance(index, tuple) and len(index) == 1:
         index = index[0]
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -107,7 +107,9 @@ def _normalize_index(
         raise IndexError(f"Unknown indexer {indexer!r} of type {type(indexer)}")
 
 
-def unpack_index(index: Index, axis: Literal["obs", "vars"] = "obs"):
+def unpack_index(
+    index: Index, axis: Literal["obs", "vars"] = "obs"
+) -> Tuple[Index1D, Index1D]:
     if not isinstance(index, tuple):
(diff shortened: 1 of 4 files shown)


Zethson:
@Koncopd could you have a look, please?

gszep:
update on this: it's too painful to do this with annloader. I have resorted to defining my own torch `Loader` and `Dataset` classes where `Dataset` wraps the `AnnData` object. Happy to close without merging

gszep:
for reference here is my solution
```python

class AnnDataset(Dataset):
    """``torch.Dataset`` subclass that encapsulates ``AnnData`` for loading with ``AnnDataLoader``."""

    def __init__(
        self,
        anndata: AnnData,
        dtype: _dtype = torch.float32,
    ) -> None:
        self.dtype = dtype
        self.anndata = anndata

    @property
    def shape(self) -> tuple[int, int]:
        return self.anndata.shape  # type: ignore

    @property
    def obs(self) -> pd.DataFrame:
        return self.anndata.obs

    @property
    def obsm(self) -> AxisArrays:
        return self.anndata.obsm

    @property
    def var(self) -> pd.DataFrame:
        return self.anndata.var

    @property
    def varm(self) -> AxisArrays:
        return self.anndata.varm

    @property
    def X(self) -> scipy.sparse.base.spmatrix:
        return self.anndata.X

    @property
    def anndata(self) -> AnnData:
        return self._anndata

    @anndata.setter
    def anndata(self, anndata: AnnData) -> None:
        self._anndata = anndata
        self.tensors: dict[str, torch.Tensor] = {}
        
        if anndata.X is not None:
            self.tensors.update({"X": torch.tensor(anndata.X.todense(), dtype=self.dtype)})
            
        # conversions of obs/var to tensors go here
        #
        #

    def __getitem__(self, idx: tuple[list[int], list[int]]) -> Item:
        obs_idx, var_idx = idx
        tensors: dict[str, torch.Tensor] = {}

        if self.X is not None:
            tensors.update(
                {
                    "label": self.tensors["X"][obs_idx, var_idx],
                    "X": self.tensors["X"][obs_idx, :][:, var_idx],
                }
            )
        # slicing of obs/var tensors goes here
        #
        #

        return tensors
```

PR #923, opened 2023-02-22
Fix image in github rendered readme

ivirshup:
Casualty of rst2myst


flying-sheep:
Why can MyST not handle image directives?

ivirshup:
Myst handles it, github can't:

image

PR #924, opened 2023-02-27
(feat): `read_remote` via `read_dispatched`

ilan-gold:
so you can see @ivirshup


anndata/experimental/read_remote/read_remote.py
```diff
@@ -0,0 +1,335 @@
...
+            a = a.astype(args[0])
+        return a
+
+    def __getitem__(self, selection):
+        result = super().__getitem__(selection)
+        return pd.Categorical.from_codes(
+            codes=result,
+            categories=self.categories,
+            ordered=self.ordered,
+        )
```
ivirshup, line 54: Looking again, I think you more or less want to do what was suggested. It's just a very minimal implementation of an array, that supports indexing.

I think the zarr array class is not appropriate to subclass like this. It has a ton of assumptions about how it's structured, which just aren't met by the categorical encoding here.
```suggestion
import numpy as np
import pandas as pd

# Doesn't need to use this, just using it for a couple inherited methods
from xarray.core.indexing import ExplicitlyIndexedNDArrayMixin

class LazyCategoricalArray(ExplicitlyIndexedNDArrayMixin):
    __slots__ = ("codes", "categories", "attrs")

    def __init__(self, group, *args, **kwargs):
        self.codes = group["codes"]
        self.categories = group["categories"][...]  # slots don't mix with cached_property, ExpicitlyIndexedArray uses slots
        self.attrs = dict(group.attrs)

    @property
    def dtype(self) -> pd.CategoricalDtype:
        return pd.CategoricalDtype(self.categories, self.ordered)
    
    @property
    def shape(self) -> tuple[int, ...]:
        return self.codes.shape

    @property
    def ordered(self):
        return bool(self.attrs["ordered"])

    def __array__(self, *args) -> np.ndarray:  # may need to override this, copied for now
        a = self[...]
        if args:
            a = a.astype(args[0])
        return a

    def __getitem__(self, selection) -> pd.Categorical:
        return pd.Categorical.from_codes(
            codes=self.codes.oindex[selection],
            categories=self.categories,
            ordered=self.ordered,
        )

    def __repr__(self) -> str:
        return f"LazyCategoricalArray(codes=..., categories={self.categories}, ordered={self.ordered})"
```

anndata/_core/sparse_dataset.py
```diff
@@ -376,9 +376,9 @@ def to_backed(self) -> BackedSparseMatrix:
     def to_memory(self) -> ss.spmatrix:
         format_class = get_memory_class(self.format_str)
         mtx = format_class(self.shape, dtype=self.dtype)
-        mtx.data = self.group["data"][...]
-        mtx.indices = self.group["indices"][...]
-        mtx.indptr = self.group["indptr"][...]
+        mtx.data = self.group["data"]
+        mtx.indices = self.group["indices"]
+        mtx.indptr = self.group["indptr"]
```
ivirshup, line 381: Is this neccesary? Probably leading to test failures. Also basically what `to_backed` is for?

anndata/experimental/read_remote/read_remote.py
ilan-gold: Sounds good.  Wasn't a huge fan of adding another dependency, but I guess `xarray` is coming at some point anyway.

ilan-gold:
Closing this because it is on my fork.

anndata/experimental/read_remote/read_remote.py
ivirshup: Could also do it without xarray, though it's a very light dependency.

ivirshup: I think you'd just need to define things like `__length__`, `.ndim`, etc.

PR #928, opened 2023-03-01
Add sphinx-opengraph

ivirshup:
Adding open graph support to our docs so we get previews on social sites.


ivirshup:
I'm not 100% sure if the site url was necessary, or if adding that happened to coincide with the cache updating so https://www.opengraph.xyz/ would show the image.

PR #929, opened 2023-03-01
Try checking against upstream release candidates

ivirshup:
Adding a test job that uses `pip install --pre` to test against upstream release candidates.

Could use some consideration on how exactly this works.

- [ ] Decide if this should be required
- [ ] Check that installation is happening correctly


ivirshup:
Included in #931

PR #946, opened 2023-03-08
(feat): swap out `zarr` categorical array for `xarray`

ilan-gold:


ilan-gold:
@ivirshup do you mind if I leave the xarray dependency for now and then we can double back to this?

ivirshup:
Fine to leave for now. Can this get tests?

PR #947, opened 2023-03-08
(feat): experimental `read_backed` method for `zarr` + `hdf5` via `read_dispatched`

### LINKED ISSUE #951: Dask and Zarr not loading obsp and obsm from remote s3
Hi,

I'm using @ilan-gold's nice sample code at
https://anndata.readthedocs.io/en/latest/tutorials/notebooks/%7Bread,write%7D_dispatched.html
to read remote anndata, which is working great when I'm serving data locally via http.

But when I'm using minio to serve the data, I'm not getting the obsp, obsm or uns parts of the AnnData object. See code sample below.

A UI view of the data is at https://deploy-preview-20--ome-ngff-validator.netlify.app/?source=https://minio-dev.openmicroscopy.org/idr/temp_table/test_segment.zarr/tables/regions_table/
(from https://github.com/ome/ome-ngff-validator/pull/20) where I'm using the extra "keys" in e.g. https://minio-dev.openmicroscopy.org/idr/temp_table/test_segment.zarr/tables/regions_table/obsm/.zattrs to load those groups:

```json
{
    "encoding-type": "dict",
    "encoding-version": "0.1.0",
    "keys": [
        "X_scanorama",
        "X_umap",
        "spatial"
    ]
}
```

Is there any way I can use those 'keys' to load `obsp` and `obsm` data?

 Thanks!

```py
from typing import Callable, Union
import dask.array as da
import zarr
from anndata import AnnData
from anndata._io.specs import IOSpec
from anndata.compat import H5Array, H5Group, ZarrArray, ZarrGroup

# ** requires anndata==0.9.0.rc1
from anndata.experimental import read_dispatched, read_elem
from zarr.storage import FSStore
from ome_zarr.io import parse_url
StorageType = Union[H5Array, H5Group, ZarrArray, ZarrGroup]


def read_remote_anndata(store: FSStore, name: str) -> AnnData:
    table_group = zarr.group(store=store, path=name)

    def callback(
        func: Callable, elem_name: str, elem: StorageType, iospec: IOSpec
    ) -> AnnData:
        if iospec.encoding_type in (
            "dataframe",
            "csr_matrix",
            "csc_matrix",
            "awkward-array",
        ):
            # Preventing recursing inside of these types
            return read_elem(elem)
        elif iospec.encoding_type == "array":
            return da.from_zarr(elem)
        else:
            return func(elem)

    adata = read_dispatched(table_group, callback=callback)
    return adata

url = "https://minio-dev.openmicroscopy.org/idr/temp_table/test_segment.zarr/tables/"
name = "regions_table"
store = parse_url(url, mode="r").store
anndata_obj = read_remote_anndata(store, name)

print('anndata_obj', anndata_obj)

# anndata_obj AnnData object with n_obs × n_vars = 1045 × 36
#     obs: 'row_num', 'point', 'cell_id', 'X1', 'center_rowcoord', 'center_colcoord', 'cell_size', 'category', 'donor', 'Cluster', 'batch', 'library_id'
#     var: 'mean-0', 'std-0', 'mean-1', 'std-1', 'mean-2', 'std-2'


url = "http://localhost:8000/test_segment.zarr/tables/"
store = parse_url(url, mode="r").store
anndata_obj = read_remote_anndata(store, name)

print('anndata_obj', anndata_obj)

# anndata_obj AnnData object with n_obs × n_vars = 1045 × 36
#     obs: 'row_num', 'point', 'cell_id', 'X1', 'center_rowcoord', 'center_colcoord', 'cell_size', 'category', 'donor', 'Cluster', 'batch', 'library_id'
#     var: 'mean-0', 'std-0', 'mean-1', 'std-1', 'mean-2', 'std-2'
#     uns: 'Cluster_colors', 'batch_colors', 'neighbors', 'spatial', 'umap'
#     obsm: 'X_scanorama', 'X_umap', 'spatial'
#     obsp: 'connectivities', 'distances'

```

### LINKED ISSUE #981: lazy `.obs` and `.var` with `backed="r"` mode

ilan-gold:
Creates a new experimental `AnnDataBacked` class as well as a helper `read_backed` method for reading of on-disk `zarr` and `hdf5` datasets lazily, with a focus on usage over the internet.  The primary focus is less on optimizing for performance of analysis than on reading metadata/making fetching of data quick.

Fixes #951 as well
Probably fixes #981 if an "experimental" PR can do that.

The main highlights of this PR in `experimental.read_backed`:
- `xarray` for all dataframes
- by-default lazy loading of all elements (dense array, sparse array, dataframes) except `AwkwardArray`s
- Out-of-the-box, fully backed, works with both `h5ad` and `zarr` while also being compatible with both on-disk/remote storage (i.e., changing subelements __should__ work if you want them to be e.g., `numpy` arrays although this isn't tested because the focus here is really on reading)
- simple `to_memory` function on the `AnnDataBacked` class that brings whatever you want locally for use in `scanpy` including optional `exclude` keys for making the download faster by restricting it to exactly what you need

Outside of this PR but included in the overall work:
- views of views for backed sparse matrices (and more targeted/efficient reading)
- common abstract `AnnData` class beginning to define a sensible contract for new classes to build upon
- backed `X` for `zarr`


ilan-gold:
We should probably figure out a different name than `read_remote` also, since this could work with local zarr stores as well.

ilan-gold:
Some things to do from working with a larger dataset now:
1. Merging sparse support for zarr in to this PR and then improving on access patterns there for backed data
2. Deferring loading to as late as possible for backed sparse data
3. Deferring `categories` load in the categorical array as well
4. Concatenation

ilan-gold:
5. Caching requests via a special Zarr store

ilan-gold:
@ivirshup Here's a code snippet:

```python
import fsspec, zarr
from anndata.experimental.read_remote import read_remote
mapper = fsspec.get_mapper('https://storage.googleapis.com/vitessce-demo-data/cxg_examples/BALF_VIB-UGent_processed_cleaned_080.zarr/BALF_VIB-UGent_processed_cleaned.zarr/')
cache = zarr.LRUStoreCache(mapper, max_size=2**28)
adata = read_remote(cache)
```

From there you should be able to try out different ops.

ilan-gold:
P.S @ivirshup I was planning on splitting this out from one monolith file as well as renaming it.  You will also need additional dependencies: `requests`, `aiohttp`, `fsspec` and zarr `

ivirshup (commented):
Notes from review discussion

## Major points

* Change reprs so it's obvious when things are different
* Consistent way of getting to in memory representation. `.to_memory` method?
* Cache seems to not be hitting everything, this should get a test

## Minor points

* Bug in .iloc
* Some issue with plotting, cpu maximization
* Instead of calling `read_dispatched(x, callback)` within the callback, shouldn't we be able to call `func(x)`?

ilan-gold:
> Change reprs so it's obvious when things are different

Should be done

> Consistent way of getting to in memory representation. .to_memory method?

Also done, it now exists on the `AnnDataBacked` object.  For example, I added a `X_csc` to `layers` for fast per-gene fetching so that the following works relatively quick:

```python
import fsspec, zarr
from anndata.experimental.read_backed import read_backed
import scanpy as sc

mapper = fsspec.get_mapper('http://127.0.0.1:8080/BALF_VIB-UGent_processed_cleaned.zarr')
cache = zarr.LRUStoreCache(mapper, max_size=2**28)
adata = read_backed(cache)
local = adata[:, ['CD40', 'CD5']].to_memory(exclude_X=True)
sc.pl.umap(local, color=['28-day outcome', 'Annotation', 'CD40', 'CD5'], layer='X_csc')
```
> Cache seems to not be hitting everything, this should get a test

Bug fixed!  It was a bit of a weird bug so I'll think on how to add a test for this.

> Bug in .iloc

So this gets to something I need some clarification on.  How do we hande `subset_idx`?  I've seen things like `self._view(self.parent, (idx,))` but I don't really understand when to use a tuple around `idx` and not.  And then I am a little lost on differences in data access patterns like `arr[()]` vs. `arr[...]`, and this seems related.  But this gets tot he heart of the bug!

> Some issue with plotting, cpu maximization

`to_memory` should be called before plotting, tests added for that.

> Instead of calling read_dispatched(x, callback) within the callback, shouldn't we be able to call func(x)?

Still need to look into this

ivirshup (commented):
> So this gets to something I need some clarification on. How do we hande subset_idx? I've seen things like self._view(self.parent, (idx,)) but I don't really understand when to use a tuple around idx and not. And then I am a little lost on differences in data access patterns like arr[()] vs. arr[...], and this seems related. But this gets tot he heart of the bug!

So, it looks like `(idx,)` is used for `obsm` and `varm` but `idx` is used for `obsp` and `varp`. It looks like this has to do with their `.__init__` methods.

E.g.:

https://github.com/scverse/anndata/blob/14cadf18ce1baaa76f8598de6526a479e76ed14b/anndata/_core/aligned_mapping.py#L376

vs

https://github.com/scverse/anndata/blob/14cadf18ce1baaa76f8598de6526a479e76ed14b/anndata/_core/aligned_mapping.py#L277

 I would be happy to change this in their implementations so this would be more consistent. Or to at least fix their signatures.

However, I looked into this and think I found another bug:

https://github.com/scverse/anndata/blob/9e07da21e43a6595d912f04a9ac08d461c8fb36e/anndata/_core/aligned_mapping.py#L45-L46

The repr for aligned mappings that you've written accesses each value. For a view, this is expensive.

> And then I am a little lost on differences in data access patterns like arr[()] vs. arr[...], and this seems related. But this gets tot he heart of the bug!

I think this is array class specific, though I believe they *should* be equivalent. I don't think this is related, as this has less to do with code in `anndata`

-------

Could you merge main so the CI works?

ilan-gold:
> However, I looked into this and think I found another bug: The repr for aligned mappings that you've written accesses each value. For a view, this is expensive.

I think this is only for categorical values.  The issue is that accessing the `dtype` for a categorical array requires reading in the categories.  We can stop doing that, but then we lose the `categories` in the `dtype` i.e., we have 
```
@property
def dtype(self) -> pd.CategoricalDtype:
    return pd.CategoricalDtype(self.categories, self.ordered)
```
on the `LazyCategoricalArray` right now.

Up to you, we could just not return categories here.

pyproject.toml
```diff
@@ -57,6 +57,10 @@ Home-page = "https://github.com/scverse/anndata"
 dev = [
     # dev version generation
     "setuptools_scm",
+    # static checking
+    "black>=20.8b1",
```
Zethson, line 61: I guess people kind of get black through pre-commit? Do we need to specify it here as well?

ilan-gold: Thanks!

ilan-gold:
Before this can be merged/reviewed, there are several blocking PRs:
#949 
#765 
https://github.com/ivirshup/anndata/pull/4

ilan-gold:
Open question: If we want no-load `indices` on load (i.e., `obs_names`) and we decide to make `obs` a generator we need to figure out
1. Is it immutable or mutable? (`cached_property` or just `property`)
2. If mutable, how do we do caching?
See https://github.com/pydata/xarray/issues/1650

ilan-gold:
**Note to self**: we need the `sparse_dataset` changes (I think) because they allow for getting a representation of the matrix without actually reading it into memory.  That is, for `X` in the current `AnnData` object, the following happens
https://github.com/scverse/anndata/blob/3e340e1b266477c030606f5025972aba45876f57/anndata/_core/anndata.py#L667-L685
This is problematic because it means that accessing `X` immediately reads it into memory.  This is why I moved the indexing on to the `BaseCompressedSparseDataset` class here.

ilan-gold:
No need as #1247 seems like it will be the way to go.

PR #950, opened 2023-03-15
bugfix: read fixed-length string arrays as bytes

gszep:
There is a bug in pandas that prevents fixed-length string arrays `np.dtype("S")` from being read properly. For some reason the dataframe does the casting properly only if the array is set after dataframe creation.
```python
x = np.array([b"123",b"e23423"], dtype="S")
df = pd.DataFrame({"a": x})
df.types # sees x as object
df = pd.DataFrame({})
df["a"] = x
df.dtypes # correct type
```
This PR is a workaround until the pandas bug is addressed


gszep:
reported in pandas https://github.com/pandas-dev/pandas/issues/51994

gszep:
moved to #982 so I can delete `master` in favor of `main`

PR #955, opened 2023-03-17
Out of core concatenation support

syelman:
Hi,

Now that I have an idea how this might work, I decided to copy ```concat``` function's signature almost 1-1 was the best option for me. This way I could get familiar with the existing code easier using unit tests.

### Signature

```python 
def concat_on_disk(
	in_files: Union[Collection[str], typing.MutableMapping],
	out_file: Union[str, typing.MutableMapping],
	overwrite: bool = False,
	*,
	axis: Literal[0, 1] = 0,
	join: Literal["inner", "outer"] = "inner",
	merge: Union[StrategiesLiteral, Callable, None] = None,
	uns_merge: Union[StrategiesLiteral, Callable, None] = None,
	label: Optional[str] = None,
	keys: Optional[Collection] = None,
	index_unique: Optional[str] = None,
	fill_value: Optional[Any] = None,
	pairwise: bool = False,
):

"""Concatenates multiple AnnData objects along a specified axis using their
corresponding stores or paths, and writes the resulting AnnData object
to a target location on disk.

Unlike the `concat` function, this method does not require
loading the input AnnData objects into memory,
making it a memory-efficient alternative for large datasets.
The resulting object written to disk should be equivalent
to the concatenation of the loaded AnnData objects using
the `concat` function."""
```

Some notes and doubts on the signature decision:

- I made the file related parameters positional to somewhat make the signature easier to read. 
- regarding the above item, I don't know if it might be a bad or good thing in the future.
- I copied all the parameters of the original function and I figured even if we don't support every variation we might raise an error when not

### Functionality
Nothing yet. I indent to implement the case when the format is zarr and the others are the default params first.

### Unit Tests for Equivalence to Concat
I believe I added all the tests that should ensure `concat` and `concat_on_disk` are giving equivalent results. From the unit tests of `concat` I added all the adatas that are given to that function and 
wrote them to disk, gave it to `concat_on_disk` with same arguments except the filenames. Called `assert_equal` on both results.

### Unit Tests for Memory and Disk stuff
Not done and probably should be done. (Memory leaks, filenames etc.)

@ivirshup


anndata/_io/merge.py
```diff
@@ -0,0 +1,506 @@
...
+    # TODO: Generalize this to more than zarr
+    groups = [zarr.open(store=p) for p in in_files]
+    output_group = zarr.open(store=out_file)
+    assert len(groups) > 1
+
+    # var or obs
+    _, dim = _resolve_dim(axis=axis)
+    alt_axis, alt_dim = _resolve_dim(axis=1 - axis)
+
+    # TODO: This is just a temporary assertion
```
ivirshup, line 319: Why is this temporary and when would it be removed?

Right now, it looks like this is loading in the obs_names/ var_names multiple times, but it would be nice to load them just once.

anndata/_io/merge.py
```diff
@@ -0,0 +1,506 @@
...
+@_write_concat_elem.register
+def _(elem: SparseDataset, output_group, path, axis=0):
+    write_elem(output_group, path, elem)
+    return SparseDataset(output_group[path])
+
+
+@_write_concat_elem.register
+def _(elem: zarr.Array, output_group: zarr.Group, path, axis=0):
+    res = output_group.create_dataset(path, data=elem)
+    return res
```
ivirshup, line 119: How is this different from `write_elem`?

anndata/_io/specs/registry.py
```diff
@@ -326,6 +383,20 @@ def read_elem(elem: StorageType) -> Any:
...
+    modifiers: frozenset(str) = frozenset(),
+) -> Any:
+    """
+    Read an element from as Zarr group
+    """
+
+    reader = GroupReader(_REGISTRY)
+    return _REGISTRY.get_read_groups(type(elem), get_spec(elem), frozenset(modifiers))(
+        elem, _reader=reader
+    )
```
ivirshup, line 397: I see what you mentioned in the meeting now. 

For now, I think I would accomplish this with a `read_dispatched` function.

But, if I were to do this through the registries, instead of creating a new method on the registry class, I would create a new registry.

anndata/tests/test_concatenate_disk.py
```diff
@@ -0,0 +1,853 @@
...
+        del orig.varm
+        del orig.varp
+        result.obs.drop(columns=["batch"], inplace=True)
+
+    # Possibly need to fix this, ordered categoricals lose orderedness
+    for k, dtype in orig.obs.dtypes.items():
+        if is_categorical_dtype(dtype) and dtype.ordered:
+            result.obs[k] = result.obs[k].astype(dtype)
+
+    return orig, result
```
ivirshup, line 135: I don't think this exact function will be relevant here, since I don't think we'll implement the `AnnData.concatenate` API

ivirshup (changes requested):
Looks like a good start! I've made a few requests for changes. A few more points:

## `read_groups`

The biggest one is probably the request to use either `read_dispatched` instead of adding new methods to the registry class. However, I'm not sure I understand what exactly the `read_groups` method is supposed to do. Could you explain that?

## Commented + unused code

 Could you do a pass and try to remove commented out code and unused code paths that you're no longer using? I think it would increase the readability here quite a bit.

## Expanding tests

This would be the other big point. Do you think you could expand the test suite to cover more cases? I've given some suggestions in my comments.

anndata/tests/test_concatenate_disk.py
```diff
@@ -0,0 +1,853 @@
...
+    res1 = concat(adatas, *args, **kwargs)
+    # create one from the on disk concat function
+    paths = adatas_to_zarr_paths(adatas, tmp_path)
+    out_name = tmp_path / "out.zarr"
+    concat_on_disk(paths, out_name, *args, **kwargs)
+    res2 = read_zarr(out_name)
+    assert_equal(res1, res2)
+
+
+def test_concatenate_dense(join_type, tmp_path):
```
ivirshup, line 181: Could this get parameterized on data types? I would like to know this works with CSR arrays as well.

Could you also add more types into the data frame columns and mapping attributes of the AnnDatas? It would be good to be verifying those were behaving as expected.

anndata/_io/specs/registry.py
syelman: fixed that here https://github.com/scverse/anndata/pull/967

syelman:
Hi @ivirshup ,

So I decided trying to implement inner concat with reindexing before creating more unit tests for the "no-reindex" version. Because creating test cases for it seemed easy and most of the tests would be checking if it failed for a bad input and when I implemented reindexing I would had to rewrite those cases and add cases for reindexing.

Firstly, I'd like to ask what you prefer. Adding the unit cases and refactoring, so we have something to merge, or should I also add the inner with reindexing case?

If I should go for reindexing, I'd like to ask if supporting h5ad is worth it or not? Because dask seems to make many things easier with reindexing related stuff.

I have progressed on inner with reindexing case (supports sparse and zarr arrays) in this branch https://github.com/syelman/anndata/tree/reindexing-ooc-concat

flying-sheep:
Tell us when you’re ready for another round of reviewing!

syelman:
> Please feel free to tell us when you’re ready for another round of reviewing!

Thanks @flying-sheep , I think it's ready for another round.

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+)
+import typing
+from pathlib import Path
+import numpy as np
+
+from ..compat import ZarrGroup, ZarrArray, H5Group, H5Array
+
+from scipy.sparse import csr_matrix, csc_matrix
+
+MAX_LOAD_SIZE = 60_000_000
```
ivirshup, line 41: Could this get a comment describing what the units are (e.g. elements, bytes) and why this value?

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+            axis=axis,
+            index=index,
+            reindexers=reindexers,
+            fill_value=fill_value,
+        )
+
+
+SPARSE_MATRIX = {"csc_matrix", "csr_matrix"}
+
+EAGER_TYPES = {"dataframe", "awkward-array"}
```
ivirshup, line 98: Could these be moved to the top of the file?

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+
+
+def read_group(group: Union[ZarrGroup, H5Group]):
+    """Read the group until
+    SparseDataset, Array or EAGER_TYPES are encountered."""
```
ivirshup, line 103: Could this be called `read_as_backed`? I think `read_group` is a bit too generic.

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+    reindexers = [
+        gen_reindexer(pd.RangeIndex(min_ind), pd.RangeIndex(a.shape[1 - axis]))
+        for a in arrays
+    ]
+
+    return reindexers
+
+
+def _gen_slice_to_append(
+    datasets: Sequence[SparseDataset],
```
ivirshup, line 182: This is also called on hdf5 arrays, so I don't think this typing is accurate.

Could this either be changed, or could we use `dask` for hdf5 as well as zarr?

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+                axis=axis,
+                reindexers=reindexers,
+                fill_value=fill_value,
+            ),
+            axis=axis,
+        )
+        write_elem(output_group, out_path, res)
+        output_group[out_path].attrs.update(
+            {"encoding-type": "array", "encoding-version": "0.2.0"}
+        )
```
ivirshup, line 299: Could these cases also be wrapped in functions, like the sparse arrays are?

ivirshup, line 299: Should this be necessary? I would think `write_elem` would take care of this.

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+        idx = 0
+        for arr in _gen_slice_to_append(
+            arrays, axis=axis, reindexers=reindexers, fill_value=fill_value
+        ):
+            added_size = arr.shape[axis]
+            if axis == 0:
+                out_array[idx : idx + added_size, :] = arr
+            else:
+                out_array[:, idx : idx + added_size] = arr
+            idx += added_size
```
ivirshup, line 278: Why don't hdf5 and zarr use the same logic for dense arrays?

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+    axis=0,
+    index=None,
+    reindexers=None,
+    fill_value=None,
+    join="inner",
+):
+    """
+    array, dataframe, csc_matrix, csc_matrix
+    """
+    arrays: Union[pd.DataFrame, SparseDataset, H5Array, ZarrArray] = [
```
ivirshup, line 315:
```suggestion
    arrays: list[Union[pd.DataFrame, SparseDataset, H5Array, ZarrArray]] = [
```

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+@_get_group.register
+def _(store: H5Group, *args, **kwargs) -> H5Group:
+    return store
+
+
+def _get_groups_from_paths(
+    in_files: Union[
+        Collection[str],
+        Collection[Path],
+        MutableMapping,
```
ivirshup, line 402: I don't think this case will work:  you get keys, not values, when you iterate.

You're already extracting the values below, and never passing a mapping here – so I think we could just remove this.

```python
    if isinstance(in_files, Mapping):
        if keys is not None:
            raise TypeError(
                "Cannot specify categories in both mapping keys and using `keys`. "
                "Only specify this once."
            )
        keys, in_files = list(in_files.keys()), list(in_files.values())
```

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+
+
+def _get_group_from_hdf5(store, *args, **kwargs) -> H5Group:
+    import h5py
+
+    return h5py.File(store, *args, **kwargs)
+
+
+@singledispatch
+def _get_group(store, *args, **kwargs) -> Union[ZarrGroup, H5Group]:
```
ivirshup, line 359: I rename this. Something like`"as_group"` to highlight that it's does pretty heavy coersion, and is quite dynamic.

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+def _(store: ZarrGroup, *args, **kwargs) -> ZarrGroup:
+    return store
+
+
+@_get_group.register
+def _(store: H5Group, *args, **kwargs) -> H5Group:
+    return store
+
+
+def _get_groups_from_paths(
```
ivirshup, line 398: I think this may be simple enough to just go inline.

Otherwise I think it should at least be renamed, like `_get_group`.

flying-sheep (changes requested):
Some small changes I’d like to see in addition to Isaac’s review

anndata/tests/helpers.py
```diff
@@ -163,6 +163,7 @@ def gen_adata(
         AwkArray,
     ),
     layers_types: "Collection[Type]" = (sparse.csr_matrix, np.ndarray, pd.DataFrame),
+    sparse_fmt: str = "csr",
```
ivirshup, line 166: Is this different than what `X_type` is doing?

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+def gen_reindexers_array_inner(
+    arrays: Sequence[Union[ZarrArray, H5Array, SparseDataset]], axis: Literal[0, 1] = 0
+):
+    min_ind = min(a.shape[1 - axis] for a in arrays)
+    reindexers = [
+        gen_reindexer(pd.RangeIndex(min_ind), pd.RangeIndex(a.shape[1 - axis]))
+        for a in arrays
+    ]
+
+    return reindexers
```
ivirshup, line 178: I think this might be duplicated from the in-memory concat. Let me check in on that, and see if there's an easy way to use those methods.

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+                    ds_part = ds[idx : idx + slice_size, :]
+                elif axis == 1:
+                    ds_part = ds[:, idx : idx + slice_size]
+
+                yield ri(ds_part, axis=1 - axis, fill_value=fill_value)
+                rem_slices -= slice_size
+                idx += slice_size
+
+
+def _write_concat_sparse(
```
ivirshup, line 209: It's not clear to me that special casing is happening for concatenation of sparse arrays where no reindexing needs to happen. This would be important for performance.

ivirshup (commented):
Hey, I think this is looking pretty good!

A fair number of comments, but I think they are mainly small things. A few points that didn't fit as comments:

### Reorganization

I would like to see a little more organization here. This means:

* Global variables at top
* Order and group functions by topic. Plus section headers, like in `merge.py`. Sections would include:
  * Write methods
  * Read methods (including `get_encoding_type`, `get_shape`, `df_index`)
  * Indexing methods

### Experimental

I think this should go in experimental, not IO, for now.

### read_dispatched

Nice to see the read dispatched being used! Good testing for the method.

anndata/tests/test_concatenate_disk.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    Gets list of adatas, writes them and returns their paths as zarr
+    """
+    paths = None
+
+    def write_func(adata, path):
+        if file_format == "h5ad":
+            adata.write(path)
+        else:
+            adata.write_zarr(path)
```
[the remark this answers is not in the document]
syelman: But wouldn't that require checking for file format, opening a group according to the format, and then using write elem? Asking because I also would like to avoid that

anndata/_io/merge.py
syelman: Ok, thanks, until you wrote this, I overlooked the fact that dask supported hdf5 reading. Apparently, it's the ```from_array``` function.

anndata/tests/helpers.py
syelman: yes as the other datatypes that use sparse also need it
```python
obsm = dict(
      array=np.random.random((M, 50)),
      sparse=sparse.random(M, 100, format=sparse_fmt),
      df=gen_typed_df(M, obs_names),
      awk_2d_ragged=gen_awkward((M, None)),
      da=da.random.random((M, 50)),
  )
```

anndata/_io/merge.py
```diff
@@ -0,0 +1,647 @@
...
+    )
+    concat_annot.index = concat_indices
+    if label is not None:
+        concat_annot[label] = label_col
+    write_elem(output_group, dim, concat_annot)
+
+
+def concat_on_disk(
+    in_files: Union[
+        Collection[str],
```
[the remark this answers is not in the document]
syelman: MutableMapping should stay, right? Also maybe I shouldn't add ZarrGroup and H5Group because it's an internal wrapper class.

anndata/_io/merge.py
syelman: It is currently elements; I was going to ask if this was ok or if I should use bytes instead. For now, I am writing it as elements.

anndata/_io/merge.py
syelman: true, solved it

anndata/_io/merge.py
syelman: So I did this when there was no reindexing required at all:

```python
class IdentityReindexer:
    def __init__(self):
        self.no_change = True

    def __call__(self, x, *args, **kwargs):
        return x
```

But I wasn't sure if I wanted to do it for the cases where some matrices don't require reindexing. If it is ok to do that, then I think this would be better done by changing the Reindexer class.

syelman:
Thanks for the reviews, guys @ivirshup @flying-sheep! Many duplicate/unnecessary code was deleted. I covered most of the points besides the unresolved conversations. 

But I didn't do much about sparse concatenation because of the global variable problem (Max load limit). Because I think this should be solved on the ```SparseDataset``` level. It would make more sense to add on-disk concatenation to the ```SparseDataset``` class and then use it in our concatenation function.

ivirshup:
> Because I think this should be solved on the SparseDataset level. It would make more sense to add on-disk concatenation to the SparseDataset class and then use it in our concatenation function

Taking a look now, but one quick note: `SparseDataset` does have on-disk concatenation implemented: https://github.com/scverse/anndata/blob/7a9856db56e09508b93d29265ac0e122bb359b8c/anndata/tests/test_backed_sparse.py#L45-L92

Is this what you meant?

syelman:
> Is this what you meant?

Yes, sorry, I am already using that. I just thought since I wrote the slice-by-slice writing thing for only `SparseDataset`, wouldn't it make sense for it to support that overall regardless of AnnData? But I also now realize if we did this, we would also need to support reindexing to make it useful for concatenation. It seemed like a good idea for a while but seems like a side track (probably involves using dask inside `SparseDataset`).

For now, I will focus on adding a `max_elems_on_sparse` argument.

anndata/_io/merge.py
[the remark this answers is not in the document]
flying-sheep: Yes, the type I’m referring to is just for file paths.

Either use just `Path` or what I said for file paths, never `str`.

anndata/tests/test_concatenate_disk.py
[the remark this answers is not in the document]
ivirshup: Why?

Though this isn't terribly important as it's just part of the test.

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+
+    def _gen_elem_to_append(elems):
+        for elem, ri in zip(elems, reindexers):
+            yield ri(elem, axis=1 - axis, fill_value=fill_value)
+
+    darrays = None
+    init_elem = arrays[0]
+
+    if isinstance(init_elem, H5Array):
+        darrays = (da.from_array(a, chunks="auto") for a in arrays)
```
ivirshup, line 185: I think `from_array` isn't 100% equivalent to `from_zarr`. Could this try to use the chunk size the array already has?

Here's from `from_zarr` is defined: https://docs.dask.org/en/stable/_modules/dask/array/core.html#from_zarr

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+    output_path: Union[ZarrGroup, H5Group],
+    max_loaded_sparse_elems: int,
+    axis: Literal[0, 1] = 0,
+    reindexers: Reindexer = None,
+    fill_value=None,
+):
+    elems = _gen_slice_to_append(
+        datasets, reindexers, max_loaded_sparse_elems, axis, fill_value
+    )
+    init_elem = (csr_matrix, csc_matrix)[axis](next(elems))
```
ivirshup, line 213: Does this load the first element into memory if it's not already?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+        darrays = (da.from_array(a, chunks="auto") for a in arrays)
+    elif isinstance(init_elem, ZarrArray):
+        darrays = (da.from_zarr(a) for a in arrays)
+
+    res = da.concatenate(
+        _gen_elem_to_append(
+            darrays,
+        ),
+        axis=axis,
+    )
```
ivirshup, line 194: I think it's a bit easier to read inline. I don't love having the reindexers and the fill_values being closed over in a function that's called once.

```suggestion

    darrays = None
    init_elem = arrays[0]

    if isinstance(init_elem, H5Array):
        darrays = (da.from_array(a, chunks="auto") for a in arrays)
    elif isinstance(init_elem, ZarrArray):
        darrays = (da.from_zarr(a) for a in arrays)

    res = da.concatenate(
        [ri(a, axis=1 - axis, fill_value=fill_value) for a, ri in zip(darrays, reindexers)],
        axis=axis,
    )
```

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+    elems = _gen_slice_to_append(
+        datasets, reindexers, max_loaded_sparse_elems, axis, fill_value
+    )
+    init_elem = (csr_matrix, csc_matrix)[axis](next(elems))
+    write_elem(output_group, output_path, init_elem)
+    del init_elem
+    out_dataset: SparseDataset = read_as_backed(output_group[output_path])
+    for temp_elem in elems:
+        out_dataset.append((csr_matrix, csc_matrix)[axis](temp_elem))
+        del temp_elem
```
ivirshup, line 219: For special casing when no indexing is needed, this function should end up running:

```python
arrays: list[SparseDataset]  # Assuming they are all CSR, and we are concatenating along rows

write_elem(output_group, ouput_path, arrays[0])
target = read_as_backed(output_group[output_path])
for a in arrays[1:]:
    target.append(a)
```

This is what I meant by special casing when no reindexing needs to occur for sparse arrays. This way we can fill in the stored array by storage chunks.

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+        concat_annot[label] = label_col
+    write_elem(output_group, dim, concat_annot)
+
+
+def concat_on_disk(
+    in_files: Union[
+        Collection[Union[str, os.PathLike]],
+        typing.MutableMapping,
+    ],
+    out_file: Union[str, os.PathLike, typing.MutableMapping],
```
ivirshup, line 391: Why mutable mapping for outfile? Should it be more like: `ZarrGroup | H5Group`?

anndata/experimental/__init__.py
```diff
@@ -17,6 +18,7 @@
     "read_dispatched",
     "write_dispatched",
     "IOSpec",
+    "concat_on_disk",
```
ivirshup, line 21: This needs to be imported

ivirshup: And added to the documentation

ivirshup (commented):
Looking good! Mostly minor.

Have you done any benchmarking with this? E.g showing lower peak memory?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,602 @@
...
+    """
+    Read the group until
+    SparseDataset, Array or EAGER_TYPES are encountered.
+    """
+
+    def callback(func, elem_name: str, elem, iospec):
+        if iospec.encoding_type in SPARSE_MATRIX:
+            return SparseDataset(elem)
+        elif iospec.encoding_type in EAGER_TYPES:
+            return read_elem(elem)
```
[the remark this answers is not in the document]
syelman: It reads the dataframes so it would. Wdym by nested types? I have a special case for dict. Could you give some examples?

anndata/experimental/merge.py
syelman: The docs say they already use it but I will put it explicitly anyways.

from the docs: https://docs.dask.org/en/stable/generated/dask.array.from_array.html
```
If your underlying datastore has a .chunks attribute (as h5py and zarr datasets do) then a multiple of that chunk shape will be used if you do not provide a chunk shape.
```

anndata/experimental/merge.py
syelman: Yes, but they will be in memory; the problem is when I take a slice of the sparse dataset (i.e., read it partially), it sometimes returns it as other types of sparse format even if the SparseDataset itself has a determined format. I already checked if SparseDataset's are in the correct format and if they all have the same format. You can try to see this

```python
def write_concat_sparse(
    datasets: Sequence[SparseDataset],
    output_group: Union[ZarrGroup, H5Group],
    output_path: Union[ZarrGroup, H5Group],
    max_loaded_sparse_elems: int,
    axis: Literal[0, 1] = 0,
    reindexers: Reindexer = None,
    fill_value=None,
):
    elems = _gen_slice_to_append(
        datasets, reindexers, max_loaded_sparse_elems, axis, fill_value
    )
    init_elem = next(elems)
    write_elem(output_group, output_path, init_elem)
    del init_elem
    out_dataset: SparseDataset = read_as_backed(output_group[output_path])
    for temp_elem in elems:
        out_dataset.append(temp_elem)
        del temp_elem
```

I am still not sure why this happens I just used that as a workaround

anndata/experimental/merge.py
syelman: I assumed it would cover both because I saw such annotations for the specs library anyways if I am removing it if it doesn't fit.

anndata/experimental/__init__.py
syelman: Leads to a circular import because I use read_dispatch. Should I use some wrapper function etc?

anndata/experimental/merge.py
syelman: I got it now, I added support for it

anndata/tests/test_concatenate_disk.py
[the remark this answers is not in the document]
syelman: I was worried that is was going to look cumbersome but I found a way to deal with it
```python
write_elem(as_group(p, mode='a'), '', a)
```
Which is good I suppose to test `as_group` too

anndata/experimental/merge.py
```diff
@@ -0,0 +1,604 @@
...
+    MissingVal,
+    Reindexer,
+)
+from .._core.sparse_dataset import SparseDataset
+from .._core.file_backing import to_memory
+from . import read_dispatched
+from ..compat import ZarrGroup, ZarrArray, H5Group, H5Array
+
+
+MAX_LOAD_SIZE = 60_000_000
```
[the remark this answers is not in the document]
syelman: I added it as an argument

anndata/experimental/merge.py
ivirshup: Could we not have a branch then, and just do `ad.from_array(a, chunks="auto")` for both?

anndata/experimental/__init__.py
ivirshup: As discussed, could you move the `read_dispatched`, `write_dispatched` function into their own file so you can avoid the circular import?

anndata/experimental/merge.py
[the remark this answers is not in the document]
syelman: I am not sure if I understand read_dispatch very good. I replaced `read_elem(elem)` with `func(elem)` and it gives errors, shouldn't they be the same? I will look into it once I got other things figured

syelman:
@ivirshup 
> Looking good! Mostly minor.
> 
> Have you done any benchmarking with this? E.g showing lower peak memory?

Yes I just updated the branch [concat-on-disk-benchmark](https://github.com/syelman/anndata/tree/concat-on-disk-benchmark)

I am trying to do a setting for limiting memory usage for dask, but it is filling my tmp directory currently.

anndata/experimental/merge.py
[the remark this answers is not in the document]
syelman: Update: I checked and the documentation also uses it this way: https://anndata.readthedocs.io/en/latest/tutorials/notebooks/%7Bread%2Cwrite%7D_dispatched.html.

ivirshup:
> Yes I just updated the branch [concat-on-disk-benchmark](https://github.com/syelman/anndata/tree/concat-on-disk-benchmark)

I'm not sure I see what you're talking about here on that branch

ivirshup (approved):

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    concat_annot.index = concat_indices
+    if label is not None:
+        concat_annot[label] = label_col
+    write_elem(output_group, dim, concat_annot)
+
+
+def concat_on_disk(
+    in_files: Union[
+        Collection[Union[str, os.PathLike]],
+        typing.MutableMapping,
```
flying-sheep, line 402: Mapping from which type to which type?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    ],
+    out_file: Union[str, os.PathLike],
+    overwrite: bool = False,
+    max_loaded_sparse_elems: int = 100_000_000,
+    *,
+    axis: Literal[0, 1] = 0,
+    join: Literal["inner", "outer"] = "inner",
+    merge: Union[StrategiesLiteral, Callable, None] = None,
+    uns_merge: Union[StrategiesLiteral, Callable, None] = None,
```
flying-sheep, line 411: Callables with which arguments and which return type?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    label: Optional[str] = None,
+    keys: Optional[Collection] = None,
```
flying-sheep, line 413: Collection of which item type?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    index_unique: Optional[str] = None,
+    fill_value: Optional[Any] = None,
+    pairwise: bool = False,
+):
```
flying-sheep, line 417: Return type?

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+        * `"same"`: Elements that are the same in each of the objects.
+        * `"unique"`: Elements for which there is only one possible value.
+        * `"first"`: The first element seen at each from each position.
+        * `"only"`: Elements that show up in only one of the objects.
+    uns_merge
+        How the elements of `.uns` are selected. Uses the same set of strategies as
+        the `merge` argument, except applied recursively.
+    label
+        Column in axis annotation (i.e. `.obs` or `.var`) to place batch information in.
+        If it's None, no column is added.
```
flying-sheep, line 472: Is the “batch” the same thing as the values of the `keys` argument? If so that should be mentioned both in the `in_files` description and the `keys` description.

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    write_elem(output_group, dim, concat_annot)
+
+
+def concat_on_disk(
+    in_files: Union[
+        Collection[Union[str, os.PathLike]],
+        typing.MutableMapping,
+    ],
+    out_file: Union[str, os.PathLike],
+    overwrite: bool = False,
```
flying-sheep, line 405: bool arguments should only be positional when they’re the only poitional argument, so the `*` should be moved before this parameter.

maybe the name `max_loaded_sparse_elems` should be shortened so it can be more easily specified by name too then.

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    to the concatenation of the loaded AnnData objects using
+    the `concat` function.
+
+    To adjust the maximum amount of data loaded in memory; for sparse
+    arrays use the max_loaded_sparse_elems argument; for dense arrays
+    see the Dask documentation, as the Dask concatenation function is used
+    to concatenate dense arrays in this function
+
+
+
```
flying-sheep, line 435:
```suggestion

```

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    # Argument normalization
+    if pairwise:
+        raise NotImplementedError("pairwise concatenation not yet implemented")
+    if join != "inner":
+        raise NotImplementedError("only inner join is currently supported")
+
+    merge = resolve_merge_strategy(merge)
+    uns_merge = resolve_merge_strategy(uns_merge)
+    if len(in_files) <= 1:
+        raise ValueError("Must pass at least two files to concatenate.")
```
flying-sheep, line 507: Is there a reason to not just do sth. like this here? (needs `shutil` import)

```suggestion
    if len(in_files) <= 1:
        if len(in_files) == 1:
            if not overwrite and Path(out_file).is_file():
                raise FileExistsError(f"File “{out_file}” already exists and `overwrite` is set to False")
            shutil.copy2(in_files[0], out_file)
        return
```

I imagine someone might use this on collections of files of varying length.

anndata/experimental/merge.py
```diff
@@ -0,0 +1,615 @@
...
+    Params
+    ------
+    in_files
+        The corresponding stores or paths of AnnData objects to
+        be concatenated. If a Mapping is passed, keys are used for the `keys`
+        argument and values are concatenated.
+    out_file
+        The target path or store to write the result in.
+    overwrite
+        If False the and if a file already exists it will raise an error,
```
flying-sheep, line 445:
```suggestion
        If `False` while a file already exists it will raise an error,
```

flying-sheep (changes requested):
Love it! I looked at its surface (docs, API) there’s a few minor improvements in typing and docs I’d like to see and one possible change in API. Maybe 15 minutes of work.

flying-sheep (approved):
perfect!

Issue #956, opened 2023-03-17
URGENT: Anndata write is broken.


Koncopd:
@ivirshup 
I am on main, i also observed the same problem in the latest release.
Reproduce:
```
import scanpy as sc
pbmc68k = sc.datasets.pbmc68k_reduced()
pbmc68k.write("test.h5ad")
sc.read("test.h5ad")
```
And the error is
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~\Apps\Miniconda3\envs\nbproject\lib\site-packages\anndata\_io\utils.py:202, in report_read_key_on_error.<locals>.func_wrapper(*args, **kwargs)
    201 try:
--> 202     return func(*args, **kwargs)
    203 except Exception as e:

File ~\Apps\Miniconda3\envs\nbproject\lib\site-packages\anndata\_io\specs\registry.py:235, in Reader.read_elem(self, elem, modifiers)
    234 if self.callback is not None:
--> 235     return self.callback(read_func, elem.name, elem, iospec=get_spec(elem))
    236 else:

File ~\Apps\Miniconda3\envs\nbproject\lib\site-packages\anndata\_io\h5ad.py:241, in read_h5ad.<locals>.callback(func, elem_name, elem, iospec)
    240     return read_dataframe(elem)
--> 241 return func(elem)

File ~\Apps\Miniconda3\envs\nbproject\lib\site-packages\anndata\_io\specs\methods.py:590, in read_dataframe(elem, _reader)
    588 idx_key = _read_attr(elem.attrs, "_index")
    589 df = pd.DataFrame(
--> 590     {k: _reader.read_elem(elem[k]) for k in columns},
    591     index=_reader.read_elem(elem[idx_key]),
    592     columns=list(columns),
    593 )
    594 if idx_key != "_index":

File ~\Apps\Miniconda3\envs\nbproject\lib\site-packages\anndata\_io\specs\methods.py:590, in <dictcomp>(.0)
    588 idx_key = _read_attr(elem.attrs, "_index")
    589 df = pd.DataFrame(
--> 590     {k: _reader.read_elem(elem[k]) for k in columns},
    591     index=_reader.read_elem(elem[idx_key]),
    592     columns=list(columns),
    593 )
    594 if idx_key != "_index":

File h5py\_objects.pyx:54, in h5py._objects.with_phil.wrapper()

File h5py\_objects.pyx:55, in h5py._objects.with_phil.wrapper()

[...]

PR #957, opened 2023-03-17
Fix anndata write on Windows

Koncopd:
https://github.com/scverse/anndata/issues/956

### DIFF
--- a/anndata/_io/specs/registry.py
+++ b/anndata/_io/specs/registry.py
@@ -278,7 +278,7 @@ class Writer:
 
         # Normalize k to abosulte path
         if not PurePath(k).is_absolute():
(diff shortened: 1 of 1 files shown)


anndata/_io/specs/registry.py
```diff
@@ -278,7 +278,7 @@ def write_elem(
 
         # Normalize k to abosulte path
         if not PurePath(k).is_absolute():
-            k = str(PurePath(store.name) / k)
+            k = (PurePath(store.name) / k).as_posix()
```
flying-sheep, line 281: This should probably just be `PurePosixPath`

Koncopd: Thanks, this is indeed more elegant.

PR #962, opened 2023-03-21
Use mmread from fast_matrix_market if it's installed

grst:
I tried out https://pypi.org/project/fast-matrix-market and it works really well and is by 1-2 orders or magnitude faster on big matrices. It would be nice to support it as an optional dependency. 

WDYT @ivirshup ?

### DIFF
--- a/anndata/_io/read.py
+++ b/anndata/_io/read.py
@@ -310,7 +310,10 @@ def read_mtx(filename: PathLike, dtype: str = "float32") -> AnnData:
     dtype
         Numpy data type.
     """
-    from scipy.io import mmread
+    try:
+        from fast_matrix_market import mmread
+    except ImportError:
+        from scipy.io import mmread
 
     # could be rewritten accounting for dtype to be more performant
     X = mmread(fspath(filename)).astype(dtype)


grst:
Benchmark of all methods on a slightly larger file: 

```
!du -hs {path}
3.3G	processed/SCP1884/expression/62a7ba8b81f3f4722a7ba4e0/TI_EPI.scp.matrix.mtx
```
```
%%time
scipy.io.mmread(
    path
)
CPU times: user 5min 40s, sys: 4.64 s, total: 5min 45s
Wall time: 5min 44s
```
```
%%time
# I had to explicitly set the number of threads to see any effect
pa.set_io_thread_count(8)
pa.set_cpu_count(8)
pa.csv.read_csv(
    path,
    read_options=pa.csv.ReadOptions(
        skip_rows=2, use_threads=True, column_names=["row", "col", "data"]
    ),
    parse_options=pa.csv.ParseOptions(delimiter=" "),
)
CPU times: user 28.3 s, sys: 11.5 s, total: 39.7 s
Wall time: 4.73 s
```
```
%%time
pd.read_table(
    path,
    skiprows=2,
    sep=" ",
)
CPU times: user 41.5 s, sys: 7.51 s, total: 49 s
Wall time: 49 s
```
```
%%time
fast_matrix_market = mmread(path, parallelism=8)
CPU times: user 25.8 s, sys: 3.55 s, total: 29.3 s
Wall time: 4.48 s
```

In summary: 
 * pandas is already a huge gain over scipy.mmread
 * pyarrow and fast_matrix_market are comparable in terms of speed

ivirshup:
>  had to explicitly set the number of threads to see any effect

Interesting. I don't think I did, but it looked like I was using multiple threads. Maybe an OS difference?

I'd personally lean towards using pyarrow here since I trust it's stability more. I believe pandas also exposes the pyarrow csv reader through the `engine` kwarg, so maybe we could get a speedup in all cases with with minimal code.

What do you think?

grst:
> Interesting. I don't think I did, but it looked like I was using multiple threads. Maybe an OS difference?

On our system the `OMP_NUM_THREADS` env var is set to 1. This affects pyarrow as well. I initially tried to use
`threadpoolctl` to set a higher CPU count, but that had no effect. Setting the env var before importing pyarrow, or using `pa.set_cpu_count` worked. 

Do we expose parallelism through a `n_jobs` parameter, or just leave it to the user to deal with pyarrow/OMP?
I think it would be fine to leave it, but maybe mention it in the docstring. 

> I believe pandas also exposes the pyarrow csv reader through the engine kwarg, so maybe we could get a speedup in all cases with with minimal code.

Sounds good to me. Here's the test: 
```
%%time
pd.read_csv(
    path,
    skiprows=2,
    sep=" ",
    engine='pyarrow',
    names=["row", "col", "data"]
)
CPU times: user 33.1 s, sys: 15.5 s, total: 48.6 s
Wall time: 6.73 s
```

How do we enable pyarrow? Use it by default if it's installed? A function parameter? Or a global setting?

ivirshup:
> Do we expose parallelism through a n_jobs parameter, or just leave it to the user to deal with pyarrow/OMP?

I think we leave it to the user. I've opened an issue on threadpoolctl about this though

* https://github.com/joblib/threadpoolctl/issues/134

> How do we enable pyarrow? Use it by default if it's installed? A function parameter?

I think we need a way to explicitly test both cases, which can be a keyword argument that defaults to `pyarrow` if it's installed.

-----------

> threadpoolctl to set a higher CPU count, but that had no effect. Setting the env var before importing pyarrow, or using pa.set_cpu_count worked.

[Arrow has two thread pools, one for IO and one for compute](https://arrow.apache.org/docs/cpp/threading.html#cpu-vs-i-o). Interestingly, it looks like the csv reading speed was only effected by the compute thread pool for me. But I was only looking at small, compressed files.

grst:
It might be not as trivial as it seems. For instance, this is an example of a perfectly valid `mtx` file: 

image

Source: https://math.nist.gov/MatrixMarket/formats.html (more specifically this [postscript file](https://math.nist.gov/MatrixMarket/reports/MMformat.ps)!)

Already the variable number of comments poses a problem to `pyarrow`, which doesn't support the `comment="%"` parameter. Not to speak of empty lines and variable number of spaces. 

`fast_matrix_market.mmread` and `scipy.io.mmread` handle this just fine.

ivirshup:
I think the lines starting with `%` is manageable. The `expression_atlas` function in `scanpy` has some code for this I believe.

While one could write a file like that example, have you ever seen one in the wild? I'm wondering if we can use pyarrow, and fallback to `scipy` with a warning if the file has a weird format. Then we can get good performance on the vast majority of cases.

grst:
> I think the lines starting with % is manageable. The expression_atlas function in scanpy has some code for this I believe.

yeah, that could work. I'm just afraid this is all getting overly complex now. 

How would you solve the fallback to `scipy.mmread`? Just catch any exception and try again?

ivirshup:
>  I'm just afraid this is all getting overly complex now.

Maybe, but I'm more worried about what happens if fast_memory_market stops uploading binaries or introduces a bug.

> fast_matrix_market.mmread and scipy.io.mmread handle this just fine.

It looks like there are still some cases where they differ:

* https://github.com/alugowski/fast_matrix_market/issues/10

(I think this is the same example that you pointed to, but with a new line?)

> How would you solve the fallback to scipy.mmread? Just catch any exception and try again?

I think so, including validation on number of entries and shape of array

grst:
Maybe we anyway should address this more holistically? At some point we discussed about making a separate IO package... we could bring that up at the next community meeting.

alugowski:
Author of fast_matrix_market here.

> While one could write a file like that example, have you ever seen one in the wild?

Agreed that's a pedantic example not seen in the wild, hence why it wasn't supported until the latest release. 

>  I'm just afraid this is all getting overly complex now.

You're just getting started :) There's also:
* dense matrices (no index columns)
* pattern matrices (no values)
* complex values (two value columns)
* symmetric matrices (3 different types of symmetries supported, and `scipy.io.mmwrite()` *does* emit these if it can)
* integer/real values (always using float works in most cases, subtle bug in others)

You could use `scipy.io.mminfo` to read the header and know what to expect then read the body with pyarrow. But there's a lot of cases to handle. In general you can have 1, 2, 3, or 4 columns (column count does not uniquely identify matrix type) and yes all are widely used out in the wild.

> Maybe, but I'm more worried about what happens if fast_memory_market stops uploading binaries or introduces a bug.

Valid concern. A big reason why fmm targets a direct scipy replacement is so you're free to switch back and forth with zero cost, exactly as in this PR. Any difference is a priority bug, including stuff like exception types on esoteric errors. Use the package for speed, or just don't install it if you don't trust it and everything still works.

We do test against the SciPy test suite: https://github.com/alugowski/fast_matrix_market/blob/main/python/testsuite_scipy/test_scipy_suite.py

There's also that SciPy uses ints for indices, so on some platforms you're limited to 32-bit matrix dimensions. FMM auto switches to 64-bit if the matrix dimensions are too big for 32-bit. Though it sounds like you're largely targeting platforms that have 64-bit ints and that's not an issue.

grst:
Hi @alugowski, 

thanks for reaching out! Are there any plans trying to integrate `fast_matrix_market` into scipy directly?

alugowski:
> Hi @alugowski,
> 
> thanks for reaching out! Are there any plans trying to integrate `fast_matrix_market` into scipy directly?

Not at the moment.

ivirshup:
@alugowski, I asked the scipy maintainers if they would in principle be up for vendoring / bundling `fast_matrix_market`, and they're not opposed: https://discord.com/channels/786703927705862175/1020054666433802351/1095285619300765787

> dense matrices (no index columns)
> pattern matrices (no values)
> complex values (two value columns)
> symmetric matrices (3 different types of symmetries supported, and scipy.io.mmwrite() does emit these if it can)
> integer/real values (always using float works in most cases, subtle bug in others)

Btw, I have been discussing a new format for sparse matrices based on array storage libraries with the graph blas people over at [GraphBLAS/binsparse-specification](https://github.com/GraphBLAS/binsparse-specification). The idea there is to move on from text based formats. Notably, there was a decision to not deal with dense or symmetric matrices with this specification. Would be interesting to hear your thoughts on this!

alugowski:
> @alugowski, I asked the scipy maintainers if they would in principle be up for vendoring / bundling `fast_matrix_market`, and they're not opposed: https://discord.com/channels/786703927705862175/1020054666433802351/1095285619300765787
> 
> I would be very happy with this as a solution, if it's something you'd be up for. Maybe we can follow up in an issue on your repo?

Yeah, let's follow up there. Would you mind opening an issue?

> 
> > dense matrices (no index columns)
> > pattern matrices (no values)
> > complex values (two value columns)
> > symmetric matrices (3 different types of symmetries supported, and scipy.io.mmwrite() does emit these if it can)
> > integer/real values (always using float works in most cases, subtle bug in others)
> 
> Btw, I have been discussing a new format for sparse matrices based on array storage libraries with the graph blas people over at [GraphBLAS/binsparse-specification](https://github.com/GraphBLAS/binsparse-specification). The idea there is to move on from text based formats. Notably, there was a decision to not deal with dense or symmetric matrices with this specification. Would be interesting to hear your thoughts on this!

Is it alright if I contact you via email?

grst:
Closing as this will be in scipy soon-ish: https://github.com/scipy/scipy/pull/18631

PR #970, opened 2023-04-11
Make awkward arrays helper optional

flying-sheep:
The module `anndata.compat` has no member `awkward` if the package is not installed


flying-sheep:
@meeseeksdev backport to 0.9.x

PR #980, opened 2023-04-13
Dask Array Reindexing Issue Possible Fix

syelman:
Hi,

I am trying to implement concatenation on disk and I ran into an issue and this problem came up
```python
        if not is_dask_collection(value) and np.isnan(value).any():
            if issubclass(self.dtype.type, Integral):
>               raise ValueError("cannot convert float NaN to integer")

```

I made the changes in my commit and it worked.

Although this seems to work fine:
```python
import dask.array as da

a = da.zeros((4,2),chunks=(4,2))
a[slice(None, None, None),[False,False]] = np.inf
a.compute()
```
this fails
```python
import dask.array as da

a = da.zeros((4,2),chunks=(4,2))
a[slice(None, None, None),[False,False]] = None
a.compute()
```

And in this code I am sure we are giving it np.inf. I am not sure of the problem yet but this seems to work for my cases so far

Notify: @ivirshup


syelman:
Ok not very minimal but reproducible case: the checks of this commit should give where it fails: https://github.com/scverse/anndata/pull/955/commits/18ae5bb5f5202d546951830160e6a833c162bb6c

syelman:
Not very minimal reproducable case: I merged my work to this branch with the exception of the index fix so we can see where it fails

PR #983, opened 2023-04-16
lazy dataframes in `.obs` and `.var` with `backed="r"` mode

### LINKED ISSUE #981: lazy `.obs` and `.var` with `backed="r"` mode

gszep:
closes #981


gszep:
closes #981

gszep:
close #981

PR #986, opened 2023-04-21
Testing on views doesn't seem to work for Dask

syelman:
Since this unit test compares hashes but not content, it seems to fail. Meanwhile, I used assert_equal to show it works as intended.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
     -   id: black
 - repo: https://github.com/charliermarsh/ruff-pre-commit
   # Ruff version.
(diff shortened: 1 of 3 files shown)


ivirshup:
https://docs.dask.org/en/stable/changelog.html

Could this get a bit more investigation? It's not obvious to me if this is fine to change. Hopefully we are not propagating some change into the original dask array, which is what this test is meant to check

syelman:
The dask array hash changes after the first copy call for some reason. I opened this issue https://github.com/dask/dask/issues/10240.

The tests seem to be fine though if we do a dummy copy call before getting the hash

Issue #979, opened 2023-04-13
Can't write bool that's nulled by concatenation


syelman:
Hi,
I tried to write an adata which is a result of outer concat and I get the following error.
```
TypeError: Can't implicitly convert non-string objects to strings

Above error raised while writing key 'bool' of  to /
```

Reproducible code snippet:
```python
from anndata.tests.helpers import gen_adata
import pandas as pd
import anndata as ad


a = gen_adata((5, 5), obsm_types=(pd.DataFrame,))
b = gen_adata((3, 5), obsm_types=(pd.DataFrame,))

del b.obsm["df"]
    
# del a.obsm["df"]["bool"] # this works fine

adatas = ad.concat({"a": a, "b": b}, join="outer", label="group")

adatas.write("test.h5ad")
```

Is this a bug or should we not write nulled bools like this?

ivirshup:
I think we should create a pandas boolean column during creation, so that we can handle the null values.

syelman:
Solved with #992

PR #992, opened 2023-05-05
Bool numpy arrays now merge as pandas bool arrays

syelman:
Hi,

When the user gave bool numpy arrays with outer concatenation the output wasn't a pandas boolean array so it made it unwritable to disk.

For other numeric types this wasn't a problem since they can have NaN values but for bools this gave an error when trying to write the output, so I wrote a fix

Related issue: https://github.com/scverse/anndata/issues/979


ivirshup:
I made changes to unify handling of booleans and categoricals. Should this also handle integers?

syelman:
I think it shouldn't since it assigns NaN and it is in valid domain of integers probably because it didn't gave the error it but it gave for booleans.

flying-sheep (approved):
Looks good to me!

PR #996, opened 2023-06-02
Fix for columns dtype changes in pandas 2.0.1

### LINKED ISSUE #991: Fix compatibility with pandas 2.0.1
Pandas 2.0.1 has changed the behavior column dtypes again, which breaks a lot of our tests. I'm going to pin our version of pandas for now until the upstream issue is resolved:

* https://github.com/pandas-dev/pandas/issues/53100

flying-sheep:
In pandas 2.0.1, empty data frames have a `RangeIndex(start=0, end=0)` as `.columns` attribute. This PR changes our behavior:

- `obs` and `var` coerce the `.column` attribute of empty dataframes to be a string `Index`. I decided not to throw `ImplicitModificationWarning`s here since this empty data frames are so inconsequential.
- read functions for `IOSpec('dataframe', …)` now pass `columns=columns if len(columns) else None` instead of effectively `columns=list(columns)`, which makes `uns` entries that are data frames conform to default pandas behavior. 

Questions:

- [x] Is the implicit modification in `_set_dim_df` OK or are there cases where an empty data frame must remain untouched?
- [x] We don’t actually need all the `columns=EMPTY_IDX` in anndata.py, as the `_set_dim_df` method takes care of it. Should we do away with it?

Fixes #991

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -54,8 +54,6 @@ from ..compat import (
     _move_adj_mtx,
 )
 
-EMPTY_IDX = pd.Index([], dtype="O")
-
 
 class StorageType(Enum):
     Array = np.ndarray
@@ -112,7 +110,7 @@ def _gen_dataframe(anno, length, index_names):
     return pd.DataFrame(
         anno,
         index=pd.RangeIndex(0, length, name=None).astype(str),
-        columns=None if len(anno) else EMPTY_IDX,
+        columns=None if len(anno) else [],
     )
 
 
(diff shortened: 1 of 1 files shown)


Zethson:
Nice!
1. I can't think of a case where an empty DataFrame must be untouched.
2. If we don't need them I don't see why we'd keep them?

pyproject.toml
```diff
@@ -110,7 +110,7 @@ omit = [
 [tool.pytest.ini_options]
 addopts = "--doctest-modules"
 python_files = "test_*.py"
-testpaths = ["anndata", "docs/concatenation.rst"]
+testpaths = ["anndata/tests", "docs/concatenation.rst"]
```
ivirshup, line 113: @flying-sheep, this disabled ~8 doctests

ivirshup: From differences in `pytest --collect-only`

ivirshup: Fixed in #1004.

What were the intended effects of this change?

flying-sheep: The VS Code test runner doesn’t work, but doing this right is of course more important.

PR #999, opened 2023-06-07
Serialize Nones

### LINKED ISSUE #673: `.write` does not save `None` values
## Description

When saving an AnnData object to disk, keys of a dictionary whose value is `None` seem not to be saved.

```python
import numpy as np
import scanpy as sc
from anndata import AnnData

adata = AnnData(X=np.eye(3), uns={'key_1': 0, 'key_2': None})
adata.write('adata.h5ad')

_adata = sc.read('adata.h5ad')
```

gives

```pytb
>>> _adata
AnnData object with n_obs × n_vars = 3 × 3
    uns: 'key_1'
```

flying-sheep:
- [x] Fixes #673
- [x] Added tests
- [x] Added release notes 

[Rendered release notes](https://icb-anndata--999.com.readthedocs.build/en/999/release-notes/index.html)


ivirshup:
As discussed, I don't want to change the on-disk format in the next release. So I'd hold off on this until at least 0.11.

flying-sheep:
Since 0.10 is released, we can document and then merge this, right?

ivirshup:
I would like to have proposals for all disk format changes written up and discussed with other implementers before merging them to main.

flying-sheep:
Sure, we can use this PR as reference implementation.

Which other implementers are there? Is there a channel with the relevant people on some communication medium?

alam-shahul:
Hi there, I feel that this is an important feature/bug fix, I was wondering if there were plans to merge this anytime soon/if it's possible to help.

tests/test_readwrite.py
```diff
@@ -831,6 +831,31 @@ def test_adata_in_uns(tmp_path, diskfmt, roundtrip):
...
 
+@pytest.mark.parametrize(
+    "uns_val",
+    [
+        pytest.param(dict(base=None), id="dict_val"),
+        pytest.param(
+            pd.DataFrame(dict(col_0=["string", None])),
+            id="df",
+            marks=pytest.mark.xfail(
+                reason="Nullable string arrays not yet implemented"
```
ilan-gold, line 842: Is that true anymore? https://github.com/scverse/anndata/blob/4bcd98936395421c8759e0e30016483786cca2ef/src/anndata/_settings.py#L424-L430

src/anndata/_io/specs/methods.py
```diff
@@ -326,6 +326,30 @@ def write_raw(
...
+
+@_REGISTRY.register_read(H5Array, IOSpec("null", "0.1.0"))
+@_REGISTRY.register_read(ZarrArray, IOSpec("null", "0.1.0"))
+def read_null(_elem, _reader) -> None:
+    return None
+
+
+@_REGISTRY.register_write(H5Group, type(None), IOSpec("null", "0.1.0"))
+def write_null_h5py(f, k, _v, _writer, dataset_kwargs=MappingProxyType({})):
+    f.create_dataset(k, data=h5py.Empty("f"), **dataset_kwargs)
```
ilan-gold, line 342: I see in https://github.com/zarr-developers/zarr-specs/issues/318 you mention that `h5py.Empty` has no dtype but:

```python
>>> import h5py
>>> h5py.Empty("f")
Empty(dtype=dtype('float32'))
```
Am I missing something?

ilan-gold: This is inconsequential as far as I can tell but just wanted to make sure I understand.  It could also be relevant for the discussion

ilan-gold (commented):
Two things:

1. Should we put this feature behind a flag initially?  I know we had spoken about that as a way forward before.  However, in this case, I'm no really sure what we'd gain by doing that.
2. This PR is simple enough that it can be merged without much comment.  I think we should try to make it clear that this PR is planned for the 0.12 release and that we are looking for feedback (specifically from other languages).

PR #1001, opened 2023-06-12
Remove build hacks

flying-sheep:
Also switches from flit-core to hatchling, but at this point, that’s just a backend change.

It uses `setuptools-scm` (via the `hatch-vcs plugin`) as a build hook, so the version will ship in a generated `_version.py` file.


Koncopd:
Hi. 
What is this `hatch`? Is it something like `poetry`?
I would really prefer to keep flit as it is pretty mininal.

pyproject.toml
```diff
@@ -92,11 +87,15 @@ test = [
...
 exclude = [
-    'anndata/tests/test_*.py',
-    'anndata/tests/data',
+    "anndata/tests/test_*.py",
+    "anndata/tests/data",
 ]
+[tool.hatch.version]
+source = "vcs"
+[tool.hatch.build.hooks.vcs]
+version-file = "anndata/_version.py"
```
flying-sheep, line 98: This is all that’s added

anndata/_metadata.py
```diff
@@ -1,46 +0,0 @@
...
-    """\
-    Checks if we are being imported by flit.
-    This is necessary so flit can import __version__ without all depedencies installed.
-    There are a few options to make this hack unnecessary, see:
-    https://github.com/takluyver/flit/issues/253#issuecomment-737870438
-    """
-    for frame in traceback.extract_stack():
-        if frame.name == "get_docstring_and_version_via_import":
-            return True
-    return False
```
flying-sheep, line 46: this is all removed

pyproject.toml
```diff
@@ -18,12 +15,11 @@ authors = [
 ]
 maintainers = [
     {name = "Isaac Virshup", email = "ivirshup@gmail.com"},
-    {name = "Philipp Angerer", email = "philipp.angerer@helmholtz-muenchen.de"},
+    {name = "Philipp Angerer", email = "philipp.angerer@helmholtz-munich.de"},
     {name = "Alex Wolf", email = "f.alex.wolf@gmx.de"},
 ]
-readme = {file = "README.md", content-type="text/markdown"}
+readme = "README.md"
```
ivirshup, line 21: I recall this throwing errors if I didn't specify the content-type. Was that flit specific?

ivirshup:
It's more that I don't see how setuptools-scm is being used from the changes.

How does this not have the behavior described [here](https://github.com/ofek/hatch-vcs#editable-installs):

> Editable installs
> 
> The version file is only updated upon install or build. Thus the version number in an [editable install](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs) (Hatch's [dev mode](https://hatch.pypa.io/latest/config/build/#dev-mode)) will be incorrect if the version changes and the project is not rebuilt. An unsupported workaround for keeping the version number up-to-date can be found at [hatch-vcs-footgun-example](https://github.com/maresb/hatch-vcs-footgun-example).

flying-sheep:
To summarize our in-person discussion:

There’s no way to have the version in the package metadata automatically up-to-date. It can only be updated when (dev-)installing your package anew. There’s a discussion about adding dynamic version metadata [here](https://discuss.python.org/t/dynamic-versions-in-editable-installations/15220)

The `__version__` convention is of dubious value (packages usually look things up from package metadata using `importlib.metadata`), but of course it can be calculated using runtime code. That’s explained and demonstrated [here](https://github.com/maresb/hatch-vcs-footgun-example).

pyproject.toml
flying-sheep: not for many years: https://github.com/pypa/flit/pull/169

ivirshup: Pretty sure I was running into this more recently than that. But I believe `twine check` was the one complaining, so it's fine as long as the build check is passing.

ivirshup (commented):
Could we add a step to the build check where we check that the built package has a `__version__`?

.azure-pipelines.yml
```diff
@@ -101,3 +101,8 @@ jobs:
           python -m build --sdist --wheel .
           twine check dist/*
         displayName: "Build & Twine check"
+
+      - script: |
+          pip install dist/*.whl
+          python -c 'import anndata; import anndata._version; print(anndata.__version__)'
```
ivirshup, line 107:
```suggestion
          python -c 'import anndata; print(anndata.__version__)'
```

Zethson (approved):

PR #1006, opened 2023-06-15
Add issue templates

flying-sheep:
Also has a placeholder for version info so people know we care about anndata version info

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -19,12 +19,29 @@ body:
 - type: markdown
   attributes:
     value: |
-      **Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.
+      **Note**: Please read [this guide][] detailing how to provide the necessary information for us to reproduce your bug.
+
+      [this guide]: https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports
 - type: textarea
   id: Report
   attributes:
     label: Report
-    description: Describe the bug you encountered, and what you were trying to do. Please use [github markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) features for readability.
+    description: |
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,54 @@
...
+  validations:
+    required: true
+- type: textarea
+  id: traceback
+  attributes:
+    label: Error output
+    description: Produced by the code above. Leave empty if the issue is unexpected behavior instead of an error.
+    render: pytb
+  validations:
+    required: false
```
ivirshup, line 45:
```suggestion
- type: textarea
  id: Report
  attributes:
    label: Report
    description: Describe the bug you encountered, and what you were trying to do. Please use [github markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) features for readability.
  validations:
    required: true
```

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,54 @@
...
+- type: textarea
+  id: versions
+  attributes:
+    label: Versions
+    description: Which version of anndata and other related software you used.
```
ivirshup, line 50: Can we specifically request the output of `session_info` here?

ivirshup (commented):
Thanks!

Made a suggestion and requested a change.

I would like to keep the template for bugs more simple, just because I don't think the more structured approach works for many of the bugs I report.

.github/ISSUE_TEMPLATE/bug-report.yml
flying-sheep: In my experience, that leads to too unspecific reports. I’d rather have people write “nothing” in the code field in the rare case that it’s not about code but installation or so than have people forget to paste a stacktrace.

flying-sheep:
> My idea would look like this: [flying-sheep/playground/issues/new?assignees=&labels=Bug+%F0%9F%90%9B&projects=&template=anndata-bug-report.yml](https://github.com/flying-sheep/playground/issues/new?assignees=&labels=Bug+%F0%9F%90%9B&projects=&template=anndata-bug-report.yml)
> 
> I’d prefer that to your version. Can you live with that? 😉

flying-sheep:
@ivirshup said he’s fine with this

PR #1007, opened 2023-06-15
Add prettier pre-commit

flying-sheep:


flying-sheep:
pre-commit.ci autofix

flying-sheep:
Huh, the changes in 65ce47ee1c732ba86c407d39070dfb93929509a5 are broken. We need to figure out what does that

flying-sheep:
Ah, we need to use colon fences

ivirshup (commented):
Looks mostly good. Can we configure to not remove the trailing commas in json lists? (so long as that's valid)

ivirshup (approved):
Ah nvm, looks like trailing commas are only allowed by whatever json superset also allows comments in the asv config.

flying-sheep:
I figured it out, that bothered me as well: https://github.com/prettier/prettier/pull/10323

PR #1008, opened 2023-06-16
Allow reading and writing OrderedDicts

### LINKED ISSUE #796: Allow for more mappings in the I/O registry
Objects than contain for instance `OrderedDict` in `.uns` elements fail to be serialised since `anndata==0.8`:

```python
from collections import OrderedDict
import numpy as np
from jax import random
from anndata import AnnData

x = random.normal(key=random.PRNGKey(1), shape=(1000,500))
adata = AnnData(np.array(x))

adata.uns["ord_dict"] = OrderedDict({"key1": "value1"})
adata.write("anndata_with_ordered_dict.h5ad")
# => TypeError: No method has been defined for writing <class 'collections.OrderedDict'> elements to <class 'h5py._hl.group.Group'>
```

Currently users need to identify items in `.uns` that cause this error and convert them to `dict`.

While there's probably no reason to use `OrderedDict` in practice, it would be great to have support for it as well as for other dictionaries such as [`ConfigDict`](https://ml-collections.readthedocs.io/en/latest/config_dict.html).

flying-sheep:
Fixes #796

If we add more supported classes, we should probably make the code more generic

### DIFF
--- a/anndata/_io/specs/methods.py
+++ b/anndata/_io/specs/methods.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from collections import OrderedDict
 
 from os import PathLike
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


ivirshup (changes requested):
I don't think we should add this feature. 

Conversation happening here: https://github.com/scverse/anndata/issues/796#issuecomment-1594619549

ivirshup:
Okay to close @flying-sheep?

Issue #1011, opened 2023-06-19
Show a warning when updating `obsm` on a subset of rows


quentinblampey:
Hello,

I want to update values of an `obsm` array at a specific location (using `obs_name` labels, not row indices).
The following naive approach doesn't work (which is normal, since it updates a view), but I think that it should provide a warning:
```python
>>> import anndata
>>> import numpy as np

>>> adata = anndata.AnnData(X=np.ones((3, 4)), obsm={"key": np.ones((3, 2))})

>>> adata[["1", "2"]].obsm["key"] = np.zeros((2, 2))
>>> adata[["1", "2"]].obsm["key"]
ArrayView([[1., 1.],
           [1., 1.]])
```
In the other hand, when doing something similar with `adata.obs`, it provides a warning. But anyway, with `adata.obs`, we can simply use `adata.obs.loc[["1", "2"], "key"] = ...` ; this is something we can't do with obsm when it's a `np.ndarray`.

Maybe I missed an easy way to be able to do what I want? The two solutions I currently see are the following:
- use a DataFrame with`index=adata.obs_names` instead of a `ndarray`: this way, I can do `adata.obsm["key"].loc[["1", "2"]] = ...`, but it requires updating all my current `h5ad` files
- get the row indices of my labels so that I can do `adata.obsm["key"][[1, 2]] = ...`, but I think this is ugly and error-prone, mostly when the indices are not just numeric values formatted as strings

In brief, did I miss something? And, if I didn't, could the above code show a warning?
> NB: I can contribute to this if you believe this is a relevant enhancement

flying-sheep:
We actualize views when a user updates it, so this is a bug. `.obsm` on a AnnData view is a `AxisArraysView`, which inherits from `AlignedViewMixin`, which means this code should run when you do what you do `adata[["1", "2"]].obsm["key"] = ...`:

https://github.com/scverse/anndata/blob/7a9856db56e09508b93d29265ac0e122bb359b8c/anndata/_core/aligned_mapping.py#L147-L152

PR #1016, opened 2023-06-21
Fix view setitem warnings (alt)

ivirshup:
Fixes https://github.com/scverse/anndata/issues/1011

Supersedes https://github.com/scverse/anndata/pull/1015

Alternative to #1015

* Adds a warning for `delitem` as well as `setitem`
* Doesn't subclass from views code
* Improved tests (check that the right element is mentioned.
* Improved warnings (now reports which element was changed)

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -136,8 +136,6 @@ class AlignedViewMixin(_SetItemMixin):
     parent_mapping: Mapping[str, V]
     """The object this is a view of."""
 
-    _view_args: ElementRef
-
     is_view = True
 
     def __getitem__(self, key: str) -> V:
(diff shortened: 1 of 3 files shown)


flying-sheep:
I like the approach in https://github.com/scverse/anndata/pull/1015 better

ivirshup:
Difference in lines may be attributable to correcting the stack trace level, and covering an additional case. Opened #1026 to finish review.

flying-sheep (changes requested):
I’m Ok with this approach, if we reuse more code that exists:

https://github.com/ivirshup/anndata/pull/3

ivirshup:
I've slightly altered your approach. I would like to limit how ingrained the circular references are unless we can solve their issues.

Also, why did you label the initial PR for 0.9.2 when the issue is an enhancement not a bug fix?

flying-sheep:
good point, feel free to update the milestone

flying-sheep (approved):

PR #1026, opened 2023-06-26
Fix axisarray view set

ivirshup:
Re-opening #1015 to finish review (@flying-sheep)

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -276,8 +276,9 @@ class AxisArraysView(AlignedViewMixin, AxisArraysBase):
         self.subset_idx = subset_idx
         self._axis = parent_mapping._axis
 
+    @property
     def _view_args(self) -> ElementRef:
-        return ElementRef(self._parent, self._axis, (self.attrname,))
+        return ElementRef(self._parent, self.attrname, ())
 
 
 AxisArraysBase._view_class = AxisArraysView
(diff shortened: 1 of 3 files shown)


anndata/_core/aligned_mapping.py
```diff
@@ -277,6 +276,10 @@ def __init__(
         self.subset_idx = subset_idx
         self._axis = parent_mapping._axis
 
+    @property
```
ivirshup, line 279: This should probably be defined on the AlignedViewMixin instead of on two subclasses, right?

ivirshup (commented):
@flying-sheep

I'm not sure I love the approach. It re-uses code, but puts it through a more complicated code path. I'm not sure I understand the implications of all the code.

I'm also not sure I want to "enshrine" the code in views any more than it is already.

## Views

I'd also be interested in completley redoing how views work.

One point that only really effects the `AlignedMapping` code is: we don't necessarily need these to be view in the same way. We could allow adding and removing elements, without needing to actualize the parent object.

## warning message

I don't even know that we want to reuse the exact warning message. We should probably be using the key name in the warning for individual elements.

Also, I believe the stack level should be 1 level higher for this case than the case in _SetItemMixin.

## Delitem

This should also apply for deleting items.

## Change log

This PR would also need a change log entry.

# Alternative

PR #1016 addresses these things.

anndata/_core/aligned_mapping.py
flying-sheep: true, that’s even better!

flying-sheep:
`_SetItemMixin` exists for when someone tries to set an attribute on a view. It will:

1. throw a ImplicitModificationWarning
2. create a non-view copy of the anndata view object
3. replace the object
4. call `_init_as_actual`

That’s exactly what we want `AlignedViewMixin` to do, so this PR corrects the oversight of not having that one inherit from `_SetItemMixin` in the first place.

We can rename or rework the hierarchy, but that’s out of scope. This is a minimal PR to make the warning happen when it’s supposed to happen.

flying-sheep:
Closing in favor of #1026, I added a review there

PR #1029, opened 2023-06-28
Better error message from ad.concat with duplicated columns

### LINKED ISSUE #483: Non unique column names throw obscure error in concat
The following error is thrown as one AnnData object has two columns with the same name in `.obs`. It'd be great to have a better error message ("non-unique column names do not work with concat outer"):
image

ivirshup:
Fixes #483

### DIFF
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -250,12 +250,12 @@ def check_combinable_cols(cols: list[pd.Index], join: Literal["inner", "outer"])
 
     Looks for if there are duplicated column names that would show up in the result.
     """
+    repeated_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
     if join == "inner":
-        repeated_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
         intersecting_cols = intersect_keys(cols)
         problem_cols = repeated_cols.intersection(intersecting_cols)
     elif join == "outer":
-        problem_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
+        problem_cols = repeated_cols
     else:
         raise ValueError()
 
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -245,6 +245,29 @@ def try_unifying_dtype(col: list) -> pd.core.dtypes.base.ExtensionDtype | None:
         return None
 
 
+def check_combinable_cols(cols: list[pd.Index], join: Literal["inner", "outer"]):
+    """Given columns for a set of dataframes, checks if the can be combined.
+
+    Looks for if there are duplicated column names that would show up in the result.
+    """
+    if join == "inner":
+        repeated_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
+        intersecting_cols = intersect_keys(cols)
+        problem_cols = repeated_cols.intersection(intersecting_cols)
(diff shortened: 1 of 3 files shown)


anndata/_core/merge.py
```diff
@@ -245,6 +245,29 @@ def try_unifying_dtype(col: list) -> pd.core.dtypes.base.ExtensionDtype | None:
         return None
 
 
+def check_combinable_cols(cols: list[pd.Index], join: Literal["inner", "outer"]):
```
flying-sheep, line 248: Would be nice to get into the habit of using permissive types (would this really just work with lists?) and reduce the number of positional parameters in APIs.

Not important here, but a good habit for public APIs.

```suggestion
def check_combinable_cols(cols: Iterable[pd.Index], *, join: Literal["inner", "outer"]):
```

anndata/_core/merge.py
```diff
@@ -245,6 +245,29 @@ def try_unifying_dtype(col: list) -> pd.core.dtypes.base.ExtensionDtype | None:
...
+    Looks for if there are duplicated column names that would show up in the result.
+    """
+    if join == "inner":
+        repeated_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
+        intersecting_cols = intersect_keys(cols)
+        problem_cols = repeated_cols.intersection(intersecting_cols)
+    elif join == "outer":
+        problem_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
+    else:
+        raise ValueError()
```
flying-sheep, line 260: Please don’t duplicate complex chunks of code. I played “find the difference” for a few seconds because two identical variable assignments should mean “there’s a subtle difference between those lines”

```suggestion
    repeated_cols = reduce(lambda x, y: x.union(y[y.duplicated()]), cols, set())
    if join == "inner":
        intersecting_cols = intersect_keys(cols)
        problem_cols = repeated_cols.intersection(intersecting_cols)
    elif join == "outer":
        problem_cols = repeated_cols
    else:
        raise ValueError()
```

anndata/tests/test_concatenate.py
```diff
@@ -1433,3 +1433,20 @@ def test_outer_concat_outputs_nullable_bool_writable(tmp_path):
 
     adatas = concat({"a": a, "b": b}, join="outer", label="group")
     adatas.write(tmp_path / "test.h5ad")
+
+
+def test_concat_duplicated_columns(join_type):
+    # https://github.com/scverse/anndata/issues/483#issuecomment-1610579577
```
flying-sheep, line 1439: that links to the “stale” comment, did you mean this?

```suggestion
    # https://github.com/scverse/anndata/issues/483#issuecomment-749321435
```

flying-sheep (changes requested):
Looks good, only that comment link is wrong.

anndata/tests/test_concatenate.py
ivirshup: I think I actually meant this

```suggestion
    # https://github.com/scverse/anndata/issues/483
```

anndata/_core/merge.py
ivirshup: I think I'd leave this for now.

Fewer imports, and it is always a list. If type checking were on, I think I'd even want to know if this was receiving a more exotic type. Like, accidentally passing just a single index, or passing a dataframe, or passing a dict.

flying-sheep (approved):

PR #1040, opened 2023-07-06
Fix for awkward 2.3

### LINKED ISSUE #1035: Future changes to Awkward Array behavior class resolution
This is not an existing bug, but I'm raising it now so that we're on the same page.

Over at Awkward Array, we've come to the conclusion that the `__array__` mechanism, used to facilitate categorical and string arrays, should be a built-in feature, rather than implemented using Awkward behaviors. We were planning upon introducing a new parameter that mirrors `__record__` to indicate the nominal type of an array. This change would mean `__array__` no longer affects the loaded class (`type(arr)`) of an Awkward Array. Anndata uses the `__array__` parameter to wrap Awkward Arrays in [custom view wrappers](https://github.com/scverse/anndata/blob/b2965fc407122bdf41d6fa8b0335ce8eecc15c78/anndata/_core/views.py#L239-L257).

The mechanism that we were originally thinking of was a new parameter `__list__` that can _only_ attach to list-types (arrays with >1 dimension). Clearly, this is less general than the existing functionality, so I was wondering if you'd be able to clarify the kinds of Awkward Arrays that anndata needs to support.

Before we get into the details, the most important point is that we _do not_ want to make existing use cases impossible. So, I'm hoping that we can figure out what anndata needs, and rework our proposal.

When we were conceiving of these changes, we did not have any examples of custom arrays in the forefront of our minds. We did know of the use of `__array__` in anndata, but it certainly slipped my mind during these discussions. Pinging @jpivarski for visiblity.

ivirshup:
TODO

- [x] Closes #1035 (or at least addresses it)
- [x] Tests ~~added~~ refactored
- [x] Release note added
- [x] Bump awkward array version or have backward compatibility
- [ ] Figure out right way to fix this, e.g. `with_parameter`, `with_name`

### DIFF
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -255,7 +255,7 @@ try:
             array = self
             # makes a shallow copy and removes the reference to the original AnnData object
             array = ak.with_parameter(self, _PARAM_NAME, None)
-            array = ak.with_parameter(array, "__array__", None)
+            array = ak.with_parameter(array, "__list__", None)
             return array
 
     @as_view.register(AwkArray)
@@ -273,7 +273,7 @@ try:
                 "Please open an issue in the AnnData repo and describe your use-case."
             )
         array = ak.with_parameter(array, _PARAM_NAME, (parent_key, attrname, keys))
-        array = ak.with_parameter(array, "__array__", "AwkwardArrayView")
+        array = ak.with_parameter(array, "__list__", "AwkwardArrayView")
         return array
 
     ak.behavior["AwkwardArrayView"] = AwkwardArrayView
--- a/anndata/tests/test_awkward.py
+++ b/anndata/tests/test_awkward.py
@@ -147,8 +147,20 @@ def test_view(key):
 def test_view_of_awkward_array_with_custom_behavior():
     """Currently can't create view of arrays with custom __name__ (in this case "string")
     See https://github.com/scverse/anndata/pull/647#discussion_r963494798_"""
+
+    from uuid import uuid4
+
+    BEHAVIOUR_ID = str(uuid4())
+
+    class ReversibleArray(ak.Array):
+        def reversed(self):
+            return self[..., ::-1]
+
+    ak.behavior[BEHAVIOUR_ID] = ReversibleArray
     adata = gen_adata((3, 3), varm_types=(), obsm_types=(), layers_types=())
-    adata.obsm["awk_string"] = ak.Array(["AAA", "BBB", "CCC"])
+    adata.obsm["awk_string"] = ak.with_parameter(
+        ak.Array(["AAA", "BBB", "CCC"]), "__list__", BEHAVIOUR_ID
+    )
     adata_view = adata[:2]
 
     with pytest.raises(NotImplementedError):
--- a/docs/release-notes/0.9.2.md
+++ b/docs/release-notes/0.9.2.md
@@ -2,3 +2,5 @@
 
 ```{rubric} Bugfix
 ```
+
+* Views of `awkward.Array`s now work with `awkward>=2.3` {pr}`1040` {user}`ivirshup`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ test = [
     "boltons",
     "scanpy",
     "dask[array]",
-    "awkward>=2.0.8,<2.3",
+    "awkward>=2.3",
     "pytest_memray",
 ]
 


anndata/_core/views.py
```diff
@@ -273,7 +273,7 @@ def as_view_awkarray(array, view_args):
                 "Please open an issue in the AnnData repo and describe your use-case."
             )
         array = ak.with_parameter(array, _PARAM_NAME, (parent_key, attrname, keys))
-        array = ak.with_parameter(array, "__array__", "AwkwardArrayView")
+        array = ak.with_parameter(array, "__list__", "AwkwardArrayView")
         return array
 
     ak.behavior["AwkwardArrayView"] = AwkwardArrayView
```
agoose77, line 279:
```suggestion
    ak.behavior["AwkwardArrayView"] = AwkwardArrayView
    ak.behavior["*", "AwkwardArrayView"] = AwkwardArrayView
```

anndata/_core/views.py
```diff
@@ -273,7 +273,7 @@ def as_view_awkarray(array, view_args):
                 "Please open an issue in the AnnData repo and describe your use-case."
             )
         array = ak.with_parameter(array, _PARAM_NAME, (parent_key, attrname, keys))
-        array = ak.with_parameter(array, "__array__", "AwkwardArrayView")
+        array = ak.with_parameter(array, "__list__", "AwkwardArrayView")
```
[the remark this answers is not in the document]
ivirshup: I'm not sure if we need to bother applying this recursively, as we can only mutate the top level array, right?

I've given an example of this an asked a little more here: https://github.com/scverse/anndata/pull/1040#discussion_r1254529281

anndata/_core/views.py
agoose77: > (a) Would `ak.with_parameter(array, "__list__", None)`

No, the same function `with_list` should be used, with a value of `None`.

> (b) What behavior would this change allow? E.g. does __setitem__ even work for the children of this array?

The main change in Awkward is to replace the powerful `__array__` with a much tighter-scoped `__list__` parameter. This means that if your top-level layout is not an `is_list` type, it won't do anything (and in future, would error). The `with_list` function behaves like `with_name`, in that it descends until if finds a list, sets the parameter, and exits. 

Now that I think about it, the most comprehensive solution would set _either_ `__list__` or `__record__`, whichever appears first. That would look like
```python
def with_any_name(array, name):
    def apply(layout, **kwargs):
        if layout.is_list:
            return layout.with_parameter("__list__", name)
        elif layout.is_record:
            return layout.with_parameter("__record__", name)
    return ak.transform(apply, array)
```

agoose77: Hmm, actually, I suppose that you are only worried about mutability from fields (which is the only kind Awkward Arrays support). Therefore, we should just use `with_name` here: if the layout doesn't have any records, it can't be mutated. So that's the proper solution after all.

ivirshup: > That said, I suppose that if you only worry about mutability from fields (which is the only kind Awkward Arrays support), then we should just use with_name here: if the layout doesn't have any records, it can't be mutated.

:+1:

Make sense to me. So, something like:

```python
ak.with_name(array, "AwkwardArrayView")
```

instead of:

```python
ak.with_parameter(array, "__list__", "AwkwardArrayView")
```

?

This does seem to be causing some test failures, but I haven't looked too deeply at that yet.

----

Btw, the current approach works even if there are no lists. E.g.:

```python
a = ad.AnnData(
    np.ones((3, 2)),
    obsm={
        "awk": ak.Array(
            [
                {"a": {"b": 1}},
                {"a": {"b": 2}},
                {"a": {"b": 3}}
            ]
        )
    }
)
```

Is that intentional?

ivirshup: So, if I do `with_name` I seem to need `ak.behavior["*", "AwkwardArrayView"] = AwkwardArrayView`

ivirshup: @agoose77, could you provide an example that would break given the current behavior?

agoose77: Gah, you're quicker than I can type!

> `ak.with_name(array, "AwkwardArrayView")`

Yes, this. Explicitly, what this means is that only arrays with records *somewhere* in the layout will be given your special behavior class. It so happens that only arrays with records can be mutated in-place, though, so that's exactly what we want.

I took a look at `views.py`. We still want to set the `_PARAM_NAME` parameter on the top-level node, because any operation that drops these parameters will involve losing the original `ak.Array` anyway.

Setting the record name means that users won't be able to set their own record names without losing the anndata view. I think that's actually OK - behavior classes are somewhat at odds with anndata's needs here, and I think it's fine for a user to need to write
```python
ak.with_name(anndata['my_vector'], "Vector3D")
```
to get the vector behavior on an anndata-stored Awkward Array. Pinging @jpivarski to have an opinion here, though.

If Jim's happy with this trade-off, then I think this should work. The only thing I also noticed; does the registry store a weakref to the `ak.Array` object? We should test for whether `ak.Array(array)` retains the link to the original view - it shouldn't in my view. Storing the original array in the parameters as a weakref would mean we could check that, and avoid triggering a copy-on-write in such circumstances.

ivirshup: > Setting the record name means that users won't be able to set their own record names without losing the anndata view. I think that's actually OK - behavior classes are somewhat at odds with anndata's needs here, and I think it's fine for a user to need to write

This was true before (I think?)

Would it be true with the `ak.with_parameter(array, "__list__", "AwkwardArrayView")` approach?

Playing around a bit:

```python
two = ak.Array([[{"x": 0.9, "y": 1}, {"x": 2, "y": 2.2}, {"x": 2.9, "y": 3}],
                [],
                [{"x": 3.9, "y": 4}, {"x": 5, "y": 5.5}],
                [{"x": 5.9, "y": 6}],
                [{"x": 6.9, "y": 7}, {"x": 8, "y": 8.8}, {"x": 8.9, "y": 9}]],
               with_name="point")

ak.with_name(two, "AwkwardArrayView").typestr
#  '5 * var * AwkwardArrayView[x: float64, y: float64]'

ak.with_parameter(two, "__list__", "AwkwardArrayView").typestr
# '5 * [var * point[x: float64, y: float64], parameters={"__list__": "AwkwardArrayView"}]'
```

I think the `with_parameter` result better fits my mental model of what we're trying to do.

agoose77: `__list__` is defined only for `is_list` types. Here's an example of how fragile that is if we only set the top-most parameter:
```
>>> array = ak.Array([[1,2,3], [4]])
>>> result = array.mask[[True, False]]
>>> ak.with_parameter(result, "__list__", "View").layout
<ByteMaskedArray valid_when='true' len='2'>
    <parameter name='__list__'>'View'</parameter>
    <mask><Index dtype='int8' len='2'>
        [1 0]
    </Index></mask>
    <content><ListOffsetArray len='2'>
        <offsets><Index dtype='int64' len='3'>
            [0 3 4]
        </Index></offsets>
        <content><NumpyArray dtype='int64' len='4'>[1 2 3 4]</NumpyArray></content>
    </ListOffsetArray></content>
</ByteMaskedArray>
```

This currently works, but it won't forever; `__list__` is defined only for `is_list` types (i.e. the `ListOffsetArray` here, not the `ByteMaskedArray`). We're just gradually introducing these rules, so the sanity checks will come next.

But, you can ensure that we only set the parameter on `__list__` nodes using the recursive transform above. The question is what else breaks?

If you support single-records in AnnData, i.e. `ak.record.Record` objects that wrap an array, then these are top-level nodes that do not have `is_list=True`, so these would not be supported. Records _can_ be mutated just like record _arrays_.

grst: Just catching up here...

> That said, I suppose that if you only worry about mutability from fields (which is the only kind Awkward Arrays support), then we should just use with_name here: if the layout doesn't have any records, it can't be mutated.

this makes sense to me. 

I understand the final discussion is whether to use `with_name`, or `with_parameter` here?
 * using `with_name` doesn't have any obvious downsides except that the `name` parameter is lost? In that case we'd probably want to update [this code](https://github.com/scverse/anndata/blob/98ba35999276d2bf15fd9b9e2d86f6df87dcc6ab/anndata/_core/views.py#L270-L274) to raise a `NotImplementedError` when a name is detected instead. 
 * Using `with_parameter` is bound to break when there is no `ListType` in the array. Ist this something that can actually happen, as we require at least one dimension to be aligned to AnnData (except for `adata.uns` ofc.). 
 * Any updates on  https://github.com/scikit-hep/awkward/issues/1391#issuecomment-1412297114? Back then we discussed this as the ideal solution once it is ready.

ivirshup:
@agoose77 @grst @jpivarski @ilan-gold

Had a discussion about this, with the following conclusions:

* We'll merge as is
  * So we don't put an upper bound on awkward
* @grst will try to see if he can figure out `with_name` later in the week
* @jpivarski @agoose77, if we can't figure out `with_name`, would you be up for making a release candidate before making a change that changes this?

PR #1043, opened 2023-07-10
Fix viewness propagation

### LINKED ISSUE #191: np.allclose modifies ArrayView if it contains nan
This is weird, and probably not too common, but it suggests a bigger problem with ArrayView behavior:

```python
import scanpy as sc
import anndata as ad
from anndata.tests.helpers import assert_equal
pbmc = sc.datasets.pbmc68k_reduced()

import scanpy as sc
import numpy as np
# This dataset has nans in .varm["PCs"]
pbmc = sc.datasets.pbmc68k_reduced()
a = pbmc[:100].copy()
b = pbmc[:100]

# .copy() turns view to ndarray, so this is fine:
assert np.allclose(a.varm["PCs"], b.varm["PCs"].copy(), equal_nan=True)
# Next line triggers the mutation:
assert not np.allclose(a.varm["PCs"], b.varm["PCs"], equal_nan=True)
# Showing that the mutation happened:
assert not np.allclose(a.varm["PCs"], b.varm["PCs"].copy(), equal_nan=True)
```

It seems to me that the only way this could be happening is through `__setitem__` as making a dummy subclass of `np.ndarray` doesn't trigger the behavior. But why would `np.allclose` – also `np.isclose` – call `__setitem__`?

### LINKED ISSUE #239: indexing error on sliced X
```
import anndata
import numpy as np

adata = anndata.AnnData(np.random.random((10,10)))
adata = adata[:, [0,2,4]]
v = adata.X.var(axis=0)
v[np.isnan(v)] = 0
```

raises IndexError: boolean index did not match indexed array along dimension 0; dimension is 10 but corresponding boolean dimension is 3

flying-sheep:
Implementation mainly follows the example in https://numpy.org/devdocs/user/basics.subclassing.html#array-ufunc-for-ufuncs

@ivirshup: should we implement this for other classes?

Fixes #191, fixes #239

### DIFF
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -139,7 +139,7 @@ class ArrayView(_SetItemMixin, np.ndarray):
         if ufunc.nout == 1:
             results = (results,)
         results = tuple(
-            (np.asarray(result).view(np.ndarray) if output is None else output)
+            (np.asarray(result) if output is None else output)
             for result, output in zip(results, outputs)
         )
         return results[0] if len(results) == 1 else results
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 from contextlib import contextmanager
 from copy import deepcopy
-from collections.abc import Sequence, KeysView
+from collections.abc import Sequence, KeysView, Callable, Iterable, Mapping
 from functools import reduce, singledispatch, wraps
-from typing import Any
+from typing import Any, Literal
 import warnings
 
 import numpy as np
@@ -86,6 +86,9 @@ class _ViewMixin(_SetItemMixin):
         return deepcopy(getattr(parent._adata_ref, attrname))
 
 
+_UFuncMethod = Literal["__call__", "reduce", "reduceat", "accumulate", "outer", "inner"]
+
+
 class ArrayView(_SetItemMixin, np.ndarray):
     def __new__(
         cls,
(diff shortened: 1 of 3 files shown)


flying-sheep:
Huh, weird, locally I get no errors at all.

ivirshup:
I've updated the branch, there's a dependency issue. But I've just pinned it so it won't block CI.

anndata/_core/views.py
```diff
@@ -103,6 +106,44 @@ def __array_finalize__(self, obj: np.ndarray | None):
...
+        results = super().__array_ufunc__(
+            ufunc, method, *convert_all(inputs), out=out, **kwargs
+        )
+        if results is NotImplemented:
+            return NotImplemented
+
+        if ufunc.nout == 1:
+            results = (results,)
+        results = tuple(
+            (np.asarray(result).view(np.ndarray) if output is None else output)
```
ivirshup, line 142: Why do the type coercion of the result?

ivirshup (commented):
> should we implement this for other classes?

Would this fix bugs in behavior with other classes? I'd lean toward no, if we don't have examples of that.

- [ ] It would probably also be good to run the scanpy test suite against this PR to check that it doesn't introduce other unintentional behavior.

> Fixes https://github.com/scverse/anndata/issues/191

- [x] Could you add tests for this?

anndata/tests/test_views.py
```diff
@@ -545,12 +545,12 @@ def test_invalid_scalar_index(adata, index):
         _ = adata[index]
 
 
-@pytest.mark.parametrize("obs", [False, True])
+@pytest.mark.parametrize("axis", ["obs", "vars"])
 @pytest.mark.parametrize("index", [-100, -50, -1])
-def test_negative_scalar_index(adata, index: int, obs: bool):
-    pos_index = index + (adata.n_obs if obs else adata.n_vars)
+def test_negative_scalar_index(adata, axis, index):
+    pos_index = index + getattr(adata, f"n_{axis}")
```
ivirshup, line 551: I'm not sure the changes here make the test better. Maybe lets revert?

anndata/tests/test_views.py
```diff
@@ -565,6 +565,15 @@ def test_negative_scalar_index(adata, index: int, obs: bool):
     )
 
 
+def test_viewness_propagation():
```
ivirshup, line 568:
```suggestion
def test_viewness_propagation():
    # https://github.com/scverse/anndata/issues/239
```

It's good to link the relevant issue for tests so we can go back for context.

anndata/tests/test_views.py
flying-sheep: I just need to stop doing aesthethic things like this in unrelated PR and do them independently

flying-sheep:
> Could you add tests for this?

the test was the first commit.

flying-sheep:
This is a bugfix, so 0.9.1 should be the correct milestone

flying-sheep:
all done, merging

docs/release-notes/0.9.2.md
```diff
@@ -1,4 +1,5 @@
 ### 0.9.2 {small}`the future`
 
 ```{rubric} Bugfix
+* Fix ufuncs of views like ``adata.X[:10].cov(axis=0) returning views {pr}`1043` {user}`flying-sheep`
```
ivirshup, line 4: Look at the other release notes for usage

anndata/_core/views.py
```diff
@@ -103,6 +106,44 @@ def __array_finalize__(self, obj: np.ndarray | None):
...
+        results = super().__array_ufunc__(
+            ufunc, method, *convert_all(inputs), out=out, **kwargs
+        )
+        if results is NotImplemented:
+            return NotImplemented
+
+        if ufunc.nout == 1:
+            results = (results,)
+        results = tuple(
+            (np.asarray(result) if output is None else output)
```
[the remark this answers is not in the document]
flying-sheep: I did

flying-sheep:
everything was resolved

docs/release-notes/0.9.2.md
flying-sheep: ???

ivirshup: Text doesn't go inside the ticks:

https://raw.githubusercontent.com/scverse/anndata/main/docs/release-notes/0.8.1.md

anndata/_core/views.py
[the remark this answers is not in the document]
ivirshup: `np.asarray`?

ivirshup:
But resolved by you, not by the reviewer. Clearly we understood the coercion change differently, and there was a bug in the addition of the release notes.

PR #1046, opened 2023-07-10
Add new DON'T MERGE check

### LINKED ISSUE #1025: PR checks for dev process PRs?

I don't think we need to have a milestone on dev process PRs like #1024

Maybe we allow milestone OR dev process tag?

flying-sheep:
- [x] Closes  #1025
- [x] Tests added

### DIFF
--- a/.github/workflows/check-pr-milestoned.yml
+++ b/.github/workflows/check-pr-milestoned.yml
@@ -20,12 +20,16 @@ on:
       - synchronize
 
 env:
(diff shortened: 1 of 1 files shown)


ivirshup (changes requested):
The check isn't running. Looking at that, it's complaining that the workflow file is invalid.

flying-sheep:
whu. why do double quotes not work in GitHub workflows? not a fan of these made up DSLs …

ivirshup (approved):

PR #1048, opened 2023-07-10
 Backport PR #1043: Fix viewness propagation 

flying-sheep:


flying-sheep:
@ivirshup any idea why that test fails?

ivirshup:
Not 100% sure. Maybe flaky test? We had bumped the memory requirement to avoid this, but I'll rerun the test to check.

PR #1049, opened 2023-07-10
Fix 0.9.2 release notes

flying-sheep:
yikes, MyST makes it really hard to see if your syntax is correct.

This fix is already included in the backport PR #1048


ivirshup:
Eh, I think we've got a weird directive here. It should probably be a role or just bold text.

PR #1054, opened 2023-07-11
Update bug-report template to include dependencies

ivirshup:


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -48,9 +48,14 @@ body:
     id: versions
     attributes:
       label: Versions
-      description: Which version of anndata and other related software you used. Please install and use `session-info`
-      render: python
-      value: |
-        >>> import anndata, session_info; session_info.show()
```
flying-sheep, line 54: please add `value` back. it seems to work well enough in the scanpy template, even inexperienced people seem to paste their output into the right spot

ivirshup: I removed it because it made it so you could submit the issue without actually putting any info in there. Since there's already a value, `require` was satisfied

flying-sheep: hmm. makes sense in this case, as we don’t use more complex syntax like `<details>` here

PR #1061, opened 2023-07-17
Restrict files written by tests

### LINKED ISSUE #1060: Stop tests from writing files everywhere

AnnData’s tests are not well behaved and write `test.h5ad` and other things into the work directory.

We should set the working directory to read only in CI during test runs to prevent this.

flying-sheep:
- [x] Closes #1060
- [x] Tests added
- [x] Release note unnecessary

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -4,7 +4,7 @@ trigger:
 variables:
   PIP_CACHE_DIR: $(Pipeline.Workspace)/.pip
   RUN_COVERAGE: no
-  PYTEST_ADDOPTS: --color=yes --nunit-xml=test-data/test-results.xml
+  PYTEST_ADDOPTS: --color=yes --junitxml=test-data/test-results.xml
   PRERELEASE_DEPENDENCIES: no
 
 jobs:
@@ -72,7 +72,7 @@ jobs:
         condition: succeededOrFailed()
         inputs:
           testResultsFiles: "test-data/test-results.xml"
-          testResultsFormat: NUnit
+          testResultsFormat: JUnit
           testRunTitle: "Publish test results for Python $(python.version)"
 
       - script: bash <(curl -s https://codecov.io/bash)
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -54,7 +54,7 @@ jobs:
 
       - script: |
           mkdir -p cache
-          find . -type d \( -path cache -prune -o -exec chmod -w {} \; \)
+          find . -type d \( -path "./cache" -prune -o -exec chmod -w {} \; \)
         displayName: "Ensure that tests run without randomly writing files"
 
       - script: |
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -53,8 +53,8 @@ jobs:
         displayName: "Display installed versions"
 
       - script: |
-          mkdir -p cache
-          find . -type d \( -path "./cache" -prune -o -exec chmod -w {} \; \)
+          mkdir -p cache .pytest_cache
+          find . -type d \( -path "./cache" -prune -o -path "./.pytest_cache" -prune -o -exec chmod -w {} \; \)
         displayName: "Ensure that tests run without randomly writing files"
 
       - script: |
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -13,13 +13,13 @@ jobs:
       vmImage: "ubuntu-22.04"
     strategy:
       matrix:
-        Python310:
-          python.version: "3.10"
+        Python3.11:
+          python.version: "3.11"
           RUN_COVERAGE: yes
-        Python38:
+        Python3.8:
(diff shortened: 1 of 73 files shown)


flying-sheep:
@ivirshup implemented your suggestions from the call!

pyproject.toml
```diff
@@ -81,6 +81,7 @@ test = [
     "loompy>=3.0.5",
     "pytest>=6.0",
     "pytest-cov>=2.10",
+    "pytest-nunit",
```
ivirshup, line 84: Why are we adding a dependency? Is this doing anything, and if not, can we remove this?

conftest.py
```diff
@@ -1,15 +1,37 @@
...
+
+
+doctest_marker = pytest.mark.usefixtures("doctest_env")
+
+
+@pytest.fixture
+def doctest_env(cache: pytest.Cache, tmp_path: Path) -> None:
+    from scanpy import settings
+
+    old_dd, settings.datasetdir = settings.datasetdir, cache.mkdir("scanpy-data")
```
ivirshup, line 20: Where does the cached `scanpy-data` get made?

.azure-pipelines.yml
```diff
@@ -37,15 +37,13 @@ jobs:
         displayName: Cache pip packages
 
       - script: |
-          python -m pip install --upgrade pip
-          pip install pytest-cov wheel
+          python -m pip install --upgrade pip wheel
```
ivirshup, line 40: Why is this changing? I think we intentionally did this to make sure we used an upgraded pip to install everything.

anndata/compat/__init__.py
```diff
@@ -24,7 +29,31 @@ class Empty:
...
+        _old_cwd: list[Path] = field(default_factory=list)
+
+        def __enter__(self) -> None:
+            self._old_cwd.append(Path())
+            os.chdir(self.path)
+
+        def __exit__(self, *_exc_info) -> None:
+            os.chdir(self._old_cwd.pop())
+
+
```
ivirshup, line 53: Wouldn't this make more sense in `utils` than `compat`?

ivirshup (changes requested):
Can you give a summary of what goes where? Is it only coverage that's going into `test-data`?

ivirshup:
Also, I get why you want to make the data files not be written to the root. Why move test outputs like reports and coverage from their default locations?

conftest.py
flying-sheep: in .pytest-cache

.azure-pipelines.yml
flying-sheep: still the case, this only removes `pytest-cov`

anndata/compat/__init__.py
flying-sheep: it’s for backwards compat with Python versions that don’t have `contextlib.chdir`.

flying-sheep:
> Why move test outputs like reports and coverage from their default locations?

so test outputs end up in one location that can be .gitignored instead of polluting the project directory.

pyproject.toml
```diff
@@ -107,13 +107,16 @@ source = "vcs"
 version-file = "anndata/_version.py"
 
 [tool.coverage.run]
+data_file = "test-data/coverage"
```
ivirshup, line 110:
```suggestion
```

pyproject.toml
```diff
@@ -107,13 +107,16 @@ source = "vcs"
...
 source = ["anndata"]
 omit = [
 	"setup.py",
 	"versioneer.py",
 	"anndata/_version.py",
 	"**/test_*.py",
 ]
+[tool.coverage.xml]
+output = "test-data/coverage.xml"
```
ivirshup, line 119:
```suggestion
```

.gitignore
```diff
@@ -2,26 +2,20 @@
...
-# Tests and coverage
-/*cache/
-/data/
-/tmp.zarr/
-test.h5ad
-test.loom
-test.zarr
-.coverage
+# Test results and coverage
+/test-data/
```
ivirshup, line 18:
```suggestion
# Test results and coverage
/test-data/
.coverage
```

ivirshup (changes requested):
I don't want to change local behavior, so have made a set of suggestions reverting that. Otherwise looks good.

> documented in .gitignore:

I'm not completley clear on what counts as a "test-result".

flying-sheep (commented):
&nbsp;

ivirshup (changes requested):

.gitignore
```diff
@@ -2,26 +2,21 @@
...
-/*cache/
-/data/
-/tmp.zarr/
-test.h5ad
-test.loom
-test.zarr
-.coverage
+# Test results (nunit/junit) and coverage
+/test-data/
+*coverage*
```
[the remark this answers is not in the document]
flying-sheep: somewhere, yes, in the project dir, no.

ivirshup (approved):

PR #1066, opened 2023-07-21
Add cupy support + CI

Zethson:
-----------------------------------

@ivirshup says:

TODO:

- [x] Figure out if pytest logic can be simplified: https://github.com/scverse/anndata/pull/1066#discussion_r1273677381
- [x] Concatenation
- [x] IO (at least writing, probably via CPU memory)
- [x] Indexing
- [x] Views
- [x] Release note
~~- [ ] Consider how much can be done with array_api~~
- [x] Benchmark concatenation to be sure CPU stuff didn't get slower


.cirun.yml
```diff
@@ -0,0 +1,9 @@
+runners:
+  - name: aws-gpu-runner
+    cloud: aws
+    instance_type: g4dn.xlarge
+    machine_image: ami-0678adbdcb4c3a662
+    preemptible: false
+    workflow: .github/workflows/build_gpu.yml
+    labels:
+      - aws-gpu
```
ivirshup, line 9:
```suggestion
runners:
  - name: aws-gpu-runner
    cloud: aws
    # Cheapest machine we can attach a GPU to
    instance_type: g4dn.xlarge
    # Machine image, as used in https://docs.cirun.io/custom-images/cloud-custom-images#using-prebuilt-images-nvidia-deep-learning-ami
    # https://aws.amazon.com/marketplace/pp/prodview-7ikjtg3um26wq?sr=0-1&ref_=beagle&applicationId=AWS-EC2-Console
    machine_image: ami-0678adbdcb4c3a662
    preemptible: false
    workflow: .github/workflows/build_gpu.yml
    labels:
      - cirun-aws-gpu
```

The label may need to start with `cirun`:  https://docs.cirun.io/reference/yaml#labels-labels ?

.github/workflows/build_gpu.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: [self-hosted, aws-gpu] # aws-gpu == label in cirun.yml
```
ivirshup, line 11:
```suggestion
    runs-on: [self-hosted, aws-gpu] # aws-gpu == label in cirun.yml
    defaults:
      run:
        shell: bash -l {0}
```

.cirun.yml
ivirshup: But it seems like it's working without?

.github/workflows/build_gpu.yml
```diff
@@ -23,19 +23,16 @@ jobs:
...
-        shell: bash -el {0}
 
       - name: Mamba list
         run: micromamba list
-        shell: bash -el {0}
 
       - name: Run test
-        run: python -c 'import anndata, cupy; cupyx.scipy.sparse.random(100, 50, format="csr")'
-        shell: bash -el {0}
+        run: python3 -c 'import anndata, cupyx; cupyx.scipy.sparse.random(100, 50, format="csr")'
```
flying-sheep, line 37:
```suggestion
        run: |
          import anndata, cupyx
          cupyx.scipy.sparse.random(100, 50, format="csr")
        shell: python3 -c {0}
```

ivirshup:
So, it runs. But I see some issues:

* `micromamba list` does not report the names of `pip installed` packages. I do not actually know if this feature is supported.
* `pip install` is trying to overwrite the `micromamba` installed `numpy`. This is probably bad, but does not seem to be causing an import issue...

flying-sheep:
> pip install is trying to overwrite the `micromamba` installed `numpy`

Both are dependency resolvers. Pip should actually avoid upgrading packages when `--upgrade`/`-U` is not specified:

> ## [Controlling what gets installed](https://pip.pypa.io/en/stable/development/architecture/upgrade-options/#controlling-what-gets-installed)
>
> […]
> 
> the “default” upgrade strategy when `--upgrade` is not set [is that] packages are not upgraded (not even direct requirements) unless the currently installed version fails to satisfy a requirement (either explicitly specified or a dependency).

If we want to force pip to only consider pre-installed versions, as there’s no option for this (https://github.com/pypa/pip/issues/4745), we could do something like

```console
$ micromamba create -n ... python=3.10 cupy  # except via action, not CLI
$ pip install -r <(pip freeze) .[dev,test]  # pin everything that’s already installed
$ pip check  # verify that the last step didn’t introduce resolution errors
```

.cirun.yml
aktech: It works for now, because you have `self-hosted` in your workflow yaml, but its deprecated. I would recommend using label name starting with cirun to avoid breaking it in future.

I would also recommend suffixing the `runs-on` param with `--${{ github.run_id }}"` to ensure that the runners are consumed by the workflow which created it.

See: https://docs.cirun.io/reference/unique-runner-labels

TLDR:
```yml
runs-on: "cirun-aws-gpu--${{ github.run_id }}"
```

In your `.cirun.yml`:
```yml
    labels:
      - cirun-aws-gpu
```

Zethson: Thank you!

ivirshup:
Because pip can't actually uninstall conda packages.

flying-sheep:
> Because pip can't actually uninstall conda packages.

In that case, there simply doesn’t seem to be a way to do things correctly* and we’re left with a hack.

A hack that works for a lot of people, but a hack that can lead to problems nontheless (adding to an environment.yml some `pip: [some-dep]` entry does the exact same as just invoking pip manually afterwards)

*Is there a way to not use conda?

conftest.py
```diff
@@ -1,3 +1,28 @@
...
 import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--only-gpu",
+        action="store_true",
+        default=False,
+        help="Only run GPU tests.",
+    )
```
flying-sheep, line 12: just use `-m gpu` instead

conftest.py
```diff
@@ -1,3 +1,28 @@
...
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "gpu: mark test as requiring GPU")
+
+
```
flying-sheep, line 18:
```suggestion
```

we’re already doing this in `pyproject.toml`, no need to repeat ourselves.

flying-sheep (changes requested):

conftest.py
```diff
@@ -1,3 +1,28 @@
...
+def pytest_runtest_setup(item):
+    gpu = len([mark for mark in item.iter_markers(name="gpu")]) > 0
+
+    if not gpu:  # Skip non gpu tests on gpu CI
+        if item.config.getoption("--only-gpu"):
+            pytest.skip("Only running GPU tests.")
+    else:  # Skip GPU tests if cupy not installed
+        from importlib.util import find_spec
+
+        pytest.mark.skipif(not find_spec("cupy"), reason="No cupy installed")
```
flying-sheep, line 28:
```suggestion
```

not necessary when simply using `-m gpu`

conftest.py
```diff
@@ -25,4 +25,7 @@ def pytest_runtest_setup(item):
     else:  # Skip GPU tests if cupy not installed
         from importlib.util import find_spec
 
-        pytest.mark.skipif(not find_spec("cupy"), reason="No cupy installed")
+        if not find_spec("cupy"):
+            pytest.skip("Cupy not installed.")
+        # why is this different?
+        # pytest.mark.skipif(not find_spec("cupy"), reason="No cupy installed")
```
flying-sheep, line 31: Because marks can be either used as decorators or by passing them to the `marks=` kwarg of `pytest.param`.

They can’t be used imperatively. You can do `item.add_marker(pytest.mark.skipif(...))`, but as said, instead of fiddling with this, you can just do this for the tests

```py
@needs('cupy')
@pytest.mark.gpu
def test_...(): ...
```

and this in pyproject.toml to deselect them by default:

```toml
[tool.pytest.ini_options]
addopts = [
    ...
    '-m', 'not gpu',
]
```

all tests can be run with `pytest -m 'gpu and not gpu'`

ivirshup: I don't think we should deselect these by default. They should be tested if the current environment has the ability to test them. We may be able to define this just a marker then, but will look into it later.

ivirshup:
From discussion w/ @Zethson:

> I think it makes sense to get it set up so that it also has the pytest stuff, but also actually checking the pytest stuff works probably requires actual implementation of cupy support

So I think we may just go for initial cupy support in this PR as well.

conftest.py
ivirshup: As discussed, the behavior would not be quite the same with `-m gpu`

Intron7:
PR #1078 makes Raw RAM only. I would also move away from automatic GPU testing. cuML allows testing after somebody sign off on it.

Intron7:
I tested the preprocessing from *rapids-singlecell* with cupy anndata. There are no issues expect the way views interact with .X replacements. This happens when i want to put a dense matrix in after `regress_out`.
In anycase it works super well so far.

Intron7:
So I have some more small Ideas that I think would be good. I can also start implementing some of them:
* Check if .nnz for cpx is less than 2^31-1 since cupy only supports int32 indptr
* make a .todevice() like torch. To transform X or .layers from and to GPU. Maybe add and `all` to dump everything into RAM. 
* add a flag/property for .X and layers if its in RAM or VRAM

ivirshup:
I think this is getting pretty close to mergable, so I think I'd leave extra features on the actual cupy support out for now. I've opened https://github.com/scverse/anndata/issues/1080 to discuss follow up PRs.

ivirshup:
@Intron7, you had mentioned some rules for controlling when this CI was run. Do you think you could link out to this/ maybe help set this up?

Intron7:
So far I only know that cuML uses that solution. I didnt check out how this works. But I will investigate this.

Intron7:
If you try to set a dense array for a sparse matrix in a view cuda this happens. Can we handle this a bit more gracefully?

```
MemoryError                               Traceback (most recent call last)
File cupy/cuda/memory.pyx:742, in cupy.cuda.memory.alloc()

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/rmm/allocators/cupy.py:37, in rmm_cupy_allocator(nbytes)
     34     raise ModuleNotFoundError("No module named 'cupy'")
     36 stream = Stream(obj=cupy.cuda.get_current_stream())
---> 37 buf = librmm.device_buffer.DeviceBuffer(size=nbytes, stream=stream)
     38 dev_id = -1 if buf.ptr else cupy.cuda.device.get_device_id()
     39 mem = cupy.cuda.UnownedMemory(
     40     ptr=buf.ptr, size=buf.size, owner=buf, device_id=dev_id
     41 )

File device_buffer.pyx:85, in rmm._lib.device_buffer.DeviceBuffer.__cinit__()

MemoryError: std::bad_alloc: out_of_memory: CUDA error at: /home/sdicks/miniconda3/envs/anndata_test/include/rmm/mr/device/cuda_memory_resource.hpp:70: cudaErrorMemoryAllocation out of memory
Exception ignored in: 'cupy.cuda.thrust.cupy_malloc'
  File "cupy/cuda/memory.pyx", line 742, in cupy.cuda.memory.alloc
  File "/home/sdicks/miniconda3/envs/anndata_test/lib/python3.10/site-packages/rmm/allocators/cupy.py", line 37, in rmm_cupy_allocator
    buf = librmm.device_buffer.DeviceBuffer(size=nbytes, stream=stream)
  File "device_buffer.pyx", line 85, in rmm._lib.device_buffer.DeviceBuffer.__cinit__
MemoryError: std::bad_alloc: out_of_memory: CUDA error at: /home/sdicks/miniconda3/envs/anndata_test/include/rmm/mr/device/cuda_memory_resource.hpp:70: cudaErrorMemoryAllocation out of memory
RuntimeError                              Traceback (most recent call last)
File <timed eval>:1

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/rapids_singlecell/cunnData_funcs/_regress_out.py:115, in regress_out(cudata, keys, layer, inplace, batchsize, verbose)
    113         cudata.layers[layer] = outputs
    114     else:
--> 115         cudata.X = outputs
    116 else:
    117     return outputs

File ~/git/anndata/anndata/_core/anndata.py:682, in AnnData.X(self, value)
    678     if sparse.issparse(self._adata_ref._X) and isinstance(
    679         value, np.ndarray
    680     ):
    681         value = sparse.coo_matrix(value)
--> 682     self._adata_ref._X[oidx, vidx] = value
    683 else:
    684     self._X = value

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/cupyx/scipy/sparse/_index.py:446, in IndexMixin.__setitem__(self, key, x)
    444     return
    445 x = x.reshape(i.shape)
--> 446 self._set_arrayXarray(i, j, x)

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/cupyx/scipy/sparse/_compressed.py:480, in _compressed_sparse_matrix._set_arrayXarray(self, row, col, x)
    478 def _set_arrayXarray(self, row, col, x):
    479     i, j = self._swap(row, col)
--> 480     self._set_many(i, j, x)

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/cupyx/scipy/sparse/_compressed.py:557, in _compressed_sparse_matrix._set_many(self, i, j, x)
    555 j = j[mask]
    556 j[j < 0] += N
--> 557 self._insert_many(i, j, x[mask])

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/cupyx/scipy/sparse/_compressed.py:616, in _compressed_sparse_matrix._insert_many(self, i, j, x)
    607 def _insert_many(self, i, j, x):
    608     """Inserts new nonzero at each (i, j) with value x
    609     Here (i,j) index major and minor respectively.
    610     i, j and x must be non-empty, 1d arrays.
   (...)
    613     Modifies i, j, x in place.
    614     """
--> 616     order = cupy.argsort(i)  # stable for duplicates
    617     i = i.take(order)
    618     j = j.take(order)

File ~/miniconda3/envs/anndata_test/lib/python3.10/site-packages/cupy/_sorting/sort.py:116, in argsort(a, axis, kind)
    114 if kind is not None and kind != 'stable':
    115     raise ValueError("kind can only be None or 'stable'")
--> 116 return a.argsort(axis=axis)

File cupy/_core/core.pyx:879, in cupy._core.core._ndarray_base.argsort()

File cupy/_core/core.pyx:896, in cupy._core.core._ndarray_base.argsort()

File cupy/_core/_routines_sorting.pyx:96, in cupy._core._routines_sorting._ndarray_argsort()

File cupy/cuda/thrust.pyx:117, in cupy.cuda.thrust.argsort()

RuntimeError: transform: failed to synchronize: cudaErrorIllegalAddress: an illegal memory access was encountered
```

Intron7:
> Could you share some code that throws this?
```
from anndata import AnnData
from cupyx.scipy import sparse as cpsparse
from scipy import sparse
import cupy as cp
import numpy as np

rand = sparse.random(100000, 20000, density=0.05,dtype=np.float32, format="csr")
adata = AnnData(X= cpsparse.csr_matrix(rand))
adata = adata[:,:5000]
X = cp.random.rand(100000,5000, dtype= cp.float32)
adata.X = X
```

This works though
```
adata = AnnData(X= cpsparse.csr_matrix(rand))
adata = adata[:,:5000].copy()
X = cp.random.rand(100000,5000, dtype= cp.float32)
adata.X = X
```

ivirshup:
In this instance I think you can do:

```python
rand = sparse.random(100000, 20000, density=0.05,dtype=np.float32, format="csr")
adata = AnnData(X= cpsparse.csr_matrix(rand))
adata = adata[:,:5000]
X = cp.random.rand(100000,5000, dtype= cp.float32)

del adata.X
adata.X = X
```

But yeah the behavior is weird, but not a bug. I'm not really sure what a more graceful way to handle this would be here.

I think I could be up for a inplace conversion to actual though. Can you open an issue for this?

Intron7:
For very small matrices it works but is super slow.

Open an issue for the feature #1082

conftest.py
```diff
@@ -1,3 +1,15 @@
 # This file exists just to allow ignoring warnings without test collection failing on CI
 # TODO: Fix that
 import pytest
+
+
+def pytest_itemcollected(item):
+    # Defniing behavior of pytest.mark.gpu
```
flying-sheep, line 7:
```suggestion
    """Defining behavior of pytest.mark.gpu"""
```

flying-sheep (changes requested):

flying-sheep (approved):
Looking good! Once the TODOs are done, I think we’re good to go!

anndata/tests/test_gpu.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+
+@pytest.mark.gpu
+def test_gpu():
```
Zethson, line 5: Do you want to keep this? It looks like a dummy

ivirshup: It's testing that the gpu tests aren't being run when cupy isn't installed

PR #1070, opened 2023-07-24
Fix view behavior for AwkwardArrays

### LINKED ISSUE #1035: Future changes to Awkward Array behavior class resolution

grst:
- [x] Closes #1035
- [x] Follow up of #1040 

### DIFF
--- a/anndata/_core/views.py
+++ b/anndata/_core/views.py
@@ -262,71 +262,11 @@ def as_view_zappy(z, view_args):
     return z
 
 
-try:
-    from ..compat import awkward as ak
-    import weakref
-
-    # Registry to store weak references from AwkwardArrayViews to their parent AnnData container
-    _registry = weakref.WeakValueDictionary()
-    _PARAM_NAME = "_view_args"
-
-    class AwkwardArrayView(_ViewMixin, AwkArray):
-        @property
-        def _view_args(self):
-            """Override _view_args to retrieve the values from awkward arrays parameters.
-
-            Awkward arrays cannot be subclassed like other python objects. Instead subclasses need
-            to be attached as "behavior". These "behaviors" cannot take any additional parameters (as we do
-            for other data types to store `_view_args`). Therefore, we need to store `_view_args` using awkward's
-            parameter mechanism. These parameters need to be json-serializable, which is why we can't store
-            ElementRef directly, but need to replace the reference to the parent AnnDataView container with a weak
-            reference.
-            """
-            parent_key, attrname, keys = self.layout.parameter(_PARAM_NAME)
-            parent = _registry[parent_key]
-            return ElementRef(parent, attrname, keys)
-
-        def __copy__(self) -> AwkArray:
-            """
-            Turn the AwkwardArrayView into an actual AwkwardArray with no special behavior.
-
-            Need to override __copy__ instead of `.copy()` as awkward arrays don't implement `.copy()`
-            and are copied using python's standard copy mechanism in `aligned_mapping.py`.
-            """
-            array = self
-            # makes a shallow copy and removes the reference to the original AnnData object
-            array = ak.with_parameter(array, _PARAM_NAME, None)
-            array = ak.with_parameter(array, "__record__", None)
-            return array
-
-    @as_view.register(AwkArray)
-    def as_view_awkarray(array, view_args):
-        parent, attrname, keys = view_args
-        parent_key = f"target-{id(parent)}"
-        _registry[parent_key] = parent
-        # TODO: See https://github.com/scverse/anndata/pull/647#discussion_r963494798_ for more details and
-        # possible strategies to stack behaviors.
-        # A better solution might be based on xarray-style "attrs", once this is implemented
-        # https://github.com/scikit-hep/awkward/issues/1391#issuecomment-1412297114
-        #
-        # Allow only Arrays that have no record behavior attached, or with __record__ explicitly
-        # set to None
-        if array.layout.purelist_parameter("__record__") is not None:
-            raise NotImplementedError(
-                "Cannot create a view of an awkward array with __record__ parameter. "
-                "Please open an issue in the AnnData repo and describe your use-case."
-            )
-        array = ak.with_parameter(array, _PARAM_NAME, (parent_key, attrname, keys))
-        array = ak.with_parameter(array, "__record__", "AwkwardArrayView")
-        return array
-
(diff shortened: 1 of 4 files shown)


grst:
I think this does the trick

flying-sheep (changes requested):
Shouldn’t this have some tests?

anndata/_core/views.py
```diff
@@ -308,15 +308,19 @@ def as_view_awkarray(array, view_args):
         # possible strategies to stack behaviors.
         # A better solution might be based on xarray-style "attrs", once this is implemented
         # https://github.com/scikit-hep/awkward/issues/1391#issuecomment-1412297114
-        if type(array).__name__ != "Array":
+        #
+        # Allow only Arrays that have no record behavior attached, or with __record__ explicitly
+        # set to None
+        if ak.parameters(array).get("__record__", None) is not None:
```
agoose77, line 314: (Record) names can be set at arbitrary depths. I'd suggest checking recursively using our L2 (public-developer) API:
```suggestion
        if array.layout.purelist_parameter("__record__") is not None:
```

agoose77: [This setitem](https://github.com/scverse/anndata/blob/b6406179f50ce69d84112a5426d206a203a1181d/anndata/_core/views.py#L323) is responsible for registering the behavior class for lists containing records, rather than 1D arrays of records.

anndata/_core/views.py
```diff
@@ -295,8 +295,8 @@ def __copy__(self) -> AwkArray:
             """
             array = self
             # makes a shallow copy and removes the reference to the original AnnData object
-            array = ak.with_parameter(self, _PARAM_NAME, None)
-            array = ak.with_parameter(array, "__list__", None)
+            array = ak.with_parameter(array, _PARAM_NAME, None)
+            array = ak.with_parameter(array, "__record__", None)
```
agoose77, line 299: This should use `with_name`, so that we can traverse into the layout to find the record-node
```suggestion
            array = ak.with_name(array, None)
```

agoose77:
@grst I get the sense that it might be helpful to have a meeting to discuss this. Would you be able to find time for a zoom at some point?

anndata/_core/views.py
ilan-gold: I imagine that after our dicsussion @grst could not get that working (see https://github.com/scverse/anndata/pull/1040#issuecomment-1647878246)

Is there a reason traversing is better than what we have (assuming @grst could not get the `with_name` working)?  We only ever deal with this at a "top level."

grst:
Agreed, I sent an email with a poll to find a timeslot.

grst:
@ivirshup, @ilan-gold, @agoose77, ready for re-review. Feels good to get rid of all that code :boom: 

Currently the tests fail for `uns`, but that seems to be a more general problem, see #1118.

agoose77:
@grst it is possible to modify the original array if you don't perform any row-based slices, i.e `array["x"] = x`, then this would modify `array`.

I can't recall to what extent we touched on this on the call. If the process of pulling out the Awkward Array always returns a shallow copy, this wouldn't be a problem.

grst:
Hi @agoose77, 

returning the shallow copy (or slice) works as expected. The remaining problem is that there's no handler implemented for the `uns` slot while subsetting AnnData -- regardless of whether it's an awkward array or a numpy array. I don't think we want to always return a shallow copy on `__getitem__` -- I think there are legitimate use-cases where one wants to modify records.

ivirshup:
> The remaining problem is that there's no handler implemented for the uns slot while subsetting AnnData

Is that blocking for this PR, or tangential? I'd like to make a release candidate by the end of the week, so would like to merge soon if possible.

grst:
If you don't mind, I can mark the respective tests as `xfail`. I think it's ok as
 * the behavior is not a regression of this PR, it was like that before
 * the problem is tracked in #1118 and
 * I don't think it affects many users

ivirshup:
> I can mark the respective tests as xfail

Which tests are these?

grst:
@ivirshup, these: 
https://github.com/scverse/anndata/pull/1070/files#diff-1534693aa7219116bb36568d007d5f2116414d0c86148cfa15d859b861abafc7R143-R166

grst:
Thanks for catching this! I thought my tests cover this, but it turns out I retrieved the awkward array from the view only once and did all tests on that copy. Instead, retreiving `v.obsm["awk"]` multiple times always returns a fresh copy of the sliced awkward array which makes it impossible to modify it. 

To solve this, it is necessary to perform a shallow copy of the awkward array on view creation (which is a cheap operation). I added this behavior here: 
https://github.com/scverse/anndata/pull/1070/files#diff-af8b8ed3c983f1ed77b81030eb9d80df8136e0640324f180354e6e0d3b49e7e7R139-R159

Maybe this could also be useful for pandas data frames with copy-on-write behavior in the future?

ivirshup:
I think caching the "view"s could be a good idea. I may need to think for a bit on how this effects my current thoughts on how to fix the garbage collection problem:

* https://github.com/scverse/anndata/pull/1119

Currently, this won't work with the solution proposed there, since the AlignedMappings are created on the fly to prevent circular references from being formed.

PR #1073, opened 2023-07-25
Remove unused imports

flying-sheep:
Clarify API packages, remove unused imports

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -15,8 +15,8 @@ except (ImportError, LookupError):
 from ._core.anndata import AnnData
 from ._core.merge import concat
 from ._core.raw import Raw
-from ._io import (
-    read_h5ad,
+from ._io.h5ad import read_h5ad
+from ._io.read import (
     read_loom,
     read_hdf,
     read_excel,
@@ -35,7 +35,7 @@ from ._warnings import (
 from . import experimental
 
 # backwards compat / shortcut for default format
-from ._io import read_h5ad as read  # noqa: F401
+read = read_h5ad
 
 __all__ = [
     "__version__",
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -34,4 +34,24 @@ from ._warnings import (
 )
 
 # backwards compat / shortcut for default format
-from ._io import read_h5ad as read
+from ._io import read_h5ad as read  # noqa: F401
+
+__all__ = [
+    "__version__",
+    "AnnData",
+    "concat",
+    "Raw",
+    "read_h5ad",
+    "read_loom",
+    "read_hdf",
+    "read_excel",
+    "read_umi_tools",
+    "read_csv",
+    "read_text",
+    "read_mtx",
+    "read_zarr",
+    "OldFormatWarning",
+    "WriteWarning",
+    "ImplicitModificationWarning",
+    "ExperimentalFeatureWarning",
+]
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -15,15 +15,15 @@ except (ImportError, LookupError):
 from ._core.anndata import AnnData
 from ._core.merge import concat
 from ._core.raw import Raw
-from ._io.h5ad import read_h5ad
-from ._io.read import (
-    read_loom,
-    read_hdf,
-    read_excel,
-    read_umi_tools,
+from ._io import (
     read_csv,
-    read_text,
+    read_excel,
+    read_h5ad,
+    read_hdf,
+    read_loom,
     read_mtx,
+    read_text,
+    read_umi_tools,
     read_zarr,
 )
 from ._warnings import (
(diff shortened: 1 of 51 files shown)


anndata/_io/zarr.py
```diff
@@ -136,8 +145,6 @@ def read_dataframe_legacy(dataset: zarr.Array) -> pd.DataFrame:
 
 @report_read_key_on_error
 def read_dataframe(group) -> pd.DataFrame:
-    from .specs import _REGISTRY
```
flying-sheep, line 139: if this is necessary, it should have a comment explaining why

anndata/_io/h5ad.py
```diff
@@ -41,8 +51,6 @@ def write_h5ad(
     dataset_kwargs: Mapping = MappingProxyType({}),
     **kwargs,
 ) -> None:
-    from anndata.experimental import write_dispatched
```
flying-sheep, line 44: This is less likely to be a import with side effects, but it could be too: is it?

flying-sheep: also not necessary

anndata/__init__.py
```diff
@@ -34,4 +34,24 @@
...
 
 # backwards compat / shortcut for default format
-from ._io import read_h5ad as read
+from ._io import read_h5ad as read  # noqa: F401
+
+__all__ = [
+    "__version__",
+    "AnnData",
+    "concat",
+    "Raw",
```
ivirshup, line 43:
```suggestion
    "Raw",
    "experimental",
```

Right?

pyproject.toml
```diff
@@ -133,6 +131,8 @@ select = [
     "F",
     "W",
 ]
+# Disable autofix for unused imports (`F401`).
+unfixable = ["F401"]
```
ivirshup, line 135: Why?

anndata/_io/h5ad.py
```diff
@@ -30,6 +30,16 @@
...
+
+__all__ = [
+    "read_elem",
+    "write_elem",
+    "read_h5ad",
+    "write_h5ad",
+    "read_dataset",
+]
+
+
```
ivirshup, line 42: Why put this in a private module?

anndata/__init__.py
flying-sheep: No, if you have a submodule that nothing has imported yet, the submodule is not a member of the super module:

```console
$ mkdir mod
$ touch mod/submod.py
$ touch mod/__init__.py
$ python -c 'import mod; mod.submod'
AttributeError: Module `mod` has no member `submod`
$ python -c 'import mod.submod'  # works
```

pyproject.toml
flying-sheep: If you have your editor set up to do them for you, it can be annoying to have it remove imports that are already written but not used yet.

I’m OK with both, I’ve changed my habits to never do the above, but I thought it might be annoying for others. Your call!

anndata/_io/h5ad.py
flying-sheep: It clarifies what’s imported from elsewhere. That’s the list of functions that `anndata._io.zarr` has in an equivalent way. But we can get rid of it if you prefer.

pyproject.toml
ivirshup: I would really prefer that it makes the change on pre-commit

anndata/_io/h5ad.py
flying-sheep: he prefers

anndata/_io/specs/__init__.py
```diff
@@ -1,2 +1,13 @@
...
+
+__all__ = [
+    "methods",
+    "write_elem",
+    "get_spec",
+    "read_elem",
+    "Reader",
+    "Writer",
+    "IOSpec",
+]
```
ivirshup, line 13: Again, private module defining exports

```suggestion
```

anndata/_io/read.py
```diff
@@ -27,6 +27,19 @@ def read_zarr(*_, **__):
...
+    "read_csv",
+    "read_excel",
+    "read_umi_tools",
+    "read_hdf",
+    "read_loom",
+    "read_mtx",
+    "read_text",
+]
+
+
```
ivirshup, line 42:
```suggestion
```

anndata/_io/__init__.py
```diff
@@ -9,5 +9,22 @@
...
+    "read_loom",
+    "read_mtx",
+    "read_text",
+    "read_zarr",
+    "read_h5ad",
+    "write_csvs",
+    "write_loom",
+    "write_zarr",
+    "h5ad",
+]
```
ivirshup, line 30: So, two things here:

* I don't think we should define namespaces for private modules
* I am up for a public `io` module, but I think we should do that in a separate PR

```suggestion
```

ivirshup (changes requested):

anndata/_io/specs/__init__.py
flying-sheep: please look at the file: we’re using it as an internal API module.

are you saying that we should get rid of all of those and directly import everything from where it’s defined?

ivirshup (approved):

PR #1078, opened 2023-07-26
`.raw` copys from VRAM to RAM

Intron7:
Moves `.X` for `.raw` from GPU to RAM. This will save a lot of VRAM

### DIFF
--- a/anndata/_core/raw.py
+++ b/anndata/_core/raw.py
@@ -5,11 +5,23 @@ import numpy as np
 import pandas as pd
 from scipy import sparse
 from scipy.sparse import issparse
+from enum import Enum
 
 from . import anndata
 from .index import _normalize_index, _subset, unpack_index, get_vector
 from .aligned_mapping import AxisArrays, AxisArraysView
 from .sparse_dataset import SparseDataset
+
+from ..compat import CupyArray, CupySparseMatrix
+
+
+class CPType(Enum):
+    CupyArray = CupyArray
+    CupySparseMatrix = CupySparseMatrix
+
+    @classmethod
+    def classes(cls):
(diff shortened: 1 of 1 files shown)


anndata/_core/raw.py
```diff
@@ -5,12 +5,24 @@
...
+
+
+class CPType(Enum):
+    CupyArray = CupyArray
+    CupySparseMatrix = CupySparseMatrix
+
+    @classmethod
+    def classes(cls):
+        return tuple(c.value for c in cls.__members__.values())
+
```
ivirshup, line 25:
```suggestion
```

I probably wouldn't bother with this

anndata/_core/raw.py
```diff
@@ -5,12 +5,24 @@
 import pandas as pd
 from scipy import sparse
 from scipy.sparse import issparse
+from enum import Enum
```
ivirshup, line 8:
```suggestion
```

anndata/_core/raw.py
```diff
@@ -31,7 +43,10 @@ def __init__(
             self._var = _gen_dataframe(var, self.X.shape[1], ["var_names"])
             self._varm = AxisArrays(self, 1, varm)
         elif X is None:  # construct from adata
-            self._X = adata.X.copy()
+            if isinstance(adata.X, CPType.classes()):
```
ivirshup, line 46:
```suggestion
            # Move from GPU to CPU since it's large and not always used
            if isinstance(adata.X, (CupyArray, CupySparseMatrix)):
```

ivirshup (commented):
Minor changes: 

* I don't think we need to bother with the enum
* Added a comment explaining why we're doing this

ivirshup:
Thanks!

PR #1079, opened 2023-07-26
Dask Distributed Write Fix For Zarr

syelman:
Hi,
This issue appears when one tries to use dask distributed then write https://github.com/dask/distributed/issues/780.
I did what xarray does for this https://github.com/pydata/xarray/pull/1179/files.

Not sure why the lock is global though for now.

Ping @ivirshup @flying-sheep

### DIFF
--- a/anndata/tests/test_dask.py
+++ b/anndata/tests/test_dask.py
@@ -96,10 +96,12 @@ def test_dask_write(adata, tmp_path, diskfmt):
     assert isinstance(orig.varm["a"], DaskArray)
 
 
+@pytest.mark.dependency(depends=[""])
 def test_dask_distributed_write(adata, tmp_path, diskfmt):
     import dask.array as da
     import numpy as np
-    import dask.distributed as dd
+
+    dd = pytest.importorskip("dask.distributed")
 
     pth = tmp_path / f"test_write.{diskfmt}"
     g = as_group(pth, mode="w")
--- a/anndata/tests/test_dask.py
+++ b/anndata/tests/test_dask.py
@@ -96,7 +96,6 @@ def test_dask_write(adata, tmp_path, diskfmt):
     assert isinstance(orig.varm["a"], DaskArray)
 
 
-@pytest.mark.needs("dask.distributed")
 def test_dask_distributed_write(adata, tmp_path, diskfmt):
     import dask.array as da
     import dask.distributed as dd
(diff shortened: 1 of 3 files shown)


flying-sheep:
is there a way to test this?

syelman:
@flying-sheep  added tests, and it now fails for h5py :D. I thought I could just go with whatever xarray did at https://github.com/pydata/xarray/pull/1179/files. But now I realize this was from 2017 so I will need investigate a bit more closely.

syelman:
Hey @flying-sheep @ilan-gold, it should be ready for review once this test passes. I ended up not adding h5ad support for distributed writing and just giving an error. To fix we would need to create a file manager class or so: https://github.com/pydata/xarray/issues/4242

anndata/tests/test_dask.py
```diff
@@ -94,6 +96,46 @@ def test_dask_write(adata, tmp_path, diskfmt):
     assert isinstance(orig.varm["a"], DaskArray)
 
 
+@pytest.mark.dependency(depends=[""])
```
flying-sheep, line 99: this needs to go!

pyproject.toml
```diff
@@ -88,6 +88,7 @@ test = [
     "boltons",
     "scanpy",
     "dask[array]",
+    "dask[distributed]",
```
flying-sheep, line 91:
```suggestion
    "dask[array,distributed]",
```

anndata/tests/test_dask.py
```diff
@@ -94,6 +96,46 @@ def test_dask_write(adata, tmp_path, diskfmt):
...
+def test_dask_distributed_write(adata, tmp_path, diskfmt):
+    import dask.array as da
+    import numpy as np
+
+    dd = pytest.importorskip("dask.distributed")
```
flying-sheep:
```suggestion
@needs('dask.distributed')
def test_dask_distributed_write(adata, tmp_path, diskfmt):
    import dask.array as da
    import dask.distributed as dd
    import numpy as np
```

flying-sheep (changes requested):

syelman:
Hey @flying-sheep could you perhaps merge this? I created an issue for h5py version here: https://github.com/scverse/anndata/issues/1105. But until I create a PR for that separately, I would like to create a notebook for concat_on_disk, which uses distributed writing for tests.

flying-sheep (approved):

ivirshup:
Alright I think we can merge. But I want to follow up.

Looking through xarray, it doesn't seem like they use this kind of solution for zarr, but do for hdf5. However here, we are using this solution for zarr and it does not work for hdf5. This is a bit disconcerting and a little concerning.

I vaguely understand why this isn't working for us with hdf5, xarray has some handling for this as mentioned above.

I do not understand what the are doing as to not need this for zarr.

I will go ask xarray/ zarr people to take a look and see if they can suggest a better solution.

PR #1081, opened 2023-07-28
Make checking var_names and obs_names for uniqueness optional

niklasmueboe:
By default when creating a new AnnData instance in the `_init_as_actual` function both `var_names` and `obs_names` are checked for uniqueness. When working with AnnData objects with several million `obs`/`var `this can become a performance limiting factor taking multiple seconds or even minutes.

This PR would allow the user the option to skip uniqueness check when instantiating a new AnnData object via `ad.AnnData(X=X, ... , check_unique=False)`. This is can be useful to improve performance especially when the input data is already guaranteed to have unique `obs_names` and `var_names`.


ivirshup:
Thanks for opening the PR. 

Could this be opened as an issue instead of a PR? I believe there will be other's who have thoughts, but also think this could use some discussion.

flying-sheep:
Thanks, we implemented this in  #1507!

PR #1084, opened 2023-07-31
Update machine image for GPU CI

ivirshup:
- [ ] Closes #
- [ ] Tests added

Trying a new AMI for GPU CI

### DIFF
--- a/.cirun.yml
+++ b/.cirun.yml
@@ -2,7 +2,7 @@ runners:
   - name: aws-gpu-runner
     cloud: aws
     instance_type: g4dn.xlarge
-    machine_image: ami-0678adbdcb4c3a662
+    machine_image: ami-0139a05e194cbd9a0
     preemptible: false
     workflow: .github/workflows/test-gpu.yml
     labels:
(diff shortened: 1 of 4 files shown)


.cirun.yml
```diff
@@ -2,7 +2,7 @@ runners:
   - name: aws-gpu-runner
     cloud: aws
     instance_type: g4dn.xlarge
-    machine_image: ami-0678adbdcb4c3a662
+    machine_image: ami-0139a05e194cbd9a0
```
aktech, line 5: Note: This failed here: https://github.com/scverse/anndata/pull/1084/checks?check_run_id=15487580477
Is this AMI in `eu-west-1` (that's default region)? If not, then you need to add `region: <region-name>` in here.

ivirshup: I believe this didn't work because I didn't wait for the AMI to build to stop the instance, so the AMI failed to build.

ivirshup: How would I tell what region it's associated with? I added eu-north-1 for the new one, but it's not super clear from the UI

aktech: It's visible on the top right in the aws console, you already got it anyway. :)

PR #1094, opened 2023-08-04
Fix mandatory GPU tests

flying-sheep:
Currently when the `run-gpu-ci` label is not set (or it’s a push), the job just get skipped, which counts as success:

> image

This adds a second job (on a default runner), which simply fails if the conditions aren’t met.

This PR ensures that there’s always a fail outcome unless the GPU job is successfully run, and merging is blocked when it should be.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,8 @@
+<!-- Please:
+1. Fill in the following check boxes
+2. Make sure checks pass (Ignore “Triage” ones)
+-->
+
 - [ ] Closes #
 - [ ] Tests added
 - [ ] Release note added (or unnecessary)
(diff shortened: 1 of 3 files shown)


Zethson:
I'm not sure whether this is a good idea if I get it right. I think that it's annoying to have a failing CI job if it's not even relevant (it doesn't touch GPU code).

flying-sheep:
Well, then we e.g. need a second label, e.g. `skip-gpu-tests`.

- when none is there, the first job fails and the second gets skipped as a consequence
- when `skip-gpu-tests` is there, both jobs get skipped (or the first succeeds and the second gets skipped)
- when `run-gpu-tests` is there, the first succeeds and the second gets run

Zethson (approved):
Discussed with @flying-sheep . I would have preferred a more hands-off solution but this is totally fine.

Thanks!

flying-sheep:
To elaborate: Experience shows than without a reminder, things like release notes or working links in documentation get forgotten and devolve into a hard-to-salvage state after being forgotten long enough.

Failing PR checks might be slightly confusing for contributors who aren’t used to contributing to bigger projects*, but I’ve offset this using a terse PR template comment.

*numpy, cpython, pandas, … have many many checks

PR #1097, opened 2023-08-15
pepy fix

Intron7:
Fixes the pepy badges


flying-sheep:
Thanks! Seems like there’s a problem with some dependency’s prerelease version. Let’s see if I can figure that out

flying-sheep (approved):

PR #1099, opened 2023-08-18
Fix index comparison

### LINKED ISSUE #1098: Fix/Circumvent upcoming breakage in pandas 2.1.0

### Report

Let’s see if that’s a pandas bug: https://github.com/pandas-dev/pandas/issues/54622

Code:

```console
from anndata.tests.helpers import gen_adata, GEN_ADATA_DASK_ARGS

adata = gen_adata((20, 10), **GEN_ADATA_DASK_ARGS) 

view = adata[:, ::2] 
assert view.var.shape == (5, 8) 
assert {k: v.shape[0] for k, v in view.varm.items()} == {k: 5 for k in view.varm} 
```

Traceback:

```pytb
anndata/tests/test_views.py::test_view_subset_shapes failed: def test_view_subset_shapes():
        adata = gen_adata((20, 10), **GEN_ADATA_DASK_ARGS)
    
        view = adata[:, ::2]
        assert view.var.shape == (5, 8)
>       assert {k: v.shape[0] for k, v in view.varm.items()} == {k: 5 for k in view.varm}
E       AssertionError: assert {'array': 5, ..., 'sparse': 5} == {'array': 5, ..., 'sparse': 5}
E         Omitting 3 identical items, use -vv to show
E         Differing items:
E         {'df': 10} != {'df': 5}
E         Use -v to get more diff

anndata/tests/test_views.py:617: AssertionError
```

### Versions

pandas 2.1.0rc0

flying-sheep:
- [x] Closes #1098
- [x] Tests added

### DIFF
--- a/anndata/_core/aligned_mapping.py
+++ b/anndata/_core/aligned_mapping.py
@@ -244,7 +244,7 @@ class AxisArraysBase(AlignedMapping):
         if (
             hasattr(val, "index")
             and isinstance(val.index, cabc.Collection)
-            and not (val.index == self.dim_names).all()
+            and val.index.equals(self.dim_names)
         ):
             # Could probably also re-order index if it’s contained
             raise ValueError(
--- a/docs/release-notes/release-latest.md
+++ b/docs/release-notes/release-latest.md
@@ -1,7 +1,8 @@
-## Version 0.10.0
+## Version 0.10
 
 ```{include} /release-notes/0.10.0.md
 ```
+
 ## Version 0.9
 
 ```{include} /release-notes/0.9.3.md


anndata/_core/aligned_mapping.py
```diff
@@ -241,15 +241,14 @@ def to_df(self) -> pd.DataFrame:
...
-                f"value.index does not match parent’s axis {self.axes[0]} names"
-            )
+        if hasattr(val, "index") and isinstance(val.index, cabc.Collection):
+            try:
+                pd.testing.assert_index_equal(val.index, self.dim_names)
+            except AssertionError as e:
+                # Could probably also re-order index if it’s contained
+                raise ValueError(
+                    f"value.index does not match parent’s axis {self.axes[0]} names:\n{e}"
+                ) from None
```
ivirshup, line 251: * Why this change here?
* I would probably go with `val.index.equals(self.dim_names)`. The assertion tests may spend an arbitrary amount of time formatting a nice error message, and aren't built for being performant.

anndata/_core/merge.py
```diff
@@ -235,7 +236,9 @@ def unify_dtypes(dfs: Iterable[pd.DataFrame]) -> list[pd.DataFrame]:
     return dfs
 
 
-def try_unifying_dtype(col: list) -> pd.core.dtypes.base.ExtensionDtype | None:
+def try_unifying_dtype(
+    col: Sequence[np.dtype],
```
ivirshup, line 240: Similar to above, I don't believe pandas dtypes subclass from numpy dtypes.

ivirshup (changes requested):

anndata/_core/merge.py
```diff
@@ -211,7 +212,7 @@ def unify_dtypes(dfs: Iterable[pd.DataFrame]) -> list[pd.DataFrame]:
     df_dtypes = [dict(df.dtypes) for df in dfs]
     columns = reduce(lambda x, y: x.union(y), [df.columns for df in dfs])
 
-    dtypes = {col: list() for col in columns}
+    dtypes: dict[str, list[np.dtype]] = {col: [] for col in columns}
```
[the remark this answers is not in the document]
flying-sheep: good catch!

anndata/_core/aligned_mapping.py
flying-sheep: Because it would have saved me from jumping into the debugger to see *how* the indices differ.

Making arguments about performance that don’t include a measurement is premature optimization. But I can assure you: if you put enough matrices into a mapping that 10µs matter, you have bigger performance problems:

```
In [1]: import pandas as pd, numpy as np

In [2]: i2 = (i1 := pd.Index(np.random.rand(10000))).copy()

In [3]: %timeit pd.testing.assert_index_equal(i1, i2)
22.3 µs ± 126 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [4]: %timeit i1.equals(i2)
12.7 µs ± 41.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

ivirshup: > If you want, I can change it so it calls .equals() first and uses the output from assert_index_equals

Yes.

I wouldn't expect them to put priority on performance on testing functions in the future, and for sure in the past haven't.

anndata/_core/merge.py
```diff
@@ -14,6 +14,7 @@
     Collection,
     Iterable,
     Optional,
+    Sequence,
```
ivirshup, line 17: `typing.Sequence` is deprecated in favor of `collections.abc.Sequence`: https://docs.python.org/3/library/typing.html#typing.Sequence

flying-sheep: I thought you didn’t like unrelated changes, but sure.

ivirshup: I would also be okay with not doing the typing changes as well. But this import was added in this PR, no?

anndata/_core/aligned_mapping.py
ivirshup:
```suggestion
        if (
            hasattr(val, "index")
            and isinstance(val.index, cabc.Collection)
            and not val.index.equals(self.dim_names)
        ):
            # Could probably also re-order index if it’s contained
            raise ValueError(
                f"value.index does not match parent’s axis {self.axes[0]} names"
            )
```

flying-sheep:
@ivirshup all done

ivirshup:
About the extra change

> Because it would have saved me from jumping into the debugger to see how the indices differ.

It looks like this will only be the case for small data, since afaict the error message from the assertion is just dumping the repr of each object into the exception message:

```python
In [30]: adata.obsm["b"] = b.to_df()
Cell In[30], line 1
----> 1 adata.obsm["b"] = b.to_df()

File ~/github/anndata/anndata/_core/aligned_mapping.py:189, in AlignedActualMixin.__setitem__(self, key, value)
    188 def __setitem__(self, key: str, value: V):
--> 189     value = self._validate_value(value, key)
    190     self._data[key] = value

File ~/github/anndata/anndata/_core/aligned_mapping.py:254, in AxisArraysBase._validate_value(self, val, key)
    252 except AssertionError as e:
    253     msg = f"value.index does not match parent’s axis {self.axes[0]} names:\n{e}"
--> 254     raise ValueError(msg) from None
    255 else:
    256     msg = "Index.equals and pd.testing.assert_index_equal disagree"

ValueError: value.index does not match parent’s axis 0 names:
Index are different

Index values are different (0.03791 %)
[left]:  Index(['AAACATACAACCAC-1', 'AAACATTGAGCTAC-1', 'AAACATTGATCAGC-1',
       'AAACCGTGCTTCCG-1', 'AAACCGTGTATGCG-1', 'AAACGCACTGGTAC-1',
       'AAACGCTGACCAGT-1', 'AAACGCTGGTTCTT-1', 'AAACGCTGTAGCCA-1',
       'AAACGCTGTTTCTG-1',
       ...
       'TTTCAGTGTCACGA-1', 'TTTCAGTGTCTATC-1', 'TTTCAGTGTGCAGT-1',
       'TTTCCAGAGGTGAG-1', 'TTTCGAACACCTGA-1', 'TTTCGAACTCTCAT-1',
       'TTTCTACTGAGGCA-1', 'TTTCTACTTCCTCG-1', 'TTTGCATGAGAGGC-1',
       'TTTGCATGCCTCAC-1'],
      dtype='object', length=2638)
[right]: Index(['AAACATACAACCAC-1', 'AAACATTGAGCTAC-1', 'AAACATTGATCAGC-1',
       'AAACCGTGCTTCCG-1', 'AAACCGTGTATGCG-1', 'AAACGCACTGGTAC-1',
       'AAACGCTGACCAGT-1', 'AAACGCTGGTTCTT-1', 'AAACGCTGTAGCCA-1',
       'AAACGCTGTTTCTG-1',
       ...
       'TTTCAGTGTCACGA-1', 'TTTCAGTGTCTATC-1', 'TTTCAGTGTGCAGT-1',
       'TTTCCAGAGGTGAG-1', 'TTTCGAACACCTGA-1', 'TTTCGAACTCTCAT-1',
       'TTTCTACTGAGGCA-1', 'TTTCTACTTCCTCG-1', 'TTTGCATGAGAGGC-1',
       'TTTGCATGCCTCAC-1'],
      dtype='object', name='index', length=2638)
```

^ In this case the mismatched values are in the middle, and it's a relatively small dataset.

I'm fine to keep this for now, but also fine to remove it later without needing to keep the same level of detail.

ivirshup (approved):

ivirshup:
Oh, also should this be backported?

ivirshup:
But do feel free to merge. The points above are for your consideration, but are not blocking

flying-sheep:
OK, doesn’t seem like this would merge anywhere near cleanly. Let’s not risk it.

PR #1107, opened 2023-08-28
forces Raw move to CPU

Intron7:
Hey,
so I found that the implementation I used to force Raw to the CPU was not covering all cases. I added a second check. I should now always move `.X` to the CPU.

### DIFF
--- a/anndata/tests/test_gpu.py
+++ b/anndata/tests/test_gpu.py
@@ -1,6 +1,7 @@
 import pytest
+from scipy import sparse
+
 from anndata import AnnData, Raw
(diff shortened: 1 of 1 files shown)


Intron7:
@ivirshup  I added two tests. I hope I put them into the right place.

ivirshup (approved):
Yep, I made some minor changes

* I think it's a little confusing to have names from both `cupyx.scipy.sparse` and `scipy.sparse` imported, so I've qualified their usage
* I think `isspmatrix_csr` will be deprecated, so I've replaced it with an instance check

PR #1108, opened 2023-08-28
Deprecate anndata.read

### LINKED ISSUE #1090: Deprecate `ad.read`

`ad.read` is currently a shortcut for `ad.read_h5ad`

* This is confusing since the behavior is very different than `scanpy.read` (which tries to infer the file format)
* We shouldn't export the same thing twice under two names
* `ad.read` isn't documented

I propose we deprecate this for 0.10 for removal in 0.11

ivirshup:
- [x] Closes #1090
- [x] Tests added

### DIFF
--- a/anndata/__init__.py
+++ b/anndata/__init__.py
@@ -41,8 +41,17 @@ from ._warnings import (
 )
 from . import experimental
 
-# backwards compat / shortcut for default format
-read = read_h5ad
+
+def read(*args, **kwargs):
+    import warnings
+
+    warnings.warn(
+        "`anndata.read` is deprecated, use `anndata.read_h5ad` instead. "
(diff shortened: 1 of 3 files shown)


Zethson (approved):
Interesting that it defaulted to h5ad. I expected it to be smarter and use the reader by file extension? If the latter is the case, we can't just recommend `read_h5ad` but something like `read_[filetype]`

ivirshup:
If you look at the diff, what's being replaced is `read = read_h5ad`.

I think you/ others are right to be confused by this!

Zethson:
Yeah, I noticed which prompted me to write any comment in the first place ^_^
But hey - we're solving this now.

PR #1115, opened 2023-08-31
updated _inplace

Intron7:
Here is an updated version for the inplace function `.Layers` and `.X` are now copied inplace without copy.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -136,6 +136,20 @@ def _(anno, length, index_names):
 def _(anno, length, index_names):
     raise ValueError(f"Cannot convert {type(anno)} to DataFrame")
 
+def _slice_length(slice_obj, iterable_length):
+    start, stop, step = slice_obj.indices(iterable_length)
+    return max(0, (stop - start + (step - (1 if step > 0 else -1))) // step)
+
+
+def _get_dimensions(s_object, shape):
+    if isinstance(s_object, slice):
+        return _slice_length(s_object, shape)
+    if isinstance(s_object, int):
+        return 1
+    else:
+        return len(s_object)
+
+
 
 class AnnData(metaclass=utils.DeprecationMixinMeta):
     """\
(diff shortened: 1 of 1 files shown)


ivirshup:
Does this solve the memory issues you were facing in rapids-singlecell?  Could you share some memory measurements showing this?

anndata/_core/anndata.py
```diff
@@ -1242,23 +1256,71 @@ def strings_to_categoricals(self, df: Optional[pd.DataFrame] = None):
...
+        self._var = pd.DataFrame(var_sub)
+        self._uns = uns
 
-    def _inplace_subset_obs(self, index: Index1D):
+        if self.layers:
+            for key, matrix in self.layers.items():
+                self.layers[key] = matrix[:, var_dx].reshape(self.n_obs, y_dim)
+        self._varm = self.varm._view(self, (var_dx,)).copy()
+        self._varp = self.varp._view(self, var_dx).copy()
+        self._is_view = False
```
flying-sheep, line 1289: almost identical to the other big code block, needs to be deduplicated

flying-sheep (changes requested):

Intron7:
I reran the analysis (1Million Cells) with the this code again and some things are really confusing. So the cunndata inplace works and after some promising initial results it looks like this doesn't fix the memory issue. Even though the subsetting is exactly the same. At some point I thought it had to do with views and how `.X` is presented. So I ran some more tests to see whats happening. Boolean subsets seem to work with `.X` from `anndata`. The index somehow doesn't. This doesn't explain why `cunndata` work because I also use the `_normalize_index` function from anndata.

ivirshup:
Right now I'm -1 on this PR since it's only *temporarily* lower memory usage and because it can leave the anndata in a corrupted state on failure, which is much less likely to happen in our current implementation.

Intron7:
since we found the underlying issue im fine with closing the PR

PR #1117, opened 2023-08-31
Bump min Python to 3.9

### LINKED ISSUE #1116: Drop python 3.8 support

We're well past the window for python 3.8 support (https://scientific-python.org/specs/spec-0000/#support-window), so we should drop it for the next version.

flying-sheep:
- [x] Closes #1116
- [x] Tests added

Currently, this adds all checks I think are very useful, minus the following that require more substantive changes:

- `B028`/[no-explicit-stacklevel](https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/): we should do this for more useful warning stack locations
- `FBT`/[flake8-boolean-trap](https://beta.ruff.rs/docs/rules/#flake8-boolean-trap-fbt): very useful for a future API update
- `PT`/[flake8-pytest-style](https://beta.ruff.rs/docs/rules/#flake8-pytest-style-pt) very useful, but needs a lot of changes
- `PD`/[pandas-vet](https://beta.ruff.rs/docs/rules/#pandas-vet-pd) 

The following stylistic rulesets are useful enough to go for but also need quite some changes 

- `D`/[pydocstyle](https://beta.ruff.rs/docs/rules/#pydocstyle-d)
- `RET`/[return style] Clarifies early returns (`if thing: return\nelse: ...` → `if thing: return\...`)

Instead of `FA` we just add `from __future__ import annotations` automatically everywhere.

The idea is that before the 0.10 release, we

1. `git reset --hard <first commit in this PR>` and then `git rebase main` this branch
3. re-create the “pre-commit autofixes” commit or let pre-commit handle that
4. maybe fix up some things that got less pretty, e.g.
  - wrap braces arount this type: https://github.com/scverse/anndata/blob/b746b978f5bcaaeb412d5db2d91f5b66ee0c5768/anndata/_core/index.py#L41-L47
  - manually switch from `cabc.???` to `from collections.abc import ???`
6. ~~profit~~merge


docs/release-notes/0.10.0.md
```diff
@@ -24,6 +24,8 @@
 ```{rubric} Other updates
 ```
 
+- Bump minimum python version to 3.8 {pr}`820` {user}`flying-sheep`
```
ivirshup, line 27:
```suggestion
- Bump minimum python version to 3.9 {pr}`820` {user}`flying-sheep`
```

anndata/_core/aligned_mapping.py
```diff
@@ -1,26 +1,34 @@
...
+import warnings
 from abc import ABC, abstractmethod
 from collections import abc as cabc
+from collections.abc import Iterator, Mapping, Sequence
 from copy import copy
-from typing import Union, Optional, Type, ClassVar, TypeVar  # Special types
-from typing import Iterator, Mapping, Sequence  # ABCs
-from typing import Tuple, List, Dict  # Generic base types
-import warnings
+from typing import (  # Special types  # ABCs  # Generic base types
```
ivirshup, line 8:
```suggestion
from typing import (
```

flying-sheep: those are the changes that should only be done before we merge it because they only become relevant after autofixing

flying-sheep: or maybe not, we can fix this in the commit before. hm.

flying-sheep:
pre-commit.ci autofix

ivirshup:
I've merged a version of this PR closer to the set of changes from when I reviewed last week. I did include a few more of the rules you added here.

Could you open an issue for follow up?

From time spent investigating these changes:

* "ICN" I would actually like to go a little further than this rule set goes, and be able to define our own import patterns. I think this would be useful when dealing with all the libraries that define a `sparse` or `linalg` module. I would also like to have rules that make import from anndata in the tests more like how we recommend people use them, so it's easier to copy and past code out of the tests into a repl or notebook.
* "SIM", I think I'm against this one. Some things it made better, some things it made worse. I think these should be almost always better.
* For the things around typing, I'm a little curious about them. I think I'd like to have type checking actually run before doing too much here though.

PR #1119, opened 2023-09-04
Fix garbage collection

### LINKED ISSUE #360: Anndata not properly garbage collected

ivirshup:
- [x] Closes #360
- [x] Tests added

Still needs some tests.

## Demo

<details>
<summary> Set up </summary>

```python
import anndata as ad
import numpy as np

ANNDATA_FILENAME = 'test.h5ad'

X = np.ones((10_000, 10_000))
ad.AnnData(
    X,
    layers={"X_again": X}
).write_h5ad(ANNDATA_FILENAME, compression="lzf")
del X
```

</details>

<details>
<summary>Benchmarking script </summary>

```python
import anndata as ad
import numpy as np
import os
import tracemalloc

ANNDATA_FILENAME = 'test.h5ad'
RUNS = 10

    ))
    return total

    total = np.zeros(n)
    data = func(arg).copy()
    tracemalloc.start()
    for i in range(n):
        data = func(arg).copy()
    tracemalloc.stop()
    return total

def func(pth):
    data = ad.read_h5ad(pth)
    return data[::2].copy()

total = trace_function(func, ANNDATA_FILENAME, RUNS)
total
```

</details>

Memory usage on main:

```
array([3.20191228e+09, 4.00357741e+09, 4.80520541e+09, 5.60682786e+09,
       4.00358264e+09, 4.80520492e+09, 5.60682752e+09, 6.40845017e+09,
       7.21007283e+09, 8.01169533e+09])
```

Memory usage on this branch:

```
array([2.40190353e+09, 2.40195236e+09, 2.40196104e+09, 2.40196308e+09,
       2.40196497e+09, 2.40196565e+09, 2.40196656e+09, 2.40196712e+09,
       2.40196768e+09, 2.40196820e+09])
```


ilan-gold:
Two big outstanding problems:
1. The following sort of thing no longer works
```python
AnnData.layers
```
2.  i/o memory has spiked according to benchmarking

ilan-gold:
Looking at this further: https://github.com/scverse/anndata/pull/1119/checks?check_run_id=25779040354 there is actually no improvement over `main`.   So I am not sure what's up with that.

benchmarks/benchmarks/anndata.py
```diff
@@ -0,0 +1,33 @@
...
+
+import numpy as np
+
+from .utils import gen_adata
+
+
+class GarbargeCollectionSuite:
+    runs = 10
+
+    # https://github.com/pythonprofilers/memory_profiler/issues/402 and other backend does not pick this up
```
ilan-gold, line 13: Ok @ivirshup @flying-sheep  independent of this issue, were you aware that https://pypi.org/project/memory-profiler/ is a line-by-line profiler?  I think this is probably not very good for us, no?  When I read `get_peakmem` (i.e., the function in `utils` we wrote based on `memory_profiler`), I would think that it would really track the literal peak memory, and not the peak over individual line operations.

ilan-gold: Similarly: https://scverse.zulipchat.com/#narrow/stream/393966-scanpy-anndata-dev/topic/Benchmarking/near/442513634

src/anndata/_core/aligned_mapping.py
```diff
@@ -272,20 +278,23 @@ def dim_names(self) -> pd.Index:
         return (self.parent.obs_names, self.parent.var_names)[self._axis]
 
 
+# TODO: vals can't be None
```
ilan-gold, line 281: Did this mean the opposite?  I've handled when it might be `None`

src/anndata/_core/aligned_mapping.py
```diff
@@ -272,20 +278,23 @@ def dim_names(self) -> pd.Index:
...
     ):
         self._parent = parent
         if axis not in (0, 1):
             raise ValueError()
         self._axis = axis
-        self._data = dict()
-        if vals is not None:
-            self.update(vals)
+        if vals is None:
+            vals = {}
```
ilan-gold, line 294: The codecov is real but likely was there before since the logic did not change (in theory)

src/anndata/_core/aligned_mapping.py
```diff
@@ -396,9 +415,39 @@ def __init__(
     ):
         self.parent_mapping = parent_mapping
         self._parent = parent_view
-        self.subset_idx = (subset_idx, subset_idx)
+        self.subset_idx = subset_idx
```
ilan-gold, line 418: @ivirshup I think this was you.  Why?

src/anndata/_core/aligned_mapping.py
```diff
@@ -396,9 +415,39 @@ def __init__(
...
         self._axis = parent_mapping._axis
 
 
 PairwiseArraysBase._view_class = PairwiseArraysView
 PairwiseArraysBase._actual_class = PairwiseArrays
+
+
+class AlignedMappingProperty:
```
flying-sheep, line 426: whole class is missing types

src/anndata/_core/anndata.py
```diff
@@ -675,6 +670,52 @@ def X(self, value: np.ndarray | sparse.spmatrix | SpArray | None):
...
+        """
+        return self.__obsm__alignedmapping_property
+
+    @obsm.setter
+    def obsm(self, val):
+        self.__obsm__alignedmapping_property = val
+
+    @obsm.deleter
+    def obsm(self):
+        del self.__obsm__alignedmapping_property
```
flying-sheep, line 693: You defined an accessor, why not use it directly?

```suggestion
    obsm = AlignedMappingProperty("obsm", AxisArrays, 0)
    """\
    Multi-dimensional annotation of observations
    (mutable structured :class:`~numpy.ndarray`).
    Stores for each key a two or higher-dimensional :class:`~numpy.ndarray`
    of length `n_obs`.
    Is sliced with `data` and `obs` but behaves otherwise like a :term:`mapping`.
    """
```

flying-sheep (changes requested):
Please explain what’s going on: why the changes, why do they help?

I guess that the careful “we keep a reference to the dict that’s passed in” is somehow intentional.

If so, I don’t like it. Creating a descriptor that on access creates an ephemeral class instance that just exists to mutate a `dict` feels too indirect and fragile.

src/anndata/_core/aligned_mapping.py
```diff
@@ -272,20 +278,23 @@ def dim_names(self) -> pd.Index:
...
 
 
+# TODO: vals can't be None
 class AxisArrays(AlignedActualMixin, AxisArraysBase):
     def __init__(
         self,
         parent: AnnData | Raw,
         axis: int,
-        vals: Mapping | AxisArraysBase | None = None,
+        vals: dict[str, V] | AxisArraysBase | None = None,
```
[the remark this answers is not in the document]
ilan-gold: Also, why is `Mapping` right everywhere?  Can there ever really be an immutable `vals`?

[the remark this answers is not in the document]
ilan-gold: I guess you're fix resolves that issue, actually.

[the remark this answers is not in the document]
flying-sheep: > I was trying to match this

that’s not a parameter, that’s a field. My point is that **usually**:
- function parameters should accept the most generic collections.abc interface (e.g. `Iterable` if possible, if not, then `Collection`, …)
- function return types should be concrete types
- class fields should have concrete types

> why is `Mapping` right everywhere?

because `_data: dict[...]` is a class member and `vals: Mapping[...]` is a parameter.

flying-sheep:
OK, I clarified the way this code currently works a bit.

Now it’s more clear that the code directly wants to reuse the same object that was passed in. I still don’t like the double indirection with the descriptor and the ephemeral classes though.

src/anndata/_core/anndata.py
flying-sheep: done, needed a small correction for Sphinx, that’s all.

ilan-gold: This was the reason for the boilerplate.  I couldn't get `sphinx` to work with this without some amount of pain, and then the type hints also didn't work.

src/anndata/_core/aligned_mapping.py
```diff
@@ -419,15 +424,17 @@ def __init__(
 PairwiseArraysBase._actual_class = PairwiseArrays
 
 
-class AlignedMappingProperty(property):
-    def __init__(self, name, cls, axis):
+class AlignedMappingProperty(property, Generic[T]):
+    def __init__(
+        self, name: str, cls: T, axis: Literal[0, 1] | tuple[Literal[0], Literal[1]]
```
ilan-gold, line 429: Isn't `cls` of `Type[T]`?

flying-sheep: currently T is inferred to be `type[...]` for each descriptor, e.g.

grafik

so either would be correct.

src/anndata/_core/anndata.py
flying-sheep: Now typing works, only the types in e.g. `obsm`’s documentation doesn’t work. But I think that’s fine for now.

flying-sheep: yup, we sure can

src/anndata/_core/anndata.py
```diff
@@ -675,45 +667,11 @@ def X(self, value: np.ndarray | sparse.spmatrix | SpArray | None):
...
-        self._layers = layers
-
-    @layers.deleter
-    def layers(self):
-        self.layers = dict()
+    obsm = AlignedMappingProperty("obsm", AxisArrays, 0)
+    varm = AlignedMappingProperty("varm", AxisArrays, 1)
+    layers = AlignedMappingProperty("layers", Layers, (0, 1))
+    obsp = AlignedMappingProperty("obsp", PairwiseArrays, 0)
+    varp = AlignedMappingProperty("varp", PairwiseArrays, 1)
```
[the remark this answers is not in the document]
flying-sheep: recovered

src/anndata/_core/aligned_mapping.py
ilan-gold: Ah ok, it's because of this https://github.com/scverse/anndata/pull/1119/files#diff-30201eb006d4117775a934d53f6f3474312b98528a978c41220ffa18a3360fb9R456-R457

ilan-gold (approved):

flying-sheep:
I would still like the things in here addressed: https://github.com/scverse/anndata/pull/1119#pullrequestreview-2103780966

I fixed everything that I didn’t talk about there, but
1. I’m still unclear on why the changes help / what they’re supposed to do
2. The double indirection with its “only pass by reference“ structure seems fragile and possibly too complex for me. If someone can answer 1., the complexity argument could go away, but it’s still fragile.

flying-sheep:
Hm, maybe this is just a complex problem and can’t be simplified much more without making other parts more complex.

The issue here is the complexity of the paths that data takes:

1. when initializing a new AnnData object, `AlignedMappingProperty.__set__` creates the `AnnData._{name}` attribute as a simple `dict`. `__get__` then initializes the ephemeral `AlignedMapping & *Actual`¹ objects with it.

  here my argument applies: shouldn’t there be a source of truth where that `dict` is the canonical data container and only one reference to it is stored?

2. when copying an AnnData object, we currently `.copy()` the ephemeral `AlignedMapping & *Actual` object, temporarily making it the data holder, and only then we assign the attribute on the new AnnData object in `AlignedMappingProperty.__set__`, which works, but extends the responsibility for the AxisArrays beyond “ephemeral API wrapper around `dict` stored in AnnData object.

----

¹Classes like `Layers`, … that have a `_data` attribute they share with an `AnnData` object.

flying-sheep (approved):

PR #1121, opened 2023-09-05
Fix shape inference for X=None

### LINKED ISSUE #467: Allowing `.X` to be None
Tracking issue for allowing `X` to be `None`.

When parts get ticked off they should be marked with the PR that fixed them.

# TODO:

All of these need tests, some of them may already work.

- [ ] Getter, setter, deleter
- [ ] Initializer (seems to work, needs testing)
- [ ] Views (basically just check that `.X = None` doesn't change other view behavior, `a[idx].X` doesn't do anything weird, and copying views still works)
- [ ] `_subset_inplace`
- [ ] Copying
- [ ] Backed mode
- [ ] Concatenation
- [ ] Transpose
- [ ] Methods which default to values from `X`, e.g. `obs_vector`, `to_df`.
- [ ] IO (seems to work, needs testing)
  - [ ] Main writers (`zarr`, `hdf5`)
  - [ ] Main readers
  - [ ] Specialized writers like `to_csvs`
- [ ] Raw (pretty similar set of things to be checked)
  - [ ] Initializer
  - [ ] Backed
  - [ ] Copies
  - [ ] Concatenation
  - [ ] IO

flying-sheep:
- [x] Fixes parts of #467
- [x] Tests added

TODO:

- [x] Views (basically just check that .X = None doesn't change other view behavior, a[idx].X doesn't do anything weird, and copying views still works)
  *unclear if that’s covered* @ivirshup can you clarify the scope of this?
- [x] Figure out inference of AnnData shape when X is None

Later:
- [ ] Backed mode
- [ ] Concatenation 
- [ ] IO tests
- [ ] Specialized writers like `to_csvs`
- [ ] Raw

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -103,7 +103,9 @@ def _check_2d_shape(X):
 
 
 @singledispatch
-def _gen_dataframe(anno, length, index_names):
+def _gen_dataframe(
+    anno: Mapping[str, Any], index_names: Iterable[str], length: int | None = None
+) -> pd.DataFrame:
     if anno is None or len(anno) == 0:
         anno = {}
     for index_name in index_names:
@@ -113,15 +115,27 @@ def _gen_dataframe(anno, length, index_names):
                 index=anno[index_name],
                 columns=[k for k in anno.keys() if k != index_name],
             )
-    return pd.DataFrame(
+
+    def mk_index(l: int) -> pd.Index:
+        return pd.RangeIndex(0, l, name=None).astype(str)
+
+    df = pd.DataFrame(
         anno,
-        index=pd.RangeIndex(0, length, name=None).astype(str),
+        index=None if length is None else mk_index(length),
         columns=None if len(anno) else [],
     )
+    if length is None:
+        df.index = mk_index(len(df))
+    return df
 
 
 @_gen_dataframe.register(pd.DataFrame)
-def _(anno, length, index_names):
+def _gen_dataframe_df(
+    anno: pd.DataFrame, index_names: Iterable[str], length: int | None = None
+):
+    if length is not None and length != len(anno):
+        msg = f"`shape` is inconsistent with `{index_names[0].split('_')[0]}`"
+        raise ValueError(msg)
     anno = anno.copy(deep=False)
     if not is_string_dtype(anno.index):
         warnings.warn("Transforming to str index.", ImplicitModificationWarning)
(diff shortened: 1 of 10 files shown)


anndata/_core/raw.py
```diff
@@ -34,7 +34,10 @@ def __init__(
                 self._X = X.get()
             else:
                 self._X = X
-            self._var = _gen_dataframe(var, self.X.shape[1], ["var_names"])
+            n_var = None if self._X is None else self._X.shape[1]
+            self._var = _gen_dataframe(
+                var, ["var_names"], source="X", attr="var", length=n_var
+            )
```
flying-sheep, line 40: `_gen_dataframe` used to ignore `length` before when the first argument was already a `DataFrame`.

I assume passing `self.X.shape[1]` was therefore simply wrong (note that it’s `self._X.shape[1]` now, with an underscore)

docs/release-notes/0.10.1.md
```diff
@@ -0,0 +1,4 @@
+### 0.10.1 {small}`the future`
+
+```{rubric} Bugfix
+```
```
ivirshup, line 4: Can we hold off on putting 0.10.1 file before we do a 0.10 release? I don't see any value in the 0.10.0 docs have and empty 0.10.1 section.

anndata/_core/anndata.py
```diff
@@ -473,27 +527,20 @@ def _init_as_actual(
                     X = np.array(X, dtype, copy=False)
             # data matrix and shape
             self._X = X
-            self._n_obs, self._n_vars = self._X.shape
+            n_obs, n_vars = self._X.shape
```
ivirshup, line 530:
```suggestion
            n_obs, n_vars = X.shape
```

A little more simple even?

docs/release-notes/0.10.1.md
flying-sheep: sure!

anndata/_core/anndata.py
```diff
@@ -783,12 +830,12 @@ def raw(self):
     @property
     def n_obs(self) -> int:
         """Number of observations."""
-        return self._n_obs
+        return len(self.obs)
```
ivirshup, line 833: Is this critical to the PR? 

* I can forsee issues with delayed/ out of core objects.
* I think the `len` of a dataframe is not intuitive

```python
import pandas as pd

df = pd.DataFrame({"a": [1, 2]})

assert len([i for i in df]) != len(df)
```

* I am eyeing moving towards the dataframe API, which does not define `__len__`
* It would make sense if `obs` and `var` were also views, but then `len(self.obs)` would cause compute.

As an alternative:

```suggestion
        return len(self.obs_names)
```

Since I expect those will always be around.

anndata/_core/raw.py
ivirshup: Accesses like that are often around backed mode where `_X is None` for anndata. I think it should be fine in this case.

ivirshup (changes requested):
Mostly minor changes, otherwise good.

For the subsetting + views, I think I was concerned about aligned mappings + raw with various kinds of subsetting.

anndata/_core/raw.py
flying-sheep: OK!

ivirshup (approved):

PR #1128, opened 2023-09-08
Moving source directory to `src/anndata`

ivirshup:
**Reasoning**

* Nice consistent relative positioning in the editor side bar
* Much harder to accidentally import from the code rather than installed version
* Consistent with cookie cutter

**Transition guide**

* All devs will need to `rm -r` their old anndata directory. There will be a `_version.py` file there + empty directories. I believe `conftest.py` tried to import from it.

Ping @flying-sheep


ivirshup:
Uhh, works on my machine?

ivirshup:
If I try to run the tests in a fresh clone, but using my old install, I can reproduce. Editable install of fresh clone then running the fresh clones tests works.

Probably not following this up much more tonight.

flying-sheep:
Superseded by #1128

PR #1131, opened 2023-09-09
Optimize contiguous sparse slicing

ivirshup:
Significant speed up for contiguous slices

I had totally thought this was already implemented...

TODO:

- [x] Benchmarks

* deferred
  - ~[ ] Reverse slicing~
  - ~[ ] Steps other that 1?~

It's probably optimal to do this up to pretty large step sizes. So long as two adjacent rows are in the same slice we should probably take this approach.

cc: @ilan-gold

Benchmark:

```python
from anndata.experimental import write_elem, sparse_dataset
from scipy import sparse
import numpy as np
import zarr

z = zarr.open("test.zarr")
write_elem(
    z,
    "/",
    sparse.random(
        100_000,
        10_000,
        format="csr",
        random_state=np.random.default_rng(),
    )
)

x = sparse_dataset(z["/"])
%time x[10000:20000]
```

This PR:

```
CPU times: user 9.57 ms, sys: 5.09 ms, total: 14.7 ms
Wall time: 14.4 ms
<10000x10000 sparse matrix of type '<class 'numpy.float64'>'
	with 999208 stored elements in Compressed Sparse Row format>
```

Main:

```
CPU times: user 9.35 s, sys: 5.14 s, total: 14.5 s
Wall time: 14.8 s
<10000x10000 sparse matrix of type '<class 'numpy.float64'>'
	with 999208 stored elements in Compressed Sparse Row format>
```

### DIFF
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -105,6 +105,25 @@ def _normalize_index(
         raise IndexError(f"Unknown indexer {indexer!r} of type {type(indexer)}")
 
 
+def _normalize_slice(s: slice, length: int):
+    """Normalize a slice for a given length.
+
+    This means that the slice will be clipped to the length of the object, and the step won't be None.
+    """
+    step = s.step if s.step is not None else 1
+
+    # slice constructor would have errored if step was 0
+    if step > 0:
+        start = s.start if s.start is not None else 0
+        stop = s.stop if s.stop is not None else length
+    elif step < 0:
+        # Reverse
+        start = s.start if s.start is not None else length
+        stop = s.stop if s.stop is not None else 0
+
+    return slice(start, stop, step)
+
+
 def unpack_index(index: Index) -> tuple[Index1D, Index1D]:
     if not isinstance(index, tuple):
         return index, slice(None)
--- a/anndata/_core/index.py
+++ b/anndata/_core/index.py
@@ -105,10 +105,10 @@ def _normalize_index(
         raise IndexError(f"Unknown indexer {indexer!r} of type {type(indexer)}")
 
 
-def _normalize_slice(s: slice, length: int):
-    """Normalize a slice for a given length.
+def _fix_slice_bounds(s: slice, length: int) -> slice:
+    """The slice will be clipped to length, and the step won't be None.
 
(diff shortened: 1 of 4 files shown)


anndata/_core/index.py
```diff
@@ -105,6 +105,25 @@ def name_idx(i):
         raise IndexError(f"Unknown indexer {indexer!r} of type {type(indexer)}")
 
 
+def _normalize_slice(s: slice, length: int):
```
ilan-gold, line 108: maybe `_clip_slice_to_length`? I'm not sure what "normalize" means.

anndata/_core/sparse_dataset.py
```diff
@@ -131,13 +147,21 @@ def _get_intXslice(self, row: int, col: slice) -> ss.csr_matrix:
...
         )
+
         if out_shape[0] == 1:
             return self._get_intXslice(slice_as_int(row, self.shape[0]), col)
         elif out_shape[1] == self.shape[1] and out_shape[0] < self.shape[0]:
+            if row.step == 1:
+                return ss.csr_matrix(
+                    self._get_contiguous_compressed_slice(row), shape=out_shape
+                )
             return self._get_arrayXslice(np.arange(*row.indices(self.shape[0])), col)
```
ilan-gold, line 165: if it is -1, why can't we do this process in reverse?

ivirshup: It's a little more complicated to do this in reverse, since it's not just a contiguous slice of the underlying arrays (it could be in reverse, but then it's not sorted). I'm going to give this a shot, but may end up only optimizing forward slices in this PR.

anndata/_core/index.py
ivirshup: This also remove's any `None` values from the slice. I guess I think of a slice with all elements set as the "normal form" of the slice. 

Numba calls this: `_normalize_slice` or `fix_slice`.

anndata/_core/sparse_dataset.py
ivirshup: I'm going to leave this for now, just for expediency. Please do make a PR on this!

PR #1151, opened 2023-10-02
Moving source directory to src/anndata

flying-sheep:
Supersedes #1128


flying-sheep:
The issue is that doctests make the testing process wonky, as they import their parent modules.

And if we don’t use editable installs in testing, the parent modules are the one in the project directory while there’s a second copy installed to the site-packages directory.

Using `src` layout, editable installs are fine for testing, as the `src` directory gets added to`sys.path`, not the project directory. And the `src` directory should be clean while the project directory potentially contains a bunch of stuff that python interprets as modules.

flying-sheep:
Currently, there seems to be no way to make pytest behave, so let’s table this. https://github.com/pytest-dev/pytest/issues/11475

ivirshup (changes requested):
I would like to give this a shot locally before merging

pyproject.toml
```diff
@@ -118,8 +124,10 @@ exclude_also = [
 
 [tool.pytest.ini_options]
 addopts = [
+    "--import-mode=importlib",
     "--strict-markers",
     "--doctest-modules",
+    "--pyargs",
```
ivirshup, line 130: I'm not sure I totally get this argument. What does this do other than let us use `anndata` instead of `./src/anndata` in `testpaths`?

ivirshup (changes requested):
Found one definite bug, one possible bug trying this out locally:

## Definite bug

I seem to be collecting different numbers of tests between this, and the most recent parent commit on main: 20adb70a

In this branch I collect: 3878 tests, on 20adb70a I get 3890

I think the issue is the doctests within individual function docstrings.

## Possible bug or dep issue

I get some errors during collection. Here's the traceback

```
========================================================================================= ERRORS =========================================================================================
_____________________________________________________________________________ ERROR collecting test session ______________________________________________________________________________
../../mambaforge/envs/anndata-dev/lib/python3.12/site-packages/pluggy/_manager.py:168: in register
    hook._maybe_apply_history(hookimpl)
../../mambaforge/envs/anndata-dev/lib/python3.12/site-packages/pluggy/_hooks.py:569: in _maybe_apply_history
    res = self._hookexec(self.name, [method], kwargs, False)
../../mambaforge/envs/anndata-dev/lib/python3.12/site-packages/pluggy/_manager.py:119: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
../../repos/anndata/conftest.py:62: in pytest_addoption
    parser.addoption(
../../mambaforge/envs/anndata-dev/lib/python3.12/site-packages/_pytest/config/argparsing.py:107: in addoption
    self._anonymous.addoption(*opts, **attrs)
../../mambaforge/envs/anndata-dev/lib/python3.12/site-packages/_pytest/config/argparsing.py:386: in addoption
    raise ValueError("option names %s already added" % conflict)
E   ValueError: option names {'--strict-warnings'} already added
```

I didn't get this when I was initially trying this with an outdated env, but I was getting some other collection issue there.


------


More of a note, but switching to this branch requires:

* clearing out the old directory `rm -r anndata`
* re-installing

pyproject.toml
flying-sheep: because doctests are weird and would otherwise result in importing some modules as files and others via `sys.path`.

flying-sheep:
Without `--pyargs`, both seems to behave the same.

ivirshup:
@flying-sheep is this ready for me to look at again?

flying-sheep:
Of course

flying-sheep:
still the case

pyproject.toml
```diff
@@ -119,8 +125,10 @@ exclude_also = [
 
 [tool.pytest.ini_options]
 addopts = [
+    # "--import-mode=importlib",  # TODO: enable
```
ivirshup, line 128: Some explanation of why, link to a relevant issue here would be good

flying-sheep: because you said so, there’s no issue that I’m aware of

flying-sheep:
OK, I check if we run enough tests and link the issue.

ivirshup:
@flying-sheep just a heads up that CI is failing here

.azure-pipelines.yml
```diff
@@ -84,6 +85,16 @@ jobs:
...
         condition: eq(variables['TEST_TYPE'], 'strict-warning')
 
+      # TODO: remove: https://github.com/scverse/anndata/issues/1451
+      - task: PythonScript@0
+        inputs:
+          scriptSource: inline
+          script: |
+            import sys, xml.etree.ElementTree as ET
+            results = ET.parse("./test-data/test-results.xml").findall("./*/*")
+            sys.exit(0 if len(results) > 100 else f"Error: only {len(results)} tests run")
```
ivirshup, line 95: If you wanted to bump this, we're at 3892 tests right now. So something around there. Should be fine otherwise.

ivirshup (approved):
lgtm

Once it's merged maybe send  out a message on zulip with instructions on how to

* Update a PR
* switch over to this (I think just re-install and `rm -r anndata`?)

flying-sheep:
doesn’t matter if it’s in the release or not, but it should be backported so further backports are easier

PR #1161, opened 2023-10-05
Add-concat-on-disk-examples

### LINKED ISSUE #1160: “How to” example for concat_on_disk

Add example(s) to [anndata.experimental.concat_on_disk](https://anndata.readthedocs.io/en/latest/generated/anndata.experimental.concat_on_disk.html).

Maybe using [pooch](https://pypi.org/project/pooch/), maybe just toy examples

flying-sheep:
- [x] Closes #1160
- [x] Tests added (automatic doctest)

REST API docs: https://api.cellxgene.cziscience.com/curation/ui/#/

Uses a simpler version of https://gist.github.com/ivirshup/f1a1603db69de3888eacb4bdb6a9317a (maybe they also changed the schema since then)

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1681,7 +1681,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         ...     dict(var_names=['d', 'c', 'b'], annoA=[0, 1, 2]),
         ... )
         >>> adata3 = AnnData(
-        ... np.array([[1, 2, 3], [4, 5, 6]]),
+        ...     np.array([[1, 2, 3], [4, 5, 6]]),
         ...     dict(obs_names=['s1', 's2'], anno2=['d3', 'd4']),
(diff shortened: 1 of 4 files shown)


selmanozleyen:
> @selmanozleyen We don’t have the `overwrite` argument in other commands. Is there a specific reason why you included it?

Only to speed up development.

selmanozleyen:
> @selmanozleyen We don’t have the `overwrite` argument in other commands. Is there a specific reason why you included it?

Ignore my previous message please. Sorry I confused the branches. I just wanted it to have that option as only taking filenames seemed too restrictive. But in theory the user can give zarr/h5py objects as input too so it shouldn't be a problem.

ivirshup:
Made some minor changes:

* `concat_on_disk` now errors if there are no objects to concatenate (like `ad.concat`, `pd.concat`, `xr.concat`, `np.concatenate`)
* `concat_on_disk` now errors if `out_file`'s parent directory does not exist (like `anndata.write_h5ad`, `pd.DataFrame.to_csv`)

ivirshup (approved):

PR #1162, opened 2023-10-05
Add zarr examples to fileformat docs

### LINKED ISSUE #1155: Extend documentation on Zarr AnnData disk storage format
Following discussions on table support and the recent closure of the NGFF table spec PR, it's been noted that there isn't an "official" source detailing the AnnData disk storage specifically for the Zarr format to the needed extent.

While the current AnnData documentation has a section on disk storage, it provides a broad overview without specific focus on Zarr. 

Given that developers, especially those less familiar with Python, could greatly benefit from such a resource, it's suggested to add a new section, possibly on this page: https://anndata.readthedocs.io/en/latest/fileformat-prose.html. This section could lay out the Zarr format's tree-structure, or include scripts demonstrating how to retrieve some complex, representative example datasets.

A possibility would be to use [sphinx-tabs](https://github.com/executablebooks/sphinx-tabs) and have a tab showing HDF5 and one showing Zarr.

This would essentially help in formalizing and standardizing the Zarr AnnData format, making it more accessible and comprehensible.

CC @ivirshup @joshmoore

*Note: message co-written with ChatGPT.*

ivirshup:
- [x] Closes #1155
- [ ] ~Tests added~

Added zarr examples to file format docs. I should maybe use a smaller file for the example so it's easier to distribute.

### DIFF
--- a/docs/fileformat-prose.md
+++ b/docs/fileformat-prose.md
@@ -232,6 +232,7 @@ ensembl_id <HDF5 dataset "ensembl_id": shape (40145,), type "|O">
 feature_biotype <HDF5 group "/var/feature_biotype" (2 members)>
 feature_biotype/categories <HDF5 dataset "categories": shape (1,), type "|O">
 feature_biotype/codes <HDF5 dataset "codes": shape (40145,), type "|i1">
+feature_is_filtered <HDF5 dataset "feature_is_filtered": shape (40145,), type "|b1">
 ...
 ```
 
--- a/docs/fileformat-prose.md
+++ b/docs/fileformat-prose.md
@@ -100,7 +100,7 @@ We can see an example of this with dimensionality reductions stored in the `obsm
 :sync: hdf5
 
 ```python
->>> f["obsm/X_pca"]
+>>> store["obsm/X_pca"]
 <HDF5 dataset "X_pca": shape (164114, 50), type "<f4">
 ```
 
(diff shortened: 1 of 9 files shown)


docs/fileformat-prose.md
```diff
@@ -10,21 +10,42 @@ AnnData objects are saved on disk to hierarchical array stores like [HDF5]
 (via {doc}`H5py <h5py:index>`) and {doc}`zarr:index`.
 This allows us to have very similar structures in disk and on memory.
 
-As an example we’ll look into a typical `.h5ad` object that’s been through an analysis.
```
LucaMarconato, line 13: I'd add a link to download the dataset here as it's not immediate how to actually access the dataset being considered.

docs/fileformat-prose.md
```diff
@@ -148,14 +203,17 @@ DataFrames are saved as a columnar format in a group, so each column of a DataFr
 We save a little more information in the attributes here.
 
 ```python
```
LucaMarconato, line 205: I'd consider adding the HDF5 + Zarr example also for dataframes, but it's quite straighforward, so this could be addressed even just by saying "each column of a DataFrame is saved as a separate array (e.g. /obs/n_genes_by_counts)"

docs/fileformat-prose.md
```diff
@@ -148,14 +203,17 @@ DataFrames are saved as a columnar format in a group, so each column of a DataFr
...
->>> dict(f["obs"].attrs)
-{'_index': 'Cell',
```
LucaMarconato, line 152: is the name of the Zarr group _index or Cell?

docs/fileformat-prose.md
```diff
@@ -209,12 +292,38 @@ pca/variance_ratio <HDF5 dataset "variance_ratio": shape (50,), type "<f4">
...
+```
+
+````
+
+````{tab-item} Zarr
+:sync: zarr
+
+```python
+>>> store["uns/neighbors/params"].visititems(print)
+method <zarr.core.Array '/uns/neighbors/params/method' () <U4 read-only>
```
LucaMarconato, line 315: for HDF5 the types for method and metric are |0 and |0, while for Zarr are different. Is this correct?

docs/fileformat-prose.md
```diff
@@ -283,20 +412,56 @@ We support IO with Pandas nullable integer and boolean arrays.
 We represent these on disk similar to `numpy` masked arrays, `julia` nullable arrays, or `arrow` validity bitmaps (see {issue}`504` for more discussion).
```
LucaMarconato, line 412: Can't add the comment in the right place. I'd consider adding an example for "String array specification", maybe not in the doc and just in the actual downloadable Zarr file.

docs/fileformat-prose.md
LucaMarconato: this both for the .h5ad and the .zarr file

LucaMarconato (approved):
Hi, very nice and clear! I pre-approve. The most important comment is that I would add a link the beginning to download the .h5ad and the .zarr files. The rest of the comments are minor.

docs/fileformat-prose.md
ivirshup: I'm going to leave this to a follow up. I'd rather get this in, than wait on figuring out hosting + making a smaller dataset (this one is like 6gb).

docs/fileformat-prose.md
ivirshup: The array coressponding to `adata.obs.index` is stored at `"obs/Cell"`, but I think this is documented adequately in the specification definition below.

docs/fileformat-prose.md
ivirshup: From the specification below:

> In zarr, scalar strings MUST be stored as a fixed length unicode dtype

IIRC think this was an issue with how zarr v2 does UTF8.

PR #1169, opened 2023-10-06
Fix: Dense OOC Concat With Aligned Elements Memory Issue

selmanozleyen:
Please see https://github.com/scverse/anndata-tutorials/pull/18#issuecomment-1751284746. This PR should fix this memory issue.

Ping: @ivirshup @flying-sheep

### DIFF
--- a/anndata/experimental/merge.py
+++ b/anndata/experimental/merge.py
@@ -172,12 +172,10 @@ def write_concat_dense(
     """
     import dask.array as da
 
-    chunks = "auto"
-    if arrays[0].chunks is not None and all(
-        a.chunks == arrays[0].chunks for a in arrays
-    ):
-        chunks = arrays[0].chunks
-    darrays = (da.from_array(a, chunks=chunks) for a in arrays)
+    darrays = (
+        da.from_array(a, chunks="auto" if a.chunks is None else a.chunks)
+        for a in arrays
+    )
 
     res = da.concatenate(
         [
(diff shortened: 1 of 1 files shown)


ivirshup:
@selmanozleyen, could you make a quick demonstration of the performance benefits from this?

selmanozleyen:
Hey, I modified the fix as it seemed like it would expand to more cases (e.g. when they are all not the same).  The changes on results can be seen from the tutorial notebook https://github.com/scverse/anndata-tutorials/pull/18/commits/b12e539cd0a73aff1e304706a24bff3684756a4c#r1349746130

ivirshup:
From previous discussion about this, my impression was that `da.from_array(a, chunks="auto")` would use `a.chunks` as the chunks if it was defined.

I will try and take a look, but it looks to me like these changes could be coming from the use of `write_chunked`.

selmanozleyen:
> my impression was that `da.from_array(a, chunks="auto")` would use `a.chunks` as the chunks if it was defined.
> 
> I will try and take a look, but it looks to me like these changes could be coming from the use of `write_chunked`.

I see, I initially tried that but it didn't work for some reason. If I switch to main branch the results would be the same.

flying-sheep:
That tracks with the documentation:

> The word “auto” which acts like the above, but uses a configuration value `array.chunk-size` for the chunk size

No word of reusing the array chunks.

flying-sheep (approved):

ivirshup:
@flying-sheep, shouldn't this get a release note and a test?

PR #1182, opened 2023-10-10
Treat warnings as errors in tests

### LINKED ISSUE #1180: Catch lots of subtle bugs by setting -Werror in tests

### Report

Code:

```console
$ pytest -Werror
1600 failed
```

### Versions

<details>
<summary>Details</summary>

```
-----
anndata             0.10.0.dev93+gddba0bc2
session_info        1.0.0
-----
asciitree           NA
awkward             2.4.2
awkward_cpp         NA
cloudpickle         2.2.1
cython_runtime      NA
dask                2023.9.1
dateutil            2.8.2
entrypoints         0.4
fasteners           0.18
h5py                3.9.0
importlib_metadata  NA
jinja2              3.1.2
markupsafe          2.1.3
mpl_toolkits        NA
msgpack             1.0.5
natsort             8.4.0
numcodecs           0.11.0
numpy               1.24.4
packaging           23.1
pandas              2.1.0
psutil              5.9.5
pyarrow             13.0.0
pytz                2023.3.post1
scipy               1.11.2
six                 1.16.0
tblib               2.0.0
tlz                 0.12.0
toolz               0.12.0
yaml                6.0.1
zarr                2.16.1
zipp                NA
-----
Python 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801]
Linux-6.5.6-zen2-1-zen-x86_64-with-glibc2.38
-----
Session information updated at 2023-10-10 12:27
```

</details>

flying-sheep:
- [x] Closes #1180
- [x] Tests added

Huh, this was surprisingly low pain.

Deprecation/change fixes:

- [x] `pd.Categorical.map`’s `na_action` parameter default was changed to `None` ([link](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v2.1.0.html#new-dataframe-map-method-and-support-for-extensionarrays)). This sets it explicitly to the old default. We should probably change it to `None` and add a test.
- We now use `iloc` for `pd.Series` where appropriate
- `pd.concat` doesn’t like empty DataFrames, so we filter them before passing them now: https://github.com/pandas-dev/pandas/blob/2a65fdd227734cbda5d8e33171e857ab4aaeb9d5/pandas/core/internals/concat.py#L488-L500
- [x] `awkward.to_categorical` was deprecated in favor of `awkward.str.to_categorical`, which depends on pyarrow. Is making `pyarrow` a test dep OK or should we ignore/circuvent this?

Bugfixes

- `write_h5ad`’s `as_dense` failed to add IOSpec attrs
- `ExperimentalFeatureWarning`
  1. got silenced after the first occurrence even in tests. Now it’s only silenced in non-test code after the first occurrence.
  2. one reason why it got raised so much in the first place is that we initialize `AnnData` objects in functions and methods like `concat`, `copy`, or `read_*`. All these places now suppress the warning.
- `re_raise_error` made the traceback harder to understand, so I replaced it with `add_key_note`, which doesn’t raise the error. Raising it from inside the `except` block using plain `raise` works as expected.

Deferred

- `adata.uns[key] = ak.Array(...)` didn’t raise a warning and doesn’t do so now.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,20 +63,7 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
-      # TODO: centralize using a custom pytes CLI arg, because:
-      #       1. This allows to easily locally run with the same settings as this job
-      #       2. Otherwise we can’t filter anndata warning classes directly,
-      #          because we’re neither using editable installs nor a src/ layout.
-      #          This is better because it’s a more specific filter and we don’t need one line per message.
-      #          Setting the filters in code instead of per `-W` fixes that.
-      #          This affects:
-      #          - `anndata._warnings.ExperimentalFeatureWarning`
-      #            - “Support for Awkward Arrays”
-      #            - “Outer joins on awkward.Arrays”
-      #          - `anndata._warnings.ImplicitModificationWarning`
-      #            - “Trying to modify attribute”
-      #            - “Transforming to str index”
-      #       3. The 'ignore' lines need to be kept in sync with `filterwarnings` config in pyproject.toml
+      # TODO: Centralize, see https://github.com/scverse/anndata/issues/1204
       - script: >
           pytest
           -W error
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,12 +63,25 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
-      # replacing “:Setting element”, “:Trying to modify attribute” and “:Transforming to str index”
-      # with “::anndata._warnings.ImplicitModificationWarning” breaks,
-      # because we’re neither using editable installs nor a src/ layout
+      # TODO: centralize using a custom pytes CLI arg, because:
+      #       1. This allows to easily locally run with the same settings as this job
+      #       2. Otherwise we can’t filter anndata warning classes directly,
+      #          because we’re neither using editable installs nor a src/ layout.
+      #          This is better because it’s a more specific filter and we don’t need one line per message.
+      #          Setting the filters in code instead of per `-W` fixes that.
+      #          This affects:
+      #          - `anndata._warnings.ExperimentalFeatureWarning`
+      #            - “Support for Awkward Arrays”
+      #            - “Outer joins on awkward.Arrays”
+      #          - `anndata._warnings.ImplicitModificationWarning`
+      #            - “Trying to modify attribute”
+      #            - “Transforming to str index”
+      #       3. The 'ignore' lines need to be kept in sync with `filterwarnings` config in pyproject.toml
       - script: >
           pytest
           -W error
+          -W 'ignore:Support for Awkward Arrays is currently experimental'
+          -W 'ignore:Outer joins on awkward.Arrays'
           -W 'default:Setting element:UserWarning'
           -W 'default:Trying to modify attribute:UserWarning'
           -W 'default:Transforming to str index:UserWarning'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,12 +63,13 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
-      # replacing “:Trying to modify attribute” and “:Transforming to str index”
+      # replacing “:Setting element”, “:Trying to modify attribute” and “:Transforming to str index”
       # with “::anndata._warnings.ImplicitModificationWarning” breaks,
       # because we’re neither using editable installs nor a src/ layout
       - script: >
           pytest
           -W error
+          -W 'default:Setting element:UserWarning'
           -W 'default:Trying to modify attribute:UserWarning'
           -W 'default:Transforming to str index:UserWarning'
           -W 'default:Observation names are not unique. To make them unique:UserWarning'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,13 +63,13 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
-      # replacing “:initializing view as actual” and “:Transforming to str index”
+      # replacing “:Trying to modify attribute” and “:Transforming to str index”
       # with “::anndata._warnings.ImplicitModificationWarning” breaks,
       # because we’re neither using editable installs nor a src/ layout
       - script: >
           pytest
           -W error
-          -W 'default:initializing view as actual:UserWarning'
+          -W 'default:Trying to modify attribute:UserWarning'
           -W 'default:Transforming to str index:UserWarning'
           -W 'default:Observation names are not unique. To make them unique:UserWarning'
           -W 'default:Variable names are not unique. To make them unique:UserWarning'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -55,12 +55,12 @@ jobs:
       - script: |
           pytest
         displayName: "PyTest"
-        condition: eq(variables['RUN_COVERAGE'], 'no') && eq(variables['PRERELEASE_DEPENDENCIES'], 'no')
+        condition: and(eq(variables['RUN_COVERAGE'], 'no'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       - script: |
           pytest --cov --cov-report=xml --cov-context=test
         displayName: "PyTest (coverage)"
-        condition: eq(variables['RUN_COVERAGE'], 'yes') && eq(variables['PRERELEASE_DEPENDENCIES'], 'no')
+        condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
       - script: >
@@ -72,7 +72,7 @@ jobs:
           -W 'default::scipy.sparse._base.SparseEfficiencyWarning'
           -W 'default::dask.array.core.PerformanceWarning'
         displayName: "PyTest (treat warnings as errors)"
-        condition: eq(variables['RUN_COVERAGE'], 'no') && eq(variables['PRERELEASE_DEPENDENCIES'], 'yes')
+        condition: and(eq(variables['RUN_COVERAGE'], 'no'), eq(variables['PRERELEASE_DEPENDENCIES'], 'yes'))
 
       - task: PublishCodeCoverageResults@1
         inputs:
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -55,12 +55,24 @@ jobs:
       - script: |
           pytest
         displayName: "PyTest"
-        condition: eq(variables['RUN_COVERAGE'], 'no')
+        condition: eq(variables['RUN_COVERAGE'], 'no') && eq(variables['PRERELEASE_DEPENDENCIES'], 'no')
 
       - script: |
           pytest --cov --cov-report=xml --cov-context=test
         displayName: "PyTest (coverage)"
-        condition: eq(variables['RUN_COVERAGE'], 'yes')
+        condition: eq(variables['RUN_COVERAGE'], 'yes') && eq(variables['PRERELEASE_DEPENDENCIES'], 'no')
+
+      # TODO: fix all the exceptions here
+      - script: >
+          pytest
+          -W error
+          -W 'default::anndata._warnings.ImplicitModificationWarning'
+          -W 'default:Observation names are not unique. To make them unique:UserWarning'
+          -W 'default:Variable names are not unique. To make them unique:UserWarning'
+          -W 'default::scipy.sparse._base.SparseEfficiencyWarning'
+          -W 'default::dask.array.core.PerformanceWarning'
+        displayName: "PyTest (treat warnings as errors)"
+        condition: eq(variables['RUN_COVERAGE'], 'no') && eq(variables['PRERELEASE_DEPENDENCIES'], 'yes')
 
       - task: PublishCodeCoverageResults@1
         inputs:
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,10 +63,12 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
+      # replacing “:initializing view as actual:UserWarning“ with “::anndata._warnings.ImplicitModificationWarning”
+      # breaks because we’re not using editable installs or a src/ layout
       - script: >
           pytest
           -W error
-          -W 'default::anndata._warnings.ImplicitModificationWarning'
+          -W 'default:initializing view as actual:UserWarning'
           -W 'default:Observation names are not unique. To make them unique:UserWarning'
           -W 'default:Variable names are not unique. To make them unique:UserWarning'
           -W 'default::scipy.sparse._base.SparseEfficiencyWarning'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -63,12 +63,14 @@ jobs:
         condition: and(eq(variables['RUN_COVERAGE'], 'yes'), eq(variables['PRERELEASE_DEPENDENCIES'], 'no'))
 
       # TODO: fix all the exceptions here
-      # replacing “:initializing view as actual:UserWarning“ with “::anndata._warnings.ImplicitModificationWarning”
-      # breaks because we’re not using editable installs or a src/ layout
+      # replacing “:initializing view as actual” and “:Transforming to str index”
+      # with “::anndata._warnings.ImplicitModificationWarning” breaks,
+      # because we’re neither using editable installs nor a src/ layout
       - script: >
           pytest
           -W error
           -W 'default:initializing view as actual:UserWarning'
+          -W 'default:Transforming to str index:UserWarning'
           -W 'default:Observation names are not unique. To make them unique:UserWarning'
           -W 'default:Variable names are not unique. To make them unique:UserWarning'
           -W 'default::scipy.sparse._base.SparseEfficiencyWarning'
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -30,7 +30,6 @@ from scipy.sparse import csr_matrix, issparse
 from anndata._warnings import ImplicitModificationWarning
 
 from .. import utils
-from .._warnings import ExperimentalFeatureWarning
 from ..compat import (
     CupyArray,
     CupySparseMatrix,
@@ -896,9 +895,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
     def _set_dim_index(self, value: pd.Index, attr: str):
         # Assumes _prep_dim_index has been run
         if self.is_view:
-            with warnings.catch_warnings():
-                warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
-                self._init_as_actual(self.copy())
+            self._init_as_actual(self.copy())
         getattr(self, attr).index = value
         for v in getattr(self, f"{attr}m").values():
             if isinstance(v, pd.DataFrame):
@@ -1305,9 +1302,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         """
         adata_subset = self[:, index].copy()
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
-            self._init_as_actual(adata_subset)
+        self._init_as_actual(adata_subset)
 
     def _inplace_subset_obs(self, index: Index1D):
         """\
@@ -1317,9 +1312,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         """
         adata_subset = self[index].copy()
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
-            self._init_as_actual(adata_subset)
+        self._init_as_actual(adata_subset)
 
     # TODO: Update, possibly remove
     def __setitem__(
@@ -1357,20 +1350,18 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 "which is currently not implemented. Call `.copy()` before transposing."
             )
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
-            return AnnData(
-                X=_safe_transpose(X) if X is not None else None,
-                layers={k: _safe_transpose(v) for k, v in self.layers.items()},
-                obs=self.var,
-                var=self.obs,
-                uns=self._uns,
-                obsm=self._varm,
-                varm=self._obsm,
-                obsp=self._varp,
-                varp=self._obsp,
-                filename=self.filename,
-            )
+        return AnnData(
+            X=_safe_transpose(X) if X is not None else None,
+            layers={k: _safe_transpose(v) for k, v in self.layers.items()},
+            obs=self.var,
+            var=self.obs,
+            uns=self._uns,
+            obsm=self._varm,
+            varm=self._obsm,
+            obsp=self._varp,
+            varp=self._obsp,
+            filename=self.filename,
+        )
 
     T = property(transpose)
 
@@ -1578,9 +1569,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
         if self.isbacked:
             self.file.close()
 
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
-            return AnnData(**new)
+        return AnnData(**new)
 
     def copy(self, filename: PathLike | None = None) -> AnnData:
         """Full copy, optionally on disk."""
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -40,7 +40,7 @@ from ..compat import (
     _move_adj_mtx,
 )
 from ..logging import anndata_logger as logger
-from ..utils import convert_to_dict, dim_len, ensure_df_homogeneous
+from ..utils import convert_to_dict, deprecated, dim_len, ensure_df_homogeneous
 from .access import ElementRef
 from .aligned_mapping import (
     AxisArrays,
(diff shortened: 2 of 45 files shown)


pyproject.toml
```diff
@@ -104,13 +105,32 @@ version-file = "anndata/_version.py"
...
+    'ignore:`anndata\.read` is deprecated:FutureWarning',
+    # We test our experimental features too
+    'ignore::anndata._warnings.ExperimentalFeatureWarning',
+    # Third party issues
+    'ignore::numba.core.errors.NumbaDeprecationWarning', # we don’t import numba
+    # TODO: Catch the following ones in tests or fix them
+    'default::anndata._warnings.ImplicitModificationWarning',
+    'default:.*names are not unique. To make them unique:UserWarning',
+    'default::scipy.sparse._base.SparseEfficiencyWarning',
+    'default::dask.array.core.PerformanceWarning',
```
flying-sheep, line 132: - our `ImplicitModificationWarning` and `names are not unique` warnings should be fixed in lib code and caught in test code so we can be sure none of our functions/methods invoke them, only user/test code.
- scipy `SparseEfficiencyWarning` and dask `PerformanceWarning` should be fixed, they probably indicate real performance issues

anndata/_core/anndata.py
```diff
@@ -1685,6 +1704,14 @@ def concatenate(
...
         ...     dict(var_names=['d', 'c', 'b'], annoA=[0, 2, 3], annoB=[0, 1, 2]),
         ... )
+
+        They can now be concatenated. Since this method is deprecated, it will raise a warning:
+
+        >>> warnings.filterwarnings(
+        ...     'ignore',
+        ...     r'The AnnData\\.concatenate method is deprecated',
+        ...     FutureWarning,
+        ... )
```
ivirshup, line 1714: I don't think we should show people how to turn off the warnings in documentation for deprecated methods. It suggests that they should turn off the warnings instead of just changing their usage.

anndata/_core/merge.py
```diff
@@ -206,7 +207,7 @@ def as_cp_sparse(x) -> CupySparseMatrix:
         return cpsparse.csr_matrix(x)
 
 
-def unify_dtypes(dfs: Iterable[pd.DataFrame]) -> list[pd.DataFrame]:
+def _unify_dtypes(dfs: Iterable[pd.DataFrame]) -> list[pd.DataFrame]:
```
ivirshup, line 210: Why change the name?

anndata/_core/merge.py
```diff
@@ -240,6 +241,17 @@ def unify_dtypes(dfs: Iterable[pd.DataFrame]) -> list[pd.DataFrame]:
...
 
 
+def concat_with_unified_dtypes(
+    dfs: Iterable[pd.DataFrame],
+    *,
+    join: Literal["inner", "outer"] = "outer",
+    axis: Literal[0, 1] = 0,
+) -> pd.DataFrame:
+    dfs = _unify_dtypes(dfs)
+    nonempty_dfs = [df for df in dfs if len(df)]
```
ivirshup, line 251: TODO for me: figure out if this has any effect. Also probably better to filter *before* dtype unification.

But then this can just be:

```python
pd.concat(unify_dtypes(filter(len, dfs)), ...)
```

And doesn't need to be a function.

ivirshup: I also have not been able to reproduce this warning so far. Which tests were you getting this from @flying-sheep?

You mentioned: https://github.com/pandas-dev/pandas/blob/2a65fdd227734cbda5d8e33171e857ab4aaeb9d5/pandas/core/internals/concat.py#L488-L500 and `_concatenate_join_units`, but I'm not seeing the connection. I don't immediately see how the referenced code relates to dataframes with no rows.

anndata/_core/merge.py
```diff
@@ -1268,37 +1276,30 @@ def concat(
...
+        concat_aligned_mapping = partial(
+            outer_concat_aligned_mapping, fill_value=fill_value
         )
-        concat_mapping = outer_concat_aligned_mapping(
-            [getattr(a, f"{dim}m") for a in adatas],
-            index=concat_indices,
-            fill_value=fill_value,
+        join_keys = union_keys
+    else:
+        assert False
```
ivirshup, line 1286: This isn't useful.

If you want to validate the `join` argument, this should go with the other validation code at the top of the function.

pyproject.toml
```diff
@@ -88,9 +88,10 @@ test = [
     "joblib",
     "boltons",
     "scanpy",
-    "httpx", # For data downloading
+    "httpx",                   # For data downloading
     "dask[array,distributed]",
     "awkward>=2.3",
+    "pyarrow",                 # For non-deprecated awkward array functionality
```
ivirshup, line 94:
```suggestion
    "pyarrow",
```

I find the wording of this comment confusing. I would also be fine with no comment since I think pyarrow as a test dependency will be coming for other reasons soon anyways.

.azure-pipelines.yml
```diff
@@ -55,12 +55,28 @@ jobs:
...
+      # because we’re neither using editable installs nor a src/ layout
+      - script: >
+          pytest
+          -W error
+          -W 'default:initializing view as actual:UserWarning'
+          -W 'default:Transforming to str index:UserWarning'
+          -W 'default:Observation names are not unique. To make them unique:UserWarning'
+          -W 'default:Variable names are not unique. To make them unique:UserWarning'
+          -W 'default::scipy.sparse._base.SparseEfficiencyWarning'
+          -W 'default::dask.array.core.PerformanceWarning'
```
flying-sheep, line 77: If these were in `filterwarnings` and we wouldn’t have to deal with `ImportPathMismatchError`s, these would be

```
filterwarnings = [
    'error',
    'default::anndata._warnings.ImplicitModificationWarning',
    'default:.*names are not unique. To make them unique:UserWarning',
    'default::scipy.sparse._base.SparseEfficiencyWarning',
    'default::dask.array.core.PerformanceWarning',
]
```

All of them have at least like 60 instances (sure, some are parametrized tests) and would therefore need a bit of work to be fixed.

Not in this PR.

flying-sheep:
only the two checkmarks in the original comment left.

anndata/utils.py
```diff
@@ -311,6 +313,14 @@ def convert_dictionary_to_structured_array(source: Mapping[str, Sequence[Any]]):
     return arr
 
 
+def warn_once(msg: str, category: type[Warning], stacklevel: int = 1):
+    warnings.warn(msg, category, stacklevel=stacklevel)
+    if "PYTEST_CURRENT_TEST" not in os.environ:
```
ivirshup, line 318: Why is this bit necessary? It seems like this is throwing many errors during testing anyways.

ivirshup (commented):
> awkward.to_categorical was deprecated in favor of awkward.str.to_categorical, which depends on pyarrow. Is making pyarrow a test dep OK or should we ignore/circuvent this?

Totally fine

------

> write_h5ad’s as_dense failed to add IOSpec attrs

Good catch!

-------

> ExperimentalFeatureWarning
> * got silenced after the first occurrence even in tests. Now it’s only silenced in non-test code after the first occurrence.
> * one reason why it got raised so much in the first place is that we initialize AnnData objects in functions and methods like concat, copy, or read_*. All these places now suppress the warning.

Aren't these two points in conflict? If it was being silenced in tests, why do we see it raised so many times in the tests? Even from the same location:

<details>
<summary> Locations of ExperimentalFeatureWarning from `test_concatenate.py` on main </summary>

```python
anndata/_core/aligned_mapping.py:64: 8 warnings
anndata/tests/test_base.py: 3 warnings
anndata/tests/test_concatenate.py: 53 warnings
anndata/tests/test_awkward.py: 39 warnings
anndata/tests/test_hdf5_backing.py: 98 warnings
anndata/tests/test_inplace_subset.py: 70 warnings
anndata/tests/test_io_elementwise.py: 10 warnings
anndata/tests/test_io_conversion.py: 17 warnings
anndata/tests/test_helpers.py: 4 warnings
anndata/tests/test_readwrite.py: 15 warnings
anndata/tests/test_io_dispatched.py: 3 warnings
anndata/tests/test_views.py: 245 warnings
anndata/tests/test_transpose.py: 13 warnings
anndata/tests/test_x.py: 7 warnings
  /Users/isaac/github/anndata/anndata/_core/aligned_mapping.py:64: ExperimentalFeatureWarning: Support for Awkward Arrays is currently experimental. Behavior may change in the future. Please report any issues you may encounter!
    warnings.warn(

```

</details>

This makes me think pytest was reseting the warnings filter by itself, so we don't need to.

Also, I think we shouldn't be catching this warning so many places in the library code. These all seem like places we may want to actually throw this class of warning. I'm fine with larger changes with lots of catching in the test suite, but don't want to make large scale changes to the library code for this.

Could we add a very awkward array assignment specific filter globally to the code of the test suite, then selectively turn it off?

ivirshup:
Also I think this could get a release note for any upstream deprecation warnings it fixes.

anndata/utils.py
flying-sheep: Note the “not”. It silences itself in non-testing code

ivirshup: That's not quite what I was saying. If you take a look in my [comment here](https://github.com/scverse/anndata/pull/1182#pullrequestreview-1667382109) I show that this warning still shows up in tests on main even though an `ignore` [should be present](https://github.com/scverse/anndata/blob/26e909970c7fbb382e8addfaed6ce76cff4a4e74/anndata/_core/aligned_mapping.py#L72C9-L76).

So, why do we need test specific behavior to NOT filter during testing if it seems that this warning was totally fine showing up without it?

I also really don't like the idea of code branches that are specific to whether the tests are running.

anndata/_io/write.py
```diff
@@ -103,7 +103,11 @@ def write_loom(filename: PathLike, adata: AnnData, write_obsm_varm: bool = False
     for key in adata.layers.keys():
         layers[key] = adata.layers[key].T
 
-    from loompy import create
+    from numba.core.errors import NumbaDeprecationWarning
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", NumbaDeprecationWarning)
+        from loompy import create
```
ivirshup, line 110:
```suggestion
    from loompy import create
```

I don't know that it's our place to hide this warning. I wouldn't want people to think our implementation was in a better state than loompy's since it doesn't warn.

anndata/tests/helpers.py
```diff
@@ -256,8 +256,14 @@ def gen_adata(
...
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore", ExperimentalFeatureWarning)
+
+    ctx = (
+        pytest.warns(ExperimentalFeatureWarning)
+        if AwkArray in {*obsm_types, *varm_types}
+        else nullcontext()
+    )
+
+    with ctx:
```
ivirshup, line 266:
```suggestion
    with warnings.catch_warnings():
        warnings.filterwarnings(
            "ignore",
            "Support for Awkward Arrays is currently experimental",
            ExperimentalFeatureWarning,
        )
```

I don't think this function should error if the warning isn't thrown. We have tests specifically for this, and I don't think every test that uses `gen_adata` with an awkward array needs to fail if it doesn't throw a warning.

I do think it could make sense to be more specific with the feature warning though.

anndata/_core/anndata.py
```diff
@@ -875,23 +875,21 @@ def _prep_dim_index(self, value, attr: str) -> pd.Index:
             value = pd.Index(value)
             if not isinstance(value.name, (str, type(None))):
                 value.name = None
-        # fmt: off
         if (
-            not isinstance(value, pd.RangeIndex)
+            len(value) > 0
```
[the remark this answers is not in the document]
flying-sheep: See original comment:

> - `pd.concat` doesn’t like empty DataFrames, so we filter them before passing them now: https://github.com/pandas-dev/pandas/blob/2a65fdd227734cbda5d8e33171e857ab4aaeb9d5/pandas/core/internals/concat.py#L488-L500

anndata/_core/anndata.py
flying-sheep: what’s your suggestion?

anndata/_core/merge.py
flying-sheep: well that’s the code path that gets hit and that’s the warning that gets displayed / converted into an error.

anndata/_core/anndata.py
ivirshup: For now, I'd suggest

```suggestion
```

But I've been eyeing pytest-doctestplus for a bit. `# doctest: +SHOW_WARNINGS` or `# doctest: +IGNORE_WARNINGS` could be used here in future.

anndata/_core/merge.py
ivirshup: Can you show me which tests throw the warning and share your environment info with me?

ivirshup: Alright, it's half the `test_concat_size_0_dim` cases where this is coming up. 

I've also done some reading into the warning, and I *like* the new behavior, and think we should use it. With the change in pandas, the `.dtypes` of the output can be figured out from the `.dtypes` of the inputs. To me, this is good and expected. I don't want to opt-out of that when it becomes the default behavior.

So I'm thinking we keep this warning user visible so if it come up, people know where to look. This also makes the behavior change.

I think we can just ignore it during the test, though maybe this could be a `pytest.raises` so we know when to remove it?

ivirshup: Would be nice if there was a way to opt in to the new behavior earlier, but I guess leaving it to pandas makes it not our problem.

anndata/_core/merge.py
```diff
@@ -1239,7 +1245,9 @@ def concat(
         [pd.Series(dim_indices(a, axis=axis)) for a in adatas], ignore_index=True
     )
     if index_unique is not None:
-        concat_indices = concat_indices.str.cat(label_col.map(str), sep=index_unique)
+        concat_indices = concat_indices.str.cat(
+            label_col.map(str, na_action="ignore"), sep=index_unique
+        )
```
ivirshup, line 1250:
```suggestion
        if parse_version(pd.__version__) >= parse_version("2.0"):
            # Argument added in pandas 2.0
            concat_indices = concat_indices.str.cat(
                label_col.map(str, na_action="ignore"), sep=index_unique
            )
        else:
            concat_indices = concat_indices.str.cat(
                label_col.map(str), sep=index_unique
            )
```

It looks like `na_value` wasn't defined pre 2.0 and unfortunately I don't think we can just bump our minimum pandas version, since it looks like uptake of pandas 2.0 is [unfortunately slow](https://www.pepy.tech/projects/pandas?versions=2.*&versions=1.*&versions=*)

You could probably cherry pick this commit: https://github.com/scverse/anndata/pull/1189/commits/d7d0fa7eca2694325af1ae609116a24e069a0230

anndata/_core/merge.py
flying-sheep: > I think we can just ignore it during the test, though maybe this could be a pytest.raises so we know when to remove it?

OK!

anndata/_core/anndata.py
flying-sheep: not possible, that’ll make that doctest fail in the --pre job

anndata/_core/merge.py
flying-sheep: merged your main changes

anndata/utils.py
flying-sheep: That ignore is still active in the same test case after the first hit.

> I also really don't like the idea of code branches that are specific to whether the tests are running.

Me neither! Just as I don’t like silently changing global state like we do with that ignore filter. But since we do the latter, we also need to do the former.

Both hacks are just affecting a warning filter, so it’s not that bad, but I’d like to see us removing both (at the same time). The only way I see this happening is by removing the need for “warn once” hack

anndata/_core/anndata.py
ivirshup: I would also suggest not turning that on this PR.

anndata/utils.py
ivirshup: > Just as I don’t like silently changing global state like we do with that ignore filter

It's very targeted to our specific warning, and I believe this is fairly common practice (I believe I figured this out from an issue on an upstream library). It's not great but I think it's better than this.

I'm not up for this change.

ivirshup (changes requested):
Added a few more notes, and could you check on my previous comment? Didn't seem like it was addressed.

I think there's a lot here which is mergable immediately. I would suggest merging those parts, and moving the other parts to a follow up (as you suggest for other warnings) since they're pretty orthogonal.

flying-sheep:
> Added a few more notes, and could you check on my previous comment (https://github.com/scverse/anndata/pull/1182#pullrequestreview-1667382109)? Didn't seem like it was addressed.

That’s about `warn_once` and addressed there, right?

anndata/_core/anndata.py
flying-sheep: I invested quite some work into making it possible to do that in a way you’re happy with. Please don‘t go back on this now.

anndata/utils.py
flying-sheep: > It's not great but I think it's better than this.

I’m confused. I added the `if "PYTEST_CURRENT_TEST" not in os.environ` because it’s the only way to tame the `warnings.simplefilter('ignore', ...)` that I can think of. If you have an alternative idea how we can reliably test where that warning is thrown, I’d be happy to switch to it.

I don’t understand your objection. As said:

1. Global changes are bad. But this is just a warning filter, so I see why it’s fine.
2. Changing behavior during tests is bad. But this is the exact same warning filter from 1., so the exact same reasoning applies.

Why differentiate between the two? In both cases it’s “this would be a code smell, if it wasn’t just about warning visibility, which really isn’t a big deal”

anndata/_core/anndata.py
ivirshup: As an alternative you can add this warning to the ignore.

I just want to merge the parts of this PR that are ready to go.

I don't want to add more into this PR because it touches a lot of orthogonal things. This makes it time consuming to review since I need to figure out which code change is related to which warning.

ivirshup:
> > Also, I think we shouldn't be catching this warning so many places in the library code. These all seem like places we may want to actually throw this class of warning. I'm fine with larger changes with lots of catching in the test suite, but don't want to make large scale changes to the library code for this.
> 
> I think you’re mistaking the “Support for Awkward Arrays is currently experimental” warning with the “Outer joins on awkward.Arrays will have different return values in the future”.
> 
> The code is supposed to catch the _former_ warning
> 
> * when transforming an AnnData object: `_inplace_subset`, `transpose`, `to_memory`, `_view_as_actual`
>   The object already exists, the user doesn’t need to see the warning.
> * when reading. I’d be happy to remove the filtering here

The first time an awkward array is put inside of an AnnData, or the first time outer concatenation on an awkward array is performed the user should see a warning. Here is a demo from my machine:

```python
isaac@Mimir:~/github/anndata ‹main›
$ ipython
Python 3.9.18 (main, Aug 24 2023, 21:32:48) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import anndata as ad, numpy as np, awkward as ak, pandas as pd
   ...: 

In [2]: # This warns
   ...: a = ad.AnnData(
   ...:     np.ones((3, 3)),
   ...:     obs=pd.DataFrame(index=["a", "b", "c"]),
   ...:     obsm={"awk": ak.Array([{"a": 1}, {"a": 2}, {"a": 3}])},
   ...: )
/Users/isaac/github/anndata/anndata/_core/aligned_mapping.py:64: ExperimentalFeatureWarning: Support for Awkward Arrays is currently experimental. Behavior may change in the future. Please report any issues you may encounter!
  warnings.warn(

In [3]: # This does not, because the warning is only shown once
   ...: b = ad.AnnData(
   ...:     np.ones((3, 3)),
   ...:     obs=pd.DataFrame(index=["d", "e", "f"]),
   ...:     obsm={"awk": ak.Array([{"a": 1}, {"a": 2}, {"a": 3}])},
   ...: )

In [4]: # This does not warn:
   ...: ad.concat([a, b])
Out[4]: 
AnnData object with n_obs × n_vars = 6 × 3
    obsm: 'awk'

In [5]: # This warns the first time:
   ...: ad.concat([a, b], join="outer")
/Users/isaac/github/anndata/anndata/_core/merge.py:881: ExperimentalFeatureWarning: Outer joins on awkward.Arrays will have different return values in the future.For details, and to offer input, please see:

	https://github.com/scverse/anndata/issues/898
  warn(
Out[5]: 
AnnData object with n_obs × n_vars = 6 × 3
    obsm: 'awk'

In [6]: # But not the second:
   ...: ad.concat([a, b], join="outer")
Out[6]: 
AnnData object with n_obs × n_vars = 6 × 3
    obsm: 'awk'

In [7]: a[:2][:,:2].copy()  # Still no warning
Out[7]: 
AnnData object with n_obs × n_vars = 2 × 2
    obsm: 'awk'
```

Notably, this is behavior we already have on main. So I don't want to add warning catching blocks all over the place, since we already have this behavior.

Do you see different behavior on your machine?

---

> I should probably either subclass the ExperimentalFeatureWarning and catch the specific subclass, or make the filter specific to the message I want to catch to make sure it’s only that one. What do you think?

Specific to the message, like you were already doing:

```python
def warn_once(msg: str, category: type[Warning], stacklevel: int = 1):
    warnings.warn(msg, category, stacklevel=stacklevel)
    warnings.filterwarnings("ignore", category=category, message=re.escape(msg))
```

If you want to be even more specific, I think we could not allow setting a stack level then specify the line number.

> Could we add a very awkward array assignment specific filter globally to the code of the test suite, then selectively turn it off?

This is the “Support for Awkward Arrays is currently experimental” warning you mentioned. The warning that occurs on assignment.

anndata/_core/anndata.py
flying-sheep: > As an alternative you can add this specific warning to the ignore.

Which one? Not inline, you just said that. Not globally, we agreed that we want to be specific.

- Maybe for all doctests?
- Or by adding a `doctest_mark` or `doctest_filterwarnings` like in https://github.com/scverse/scanpy/pull/2605?

flying-sheep:
> Notably, this is behavior we already have on main. So I don't want to add warning catching blocks all over the place, since we already have this behavior.

So you’re saying that because usually people create AnnData objects with AwkwardArrays inside instead of reading, unpickling, or so, it’s fine if the one and only instance of the warning being raised happens wherever.

OK, fair enough!

> Specific to the message, like you were already doing:

I meant in all the other cases where I have to catch it or verify it with `pytest.raises`

----

what are your thoughts about the unchecked boxes in https://github.com/scverse/anndata/pull/1182#issuecomment-1759432787?

flying-sheep:
Ooof, IDK if this is better than having the checks in library code: https://github.com/scverse/anndata/pull/1182/commits/361aa054bb4ac68450244b8bb2f7749bf24a4fe7

anndata/_core/anndata.py
flying-sheep: done!

anndata/utils.py
flying-sheep: OK fixed line we talked about: Globally ignore ExperimentalFeatureWarning

.azure-pipelines.yml
```diff
@@ -63,12 +63,25 @@ jobs:
...
+      #          This is better because it’s a more specific filter and we don’t need one line per message.
+      #          Setting the filters in code instead of per `-W` fixes that.
+      #          This affects:
+      #          - `anndata._warnings.ExperimentalFeatureWarning`
+      #            - “Support for Awkward Arrays”
+      #            - “Outer joins on awkward.Arrays”
+      #          - `anndata._warnings.ImplicitModificationWarning`
+      #            - “Trying to modify attribute”
+      #            - “Transforming to str index”
+      #       3. The 'ignore' lines need to be kept in sync with `filterwarnings` config in pyproject.toml
```
ivirshup, line 79: Should this be an issue? I think it's more visible if it's on a milestone / tracker

anndata/_io/specs/methods.py
```diff
@@ -281,6 +281,7 @@ def read_anndata(elem, _reader):
     ]:
         if k in elem:
             d[k] = _reader.read_elem(elem[k])
+
```
ivirshup, line 284:
```suggestion
```

anndata/tests/helpers.py
```diff
@@ -256,19 +256,17 @@ def gen_adata(
...
+        X=X,
+        obs=obs,
+        var=var,
+        obsm=obsm,
+        varm=varm,
+        layers=layers,
+        obsp=obsp,
+        varp=varp,
+        uns=uns,
+    )
```
ivirshup, line 269:
```suggestion
        with warnings.catch_warnings():
        warnings.simplefilter("ignore", ExperimentalFeatureWarning)
        adata = AnnData(
            X=X,
            obs=obs,
            var=var,
            obsm=obsm,
            varm=varm,
            layers=layers,
            obsp=obsp,
            varp=varp,
            uns=uns,
        )
```

Would you be against just not touching this part?

anndata/tests/test_dask.py
```diff
@@ -107,21 +109,20 @@ def test_dask_distributed_write(adata, tmp_path, diskfmt):
...
-                with pytest.raises(
-                    ValueError, match="Cannot write dask arrays to hdf5"
-                ):
-                    write_elem(g, "", orig)
-                return
-            write_elem(g, "", orig)
-            curr = read_elem(g)
+    with dd.LocalCluster(
+        n_workers=1, threads_per_worker=1, processes=False
+    ) as cluster, dd.Client(cluster), warnings.catch_warnings():
```
ivirshup, line 114:
```suggestion
    ) as cluster, dd.Client(cluster):
```

What was this for?

pyproject.toml
```diff
@@ -88,9 +88,10 @@ test = [
     "joblib",
     "boltons",
     "scanpy",
-    "httpx", # For data downloading
+    "httpx",                   # For data downloading
```
ivirshup, line 91:
```suggestion
    "httpx",  # For data downloading
```

pyproject.toml
```diff
@@ -104,13 +105,20 @@ version-file = "anndata/_version.py"
...
 [tool.pytest.ini_options]
-addopts = "--doctest-modules"
+addopts = [
+    "--strict-markers",
+    "--doctest-modules",
+    "--ignore=anndata/core.py",      # deprecated
+    "--ignore=anndata/readwrite.py", # deprecated
+]
+filterwarnings = [
+    "ignore::anndata._warnings.ExperimentalFeatureWarning",
```
ivirshup: Is this meant to be in both the job definition and here?

ivirshup (commented):
In principal I approve, I've just made suggestions on a few minor things that I think were left over from stuff left to other PRs.

Mostly formatting

.azure-pipelines.yml
flying-sheep: done: https://github.com/scverse/anndata/issues/1204

anndata/tests/helpers.py
flying-sheep: I don’t see why we should double-ignore it

Its existence implies that this filter exists for a reason. But not important.

pyproject.toml
flying-sheep: yes, as discussed, if we want to ignore a warning globally, we should ignore it globally in all jobs.

flying-sheep:
@ivirshup ping

anndata/_core/anndata.py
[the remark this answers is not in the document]
ivirshup: Sorry, forgot this in last review!

@flying-sheep, we are no longer filtering out empty dataframes as discussed here: https://github.com/scverse/anndata/pull/1182#discussion_r1353380008

Is this change still necessary?

ivirshup (approved):
Apart from the one comment this is good to go

anndata/_core/anndata.py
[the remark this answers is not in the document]
flying-sheep: nope! but I’ll leave the removal of `#fmt off` in.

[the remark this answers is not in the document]
flying-sheep: yeah, of course. AnnData objects with 0 observations or 0 variables are still a thing.

PR #1183, opened 2023-10-10
Fewer warnings

ivirshup:
Reduce number of warnings thrown by the test suite

### DIFF
--- a/anndata/core.py
+++ b/anndata/core.py
@@ -1,0 +1,7 @@
+from __future__ import annotations
+
+from warnings import warn
+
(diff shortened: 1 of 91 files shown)


anndata/core.py
```diff
@@ -1,7 +0,0 @@
-from __future__ import annotations
```
ivirshup, line 1: @flying-sheep, maybe this can't be patch with these changes?

ivirshup: Do you have a good understanding of when these warnings are being thrown?

PR #1209, opened 2023-10-26
Specify deprecated theme option

flying-sheep:
See https://github.com/pydata/pydata-sphinx-theme/issues/1492


flying-sheep:
That’s correct, I initially added that fix in a backport PR.

PR #1214, opened 2023-10-30
Centralize warning config

### LINKED ISSUE #1204: Centralize testing warning filters using a custom pytest CLI option

After #1182, we should do what the title says because

1. This allows to easily locally run with the same settings as this job
2. Otherwise we can’t filter anndata warning classes directly,
   because we’re neither using editable installs nor a src/ layout.
   This is better because it’s a more specific filter and we don’t need one line per message.
   Setting the filters in code instead of per `-W` fixes that.
   This affects:
   - `anndata._warnings.ExperimentalFeatureWarning`
     - “Support for Awkward Arrays”
     - “Outer joins on awkward.Arrays”
   - `anndata._warnings.ImplicitModificationWarning`
     - “Trying to modify attribute”
     - “Transforming to str index”
3. The 'ignore' lines need to be kept in sync with `filterwarnings` config in pyproject.toml

flying-sheep:
- [x] Closes #1204
- [x] Tests added

This makes sure nothing needs to be kept in sync manually

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -58,12 +58,27 @@ def pytest_itemcollected(item: pytest.Item) -> None:
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:
+    """Hook to register custom CLI options and config values"""
     parser.addoption(
         "--strict-warnings",
         action="store_true",
         default=False,
         help="Turn most warnings into errors",
     )
+
+    parser.addini(
+        "filterwarnings_when_strict",
+        "Filters to apply after `-Werror` when --strict-warnings is active",
+        type="linelist",
+        default=[],
+    )
+
+
+def _config_get_strlist(config: pytest.Config, name: str) -> list[str]:
+    if strs := config.getini("filterwarnings"):
+        assert isinstance(strs, list)
+        return [str(f) for f in strs]
+    return []
 
 
 def pytest_collection_modifyitems(
@@ -72,19 +87,10 @@ def pytest_collection_modifyitems(
     if not config.getoption("--strict-warnings"):
         return
 
-    filters_from_config: list[str] = []
-    if fs := config.getini("filterwarnings"):
-        assert isinstance(fs, list)
-        filters_from_config = [str(f) for f in fs]
-
     warning_filters = [
         r"error",
-        *filters_from_config,
-        r"default::anndata._warnings.ImplicitModificationWarning",
-        r"default:Transforming to str index:UserWarning",
-        r"default:(Observation|Variable) names are not unique. To make them unique:UserWarning",
-        r"default::scipy.sparse.SparseEfficiencyWarning",
-        r"default::dask.array.core.PerformanceWarning",
+        *_config_get_strlist(config, "filterwarnings"),
+        *_config_get_strlist(config, "filterwarnings_when_strict"),
     ]
     warning_marks = [pytest.mark.filterwarnings(f) for f in warning_filters]
 
--- a/conftest.py
+++ b/conftest.py
@@ -74,13 +74,6 @@ def pytest_addoption(parser: pytest.Parser) -> None:
     )
 
 
-def _config_get_strlist(config: pytest.Config, name: str) -> list[str]:
-    if strs := config.getini("filterwarnings"):
-        assert isinstance(strs, list)
-        return [str(f) for f in strs]
-    return []
-
-
 def pytest_collection_modifyitems(
     session: pytest.Session, config: pytest.Config, items: Iterable[pytest.Item]
 ):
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -116,6 +116,16 @@ addopts = [
...
+    # TODO: replace both lines above with this one once we figured out how prevent ImportPathMismatchError
+    # 'ignore::anndata._warnings.ExperimentalFeatureWarning',
+]
+# When `--strict-warnings` is used, all warnings are treated as errors, except those:
+filterwarnings_when_strict = [
+    "default::anndata._warnings.ImplicitModificationWarning",
+    "default:Transforming to str index:UserWarning",
+    "default:(Observation|Variable) names are not unique. To make them unique:UserWarning",
+    "default::scipy.sparse.SparseEfficiencyWarning",
+    "default::dask.array.core.PerformanceWarning",
```
flying-sheep, line 128: Would be great to just add those to `filterwarnings` and not use a custom option, but due to some quirk in how pytest imports those object paths, that raises `ImportPathMismatchError` as described above.

So I rather have a bit more complexity by using an extra option than making the list of `default` filters as complex as they were before this PR: https://github.com/scverse/anndata/blob/00f39eba8e412639094c87a62ba935f9d94b6622/.azure-pipelines.yml#L70-L78

conftest.py
```diff
@@ -51,3 +55,47 @@ def pytest_itemcollected(item):
...
         item.add_marker(doctest_marker)
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    """Hook to register custom CLI options and config values"""
+    parser.addoption(
+        "--strict-warnings",
+        action="store_true",
+        default=False,
+        help="Turn most warnings into errors",
```
ilan-gold, line 66: This seems a little vague, maybe something like "Applies the `filterwarnings_when_strict` option from the `pyproject.toml` to a test run."  I think it would be good to make clear how this flag works.

conftest.py
```diff
@@ -51,3 +55,47 @@ def pytest_itemcollected(item):
...
+        # reverse and then prepend means essentially `marks[0:0] = warning_marks`
+        # this ensures that markers that are applied later override these
+        for mark in reversed(warning_marks):
+            item.add_marker(mark, append=False)
+
+
+def _config_get_strlist(config: pytest.Config, name: str) -> list[str]:
+    if strs := config.getini(name):
+        assert isinstance(strs, list)
+        return [str(f) for f in strs]
```
ilan-gold, line 100: Is `str(f)` necessary?  These could somehow not be strings?  If so, maybe the list shouldn't be called `strs`?

conftest.py
```diff
@@ -51,3 +55,47 @@ def pytest_itemcollected(item):
...
+
+
+def pytest_collection_modifyitems(
+    session: pytest.Session, config: pytest.Config, items: Iterable[pytest.Item]
+):
+    if not config.getoption("--strict-warnings"):
+        return
+
+    warning_filters = [
+        r"error",
```
ilan-gold, line 84: I don't think the `r` is necessary here

conftest.py
```diff
@@ -51,3 +55,47 @@ def pytest_itemcollected(item):
...
+        *_config_get_strlist(config, "filterwarnings"),
+        *_config_get_strlist(config, "filterwarnings_when_strict"),
+    ]
+    warning_marks = [pytest.mark.filterwarnings(f) for f in warning_filters]
+
+    for item in items:
+        # reverse and then prepend means essentially `marks[0:0] = warning_marks`
```
ilan-gold, line 91: What is `marks` here?

ilan-gold: Also just a little lost as to what the objective is (or rather why reversing assists in that)

ilan-gold: Ah, ok, I see.  I guess you want the users CLI options to override ours here.  But not sure I understand the reversal.

conftest.py
flying-sheep: True. I had it for consistency when things where hardcoded here.

conftest.py
flying-sheep: To preserve the order, as the comment says:

```py
before = ["last"]
to_prepend = ["first", "middle"]

correct = before.copy()
for item in reversed(to_prepend):
    correct.insert(0, item)
assert correct = ["first", "middle", "last"]

wrong = before.copy()
for item in to_prepend:
    wrong.insert(0, item)
assert wrong == ["middle", "first", "last"]
```

ilan-gold: Ok, I guess I see now.  You could have overrides within the list being prepended that you want applied in order.

ilan-gold (approved):

Issue #1224, opened 2023-11-02
Better indexing performance for backed sparse arrays via slices


ivirshup:
## The problem

Indexing statements for backed anndata objects often look like:

```python
subset = backed_adata[
    (backed_adata.obs["self_reported_ethnicity"] == "European") & \
    (backed_adata.obs["sex"] == "female")
]
```

Notably, data is often ordered such that categories of interest are actually colocalized, e.g. data is sorted by patient or condition.

The indexing statement above generates a boolean mask or integer array indexer which can unfortunately be slow, especially with zarr backed matrices. 

## Zarr benchmarks

For example:

```python
import numpy as np
import zarr
from anndata.experimental import read_elem, write_elem, sparse_dataset
from scipy import sparse

# Create data
rng = np.random.default_rng()

X = sparse.random(100_000, 10_000, density=0.01, random_state=rng, format="csr")
group = zarr.open("demo.zarr", mode="w")
write_elem(group, "X", X)

X_zarr = sparse_dataset(group["X"])
X_zarr

# Make a mask where contiguous chunks are True
inds = np.random.choice(X_zarr.shape[0], 100, replace=False)
inds.sort()

mask = np.zeros(X_zarr.shape[0], dtype=bool)
for i in range(0, len(inds) - 1, 2):
    mask[inds[i]:inds[i+1]] = True
```

```python
%time as_mask = X_zarr[mask]
# CPU times: user 19.6 s, sys: 9.06 s, total: 28.7 s
# Wall time: 28.7 s
```

```python
# converting to integer indices
%time as_indices = X_zarr[np.where(mask)]
# CPU times: user 19.8 s, sys: 8.68 s, total: 28.5 s
# Wall time: 28.5 s
```

### Potential solution

However, if we simply translate the mask into contiguous slices:

```python
# Converting to slices
def mask_to_slices(mask):
    return np.ma.extras._ezclump(mask)

%time as_slices = sparse.vstack([X_zarr[s] for s in mask_to_slices(mask)])
# CPU times: user 73.4 ms, sys: 32.5 ms, total: 106 ms
# Wall time: 105 ms
```

We see significant speed ups, even accounting for the overhead (and an admittedly inefficient quick implementation, we should be able to avoid constructing so many intermediate matrices). Of course, the results are all the same:

```python
assert (as_indices != as_mask).nnz == 0
assert (as_slices != as_mask).nnz == 0
```

## HDF5 benchmarks

[...]

PR #1233, opened 2023-11-15
(fix): indexing performance in `backed` mode for `boolean` case

ilan-gold:
Here's my shot at it.   As I explain in the comment (also in the issue), the `boolean` case is converted by `scipy` to numeric integers which is why the solution has to go before `mtx` access.  The other option would be to "solve" the `integer` case but as you point out, there are some nasty edge cases there with sorting, repeated indices etc.  So in the interest of getting something out, I have this.

I set up a little benchmarking thing.  Here's the setup:
```python
import numpy as np
import zarr
from anndata.experimental import read_elem, write_elem, sparse_dataset
from scipy import sparse
import h5py

# Create data
rng = np.random.default_rng()

X = sparse.random(100_000, 10_000, density=0.01, random_state=rng, format="csr")
group = zarr.open("demo.zarr", mode="w")
write_elem(group, "X", X)
X_zarr = sparse_dataset(group["X"])

h5_group = h5py.File("demo.h5", mode="w")
write_elem(h5_group, "X", X, dataset_kwargs={"compression": "lzf"})
X_h5 = sparse_dataset(h5_group["X"])

# randomized indices
inds = np.random.choice(X_zarr.shape[0], 100, replace=False)
inds.sort()

mask = np.zeros(X_zarr.shape[0], dtype=bool)
for i in range(0, len(inds) - 1, 2):
    mask[inds[i]:inds[i+1]] = True

# non-random indices, with alternating one false and n true
def make_alternating_mask(n):
    mask_alternating = np.zeros(X_zarr.shape[0], dtype=bool)
    for i in range(n):
        mask_alternating[i::(n + 1)] = True     
    return mask_alternating
```

Once you have this, I would set the following in order to use `make_alternating_mask` with different size `num_contiguous` contiguous slices (as opposed to random indices in `mask`):
```python
num_contiguous = 10
```

You can then profile different use cases with `X_zarr` and `X_h5`, using `np.where` on the mask to trigger "old behavior."

For example,
```python
 %prun -s cumtime X_zarr[make_alternating_mask(num_contiguous)]
```
uses the "new" behavior because `num_contiguous=10>7`, the cutoff point where the optimization is faster and
```python
 %prun -s cumtime X_zarr[np.where(make_alternating_mask(num_contiguous))]
```
 yields the old behavior.  For me, the first one is orders of magnitude faster (due to, it appears, many less `zarr` accesses).  A quick "timing" check can also be performed for different values of `num_contiguous`

```python
def timings(num_contiguous):
    print('****Proposed solution****')
    print('-------------------------')
    print('Zarr, random indices')
    print('-------------------------')
    %time X_zarr[mask]
    print('-------------------------')
    print('Zarr, alternating mask with size ', num_contiguous, ' slices')
    print('-------------------------')
    %time X_zarr[make_alternating_mask(num_contiguous)]
    print('\n****Old behavior****') # achieved using integer indices
    print('-------------------------')
    print('Zarr, random indices')
    print('-------------------------')
    %time X_zarr[np.where(mask)[0]]
    print('-------------------------')
    print('Zarr, alternating mask with size ', num_contiguous, ' slices')
    print('-------------------------')
    %time X_zarr[np.where(make_alternating_mask(num_contiguous))[0]]

    print('\n****Proposed solution****')
    print('-------------------------')
    print('h5, random indices')
    print('-------------------------')
    %time X_h5[mask]
    print('-------------------------')
    print('h5, alternating mask with size ', num_contiguous, ' slices')
    print('-------------------------')
    %time X_h5[make_alternating_mask(num_contiguous)]
    print('\n****Old behavior****') # achieved using integer indices
    print('-------------------------')
    print('h5, random indices')
    print('-------------------------')
    %time X_h5[np.where(mask)[0]]
    print('-------------------------')
    print('h5, alternating mask with size ', num_contiguous, ' slices')
    print('-------------------------')
    %time X_h5[np.where(make_alternating_mask(num_contiguous))[0]]
```

- [ ] Addresses the `boolean` type case in #1224
- [ ] Tests added

### DIFF
--- a/anndata/_core/sparse_dataset.py
+++ b/anndata/_core/sparse_dataset.py
@@ -16,6 +16,7 @@ import collections.abc as cabc
 import warnings
 from abc import ABC
 from itertools import accumulate, chain
+from math import floor
 from typing import TYPE_CHECKING, Literal, NamedTuple
 
 import h5py
@@ -348,10 +349,22 @@ class BaseCompressedSparseDataset(ABC):
         def make_slices(mask):
             return np.ma.extras._ezclump(mask)
 
+        def average_slice_length(slices):
+            return floor((slices[-1].stop - slices[0].start) / len(slices))
+
         if self.format == "csr" and np.array(row).dtype == bool:
-            sub = ss.vstack([mtx[s, col] for s in make_slices(row)])
+            slices = make_slices(row)
+            # This is the break-even point at which this optimization no longer is faster.
+            if average_slice_length(slices) > 7:
+                sub = ss.vstack([mtx[s, col] for s in make_slices(row)])
+            else:
+                sub = mtx[row, col]
         elif self.format == "csc" and np.array(col).dtype == bool:
-            sub = ss.vstack([mtx[row, s] for s in make_slices(row)])
+            slices = make_slices(col)
+            if average_slice_length(slices) > 7:
+                sub = ss.vstack([mtx[row, s] for s in make_slices(col)])
+            else:
+                sub = mtx[row, col]
         else:
             sub = mtx[row, col]
         # If indexing is array x array it returns a backed_sparse_matrix
--- a/anndata/_core/sparse_dataset.py
+++ b/anndata/_core/sparse_dataset.py
@@ -345,17 +345,13 @@ class BaseCompressedSparseDataset(ABC):
         mtx = self._to_backed()
 
         # see https://github.com/scverse/anndata/issues/1224 for special handling of boolean masks
-        def mask_to_slices_or_ints(mask):
-            slices = np.ma.extras._ezclump(mask)
-            for s in slices:
-                if s.stop - s.start < 10:
-                    yield list(range(s.start, s.stop))
-                yield s
+        def make_slices(mask):
+            return np.ma.extras._ezclump(mask)
 
         if self.format == "csr" and np.array(row).dtype == bool:
-            sub = ss.vstack([mtx[s, col] for s in mask_to_slices_or_ints(row)])
+            sub = ss.vstack([mtx[s, col] for s in make_slices(row)])
         elif self.format == "csc" and np.array(col).dtype == bool:
-            sub = ss.vstack([mtx[row, s] for s in mask_to_slices_or_ints(row)])
+            sub = ss.vstack([mtx[row, s] for s in make_slices(row)])
         else:
             sub = mtx[row, col]
         # If indexing is array x array it returns a backed_sparse_matrix
--- a/anndata/_core/sparse_dataset.py
+++ b/anndata/_core/sparse_dataset.py
@@ -343,7 +343,21 @@ class BaseCompressedSparseDataset(ABC):
     def __getitem__(self, index: Index | tuple[()]) -> float | ss.spmatrix:
         row, col = self._normalize_index(index)
         mtx = self._to_backed()
-        sub = mtx[row, col]
+
+        # see https://github.com/scverse/anndata/issues/1224 for special handling of boolean masks
+        def mask_to_slices_or_ints(mask):
+            slices = np.ma.extras._ezclump(mask)
+            for s in slices:
+                if s.stop - s.start < 10:
+                    yield list(range(s.start, s.stop))
+                yield s
+
+        if self.format == "csr" and np.array(row).dtype == bool:
+            sub = ss.vstack([mtx[s, col] for s in mask_to_slices_or_ints(row)])
+        elif self.format == "csc" and np.array(col).dtype == bool:
+            sub = ss.vstack([mtx[row, s] for s in mask_to_slices_or_ints(row)])
+        else:
+            sub = mtx[row, col]
         # If indexing is array x array it returns a backed_sparse_matrix
         # Not sure what the performance is on that operation
         if isinstance(sub, BackedSparseMatrix):
--- a/anndata/_core/sparse_dataset.py
+++ b/anndata/_core/sparse_dataset.py
@@ -155,7 +155,6 @@ class backed_csr_matrix(BackedSparseMatrix, ss.csr_matrix):
             slice_len(row, self.shape[0]),
             slice_len(col, self.shape[1]),
         )
-
         if out_shape[0] == 1:
             return self._get_intXslice(slice_as_int(row, self.shape[0]), col)
         elif out_shape[1] == self.shape[1] and out_shape[0] < self.shape[0]:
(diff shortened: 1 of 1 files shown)


anndata/_core/sparse_dataset.py
```diff
@@ -343,7 +343,29 @@ def __repr__(self) -> str:
...
         mtx = self._to_backed()
-        sub = mtx[row, col]
+
+        # see https://github.com/scverse/anndata/issues/1224 for special handling of boolean masks
+        # scipy internally converts boolean masks to integer indices so this needs to occur before access to `mtx`.
+        def make_slices(mask):
+            return np.ma.extras._ezclump(mask)
+
+        def mean_slice_length(slices):
+            return floor((slices[-1].stop - slices[0].start) / len(slices))
```
ilan-gold, line 353: Can we assume the `slices` are sorted? I think so?

anndata/_core/sparse_dataset.py
```diff
@@ -343,7 +343,29 @@ def __repr__(self) -> str:
...
+
+        if (
+            self.format == "csr"
+            and np.array(row).dtype == bool
+            and mean_slice_length(make_slices(row)) > 7
+        ):
```
ilan-gold, line 359: I will try to come up a less repetitive way of doing this, especially since `make_slices` might not be so cheap in the future with larger datasets or something....

anndata/_core/sparse_dataset.py
```diff
@@ -343,7 +343,29 @@ def __repr__(self) -> str:
     def __getitem__(self, index: Index | tuple[()]) -> float | ss.spmatrix:
         row, col = self._normalize_index(index)
         mtx = self._to_backed()
-        sub = mtx[row, col]
+
+        # see https://github.com/scverse/anndata/issues/1224 for special handling of boolean masks
+        # scipy internally converts boolean masks to integer indices so this needs to occur before access to `mtx`.
```
ivirshup, line 348: While this can't go into one of the methods in the `BackedSparseMatrix` classes, could this logic at least be factored out into a function? Mainly for readability of this method, and so it's easy to move later.

anndata/_core/sparse_dataset.py
ivirshup: Because we made them, yes.

anndata/_core/sparse_dataset.py
ivirshup: Any luck on this?

anndata/_core/sparse_dataset.py
```diff
@@ -343,7 +343,29 @@ def __repr__(self) -> str:
...
+
+        def mean_slice_length(slices):
+            return floor((slices[-1].stop - slices[0].start) / len(slices))
+
+        if (
+            self.format == "csr"
+            and np.array(row).dtype == bool
+            and mean_slice_length(make_slices(row)) > 7
+        ):
+            sub = ss.vstack([mtx[s, col] for s in make_slices(row)])
```
ivirshup, line 360: Instead of creating and concatenating sparse matrices, could we construct the underlying arrays directly? This could be done like `get_compressed_vectors`

ivirshup (commented):
> Benchmarking included here?

Some simple benchmark cases here would be great: https://github.com/scverse/anndata/blob/main/benchmarks/benchmarks/sparse_dataset.py

> Tests? What are we testing here? Can we do np.where vs. slices?

If we're not already testing for equivalency of these two cases we should definitely do that

anndata/_core/sparse_dataset.py
ilan-gold: Tied, see above comment about refactoring.

anndata/_core/sparse_dataset.py
```diff
@@ -340,10 +360,62 @@ def value(self) -> ss.spmatrix:
     def __repr__(self) -> str:
         return f"{type(self).__name__}: backend {self.backend}, shape {self.shape}, data_dtype {self.dtype}"
 
+    @property
+    def _efficient_axis(self):
+        """The axis over which the sparse matrix is efficient (i.e., row for csr)"""
+        formats = ["csr", "csc"]
+        return formats.index(self.format)
+
+    def _efficient_index(self, indices):
```
ilan-gold, line 369: Maybe a better thing to call this?  This is not my forte.  Would be great to chat IRL

ivirshup: "compressed axis" I think the more common term

ilan-gold: Oh that is so much better, yes.

anndata/_core/sparse_dataset.py
```diff
@@ -354,7 +428,7 @@ def __getitem__(self, index: Index | tuple[()]) -> float | ss.spmatrix:
     def _normalize_index(
         self, index: Index | tuple[()]
     ) -> tuple[np.ndarray, np.ndarray]:
-        if index == ():
+        if isinstance(index, tuple) and not len(index):
```
ilan-gold, line 431: See: https://github.com/scverse/anndata/issues/1254

ilan-gold: Separate PR for blame purposes: https://github.com/scverse/anndata/pull/1255

anndata/_core/sparse_dataset.py
ivirshup: Yeah, I think it may have gotten more complicated...

I was thinking more:

* Don't include definitions for `make_slices` and `mean_slice_length` in here
* Keep the flow control logic, but make it like:


```python
if (
    self.format == "csr"
    and np.array(row).dtype == bool
):
    sub = sparse.csr_matrix(_mask_major_axis(mtx, row), shape=(mask.sum(), mtx.shape[1]))[:, col]
elif (
    self.format == "csc"
    and np.array(col).dtype == bool
):
    sub = sparse.csc_matrix(_mask_major_axis(mtx, col), shape=(mtx.shape[0], mask.sum()))[row, :]
```

Where

```python
def _mask_major_axis(mtx: sparse.spmatrix, mask: np.ndarray) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
    slices = make_slices(mask)

    if mean_slice_length(slices) <= 7:
        return get_compressed_vectors(mtx.data, mtx.indices, mtx.indptr, np.where(mask)[0])
    else:
        return get_compressed_vector_for_slices(mtx.data, mtx.indices, mtx.indptr, slices)
```

anndata/_core/sparse_dataset.py
ivirshup: Yeah, I have definitely gotten very confused while writing sparse matrix code. Just a lot of offsets to keep in mind.

ilan-gold:
@ivirshup I just noticed that https://github.com/scverse/anndata/blob/main/anndata/_core/index.py#L104 has `np.where` as well.  So we might want to remove that as well, but I think it is a separate issue.  I don't know the history of it.

ivirshup:
I've moved the indexing methods out of the class since neither actually used the `self` parameter, and purely operated on the arguments. This is more like the other indexing code that live in the `get_...` functions.

ivirshup (approved):

ilan-gold:
Thanks! Will keep this in mind for next time.

ivirshup:
Then feel free to merge this

ilan-gold:
> I suspect this was because it was easier to do this than have to deal with boolean indices everywhere.

I think the context of my comment might have gotten lost - `scipy` calls `np.where` internally anywhere, so maybe it was a behavioral change on their part (we are using "non-public" APIs after all, I think).

PR #1240, opened 2023-11-21
Allow for specifying a tensor device in AnnLoader

austinv11:
Currently AnnLoader only supports transferring tensors to cuda devices. However, PyTorch as of late has much improved its support of other devices such as mps (Mac GPUs) and XLA (Google TPUs). So I added a device argument that is more semantically similar to other PyTorch APIs to defer to the user to decide which accelerator device (if any) to load the data onto.

- [x] Closes N/A
- [ ] Tests added: There are currently no tests for AnnLoader. So I am not sure how to add tests for this.

### DIFF
--- a/anndata/experimental/pytorch/_annloader.py
+++ b/anndata/experimental/pytorch/_annloader.py
@@ -4,6 +4,7 @@ from copy import copy
 from functools import partial
 from math import ceil
 from typing import TYPE_CHECKING
+from warnings import warn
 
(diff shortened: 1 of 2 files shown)


Zethson:
We now suggested to use https://github.com/scverse/annbatch/ for dataloading. Therefore, I'll close this PR now. Please let us know if annbatch doesn't cover your use case.

PR #1244, opened 2023-11-28
Add semantic dimension identifier to concat API

### LINKED ISSUE #1245: Concat API lacks semantic dimension identifier

### Report

AnnData was designed around semantic dimension names instead of implementation details.

The `concat` API lacks support for this basic design goal.

While it’s of course not entirely possible to avoid numerical axes like that, all APIs should support semantic axis names.

Code:

```python
anndata.concat([...], dim="var")
```

Traceback:

```pytb
TypeError: Function `concat` has no parameter named `dim`
```

### Versions

```python
0.10.3
```

flying-sheep:
- [x] Closes #1245
- [x] Tests added

Follows the consensus `axis`/`axis_name`


ivirshup (changes requested):
I commented on the original issue, but basically would like to discuss this.

To me, you are proposing that we do a broad API change of introducing a keyword argument `dim` which is analogous to `axis`, but refers to the axes as `"obs", "var"` instead of `0, 1`. I would like to at least discuss the implications of this, especially considering that we allow names to be set dynamically on the dimensions. Let's do this on the issue.

I also would not consider this a bug fix as it introduces a new element to the API.

anndata/tests/test_concatenate.py
```diff
@@ -98,11 +98,21 @@ def fill_val(request):
...
 
+@pytest.fixture
+def axis(axis_name) -> Literal[0, 1]:
+    return 0 if axis_name == "obs" else 1
+
+
+@pytest.fixture
+def alt_axis_name(axis_name) -> Literal["obs", "var"]:
+    return "var" if axis_name == "obs" else "obs"
+
```
ivirshup, line 114: Instead of having two parameters (which function as one) can we just have one, then derive the value in the test?

anndata/tests/test_concatenate.py
```diff
@@ -799,44 +809,42 @@ def test_awkward_does_not_mix(join_type, other):
...
+                "varp": {"arr": gen_axis_array(n)},
             }
         )
         for k, m, n in zip("abc", Ms, Ns)
     }
 
-    w_pairwise = concat(adatas, axis=axis, label="orig", pairwise=True)
-    wo_pairwise = concat(adatas, axis=axis, label="orig", pairwise=False)
+    w_pairwise = concat(adatas, axis=axis_name, label="orig", pairwise=True)
+    wo_pairwise = concat(adatas, axis=axis_name, label="orig", pairwise=False)
```
ivirshup, line 835:
```suggestion
    w_pairwise = concat(adatas, axis=axis, label="orig", pairwise=True)
    wo_pairwise = concat(adatas, axis=axis, label="orig", pairwise=False)
```

I'm not sure I see why we should change this. I think just adding another test which makes sure the two variants behave the same would be better.

anndata/_core/merge.py
```diff
@@ -991,31 +991,25 @@ def merge_outer(mappings, batch_keys, *, join_index="-", merge=merge_unique):
...
-    else:
-        return axis, _dims[axis]
+def _resolve_axis(
+    axis: Literal["obs", 0, "var", 1],
+) -> tuple[Literal[0], Literal["obs"]] | tuple[Literal[1], Literal["var"]]:
+    if axis is None:
+        raise ValueError("Must pass `axis` != None.")
+    elif axis not in (0, 1, "obs", "var"):
+        raise ValueError(f"`axis` must be either 0, 1, 'obs', or 'var', was {axis}")
+    return (0, "obs") if axis in {0, "obs"} else (1, "var")
```
ilan-gold, line 1001:
```suggestion
    if axis in {0, "obs"}:
        return (0, "obs") 
    if axis in {1, "var"}:
        return (1, "var")
    raise ValueError(f"`axis` must be either 0, 1, 'obs', or 'var', was {axis}")
```

anndata/_core/merge.py
```diff
@@ -1152,6 +1146,8 @@ def concat(
     s2     2     3
     s3     4     5
     s4     7     8
+    >>> # ad.concat([a, c], axis="var").to_df()
+    >>> # or
```
ilan-gold, line 1150: These are run, right? Why not do both here?

ivirshup (changes requested):

anndata/tests/test_concatenate.py
flying-sheep: That’s how it was before, but I found stuff like `("uns", "var)[axis]` to be too cryptic and mistakes too easily missable.

That’s why I switched to a centralized, clean alternative.

anndata/_core/merge.py
```diff
@@ -991,31 +991,25 @@ def merge_outer(mappings, batch_keys, *, join_index="-", merge=merge_unique):
...
-    elif dim is not None and dim not in _dims:
-        raise ValueError(f"`dim` must be one of ('obs', 'var'), was {dim}")
-    elif axis is not None and axis not in (0, 1):
-        raise ValueError(f"`axis` must be either 0 or 1, was {axis}")
-    if dim is not None:
-        return _dims.index(dim), dim
-    else:
-        return axis, _dims[axis]
+def _resolve_axis(
+    axis: Literal["obs", 0, "var", 1],
```
ivirshup, line 995: I feel like this typing is a little weird, cause if the typing were correct the ValueError would never be raised. But, not really important.

anndata/_core/merge.py
```diff
@@ -1146,12 +1140,17 @@ def concat(
...
     s3     4     5
     s4     7     8
+    >>> ad.concat([a, c], axis="var").to_df()
+        var1  var2  var3  var4
+    s1     0     1    10    11
+    s2     2     3    12    13
+    >>> # or, equivalently
     >>> ad.concat([a, c], axis=1).to_df()
         var1  var2  var3  var4
     s1     0     1    10    11
```
ivirshup, line 1156: I think this could be its own section, and it would be a little cleaner. It may be better lower down the page since `inner`/ `outer` functionality may be important than an alternate way to do things.

```suggestion
    >>> ad.concat([a, b]).to_df()
        var1  var2
    s1     0     1
    s2     2     3
    s3     4     5
    s4     7     8
    >>> ad.concat([a, c], axis=1).to_df()
        var1  var2  var3  var4
    s1     0     1    10    11
    s2     2     3    12    13

    Use semantic identifiers to select axis

    >>> ad.concat([a, b], axis="obs").to_df()  # Equivalent to axis=0
        var1  var2
    s1     0     1
    s2     2     3
    s3     4     5
    s4     7     8
    >>> ad.concat([a, c], axis="var").to_df()  # Equivalent to axis=1
        var1  var2  var3  var4
    s1     0     1    10    11
```

anndata/tests/test_concatenate.py
```diff
@@ -406,6 +407,21 @@ def test_concatenate_obsm_outer(obsm_adatas, fill_val):
...
+        pytest.param("obs", (20, 10), id="obs"),
+        pytest.param(0, (20, 10), id="0"),
+        pytest.param("var", (10, 20), id="var"),
+        pytest.param(1, (10, 20), id="1"),
+    ],
+)
+def test_concat_axis_param(axis_val, shape):
+    a, b = gen_adata((10, 10)), gen_adata((10, 10))
+    c = concat([a, b], axis=axis_val)
+    assert c.shape == shape
```
ivirshup, line 422:
```suggestion
@pytest.mark.parametrize(
    ("axis", "axis_name"),
    [("obs", 0), ("var", 1)],
)
def test_concat_axis_param(axis, axis_name):
    a, b = gen_adata((10, 10)), gen_adata((10, 10))
    assert_equal(
        concat([a, b], axis=axis),
        concat([a, b], axis=axis_name)
    )
```

A bit more simple and more thorough I think.

I think this basically covers all changes that are really needed to the tests in this PR too, no?

ivirshup (changes requested):
A couple points. 

It would also be good to address the missing coverage on the added code.

* `axis_size` may be able to just get deleted
* `dim_len` could probably get deleted (was it used somewhere?), or it should get a deprecation test

anndata/tests/test_concatenate.py
flying-sheep: I think so too.

flying-sheep:
all done, including the `axis_size` and `dim_len` parts

flying-sheep:
done!

PR #1247, opened 2023-11-30
(feat): `read_lazy` for whole `AnnData` lazy-loading + `xarray` reading + `read_elem_as_dask` -> `read_elem_lazy`

### LINKED ISSUE #951: Dask and Zarr not loading obsp and obsm from remote s3

### LINKED ISSUE #981: lazy `.obs` and `.var` with `backed="r"` mode

ilan-gold:
This PR is a lighter weight version of #947 that involves using the original `AnnData` object as the class to hold `obs` and `var` `xr.Dataset`.  

- [x] Closes #951 and closes #981
- [x] Tests added


ilan-gold:
@ivirshup @flying-sheep Not really looking for a thorough code review at the moment, more of a look at the structure of what we are exporting.  The big changes are

1. `read_elem_as_dask`->`read_elem_lazy` becomes a more general method
2. `read_backed` is exported as a one-stop-shop for reading everything at once

Do we want this way of doing things?  Or is there some other route?

Separately, are the changes made to the core acceptable?  After that, I think we can look into the specifics of the code I added.  Or you can review that now, but I'd rather get big changes out of the way first.

ilan-gold:
I will continue to make little changes to clean things up (this is still a draft!) but I think this structure is the way I would go.  But maybe you have different ideas!

ci/scripts/min-deps.py
```diff
@@ -62,7 +62,7 @@ def extract_min_deps(
             for extra in req.extras:
                 extra_deps = pyproject["project"]["optional-dependencies"][extra]
                 dependencies += map(Requirement, extra_deps)
-        else:
+        elif req.name != "xarray":
```
flying-sheep, line 65: I know this is just a draft, but if this got in, it would be a heinously hard-to-find hack. let’s not special-case dependencies in the middle of the logic here lol

pyproject.toml
```diff
@@ -98,12 +98,15 @@ test = [
     "awkward>=2.3",
     "pyarrow",
     "pytest_memray",
-    "pytest-mock"
+    "pytest-mock",
+    "xarray>=2024.06.0"
```
flying-sheep, line 102:
```suggestion
    "xarray>=2024.06.0",
```

src/anndata/_core/anndata.py
```diff
@@ -1014,28 +1039,12 @@ def __getitem__(self, index: Index) -> AnnData:
...
-            if np.array(color_vec).ndim == 0:
-                # Make 0D arrays into 1D ones
-                uns[color_key] = np.array(color_vec)[(None,)]
-            elif len(color_vec) != len(all_categories):
-                # Reset colors
-                del uns[color_key]
-            else:
-                idx = np.where(np.isin(all_categories, df_sub[k].cat.categories))[0]
-                uns[color_key] = np.array(color_vec)[(idx,)]
+        _remove_unused_categories(df_full, df_sub, uns)
```
flying-sheep, line 1047: is it necessary to keept the method if you make it into a function anyway? why not just change all call sites to the function?

flying-sheep (changes requested):
I really like the approach.

Would of course be better if that stuff got upstreamed, but with only the category and mask handling being done by us, this is feasible I think.

There’s one hack in there that I really don’t want us to leave in, otherwise already looks quite clean.

ci/scripts/min-deps.py
ilan-gold: That's a totally fair point.  What should be done then?  `xarray` requires pandas 2.0 which breaks our min-deps resolution.

ilan-gold: And it's not relevant for the `min-deps` testing anyway.

pyproject.toml
```diff
@@ -98,12 +98,15 @@ test = [
     "awkward>=2.3",
     "pyarrow",
     "pytest_memray",
-    "pytest-mock"
+    "pytest-mock",
+    "xarray>=2024.06.0",
```
flying-sheep, line 102:
```suggestion
    "anndata[xarray]",
```

DRY

ci/scripts/min-deps.py
flying-sheep: the exact same way as in scanpy:

```toml
[project.optional-dependencies]
test = [...]  # as before this PR
test-full = ["anndata[test]", "anndata[xarray]"]
xarray = ["xarray>=..."]
```

then modify the CI setup so it has a `TEST_EXTRA` variable defaulting to `test-full` that’s overridden for the min-deps job.

ilan-gold: cool!

ilan-gold:
> Would of course be better if that stuff got upstreamed, but with only the category and mask handling being done by us, this is feasible I think, but I don’t have a lot of xarray experience.

The ability to do this was upstreamed, but this is really just i/o and I don't think we can expect the xarray folks to take our custom i/o formats.  Would be nice though.  Eventually, they will do i/o for this and hopefully we can use their i/o completely.

ilan-gold:
Without @ivirshup 's high level review, I think @flying-sheep if you want to look at the code closer, that'd be great.  I might avoid the big `read_backed` method as I am not sure that will go in the final PR even though I would like it to.

ilan-gold:
One potential alteration would be to remove `read_backed` and create a registered method for the full `AnnData` object on `read_elem_lazy`

flying-sheep:
> @flying-sheep if you want to look at the code closer, that'd be great.

will do today!

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,155 @@
...
+        self._ordered = ordered
+        self._categories_cache = None
+        self._codes = ZarrOrHDF5Wrapper[type(codes)](codes)
+        self.shape = self._codes.shape
+        self.dtype = CategoricalDtypeAccessor(
+            categories=self._categories, ordered=self._ordered
+        )
+
+    @property
+    def categories(self):  # __slots__ and cached_property are incompatible
```
flying-sheep, line 93: I don’t see any `__slots__`. Does `BackendArray` define it? And doesn’t subclassing allow to have a `__dict__` and therefore `cached_property`s?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,155 @@
...
+        self, key: xr.core.indexing.ExplicitIndexer
+    ) -> xr.core.extension_array.PandasExtensionArray:
+        codes = self._codes[key]
+        categorical_array = pd.Categorical.from_codes(
+            codes=codes, categories=self.categories, ordered=self._ordered
+        )
+        if settings.should_remove_unused_categories:
+            return xr.core.extension_array.PandasExtensionArray(
+                categorical_array.remove_unused_categories()
+            )
```
flying-sheep, line 115:
```suggestion
            categorical_array = categorical_array.remove_unused_categories()
```

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class MaskedArray(BackendArray):
+    def __init__(
+        self,
+        values: ZarrArray | H5Array,
+        dtype_str: str,
+        mask: ZarrArray | H5Array | None = None,
+    ):
+        self._mask = ZarrOrHDF5Wrapper[type(mask)](mask)
```
flying-sheep, line 126: what are the `[]` for? if this is generics, you need to work with TypeVars instead.

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,155 @@
...
+                    pd.arrays.IntegerArray(values, mask=mask)
+                )
+            elif self._dtype_str == "nullable-boolean":
+                return xr.core.extension_array.PandasExtensionArray(
+                    pd.arrays.BooleanArray(values, mask=mask)
+                )
+        return xr.core.extension_array.PandasExtensionArray(pd.array(values))
+
+
+@_subset.register(DataArray)
```
flying-sheep, line 148: Hm, I’m not sure if it’s good practice to register more here in another file.

I think I like it …

src/anndata/experimental/backed/_xarray.py
```diff
@@ -0,0 +1,110 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from typing import Any, Literal
+
+    from ..._core.index import Index
+    from ._compat import DataArray
+
+
+def get_index_dim(ds):
```
flying-sheep, line 21: types please

src/anndata/experimental/backed/_xarray.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns
+        -------
+            The index of the of the dataframe as resolved from :attr:`~xarray.Dataset.coords`.
+        """
+        coord = list(self.coords.keys())[0]
+        return pd.Index(self.coords[coord].data)
+
+    @property
+    def shape(
+        self,
```
flying-sheep, line 42:
```suggestion
        self
```

src/anndata/experimental/backed/_xarray.py
```diff
@@ -0,0 +1,110 @@
...
+        return IlocGetter(self)
+
+
+@_subset.register(Dataset2D)
+def _(a: DataArray, subset_idx: Index):
+    key = get_index_dim(a)
+    if (
+        isinstance(subset_idx, tuple) and len(subset_idx) == 1
+    ):  # xarray seems to have some code looking for a second entry in tuples
+        return a.isel(**{key: subset_idx[0]})
```
flying-sheep, line 78:
```suggestion
        subset_idx = subset_idx[0]
```

src/anndata/tests/helpers.py
```diff
@@ -969,20 +970,33 @@ def shares_memory_sparse(x, y):
     class AccessTrackingStore(zarr.DirectoryStore):
         def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
-            self._access_count = {}
-            self._accessed_keys = {}
+            self._access_count = defaultdict(int)
```
flying-sheep, line 973: use `Counter` here

tests/test_concatenate.py
```diff
@@ -152,6 +152,9 @@ def fix_known_differences(orig, result, backwards_compat=True):
     for k, dtype in orig.obs.dtypes.items():
         if isinstance(dtype, pd.CategoricalDtype) and dtype.ordered:
             result.obs[k] = result.obs[k].astype(dtype)
+    for k, dtype in orig.obsm["df"].dtypes.items():
+        if isinstance(dtype, pd.CategoricalDtype) and dtype.ordered:
+            result.obsm["df"][k] = result.obsm["df"][k].astype(dtype)
```
flying-sheep, line 157: please dedupe. you could e.g. use an inline function?

docs/api.md
```diff
@@ -148,6 +149,9 @@ Types used by the former:
    experimental.ReadCallback
    experimental.WriteCallback
    experimental.StorageType
+   experimental.backed._lazy_arrays.MaskedArray
+   experimental.backed._lazy_arrays.CategoricalArray
+   experimental.backed._xarray.Dataset2D
```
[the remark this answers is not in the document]
ilan-gold: Hmmm, that's an interesting point.  I thought that by exporting `read_lazy` in general, I might appease @ivirshup pre-emptively by not creating a monolith function and making something more composable a la `read_elem`.  But in fact, by doing that, we create more API surface, yes.  `Dataset2D` is something that will be visible on the `AnnData` object via `adata.obs` but the others would not be if we didn't make `read_lazy` so general (i.e., because we added functionality to it from its previous state as `read_elem_as_dask`).  Tough question.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -154,11 +173,104 @@ def read_h5_array(
...
+                v, coords=[v], dims=[index_label], name=index_label
+            )
+        else:
+            d_with_xr[k] = v
+    return Dataset2D(d_with_xr)
+
+
+@_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("categorical", "0.2.0"))
+@_LAZY_REGISTRY.register_read(H5Group, IOSpec("categorical", "0.2.0"))
+def read_categorical(
```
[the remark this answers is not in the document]
ilan-gold: It's part of the required signature from `read_elem_as_dask`.  I can make the `LazyReader`/`DaskReader` class check for the `chunks` param and use it if it is needed and not otherwise.  So we could drop it here.

src/anndata/_core/views.py
```diff
@@ -305,6 +306,11 @@ def as_view_dask_array(array, view_args):
 
 @as_view.register(pd.DataFrame)
 def as_view_df(df, view_args):
+    if settings.should_remove_unused_categories:
+        for col in df.columns:
+            if isinstance(df[col].dtype, pd.CategoricalDtype):
+                with pd.option_context("mode.chained_assignment", None):
+                    df[col] = df[col].cat.remove_unused_categories()
```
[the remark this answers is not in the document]
ilan-gold: I genuinely do not know.  This _seems_ to be unintentional (i.e., missed spots to do this) but the original feature request for the setting seemed only for deduping the `__init__` and/or `obs` (and this is run on `obsm` dataframe views since dedupping happens for `obs` on `__init__`).  I would be in favor of a patch since we are moving to `0.11.0` anyway so can break whatever.

src/anndata/_core/storage.py
```diff
@@ -72,12 +72,23 @@ def coerce_array(
...
+                return "mock anndata.experimental.backed._xarray."
+
     """Coerce arrays stored in layers/X, and aligned arrays ({obs,var}{m,p})."""
     # If value is a scalar and we allow that, return it
     if allow_array_like and np.isscalar(value):
         return value
     # If value is one of the allowed types, return it
-    if isinstance(value, StorageType.classes()):
+
+    if isinstance(value, StorageType.classes()) or isinstance(value, Dataset2D):
```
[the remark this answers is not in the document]
ilan-gold: It seems like there have been a few iterations on this functionality and at one point, this was used in a way where doing so would have broken other things (also, I think, why `pd.DataFrame` is not in it as below).  But that has since changed.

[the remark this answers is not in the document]
ilan-gold: Ah, right, https://github.com/scverse/anndata/pull/1247/files#diff-3b4874efa9bc9a522fa0cc5a88c69e233d7b46f2993331162bab12c4a6371bdcR75-R77

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: Ah, so I think this is again a result of old code, when we had slots, but not 100% sure.  If it breaks, we'll see and can revert.

src/anndata/experimental/backed/_io.py
```diff
@@ -0,0 +1,98 @@
...
+                "Did not read zarr as consolidated. Consider consolidating your metadata."
+            )
+            has_keys = False
+            f = zarr.open(store, mode="r")
+    else:
+        f = h5py.File(store, mode="r")
+
+    def callback(func, elem_name: str, elem, iospec):
+        if iospec.encoding_type == "anndata" or elem_name.endswith("/"):
+            cols = [
```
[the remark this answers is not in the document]
ilan-gold: Hmmm, not sure...we don't really use it anywhere else but that's not a bad idea.  I'll have a look into it and whether it makes sense to even use it elsewhere (i.e., a different PR)

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+@pytest.fixture(params=["zarr", "h5ad"], scope="session")
+def dskfmt(request):
+    return request.param
+
+
+@pytest.fixture(scope="session")
+def adata_remote_orig(
+    tmp_path_factory, dskfmt: str, mtx_format
+) -> tuple[AnnData, AnnData]:
+    orig_path = Path(tmp_path_factory.mktemp(f"orig.{dskfmt}"))
```
flying-sheep, line 40: All `tmp_path*` fixtures already return `Path` objects

```suggestion
    orig_path = tmp_path_factory.mktemp(f"orig.{dskfmt}")
```

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+    orig = gen_adata((1000, 1000), mtx_format)
+    orig.write_zarr(orig_path)
+    remote = read_backed(orig_path)
+    return remote, orig
```
flying-sheep, line 44: This is session-scoped fixture returning a backed AnnData object.

What’s the idea here? Tests shouldn’t be interdependent, so I assume the tests want to uphold the invariant of not mutating these objects. Can guarantee somehow that we uphold this invariant? E.g. opening readonly or making this function-scoped, or so?

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@pytest.fixture
+def adata_remote_with_store_tall_skinny(tmp_path_factory, mtx_format):
+    orig_path = Path(tmp_path_factory.mktemp("orig.zarr"))
```
flying-sheep, line 49:
```suggestion
    orig_path = tmp_path_factory.mktemp("orig.zarr")
```

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+    remote.obs["cat"]
+    assert store.get_access_count("obs/int64") == 0, store.get_subkeys_accessed(
+        "obs/int64"
+    )
+    assert (
+        store.get_access_count("obs/cat/categories") == 0
+    ), store.get_subkeys_accessed("obs/cat/categories")
+    subset = remote[
+        (remote.obs["cat"] == "a").data, :
+    ]  # `.data` for xarray, but should we handle internally?
```
flying-sheep, line 94: you mean to allow indexing an AnnData containing an xarray with a boolean xarray? Good idea, if it’s not too complicated to implement.

And if it is, we could bench it and refactor our index handling until it’s easy to implement.

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+    orig.write_zarr(orig_path)
+    store = AccessTrackingStore(orig_path)
+    remote = read_backed(store)
+    return remote, store
+
+
+needs_xarray = pytest.mark.skipif(
+    not find_spec("xarray"), reason="Xarray is not installed"
+)
```
flying-sheep, line 69: you could just set [pytestmark](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=pytest_marks#globalvar-pytestmark) once at the top, then the whole file is skipped and you don’t have to annotate each individual function.

```py
pytestmark = pytest.mark.skipif(
    not find_spec("xarray"), reason="Xarray is not installed"
)
```

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+    sub_subset.obs["int64"]
+    assert store.get_access_count("X") == 0, store.get_subkeys_accessed("X")
+    assert store.get_access_count("obs/int64") == 0, store.get_subkeys_accessed(
+        "obs/int64"
+    )
+    assert store.get_access_count("var/int64") == 0, store.get_subkeys_accessed(
+        "var/int64"
+    )
+    # all codes read in for subset
+    assert store.get_access_count("obs/cat/codes") == 4, store.get_subkeys_accessed(
```
flying-sheep, line 106: that means “4” because there’s 4 codes?

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+        "obs/cat/codes"
+    )
+    remote[0:10, :].obs["int64"][0:10].compute()
+    assert store.get_access_count("obs/int64") == 1, store.get_subkeys_accessed(
+        "obs/int64"
+    )
+    # one for 0, .zmetadata handles .zarray
```
flying-sheep, line 113: I don’t understand either part of this comment, maybe expand a bit

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+    (Path(orig_pth) / ".zmetadata").unlink()
+    store = AccessTrackingStore(orig_pth)
+    store.initialize_key_trackers(["obs/.zgroup", ".zgroup"])
+    with pytest.warns(UserWarning, match=r"Did not read zarr as consolidated"):
+        remote = read_backed(store)
+    remote_to_memory = remote.to_memory()
+    assert_equal(remote_to_memory, adata)
+    assert store.get_access_count("obs/.zgroup") == 1, store.get_subkeys_accessed(
+        "obs/.zgroup"
+    )
```
flying-sheep, line 190: you know, reading this pattern again and again, maybe we should make this a helper.

It could go directly into the `AccessTrackingStore`.

That way it’s not possible to ever use two different strings on the left and right side of this statement, which would probably cause hours of debugging.

tests/test_read_backed_experimental.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@needs_xarray
+def test_unconsolidated(tmp_path, mtx_format):
+    adata = gen_adata((1000, 1000), mtx_format)
+    base_pth = Path(tmp_path)
+    orig_pth = base_pth / "orig.zarr"
+    write = lambda x: getattr(x, "write_zarr")(orig_pth)
+    write(adata)
+    (Path(orig_pth) / ".zmetadata").unlink()
```
flying-sheep, line 181: you don’t seem to trust that `Path`s stay `Path`s haha

```suggestion
    orig_pth = tmp_pth / "orig.zarr"
    adata.write_zarr(orig_pth)
    (orig_pth / ".zmetadata").unlink()
```

flying-sheep (changes requested):
Very nice!

I think there are tests missing for a bunch of moving parts like Dataset2D, but a lot of the public functionality seems to already be there and working well!

src/anndata/_core/storage.py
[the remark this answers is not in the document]
flying-sheep: We should probably fix than in your other PR then before having this hack in here. (also the comment still mentions `StorageType`, not `ArrayDataStructureType`)

docs/api.md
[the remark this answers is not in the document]
flying-sheep: As said in the “X type consolidation” PR, I think it’s fine to not export the full API surface of the classes we have and just commit to a carefully designed protocol instead.

tests/test_read_backed_experimental.py
ilan-gold: 4 because there are 4 chunks under `obs/cat/codes`

flying-sheep:
```suggestion
    # all codes read in for subset (from 4 chunks)
    assert store.get_access_count("obs/cat/codes") == 4, store.get_subkeys_accessed(
```

tests/test_read_backed_experimental.py
ilan-gold: So the idea here is that `read_backed` is read-only actually (if you look, all `mode="r"` there)

tests/test_read_backed_experimental.py
ilan-gold: Yes I avoided initially exactly because of complexity, but maybe doable now with the `compat` shim

src/anndata/experimental/backed/_compat.py
```diff
@@ -0,0 +1,42 @@
...
+    xr = None
+
+
+try:
+    from xarray.backends.zarr import ZarrArrayWrapper
+except ImportError:
+
+    class ZarrArrayWrapper:
+        def __repr__(self) -> str:
+            return "mock ZarrArrayWrapper"
```
ivirshup, line 24: This code means that you get a really confusing error if you run `read_backed` without `xarray` being installed

ilan-gold: Yes but unfortunately I think this compat layer is needed so that core is compatible.  Maybe we can add a check within `read_backed`?

src/anndata/experimental/backed/_io.py
```diff
@@ -0,0 +1,98 @@
...
+            warnings.warn(
+                "Did not read zarr as consolidated. Consider consolidating your metadata."
+            )
+            has_keys = False
+            f = zarr.open(store, mode="r")
+    else:
+        f = h5py.File(store, mode="r")
+
+    def callback(func, elem_name: str, elem, iospec):
+        if iospec.encoding_type == "anndata" or elem_name.endswith("/"):
```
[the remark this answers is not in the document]
ilan-gold: Presumably root store, we use this in all the other read methods...

ebezzi:
Issue: if you run `adata.obs["cell_type"].to_pandas()` (on a dataset from cellxgene), the output should be a categorical, but it is actually an `object`:

```
obs_names
0                                   macrophage
1                                 stromal cell
2       CD8-positive, alpha-beta memory T cell
3       CD8-positive, alpha-beta memory T cell
4        CD141-positive myeloid dendritic cell
                         ...                  
9419                              stromal cell
9420                              stromal cell
9421                              stromal cell
9422    CD8-positive, alpha-beta memory T cell
9423                              stromal cell
Length: 9424, dtype: object
```

src/anndata/_core/merge.py
```diff
@@ -757,6 +757,7 @@ def np_bool_to_pd_bool_array(df: pd.DataFrame):
     return df
 
 
+# TODO: concat for xarray
```
ivirshup, line 760: This would be nice to have with this PR. What are your thoughts on this @ilan-gold?

ilan-gold: Yea......I agree......

ilan-gold: Ok, sorry, I remember now.  I looked into this...it's a lot of work...but maybe I'll try my hand at it....sometimes you get fatigue just from how daunting something seems

ilan-gold:
@ivirshup filed https://github.com/pydata/xarray/issues/9519 but it should not block if it does not come out before this does IMO.

src/anndata/_core/merge.py
ilan-gold: Ah @dcherian I am actually almost 100% sure this is already implemented in `xarray`, but we need to get the functionality here!  I think it's all good on your end but I will let you know if we need anything!

ilan-gold:
@ivirshup Re: concatenation, I found https://github.com/pydata/xarray/issues/4628#issuecomment-2368230984

If we wanted to work around this, we could map-blocks to `object` (or even integer-code) numpy arrays and use dask.  I think this should work but it's not ideal....

ilan-gold:
Discussion point: creating dummy indices to speed up first-load time.  The real index will simply be stored in `obs_names`.

src/anndata/_core/merge.py
```diff
@@ -723,7 +727,14 @@ def default_fill_value(els):
...
     """
-    if any(isinstance(el, (sparse.spmatrix, SpArray)) for el in els):
+    if any(
+        isinstance(el, (sparse.spmatrix, SpArray))
+        or (
+            isinstance(el, DaskArray)
+            and isinstance(el._meta, (sparse.spmatrix, SpArray))
+        )
+        for el in els
+    ):
```
ilan-gold, line 737: A bug @ivirshup or @flying-sheep? Needs its own PR?  This was spitting out NaNs before for dask-sparse

ilan-gold:
> Discussion point: creating dummy indices to speed up first-load time. The real index will simply be stored in xxx_names

Done, tested, with `anndata.concat`

ilan-gold:
TODO: `uns` lazily

docs/api.md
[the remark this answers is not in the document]
flying-sheep: If we do this, we don’t need to write anything, people will only see the superclass, i.e. the stable API:

```py
qualname_overrides = {
    ...,
    "anndata.xyz.MaskedArray": "numpy.ma.Array",
}
```

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+K = TypeVar("K", H5Array, ZarrArray)
+
+
+class ZarrOrHDF5Wrapper(ZarrArrayWrapper, Generic[K]):
+    def __init__(self, array: K):
+        if isinstance(array, ZarrArray):
+            return super().__init__(array)
+        self._array = array
```
flying-sheep, line 27: you don’t call `super().__init__()` otherwise? Are you sure you’re setting all fields?

src/anndata/experimental/backed/_xarray.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Returns
+        -------
+            :class:`pandas.Index` that represents the "columns."
+        """
+        columns_list = list(self.keys())
+        return pd.Index(columns_list)
+
+
+@_subset.register(Dataset2D)
```
flying-sheep, line 95: Do we document this set of functions somewhere? seems like this is the set of functions each new array type needs support for.

tests/test_readwrite.py
```diff
@@ -847,3 +847,11 @@ def test_h5py_attr_limit(tmp_path):
         np.ones((5, N)), index=a.obs_names, columns=[str(i) for i in range(N)]
     )
     a.write(tmp_path / "tmp.h5ad")
+
+
+@pytest.mark.skipif(
+    find_spec("xarray"), reason="Xarray is installed so `read_lazy` will not error"
+)
```
flying-sheep, line 854: great idea!

pyproject.toml
```diff
@@ -109,6 +110,7 @@ cu12 = ["cupy-cuda12x"]
 cu11 = ["cupy-cuda11x"]
 # https://github.com/dask/dask/issues/11290
 dask = ["dask[array]>=2022.09.2,<2024.8.0"]
+lazy = ["xarray>=2024.06.0", "aiohttp", "requests", "zarr<3.0.0a0", "anndata[dask]"]
```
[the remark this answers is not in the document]
flying-sheep: Huh I deleted this message draft, GitHub is weird sometimes.

I realized we have no zarr extra, so this is the only way to do it. If we had one, we should use it here.

docs/conf.py
```diff
@@ -108,6 +108,7 @@
     ("py:class", "numpy.ma.core.MaskedArray"),
     ("py:class", "dask.array.core.Array"),
     ("py:class", "awkward.highlevel.Array"),
+    ("py:class", "awkward.Array"),
```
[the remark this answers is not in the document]
ilan-gold: Ah yes, we can't actually.  https://github.com/sphinx-doc/sphinx/issues/10591 prevents us from using autodoc mocking with singledispatch in our codebase.  so we could add `awkward` to docs build or do this.

src/anndata/_core/merge.py
```diff
@@ -1053,11 +1071,91 @@ def concat_Xs(adatas, reindexers, axis, fill_value):
...
+
+    annotations_with_only_dask = make_xarray_extension_dtypes_dask(annotations)
+    attrs = get_attrs(annotations_with_only_dask)
+    index_name = np.unique([a.index.name for a in annotations])[0]
+    return Dataset2D(
+        xr.concat(annotations_with_only_dask, join=join, dim=index_name), attrs=attrs
+    )
+
+
 def concat(
```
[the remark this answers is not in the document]
ilan-gold: I would say that this request is out of scope for this PR, which is already quite large.  I will file an issue

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -146,19 +167,147 @@ def read_h5_array(
...
+
+    elem_dict = {
+        k: _reader.read_elem(elem[k])
+        for k in [*elem.attrs["column-order"], elem.attrs["_index"]]
+    }
+    elem_name = get_elem_name(elem)
+    # remove end for obsm/varm
+    obs_var_match = re.findall(r"(obs|var)", elem_name)
+    if not len(obs_var_match):
+        label_based_indexing_key = "index"
```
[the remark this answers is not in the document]
ilan-gold: The match value is used below - my intention (not carried out) was to handle cases where `obs` appears as a name within the the elem_name i.e., `obsm/obs_df` or something.  Updated to handle this

src/testing/anndata/_pytest.py
```diff
@@ -55,6 +56,9 @@ def _doctest_env(
     from anndata.utils import import_name
 
     assert isinstance(request.node.parent, pytest.Module)
+    if "experimental/backed" in str(request.node.path):
```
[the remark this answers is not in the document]
ilan-gold: Can you add a mark for docstring tests? I did some digging but could not find anything like that

tests/test_read_lazy.py
```diff
@@ -0,0 +1,373 @@
...
+    if dskfmt == "h5ad":
+        orig_path = tmp_path_factory.mktemp("h5ad_file_dir") / f"orig.{dskfmt}"
+    else:
+        orig_path = tmp_path_factory.mktemp(f"orig.{dskfmt}")
+    orig = gen_adata((1000, 1100), mtx_format)
+    orig.raw = gen_adata((1000, 1100), mtx_format)
+    getattr(orig, f"write_{dskfmt}")(orig_path)
+    return lambda: read_lazy(
+        orig_path, load_annotation_index=load_annotation_index
+    ), orig
```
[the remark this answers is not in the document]
ilan-gold: This and the resetting of key trackers was due to thread safety problems - resolved both with some nice fixture scoping

tests/test_read_lazy.py
```diff
@@ -0,0 +1,373 @@
...
+    remote[0:10, :].obs["int64"].compute()
+    store.assert_access_count("obs/int64", 1)
+    # .zmetadata handles .zarray so simple access does not cause any read
+    store.assert_access_count("var/int64", 0)
+    store.assert_access_count("raw", 0)
+
+
+def test_access_count_index(adata_remote_with_store_tall_skinny):
+    _, store = adata_remote_with_store_tall_skinny
+    store.reset_key_trackers()
```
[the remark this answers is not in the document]
ilan-gold: Yup, fixed now

tests/test_read_lazy.py
```diff
@@ -0,0 +1,373 @@
...
+    store.initialize_key_trackers(["obs/_index"])
+    read_lazy(store, load_annotation_index=False)
+    store.assert_access_count("obs/_index", 0)
+    read_lazy(store)
+    # 8 is number of chunks
+    store.assert_access_count("obs/_index", 4)
```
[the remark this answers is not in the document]
ilan-gold: old comment, size of adata was halved

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: I added a check against their `__slots__` now

src/anndata/experimental/backed/_io.py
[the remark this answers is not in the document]
ilan-gold: Made as type

docs/api.md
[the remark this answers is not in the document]
ilan-gold: Back to our sphinx-autodoc-singledispatch issue, I don't think we can do this because we don't install the package unless this can work without the package installed or mocked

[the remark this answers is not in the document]
ilan-gold: hmm, right the intersphinx mapping is independent of installed packages!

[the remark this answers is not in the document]
ilan-gold: Ok @flying-sheep I am not sure `qualname_overrides` works this way - I tried adding it in but could not get the links to redirect to the xarray docs

src/anndata/_core/index.py
```diff
@@ -50,10 +50,13 @@ def _normalize_index(
     | pd.Index,
     index: pd.Index,
 ) -> slice | int | np.ndarray:  # ndarray of int or bool
-    if not isinstance(index, pd.RangeIndex):
-        msg = "Don’t call _normalize_index with non-categorical/string names"
-        assert index.dtype != float, msg
-        assert index.dtype != int, msg
+    from ..experimental.backed._compat import xr
+
+    # TODO: why is this here? All tests pass without it and it seems at the minimum not strict enough.
```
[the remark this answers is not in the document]
ilan-gold: But it would be good to understand what we even going on here @ivirshup ?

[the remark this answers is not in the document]
ilan-gold: I mean if the tests pass, we're good? And if we're not good, we should add a test?

tests/test_read_lazy.py
```diff
@@ -0,0 +1,592 @@
...
+        == orig_concatenated.var_names.tolist()
+    )
+
+
+@pytest.mark.parametrize(
+    "elem_key",
+    map(
+        lambda x: pytest.param(x, id="-".join(map(str, x))),
+        [("obs", None), ("var", None), ("obsm", "df"), ("varm", "df")],
+    ),
```
flying-sheep, line 555: please use comprehensions instead of mapping lambdas.

also please unpack the tuple into two parameters here instead of doing `elem_key[0|1]` a bunch, something like: `("attr", "key"), [pytest.param(attr, key, id=...) for attr, key in [...]]`

tests/test_read_lazy.py
```diff
@@ -0,0 +1,592 @@
...
+        pytest.param(
+            slice(500, 1500),
+            id="slice",
+        ),
+        pytest.param(
+            np.arange(950, 1050),
+            id="consecutive integer array",
+        ),
+        pytest.param(
+            np.random.choice(np.arange(800, 1100), 500),
```
flying-sheep, line 501: isn’t that [randint](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html)?

tests/test_read_lazy.py
```diff
@@ -0,0 +1,592 @@
...
+    from anndata.experimental.backed._compat import Dataset2D
+
+    remote, orig = adata_remote_orig
+
+    maybe_warning_context = (
+        pytest.warns(UserWarning, match=r"Concatenating with a pandas numeric")
+        if not load_annotation_index
+        else nullcontext()
+    )
+    with maybe_warning_context:
```
flying-sheep, line 528: thanks to Python 3.10, we can now do this:

```suggestion
    with (
        pytest.warns(UserWarning, match=r"Concatenating with a pandas numeric")
        if not load_annotation_index
        else nullcontext()
    ):
```

tests/test_read_lazy.py
```diff
@@ -0,0 +1,592 @@
...
+    store.initialize_key_trackers(["obs/.zgroup", ".zgroup"])
+    with pytest.warns(UserWarning, match=r"Did not read zarr as consolidated"):
+        remote = read_lazy(store)
+    remote_to_memory = remote.to_memory()
+    assert_equal(remote_to_memory, adata)
+    store.assert_access_count("obs/.zgroup", 1)
+
+
+# remote has object dtype, need to convert back for integers booleans etc.
+def correct_extension_dtype_differences(remote: pd.DataFrame, memory: pd.DataFrame):
```
flying-sheep, line 335: why does remote have object dtypes? can that be fixed ahead of time? will that affect users or is that just here in the tests?

also for clarity, please call it `fix_extension_dtype_differences` or `unify_extension_dtypes` and make the comment a docstring. I was unsure if `correct` was meant as a verb or something like `make_correct_...`

tests/test_read_lazy.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+@pytest.fixture
+def concatenation_objects(
+    adatas_paths_var_indices_for_concatenation,
+) -> tuple[list[AnnData], list[pd.Index], list[AccessTrackingStore], list[AnnData]]:
+    adatas, paths, var_indices = adatas_paths_var_indices_for_concatenation
+    stores = [AccessTrackingStore(path) for path in paths]
+    lazys = [read_lazy(store) for store in stores]
+    return adatas, var_indices, stores, lazys
```
flying-sheep, line 173: please split this into 4 fixtures, that way typing below is much simpler, and you can just do

```py
def mystest(objs1, objs2, objs3, objs4): ...
```

instead of

```py
def mystest(concatenation_objects):
    objs1, objs2, objs3, objs4 = concatenation_objects
    ...
```

tests/test_read_lazy.py
```diff
@@ -0,0 +1,373 @@
...
+def mtx_format(request):
+    return request.param
+
+
+@pytest.fixture(params=["zarr", "h5ad"], scope="session")
+def dskfmt(request):
+    return request.param
+
+
+@pytest.fixture(params=[True, False], scope="session")
```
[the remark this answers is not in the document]
flying-sheep: why did you mark this as solved?

flying-sheep (changes requested):
Please make a PR into the notebook repo, there are some things broken in the notebook that I’d like to properly review.

Also I’m still not a fan of the exports from private namespaces, but I’m OK with it if there are at least minimal docstrings for them.

This is very close! Thanks for taking on this huge thing!

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class ZarrOrHDF5Wrapper(ZarrArrayWrapper, Generic[K]):
+    def __init__(self, array: K):
+        if isinstance(array, ZarrArray):
+            return super().__init__(array)
+        if set(self.__slots__) != {"dtype", "shape", "_array"}:
+            msg = "Expected attributes of xarray ZarrArrayWrapper have changed - "
+            "please file an issue with anndata and consider downgrading xarray"
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
ilan-gold: Yea, probably a good idea...this doesn't really have an error at the moment, and will error anyway on its own in the future (as the previous warning showed)

tests/test_read_lazy.py
ilan-gold: I will change the name and add a better comment since it's only used for concat.  There is no way of concat-ing these lazy pandas adapter arrays we have so in order to work with them, we present a dask array wrapping each when they are concatenated.  Practically, this actually doesn't change anything.  But it means that when we make the round trip, we lose the original data type.  We could maybe store this in `uns` or something (what the original data type was) but

1. People concating remote/large datasets probably won't be reading them into memory very often
2. This gets complicated quickly with mixed data types.  So if you have two similarly named columns but different numerical data types, we need to start dealing with upcasting/downcasting.

Feels like a cross-that-bridge-when-we-come-to-it issue

ilan-gold:
@flying-sheep https://github.com/scverse/anndata-tutorials/pull/20/ notebook PR but it seems to run ok for me? Could you post your error?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -0,0 +1,143 @@
...
+        self._array = array
+        self.shape = self._array.shape
+        self.dtype = self._array.dtype
+
+    def __getitem__(self, key):
+        if isinstance(self._array, ZarrArray):
+            return super().__getitem__(key)
+        # adapted from https://github.com/pydata/xarray/blob/main/xarray/backends/h5netcdf_.py#L50-L58C13
+        # TODO: locks?
```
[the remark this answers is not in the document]
ilan-gold: This reminded me about clusters with Dask but I honestly don't even know what the use-case is here for them.  They have some locking mechanism but it has nothing to do with dask.  So I think if this comes up, we can deal with it then, but I'm not really sure what a lock would do for us.  I'll do some more digging but if I don't further respond here, I would just move on

flying-sheep (changes requested):
Looks great! This time I could actually go through it all and understand almost everything.

Apart from one spot (the `re.findall`) everything looks right!

Are there any tutorials that still use `read_elem_dask`?

tests/test_read_lazy.py
```diff
@@ -0,0 +1,667 @@
...
+def test_concat_access_count(
+    adatas_for_concat: list[AnnData],
+    stores_for_concat: list[AccessTrackingStore],
+    lazy_adatas_for_concat: list[AnnData],
+    join: Join_T,
+    elem_key: AnnDataElem,
+    sub_key: str,
+    simple_subset_func: Callable[[AnnData], AnnData],
+):
+    # track all elems except codes because they must be read in for concatenation
```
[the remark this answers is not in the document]
ilan-gold: Ah, sorry it's categories, the comment was mistaken (as you can see below, this test DOES track `codes`)

src/anndata/_core/merge.py
```diff
@@ -1337,19 +1517,49 @@ def concat(
...
+            join=join,
+            ignore_index=True,
+        )
+    else:
+        concat_annot = concat_dataset2d_on_annot_axis(annotations, join)
     concat_annot.index = concat_indices
     if label is not None:
         concat_annot[label] = label_col
 
     # Annotation for other axis
```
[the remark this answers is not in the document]
ilan-gold: Merge for `var` and concatenation for `obs` - we stack the cells, and do some merge on the gene names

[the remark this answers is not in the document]
flying-sheep: I realized that later when reading the tests, thanks!

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -176,11 +206,157 @@ def read_h5_array(
...
+) -> Dataset2D:
+    from anndata.experimental.backed._compat import Dataset2D
+
+    elem_dict = {
+        k: _reader.read_elem(elem[k])
+        for k in [*elem.attrs["column-order"], elem.attrs["_index"]]
+    }
+    elem_name = get_elem_name(elem)
+    # Determine whether we can use label based indexing i.e., is the elem `obs` or `var`
+    obs_var_matches = re.findall(r"(obs|var)", elem_name)
```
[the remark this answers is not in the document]
ilan-gold: Yeah, this was nuts convoluted anyway so I just removed it!

tests/test_read_lazy.py
```diff
@@ -0,0 +1,667 @@
...
+    # check preservation of non-categorical dtypes on the concat axis
+    assert concated_remote.obs["int64"].dtype == "int64"
+    assert concated_remote.obs["uint8"].dtype == "uint8"
+    assert concated_remote.obs["nullable-int"].dtype == "int32"
+    assert concated_remote.obs["float64"].dtype == "float64"
+    assert concated_remote.obs["bool"].dtype == "bool"
+    assert concated_remote.obs["nullable-bool"].dtype == "bool"
+
+
+def test_concat_to_memory_var(
```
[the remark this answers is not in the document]
ilan-gold: Tried to really clarify here!

src/anndata/_core/merge.py
```diff
@@ -206,6 +209,8 @@ def equal_awkward(a, b) -> bool:
 
 
 def as_sparse(x, *, use_sparse_array: bool = False):
+    if isinstance(x, DaskArray):
+        x = x.compute()
```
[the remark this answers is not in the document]
ilan-gold: Great call!

tests/lazy/test_concat.py
```diff
@@ -0,0 +1,331 @@
...
+import numpy as np
+import pandas as pd
+import pytest
+
+import anndata as ad
+from anndata._core.file_backing import to_memory
+from anndata.experimental import read_lazy
+from anndata.tests.helpers import assert_equal, gen_adata
+
+from .conftest import ANNDATA_ELEMS, get_key_trackers_for_columns_on_axis
```
flying-sheep, line 16: wait, I thought this doesn’t work. did they change that?

pyproject.toml
```diff
@@ -81,9 +81,11 @@ doc = [
     "sphinx_design>=0.5.0",
     "readthedocs-sphinx-search",
     # for unreleased changes
-    "anndata[dev-doc]",
+    "anndata[dev-doc,dask]",
+    "awkward>=2.3",
 ]
 dev-doc = ["towncrier>=24.8.0"] # release notes tool
+test-full = ["anndata[test,lazy]"]
```
ilan-gold, line 88: Maybe a different name?

src/anndata/_core/index.py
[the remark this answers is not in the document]
ilan-gold: Updated!

src/anndata/_core/merge.py
```diff
@@ -1094,11 +1114,184 @@ def concat_Xs(adatas, reindexers, axis, fill_value):
...
+        dim=DS_CONCAT_DUMMY_INDEX_NAME,
+    )
+    # prevent duplicate values
+    index.coords[DS_CONCAT_DUMMY_INDEX_NAME] = ds.coords[DS_CONCAT_DUMMY_INDEX_NAME]
+    ds[f"true_{DS_CONCAT_DUMMY_INDEX_NAME}"] = index
+    for key in set(a.attrs["indexing_key"] for a in annotations_re_indexed):
+        del ds[key]
+    if DUMMY_RANGE_INDEX_KEY in ds:
+        del ds[DUMMY_RANGE_INDEX_KEY]
+    return ds
```
ilan-gold, line 1287: This is new since last reivew, but hopefully is clear in what it is doing especially with the docstring - basically, it creates its own index to prevent loading indices into memory unecessarily or having to worry about concatenating the indices (which matches the in-memory behavior)

tests/lazy/test_concat.py
flying-sheep: yeah

pyproject.toml
flying-sheep: in scanpy we have
- `test-min` which is used in the minimum deps job,
- `test` which is a healthy subset of functionality, and
- `test-full`, which is everything (except for external I think)

ilan-gold:
> pytestmark should work and is less complex.

Putting at the top-level of the `conftest` didn't work so I could have added it to the files individually but this seems anti-scalable.

> I’m also still not happy about the deps here, it‘s pretty unclear what causes them to be there. Maybe annotate non-obvious ones so we can check if we need them and remove them at some point instead of cargo-culting them into infinity?

That's fair - maybe let's wait until zarr v3.  The reason is simply that "using remote data with zarr requires these, and otherwise you will get a `RuntimeError`"

> I’d still like to see a more local/explicit version of the "experimental/backed" in str(request.node.path)

I can redo this

ilan-gold:
I will open follow up PRs after this one to account for a few things, but, for now, I am going to leave this unmerged because the zarr v3 PR should go in first.

I am very happy with the state of things :)

ilan-gold:
@flying-sheep Here is the diff for mergin the zarr v3 into this branch - it's non-trivial so I wanted to give you a chance to peak this before merging:

https://github.com/scverse/anndata/pull/1888/commits/42264e47775770cca272d4c47d7f4ee4c9581153

Issue #890, opened 2023-01-30
Option to retain categories in columns when subsetting anndata object. 


gregjohnso:
I'm trying to figure out how to retain categorical values when subsetting an anndata object.

The issue im having is as follows:
When I create an anndata object, I can assign a categorical variable to a column in obs:
```Python
import numpy as np
import anndata as ad

n_data = 100
n_cagegories = 3

x = np.random.normal(size=(n_data, 2))
categories = np.random.choice(n_cagegories, size=n_data)

# Randomly assign categories ["c0", "c1", "c2"]
categories = np.array([f"c{c}" for c in categories])

adata = ad.AnnData(x, obs={"labels": categories})
# Make the column categorical
adata.obs["labels"] = adata.obs["labels"].astype("category")

# Observe this is a categorical column
print(adata.obs['labels'].cat.categories)
```
output:
```py
Index(['c0', 'c1', 'c2'], dtype='object')
```

If we subset the DATAFRAME to label "c0", we see that it retains all categories
```Python
# if we subset the DATAFRAME to "c0", we see that it retains all categories
c0_inds = adata.obs["labels"] == "c0"

df_obs = adata.obs[c0_inds]
print(df_obs['labels'].cat.categories)
```

output:
```py
Index(['c0', 'c1', 'c2'], dtype='object')
```

but if we subset the ANNDATA to "c0", we see that it loses all categories
```Python
# if we subset the ANNDATA to "c0", we see that it loses all categories
adata_c0 = adata[c0_inds]
print(adata_c0.obs['labels'].cat.categories)
```
output:
```py
Index(['c0'], dtype='object')
```

Is this expected? My intuition is that we should not expect different behavior between the anndata and dataframe objects. Is there a way to change this behavior easily?

gregjohnso:
I think this is the culprit: 
https://github.com/scverse/anndata/blob/master/anndata/_core/anndata.py#L345

rachelnhovde:
Fixing this issue is really important to my work as well; any way this can be prioritized?

flying-sheep:
To add the context from the duplicated issue here:

- Change was introduced in 8cabf9c86a38d6db88c664e2ea28e3fb29bdf99e
- Relevant code: https://github.com/scverse/anndata/blob/e067515887e0badfa5e02b7a75ac66679215ec2c/anndata/_core/anndata.py#L343-L346

[...]

Issue #1112, opened 2023-08-29
Make checking var_names and obs_names for uniqueness optional


niklasmueboe:
By default when creating a new AnnData instance in the `_init_as_actual` function both `var_names` and `obs_names` are checked for uniqueness. When working with AnnData objects with several million `obs`/`var` this can become a performance limiting factor taking multiple seconds or even minutes.

Making the uniqueness check optional would allow the user the option to skip it when instantiating a new AnnData object e.g. via `ad.AnnData(X=X, ... , check_unique=False)`. This could be useful to improve performance especially when the input data is already guaranteed to have unique `obs_names` and `var_names.`

This was originally raised (with an example implementation) as a PR #1081

niklasmueboe:
In my case I sometimes have up to ~200M `obs` (`var` is irrlevant in comparison) and then uniqueness checks can take around 2 mins.

I wasn't aware of the fact that pandas does uniquenesss checks when label-indexing anyway, and deferring the uniqueness check than would not necessary be beneficial (at least in my use case), because the computation will be done sooner or later.

I think that as long as pandas does not offer a solution for this (i.e. some public API for uniqueness guarantees) finding a real solution might be tricky.

Sidenote:
The issue partially arises due to the fact that only string indices are allowed while natively a MultiIndex of integers would be more suitable for me (and much faster, although not perfect). But I guess getting MultiIndex support might be even harder?

niklasmueboe:
> Integer values as indices is particularly hard because it's ambiguous with positional indexing (as discussed in previous issues on this topic). I have been wondering about having a 'label-less' dimension, where only positional indexing is allowed.

For MultiIndexes the disambiguation could be made by only allowing tuples for slicing i.e. `adata[1, :]` would be assumed to be positional indexing and `adata[(1,), :]` would be "label-based" MultiIndex of Integers. But I understand the problems. Especially because in pandas `df.loc[1, :]` is also allowed as label-based indexing.

flying-sheep:
Fixed in #1507

Issue #746, opened 2022-03-23
Support for scipy sparse arrays


ivirshup:
Apparently scipy now has [sparse arrays with array indexing](https://docs.scipy.org/doc/scipy/reference/sparse.html)?

Unfortunately they do seem to inherit from the matrix classes, which makes checking their types a bit fraught.

arogozhnikov:
actually that would be awesome if anndata was providing new sparse arrays not matrices. I find it very hard and annoying to memorize how things are working in sparse matrices. As an idea that should be easy to implement and will not break downstream projects:

```python
# setter that internally converts sparse array into a matrix
ad.X_sparr = some_sparse_array 

# getter that converts X into sparse csc_array 
ad.X_sparr
```

Eventually I believe you want to migrate internals of anndata to manipulate arrays not matrices.

ivirshup:
I think we are going to hold off on this until there is better support in the broader ecosystem for these classes. E.g. until scikit-learn can use these, it's difficult to recommend them. At that point we can quickly move to automatically reading these at IO time.

arogozhnikov:
unstale this. Still waiting for bright moment of sparse arrays

Issue #742, opened 2022-03-20
Proposal: Immutable anndata


chaichontat:
Hi! I have been using `scanpy` and `anndata` and start noticing potential difficulties as we analyze more complex datasets.

### Problem
There can be a ballooning number of implicit dependencies in `adata`. For example, running `sc.pl.pca_variance_ratio` requires running `sc.tl.pca` first, the results of which are written directly into `adata`. This may not seem like a big problem but can lead to inconsistencies down the line if the we run the first function with different parameters but forget to run the second again. The number of columns in `obs` and `var` can also grow to the point that we lose track of how each column came to be. Each step of the function could accidentally cause side effects that the user may not be aware of.

### Proposed solution
This issue could be fixed by making `anndata` and its applications more functional. By making `anndata` [immutable](https://static-frame.readthedocs.io/en/latest/articles/upgrade.html#no-3-immutable-data-efficient-memory-management-without-defensive-copies), we could prevent inadvertent modification and cache intermediate results using `lru_cache`. This way, we won't need to modify `adata` within a session. Important or expensive functions would return their results in its own container or dataclass with methods for working with those. Functions would rely less on side-effects, which would lead to more predictable outcomes in the long run. Another [benefit](https://blog.inf.ed.ac.uk/sapm/2014/03/06/enemy-of-the-state-or-why-oop-is-not-suited-to-large-scale-software/) would be the possibility of safely parallelizing pipelines in the future.

At the end of the session, expensive results could be saved along with their associated parameters as a folder in an HDF5 file, an automatic data provenance mechanism. The user could then pick which of the fields they want to save. Mappings from `obs` or `var` to this container could be created to maintain backward compatibility.

Thanks for reading! Let me know if you are interested in this approach.

### Example implementation
```python
# pyright: reportUnknownMemberType=false
#%%
from dataclasses import dataclass, make_dataclass
from functools import lru_cache, wraps
from typing import (
    Any,
    Callable,
    Collection,
    Mapping,
    ParamSpec,
    Protocol,
    TypedDict,
    TypeVar,
    cast,
)

import matplotlib.pyplot as plt
import scanpy as sc
import static_frame as sf
from matplotlib.collections import PathCollection
from scipy.sparse._csr import csr_matrix
from sklearn.decomposition import TruncatedSVD

adata = sc.read_10x_mtx("hg19", var_names="gene_symbols", cache=True)

# Make read-only.
[...]

PR #1270, opened 2023-12-18
(feat): add methods for changing settings

ilan-gold:
- [x] Towards #890 - solving the issue will simply require calling `settings.register` for the needed variable
- [x] Tests added

### DIFF
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import textwrap
+import warnings
 from collections.abc import Iterable
 from contextlib import contextmanager
 from inspect import Parameter, signature
@@ -230,9 +231,11 @@ class SettingsManager:
         """
         if option in self._deprecated_options:
             deprecated = self._deprecated_options[option]
-            raise DeprecationWarning(
-                f"{repr(option)} will be removed in {deprecated.removal_version}. "
-                + deprecated.message
+            warnings.warn(
+                DeprecationWarning(
+                    f"{repr(option)} will be removed in {deprecated.removal_version}. "
+                    + deprecated.message
+                )
             )
         if option in self._config:
             return self._config[option]
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -38,8 +38,7 @@ def update_override_function(register: Callable):
 
     Returns
     -------
-    Callable
-        The wrapped `SettingsManager.register` function.
+    The wrapped `SettingsManager.register` function.
     """
 
     @wraps(register)
@@ -121,8 +120,7 @@ class SettingsManager:
 
         Returns
         -------
-        str
-            The description.
+        The description.
         """
         if option is None:
             return self.describe(
@@ -228,8 +226,7 @@ class SettingsManager:
 
         Returns
         -------
-        object
-            Value of the option.
+        Value of the option.
         """
         if option in self._deprecated_options:
             deprecated = self._deprecated_options[option]
@@ -319,8 +316,7 @@ def check_and_get_environ_var(
 
     Returns
     -------
-    object
-        The casted value.
+    The casted value.
     """
     environ_val = os.environ.get(key, default_value)
     if allowed_values is not None and environ_val not in allowed_values:
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -26,23 +26,21 @@ class RegisteredOption(NamedTuple):
     validator: Callable[[T], None] | None
 
 
+_docstring = """
+This manager allows users to customize settings for the anndata package.
+
+Parameters
+----------
+
+{options_description}
+"""
+
+
 class SettingsManager:
-    """
-    This manager allows users to customize settings for the anndata package.
-
-    Available options:
-
-    {available_options}
-
-    Options descriptions:
-
-    {options_description}
-
-    """
-
     _registered_options: dict[str, RegisteredOption] = {}
     _deprecated_options: dict[str, DeprecatedOption] = {}
     _config: dict[str, object] = {}
+    __doc_tmpl__: str = _docstring
 
     def describe(
         self,
@@ -65,11 +63,8 @@ class SettingsManager:
             The description.
         """
         if option is None:
-            return "\n".join(
-                [
-                    self.describe(k, print_description=print_description)
-                    for k in self._registered_options
-                ]
+            return self.describe(
+                self._registered_options.keys(), print_description=print_description
             )
         if isinstance(option, Iterable) and not isinstance(option, str):
             return "\n".join(
@@ -181,7 +176,7 @@ class SettingsManager:
             )
         if option in self._config:
             return self._config[option]
-        return super().__getattr__(option)
+        raise AttributeError(f"{option} not found.")
 
     def __dir__(self) -> Iterable[str]:
         return sorted(super().__dir__() + list(self._config.keys()))
@@ -219,6 +214,13 @@ class SettingsManager:
             for attr, value in restore.items():
                 setattr(self, attr, value)
 
+    @property
+    def __doc__(self):
+        options_description = self.describe(print_description=False)
+        return self.__doc_tmpl__.format(
+            options_description=options_description,
+        )
+
 
 settings = SettingsManager()
 
@@ -226,16 +228,8 @@ settings = SettingsManager()
 # PLACE REGISTERED SETTINGS HERE SO THEY CAN BE PICKED UP FOR DOCSTRING CREATION #
 ##################################################################################
 
-
 ##################################################################################
 ##################################################################################
-
-
-options_description = settings.describe(print_description=False)
-settings.__doc__ = settings.__doc__.format(
-    options_description=options_description,
-    available_options=str(", ".join(list(settings._registered_options.keys()))),
-)
 
 
 def check_and_get_environ_var(
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -43,7 +43,9 @@ class SettingsManager:
     _deprecated_options: dict[str, DeprecatedOption] = {}
     _config: dict[str, object] = {}
 
-    def describe(self, option: str | None = None, print_description: bool = True) -> str:
+    def describe(
+        self, option: str | None = None, print_description: bool = True
+    ) -> str:
         """Print and/or return a (string) description of the option(s).
 
         Parameters
@@ -74,7 +76,9 @@ class SettingsManager:
             print(doc)
         return doc
 
-    def deprecate(self, option: str, removal_version: str, message: str | None = None) -> None:
+    def deprecate(
+        self, option: str, removal_version: str, message: str | None = None
+    ) -> None:
         """Deprecate options with a message at a version.
 
         Parameters
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -27,7 +27,7 @@ class RegisteredOption(NamedTuple):
 
 class SettingsManager:
     """
-    This manager allows user to customize settings for the anndata package.
+    This manager allows users to customize settings for the anndata package.
 
     Available options:
 
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -23,7 +23,7 @@ class RegisteredOption(NamedTuple):
     option: str
     default_value: T
     doc: str
-    validator: Callable[[T], None] | None
+    validate: Callable[[T], bool] | None
     type: object
 
 
@@ -106,7 +106,7 @@ class SettingsManager:
         option: str,
         default_value: T,
         description: str,
-        validator: Callable[[T], None],
+        validate: Callable[[T], bool],
         option_type: object | None = None,
     ) -> None:
         """Register an option so it can be set/described etc. by end-users
@@ -119,12 +119,16 @@ class SettingsManager:
             Default value with which to set the option.
         description
             Description to be used in the docstring.
-        validator
-            A function which asserts that the option's value is valid.
+        validate
+            A function which returns True if the option's value is valid and otherwise raises an exception.
         option
             Optional override for the option type to be displayed.  Otherwise `type(default_value)`.
         """
-        validator(default_value)
+        try:
+            validate(default_value)
+        except (ValueError, TypeError) as e:
+            e.add_note(f" for option {repr(option)}")
+            raise e
         option_type_str = (
             type(default_value).__name__ if option_type is None else str(option_type)
         )
@@ -135,7 +139,7 @@ class SettingsManager:
         """
         doc = textwrap.dedent(doc)
         self._registered_options[option] = RegisteredOption(
-            option, default_value, doc, validator, option_type
+            option, default_value, doc, validate, option_type
         )
         self._config[option] = default_value
         self._update_override_function_for_new_option(option)
@@ -207,7 +211,7 @@ class SettingsManager:
             )
         else:
             registered_option = self._registered_options[option]
-            registered_option.validator(val)
+            registered_option.validate(val)
             self._config[option] = val
 
     def __getattr__(self, option: str) -> object:
@@ -226,7 +230,7 @@ class SettingsManager:
         if option in self._deprecated_options:
             deprecated = self._deprecated_options[option]
             raise DeprecationWarning(
-                f"{option} will be removed in {deprecated.removal_version}. "
+                f"{repr(option)} will be removed in {deprecated.removal_version}. "
                 + deprecated.message
             )
         if option in self._config:
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+import os
+import warnings
+from typing import TYPE_CHECKING, Any, Callable, NamedTuple
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+# Heavily inspired by pandas' options mechanism, but stripped down.
+# https://github.com/pandas-dev/pandas/blob/86488f700ead42d75dae169032a010c6058e0602/pandas/_config/config.py
+
+config = {}
+
+
+class DeprecatedOption(NamedTuple):
+    key: str
+    msg: str | None
+    removal_ver: str | None
+
+
+class RegisteredOption(NamedTuple):
+    key: str
+    defval: object
+    doc: str
+    validator: Callable[[object], Any] | None
+
+
+registered_options: dict[str, RegisteredOption] = {}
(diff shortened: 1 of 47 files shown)


ilan-gold:
I am aware this was not your top choice @flying-sheep but you also didn't seem so jazzed about scanpy's solution either.  So I finished this up. If we want something different, we can go for it.  I think the best option might be something shared between scanpy and anndata.  What do you think?  Short of that, despite the unpythonic nature of this, it is quite simple IMO which is a big plus.

ilan-gold:
@flying-sheep I'm fine with imitating scanpy.  You sounded fairly negative about it, so I guess it didn't seem like it mattered if we used one "bad" solution or another.  Do you think this should be built with scanpy in mind i.e., migrating scanpy to whatever we do here (in some capacity)?

flying-sheep (changes requested):
> I'm fine with imitating scanpy. You sounded fairly negative about it, so I guess it didn't seem like it mattered if we used one "bad" solution or another.

Scanpy’s solution is … fine. The general idea is great: a perfectly Pythonic API; leniently allow setting `settings.verbosity = 'info'` and then the user can read out `setting.verbosity is Verbosity.info`.

It was made in a time when module-wide `__getattr__`/`__setattr__` wasn’t a thing, that’s why we went with properties and a singleton class. It doesn’t really matter how to implement it, as long as the API and documentation are nice and clean. For a highly user-facing API like this, complexity is very much secondary, so let’s go with the solution that has the least tradeoffs in documentation and API.

I was unenthused about the `override` mechanism (only for `verbosity`) and the `reset` mechanism (absent). Any solution with `get_` and `set_` can only ever feel out of place in a Python project.

> Do you think this should be built with scanpy in mind i.e., migrating scanpy to whatever we do here (in some capacity)?

Yeah, I think making this reusable would be nice if it becomes sufficiently complex.

docs/release-notes/0.10.4.md
```diff
@@ -14,3 +14,7 @@
 
 ```{rubric} Performance
 ```
+
+```{rubric} Other updates
+```
+* Add `describe_option`, `get_option`, `reset_option`, and `set_option` methods for altering internally-used options, like hcecking for uniqueness on `obs`' index {pr}`1270` {user}`ilan-gold`
```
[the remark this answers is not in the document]
ilan-gold: Got it.

anndata/_config.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Parameters
+        ----------
+        option
+            The option to be reset.
+        """
+        self._config[option] = self._registered_options[option].default_value
+
+    @contextmanager
+    def override(self, **overrides):
```
flying-sheep, line 173: This doesn’t support autocompletion, how can we fix that?

anndata/_config.py
```diff
@@ -0,0 +1,240 @@
...
+    Parameters
+    ----------
+    key : str
+        The environment variable name.
+    default_value : str
+        The default value for `os.environ.get`.
+    allowed_values : Sequence[str] | None, optional
+        Allowable string values., by default None
+    cast : _type_, optional
+        Casting from the string to a (potentially different) python object, by default lambdax:x
```
flying-sheep, line 226:
```suggestion
    key
        The environment variable name.
    default_value
        The default value for `os.environ.get`.
    allowed_values
        Allowable string values., by default None
    cast
        Casting from the string to a (potentially different) python object, by default lambdax:x
```

anndata/tests/test_config.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_deprecation():
+    warning = "This is a deprecation warning!"
+    version = "0.1.0"
+    settings.deprecate(test_option, version, warning)
+    described_option = settings.describe(print_description=False).split("\n")
+    assert (
+        len(described_option) == 3
+    )  # first line is message, second two from deprecation
```
flying-sheep, line 79:
```suggestion
    # first line is message, second two from deprecation
    assert len(described_option) == 3
```

anndata/tests/test_config.py
```diff
@@ -0,0 +1,91 @@
...
+    assert described_option[2] == f"{test_option} will be removed in {version}"
+
+
+def test_deprecation_no_message():
+    version = "0.1.0"
+    settings.deprecate(test_option, version)
+    described_option = settings.describe(print_description=False).split("\n")
+    assert (
+        len(described_option) == 2
+    )  # first line is message, second from deprecation version
```
flying-sheep, line 90:
```suggestion
    # first line is message, second from deprecation version
    assert len(described_option) == 2
```

anndata/tests/test_config.py
```diff
@@ -0,0 +1,91 @@
...
+        settings[test_option + "_different"] = default_val
+
+
+def test_override():
+    with settings.override(**{test_option: not default_val}):
+        assert settings[test_option] == (not default_val)
+    assert settings[test_option] == default_val
+
+
+def test_deprecation():
```
flying-sheep, line 72: Should accessing that option throw a DeprecationWarning or FutureWarning?

If so, you can maybe use https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.deprecated_call

flying-sheep (changes requested):
This looks much nicer! I think options should definitely have autocompletion, so please check if that works in Jupyter Lab.

I think we can keep the nice `register` API as long as runtime autocompletion is great.

anndata/_config.py
```diff
@@ -0,0 +1,240 @@
...
+class DeprecatedOption(NamedTuple):
+    option: str
+    message: str | None
+    removal_version: str | None
+
+
+# TODO: inherit from Generic[T] as well after python 3.9 is no longer supported
+class RegisteredOption(NamedTuple):
+    option: str
+    default_value: object
```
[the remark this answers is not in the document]
ilan-gold: We can't do `NamedTuple` and `Generic[T]` in python 3.9.  So if we did only a `Generic`, we'd have to implement an `__init__` method.  Not the worst thing in the world. But you're right about `default_val`

anndata/_config.py
ilan-gold: Ok so I have spent quite a long time on this to no avail.  I have tried many different things.  I will need to look at this again on Monday.  If this is a breaking point, we may have to go the scanpy route (i.e., explicit attributes/keyword/positional arguments).  Something I realized while working through this, though, is that this solution _is generalizable_ in that we could migrate scanpy towards this.  But with the positional/keyword args way of doing things (i.e., what scanpy does), I don't think we can really support inheritance/re-usability.  But maybe I am missing something - I think there are ways of passing documentation up an inheritance tree, but I am not sure about types.

ilan-gold:
Wow ok @flying-sheep, I finally did it.  `__signature__` was the key - `__annotations__` did not really help with the tab completion but `__signature__` did (and you can see that even that was not so straightforward).  If you want to test out tab completion here is something:

```python
import anndata as ad

test_option_doc = "doc string!"
test_option_env_var = "ANNDATA_TEST_VAR"
test_option = "test_var"
default_val = False
test_doc = f"""\
{test_option}: bool
    My doc string!
"""

test_option_doc_2 = "doc string 2!"
test_option_env_var_2 = "ANNDATA_TEST_VAR 2"
test_option_2 = "test_var_2"
default_val_2 = False
test_doc_2 = f"""\
{test_option_2}: bool
    My doc string 2!
"""


def validate_bool(val, option):
    assert val in [True, False], f"{val} not valid boolean for option {option}."


ad.settings.register(
    test_option, default_val, test_doc, lambda v: validate_bool(v, test_option)
)

ad.settings.register(
    test_option_2, default_val_2, test_doc_2, lambda v: validate_bool(v, test_option_2)
)
```

Have a look!  Now that I'm writing this, I would like to get the docstring working as well.  Will do.

ilan-gold:
From meeting: make some sort of `freeze` method so that downstream users can't edit the underlying object.

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+
+    Parameters
+    ----------
+    register
+        This is a reference to the register function of `SettingsManager`
+
+    Returns
+    -------
+    Callable
+        The wrapped `SettingsManager.register` function.
```
flying-sheep, line 42: Please put types into the annotations, not into docstrings.

```suggestion
    The wrapped `SettingsManager.register` function.
```

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+
+    @wraps(register)
+    def _impl(
```
flying-sheep, line 46: Why is `update_override_function` a decorator? Wouldn’t it be simpler if it was a function that’s called in `register`?

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+    ):
+        method_output = register(self, option, default_value, doc, validator)
+        # Update annotations for type checking.
+        self.override.__annotations__[option] = type(default_value)
+        # __signature__ needs to be updated for tab autocompletion in IPython.
+        # See https://github.com/ipython/ipython/issues/11624 for inspiration.
+        setattr(
+            self.override.__func__,  # Attributes of a method are immutable, so need to use __func__.
+            "__signature__",
+            signature(self.override).replace(
```
flying-sheep, line 61: `setattr` is when the `name` parameter isn’t fixed

```suggestion
        # Attributes of a method are immutable, so need to use __func__.
        self.override.__func__.__signature__= signature(self.override).replace(
```

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+                            annotation=type(default_value),
+                            kind=Parameter.KEYWORD_ONLY,
+                        )
+                        for k in self._registered_options
+                    ],
+                ]
+            ),
+        )
+        # Update docstring for `SettingsManager.override` as well.
+        insert_index = self.override.__doc__.find("\n        Yields")
```
flying-sheep, line 76: how does that behave if you register more than one option?

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+
+        Raises
+        ------
+        KeyError
+            If the option has not been registered, this function will raise an error.
+        """
+        if hasattr(super(), option):
+            super().__setattr__(option, val)
+        elif option not in self._registered_options:
+            raise KeyError(
```
flying-sheep, line 211: shouldn’t that be `AttributeError`?

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+                f"{option} is not an available option for anndata.\
+                Please open an issue if you believe this is a mistake."
+            )
+        else:
+            registered_option = self._registered_options[option]
+            registered_option.validator(val)
+            self._config[option] = val
+
+    def __getattr__(self, option: str) -> object:
```
flying-sheep, line 220: I guess that’s the only point where missing static typing is painful.

Using `settings.foo` anywhere in the codebase will only be `object`

anndata/_config.py
flying-sheep: you managed to do that now with `update_override_function`, no?

anndata/_config.py
```diff
@@ -0,0 +1,332 @@
...
+
+##################################################################################
+# PLACE REGISTERED SETTINGS HERE SO THEY CAN BE PICKED UP FOR DOCSTRING CREATION #
+##################################################################################
+
+##################################################################################
+##################################################################################
+
+
+def check_and_get_environ_var(
```
flying-sheep, line 301: what’s that for? I don’t see it being used in the other code

anndata/_config.py
ilan-gold: Missed that one and a few others when cleaning up.  Will fix them all.

flying-sheep (commented):
looks mostly good!

I’m still not sure if the registration way (and it’s lack of static type information) is preferable to simple properties combined with a unit test that checks that `override` and `reset` match up.

I’ll try it out a bit in an interactive session and see how I feel about it.

anndata/_config.py
ilan-gold: Huh, thought I had read this doesn't work because of `__signature__` needing to be initialized specially, but seems to work now.  Great!

anndata/_config.py
ilan-gold: What do you mean?  This looks for the spot right before "Yields" which is where you'd want to insert a new option (i.e., end of the list after "Parameters").  It should work, I checked it on the build with some test options.

anndata/_config.py
ilan-gold: This was in advance of using environment variables (i.e., for retaining categories).  I can just move it to the next PR.  It's tested and I was using for prototyping, that's why I did it here.

anndata/_config.py
ilan-gold: So yes, I think you're right.  I had not checked in VSCode for the attribute completion (only IPython kernels).  VSCode's type checker is static so this is not going to happen.  This seems like a deal breaker if I am not mistaken?

anndata/_config.py
ilan-gold: This works for non-static analyzers, like IPython, but not vscode, which uses static analysis.

anndata/_config.py
flying-sheep: `x.a = v` is exactly the same as `setattr(x, 'a', v)`.

Both go through `__setattr__` if available and so on.

anndata/_config.py
flying-sheep: This code gets called every time a setting is registered.

And every time, `options` gets inserted after `Yields`. Won’t that insert all options multiple times?

anndata/_config.py
flying-sheep: > This seems like a deal breaker if I am not mistaken?

depends on if we expect the settings mechanism to be used from library code or almost exclusively from interactive code.

ilan-gold:
List of features that might be enabled by this branch:

- https://github.com/scverse/anndata/issues/610
- https://github.com/scverse/anndata/issues/890
- https://github.com/scverse/anndata/issues/1112

I think these are all pretty advanced use-cases.

Other things I could see us moving into some sort of settings:
- If/when we transition to `xarray`, assuming we stick with the same `AnnData` object, a variable for using the "experimental" `xarray` implementation (instead of releasing via `experimental`)
- io engines (rust or tensorstore or the like for zarr?)
- mutability behavior (see https://github.com/scverse/anndata/issues/742)

Honestly, though, these first two potential use-cases should probably not be set via an environment variable and is far as the third should, it would be under the "only if you know what you're doing" banner.  But not sure it's an env variable situation anyway.

At the end of the day, most of pandas and xarray's settings are about display and/or dependencies and we just don't do either too much at the moment (and in so far as we do external dependencies, this has been handled differently).  Furthermore, display only really matters in IPython settings, where autocomplete is present as things stand.  So I guess I really only see this as being "do this if you know what you're doing" as far as VSCode audience goes.  I'd be fine leaving things as is especially given the numerous advantages of the current approach (it's extendable, has deprecation warnings, and is very clean).

ivirshup:
> List of features that might be enabled by this branch:

I would also include the path towards:

* https://github.com/scverse/anndata/issues/746

For controlling if we auto convert to these, which we should do eventually.

> pandas and xarray's settings are about display and/or dependencies and we just don't do either too much at the moment 

Pandas does do feature flags for upcoming behaviour with settings flags. For example, [copy on write](https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#copy-on-write).

ilan-gold:
I would still describe these as "advanced use-cases."  We don't have arguments like `N_PCS` from scanpy or the like which are important and could be oft used.

flying-sheep:
In that case, I fully agree with

> So I guess I really only see this as being "do this if you know what you're doing" as far as VSCode audience goes.

and like the current design. Maybe we should document that it’s only for advanced use cases.

anndata/_config.py
ilan-gold: The next line only adds the new option.  Will relabel the variable.

ilan-gold: Screenshot 2024-01-25 at 11 02 03

anndata/_config.py
ilan-gold: Removed.

anndata/_config.py
ilan-gold: From previous iterations.  Refactored.

ilan-gold:
@flying-sheep @ivirshup I addressed your comments and cleaned up the docs a bit.  Please see https://github.com/scverse/anndata/pull/1340 for the first "implementation" if it will help with the review here.  I am happy with the state of things and am ready to merge.

anndata/_config.py
```diff
@@ -0,0 +1,290 @@
...
+                        kind=Parameter.KEYWORD_ONLY,
+                    )
+                    for k in self._registered_options
+                ],
+            ]
+        )
+        # Update docstring for `SettingsManager.override` as well.
+        insert_index = self.override.__doc__.find("\n        Yields")
+        option_docstring = "\t" + "\t".join(
+            self.describe(option, print_description=False).splitlines(True)
```
flying-sheep, line 175:
```suggestion
            self.describe(option, print_description=False).splitlines(keepends=True)
```

anndata/_config.py
```diff
@@ -0,0 +1,290 @@
...
+            raise DeprecationWarning(
+                f"{option} will be removed in {deprecated.removal_version}. "
+                + deprecated.message
+            )
+        if option in self._config:
+            return self._config[option]
+        raise AttributeError(f"{option} not found.")
+
+    def __dir__(self) -> Iterable[str]:
+        return sorted(super().__dir__() + list(self._config.keys()))
```
flying-sheep, line 237:
```suggestion
        return sorted((*dir(super()), *self._config.keys()))
```

anndata/_config.py
```diff
@@ -0,0 +1,290 @@
...
+        default_value
+            Default value with which to set the option.
+        description
+            Description to be used in the docstring.
+        validator
+            A function which asserts that the option's value is valid.
+        option
+            Optional override for the option type to be displayed.  Otherwise `type(default_value)`.
+        """
+        validator(default_value)
```
flying-sheep, line 127: How about you just do something like this?

```suggestion
        try:
            validator(default_value)
        except (ValueError, TypeError) as e:
            e.add_note(f"in option {option}")
```

Then you can just do `validator=bool` instead of having to craft the message in each validator function

PS: we have some helper to add to `__notes__` in Python 3.9 I think

flying-sheep (approved):
Looks great. I can think of one simplification to the API, otherwise, this is done as far as I’m concerned.

PR #1273, opened 2023-12-20
Fix IO error reporting

### LINKED ISSUE #1272: Write error reporting got worse

### Report

There are two things working together

1. the reported paths seem wrong (always `/`), which seems to be the case since long ago
2. introduced in https://github.com/scverse/anndata/pull/1055, the following `if` condition makes it so only one key is written (so you only know the innermost one, not all of them)

https://github.com/scverse/anndata/blob/3114f6eb5986beaca096497a5da3b9d438d4c06d/anndata/_io/utils.py#L169-L175

i.e. when writing `AnnData(obs=dict(b=broken))`, instead of getting

> Error raised while writing key 'b' of `<class 'h5py._hl.group.Group'>` to **/obs**

or 

> Error raised while writing key 'b' of `<class 'h5py._hl.group.Group'>` to /
> Error raised while writing key 'obs' of `<class 'h5py._hl.group.Group'>` to /

we only get the following, without any helpful information except for the leaf key:

> Error raised while writing key 'b' of `<class 'h5py._hl.group.Group'>` to /

### Versions

AnnData master. [`session_info` is broken](https://gitlab.com/joelostblom/session_info/-/issues/12), so I can’t use it:

```pytb
Cell In[5], line 1
----> 1 import anndata, session_info; session_info.show(html=False, dependencies=True)

File ~/.local/share/hatch/env/virtual/anndata/PUvSwSLS/test.py3.11/lib/python3.11/site-packages/session_info/main.py:169, in show(na, os, cpu, jupyter, dependencies, std_lib, private, write_req_file, req_file_name, html, excludes)
    165             pass
    166     # The length of `'jupyter_client'` is 14
    167     # The spaces are added to create uniform whitespace in the output
    168     # f-strings, which is needed to clean them with inspect.cleandoc
--> 169     jup_mod_and_ver = [f'            {module.__name__:14}\t{module.__version__}'
    170                        for module in jup_modules]
    171     output_jupyter = '\n'.join(jup_mod_and_ver)
    172 else:

File ~/.local/share/hatch/env/virtual/anndata/PUvSwSLS/test.py3.11/lib/python3.11/site-packages/session_info/main.py:169, in <listcomp>(.0)
    165             pass
    166     # The length of `'jupyter_client'` is 14
    167     # The spaces are added to create uniform whitespace in the output
    168     # f-strings, which is needed to clean them with inspect.cleandoc
--> 169     jup_mod_and_ver = [f'            {module.__name__:14}\t{module.__version__}'
    170                        for module in jup_modules]
    171     output_jupyter = '\n'.join(jup_mod_and_ver)
    172 else:

AttributeError: module 'jupyter_core' has no attribute '__version__'
```

flying-sheep:
- [x] Closes #1272
- [x] Tests added

This PR makes it so group path reporting in IO errors actually works.

It also fixes `test_only_child_key_reported_on_failure`, which did nothing before:
`re.search` searches anywhere in the string, so every string can be successfully searched for `(?s)((?!Error raised while writing key '/?a').)*$`. Only with the `^` at the start will that regex successfully fail when the substring is contained.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,14 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 23.11.0
-    hooks:
-      - id: black
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version.
-    rev: "v0.1.7"
+    rev: v0.1.9
     hooks:
       - id: ruff
+        types_or: [python, pyi, jupyter]
         args: ["--fix"]
+      - id: ruff-format
+        types_or: [python, pyi, jupyter]
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v4.0.0-alpha.4
+    rev: v4.0.0-alpha.8
     hooks:
       - id: prettier
         exclude_types:
@@ -28,7 +26,6 @@ repos:
       - id: detect-private-key
       - id: no-commit-to-branch
         args: ["--branch=main"]
-
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,8 @@ repos:
     rev: v4.0.0-alpha.4
     hooks:
       - id: prettier
+        exclude_types:
+          - markdown
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
     hooks:
--- a/anndata/_io/utils.py
+++ b/anndata/_io/utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from functools import wraps
-from itertools import pairwise
 from typing import TYPE_CHECKING, Callable, Literal, Union, cast
 from warnings import warn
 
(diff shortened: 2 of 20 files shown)


.pre-commit-config.yaml
```diff
@@ -13,6 +13,8 @@ repos:
     rev: v4.0.0-alpha.4
     hooks:
       - id: prettier
+        exclude_types:
+          - markdown
```
Zethson, line 17: Why are we ignoring markdown now?

flying-sheep: because we use MyST, which is a bespoke Markdown dialect instead of standard markdown, and therefore not supported by Prettier

Zethson (approved):

anndata/_io/utils.py
```diff
@@ -151,31 +153,26 @@ class AnnDataReadError(OSError):
...
-    elif isinstance(elem, BaseCompressedSparseDataset):
-        parent = elem.group.file.name
-    else:
-        parent = elem.file.name
-    return parent
+def _get_path(elem: Elem) -> str:
+    """Return an absolute path of an element (always starts with “/”)."""
+    if isinstance(elem, BaseCompressedSparseDataset):
+        elem = elem.group
+    path = elem.name or "??"  # can be None
```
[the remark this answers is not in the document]
flying-sheep: This function is called for display purposes, so I’m handling the `None` here instead of passing it on and handling it on the call site.

I’m also not entirely sure that this case can happen at all, but I added this defensive path in case it does

PR #1302, opened 2024-01-12
Fix uns merge 3d

### LINKED ISSUE #1300: Checking array equality during merging doesn't work for >2d

### Report

Initially found by @merelkuijs

The issue happens when arrays in `uns` need to be compared, as happens with `unique`. We should use a more generic test of equality. I can't remember why we are using dataframe equality here, I suspect either due to a nice error message or pandas being faster.

Code:

```python
import anndata as ad, numpy as np

a = ad.AnnData(
    X=np.ones((1, 1)),
    uns={"x": np.ones((1, 1, 1))}
)

ad.concat([a, a], uns_merge="unique")
```

Traceback:

```pytb
Cell In[1], line 8
      1 import anndata as ad, numpy as np
      3 a = ad.AnnData(
      4     X=np.ones((1, 1)),
      5     uns={"x": np.ones((1, 1, 1))}
      6 )
----> 8 ad.concat([a, a], uns_merge="unique")

File ~/github/anndata/anndata/_core/merge.py:1327, in concat(adatas, axis, join, merge, uns_merge, label, keys, index_unique, fill_value, pairwise)
   1315 alt_mapping = merge(
   1316     [
   1317         {k: r(v, axis=0) for k, v in getattr(a, f"{alt_dim}m").items()}
   1318         for r, a in zip(reindexers, adatas)
   1319     ],
   1320 )
   1321 alt_pairwise = merge(
   1322     [
   1323         {k: r(r(v, axis=0), axis=1) for k, v in getattr(a, f"{alt_dim}p").items()}
   1324         for r, a in zip(reindexers, adatas)
   1325     ]
   1326 )
-> 1327 uns = uns_merge([a.uns for a in adatas])
   1329 raw = None
   1330 has_raw = [a.raw is not None for a in adatas]

File ~/github/anndata/anndata/_core/merge.py:438, in merge_unique(ds)
    437 def merge_unique(ds: Collection[Mapping]) -> Mapping:
--> 438     return merge_nested(ds, union_keys, unique_value)

File ~/github/anndata/anndata/_core/merge.py:415, in merge_nested(ds, keys_join, value_join)
    413 out = {}
    414 for k in keys_join(ds):
--> 415     v = _merge_nested(ds, k, keys_join, value_join)
    416     if not_missing(v):
    417         out[k] = v

File ~/github/anndata/anndata/_core/merge.py:434, in _merge_nested(ds, k, keys_join, value_join)
    432         return new_map
    433 else:
--> 434     return value_join(vals)

File ~/github/anndata/anndata/_core/merge.py:383, in unique_value(vals)
    381 unique_val = vals[0]
    382 for v in vals[1:]:
--> 383     if not equal(v, unique_val):
    384         return MissingVal
    385 return unique_val

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

File ~/github/anndata/anndata/_core/merge.py:144, in equal_array(a, b)
    142 @equal.register(np.ndarray)
    143 def equal_array(a, b) -> bool:
--> 144     return equal(pd.DataFrame(a), pd.DataFrame(asarray(b)))

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/pandas/core/frame.py:785, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    774         mgr = dict_to_mgr(
    775             # error: Item "ndarray" of "Union[ndarray, Series, Index]" has no
    776             # attribute "name"
   (...)
    782             copy=_copy,
    783         )
    784     else:
--> 785         mgr = ndarray_to_mgr(
    786             data,
    787             index,
    788             columns,
    789             dtype=dtype,
    790             copy=copy,
    791             typ=manager,
    792         )
    794 # For data is list-like, or Iterable (will consume into list)
    795 elif is_list_like(data):

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/pandas/core/internals/construction.py:314, in ndarray_to_mgr(values, index, columns, dtype, copy, typ)
    308     _copy = (
    309         copy_on_sanitize
    310         if (dtype is None or astype_is_view(values.dtype, dtype))
    311         else False
    312     )
    313     values = np.array(values, copy=_copy)
--> 314     values = _ensure_2d(values)
    316 else:
    317     # by definition an array here
    318     # the dtypes will be coerced to a single dtype
    319     values = _prep_ndarraylike(values, copy=copy_on_sanitize)

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/pandas/core/internals/construction.py:592, in _ensure_2d(values)
    590     values = values.reshape((values.shape[0], 1))
    591 elif values.ndim != 2:
--> 592     raise ValueError(f"Must pass 2-d input. shape={values.shape}")
    593 return values

ValueError: Must pass 2-d input. shape=(1, 1, 1)
```

### Versions

```python
-----
IPython             8.17.2
anndata             0.11.0.dev44+g06dfbde9.d20240112
numpy               1.23.4
session_info        1.0.0
-----
asciitree           NA
asttokens           NA
attr                23.1.0
awkward             2.5.1
awkward_cpp         NA
cloudpickle         3.0.0
cython_runtime      NA
dask                2023.10.0
dateutil            2.8.2
decorator           5.1.1
executing           2.0.1
fasteners           0.19
h5py                3.10.0
importlib_metadata  NA
jedi                0.19.1
jinja2              3.1.2
markupsafe          2.1.3
msgpack             1.0.7
natsort             8.4.0
numcodecs           0.12.1
packaging           23.2
pandas              2.1.1
parso               0.8.3
pexpect             4.8.0
prompt_toolkit      3.0.41
psutil              5.9.6
ptyprocess          0.7.0
pure_eval           0.2.2
pyarrow             14.0.1
pygments            2.16.1
pytz                2023.3.post1
rich                NA
scipy               1.11.4
setuptools          68.2.2
setuptools_scm      NA
six                 1.16.0
sphinxcontrib       NA
stack_data          0.6.3
tblib               3.0.0
tlz                 0.12.0
toolz               0.12.0
traitlets           5.13.0
wcwidth             0.2.10
yaml                6.0.1
zarr                2.16.1
zipp                NA
-----
Python 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:37) [Clang 15.0.7 ]
macOS-13.6.1-x86_64-i386-64bit
-----
Session information updated at 2024-01-12 14:22
```

ivirshup:
- [x] Closes #1300
- [x] Tests added
  - Added tests for numeric and record 3d arrays

This was a little harder than expected due to our need to handle non-numeric dtypes. We were using dataframe comparisons to handle that.

Now we are using numpy when able, but switching to dataframes constructed from reshaped arrays when we can't.

### DIFF
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -155,7 +155,7 @@ def equal_array(a, b) -> bool:
     else:
         # For non numeric types use pandas to compare
         # Reshapeing allows us to compare inputs with >2 dimensions
-        return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(asarray(b).reshape(-1)))
+        return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
 
 
 @equal.register(CupyArray)
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -141,21 +141,13 @@ def equal_dask_array(a, b) -> bool:
 
 @equal.register(np.ndarray)
 def equal_array(a, b) -> bool:
-    def _is_numeric_or_bool(x) -> bool:
-        """array_equal with equal_nan=True only works for these types."""
-        return np.issubdtype(x, np.number) or np.issubdtype(x, np.bool_)
-
+    # Reshaping allows us to compare inputs with >2 dimensions
+    # We cast to pandas since it will still work with non-numeric types
     b = asarray(b)
-
     if a.shape != b.shape:
         return False
 
-    if _is_numeric_or_bool(a.dtype) and _is_numeric_or_bool(b.dtype):
-        return np.array_equal(a, b, equal_nan=True)
-    else:
-        # For non numeric types use pandas to compare
-        # Reshapeing allows us to compare inputs with >2 dimensions
-        return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
+    return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
 
 
 @equal.register(CupyArray)
--- a/anndata/_core/merge.py
+++ b/anndata/_core/merge.py
@@ -141,7 +141,21 @@ def equal_dask_array(a, b) -> bool:
 
 @equal.register(np.ndarray)
 def equal_array(a, b) -> bool:
-    return equal(pd.DataFrame(a), pd.DataFrame(asarray(b)))
+    def _is_numeric_or_bool(x) -> bool:
+        """array_equal with equal_nan=True only works for these types."""
+        return np.issubdtype(x, np.number) or np.issubdtype(x, np.bool_)
+
+    b = asarray(b)
+
+    if a.shape != b.shape:
+        return False
+
+    if _is_numeric_or_bool(a.dtype) and _is_numeric_or_bool(b.dtype):
+        return np.array_equal(a, b, equal_nan=True)
+    else:
+        # For non numeric types use pandas to compare
+        # Reshapeing allows us to compare inputs with >2 dimensions
+        return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(asarray(b).reshape(-1)))
 
 
 @equal.register(CupyArray)
--- a/anndata/tests/test_concatenate.py
+++ b/anndata/tests/test_concatenate.py
@@ -28,6 +28,7 @@ from anndata.tests.helpers import (
     as_dense_dask_array,
     assert_equal,
     gen_adata,
+    gen_vstr_recarray,
 )
 from anndata.utils import asarray
 
@@ -1018,6 +1019,15 @@ def gen_something(n):
     return np.random.choice(options)(n)
 
 
+def gen_3d_numeric_array(n):
+    return np.random.randn(n, n, n)
+
+
+def gen_3d_recarray(_):
+    # Ignoring n as it can get quite slow
+    return gen_vstr_recarray(8, 3).reshape(2, 2, 2)
+
+
 def gen_concat_params(unss, compat2result):
     value_generators = [
         lambda x: x,
@@ -1026,6 +1036,8 @@ def gen_concat_params(unss, compat2result):
         gen_list,
         gen_sparse,
         gen_something,
+        gen_3d_numeric_array,
+        gen_3d_recarray,
     ]
     for gen, (mode, result) in product(value_generators, compat2result.items()):
         yield pytest.param(unss, mode, result, gen)
@@ -1083,16 +1095,6 @@ def gen_concat_params(unss, compat2result):
         gen_concat_params(
             [{"a": 1} for i in range(10)] + [{"a": 2}],
             {None: {}, "first": {"a": 1}, "unique": {}, "same": {}, "only": {}},
-        ),
-        gen_concat_params(  # https://github.com/scverse/anndata/issues/1300
-            [{"a": np.ones((3, 3, 3))} for _ in range(2)],
-            {
-                None: {},
-                "first": {"a": np.ones((3, 3, 3))},
-                "unique": {"a": np.ones((3, 3, 3))},
-                "same": {"a": np.ones((3, 3, 3))},
-                "only": {},
-            },
         ),
     ),
 )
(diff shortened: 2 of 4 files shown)


anndata/tests/test_concatenate.py
```diff
@@ -1018,6 +1019,15 @@ def gen_something(n):
     return np.random.choice(options)(n)
 
 
+def gen_3d_numeric_array(n):
+    return np.random.randn(n, n, n)
+
+
+def gen_3d_recarray(_):
+    # Ignoring n as it can get quite slow
```
ilan-gold, line 1027: What does this mean exactly?

anndata/_core/merge.py
```diff
@@ -141,7 +141,21 @@ def equal_dask_array(a, b) -> bool:
...
+
+    if a.shape != b.shape:
+        return False
+
+    if _is_numeric_or_bool(a.dtype) and _is_numeric_or_bool(b.dtype):
+        return np.array_equal(a, b, equal_nan=True)
+    else:
+        # For non numeric types use pandas to compare
+        # Reshapeing allows us to compare inputs with >2 dimensions
+        return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
```
ilan-gold, line 158:
```suggestion
    if _is_numeric_or_bool(a.dtype) and _is_numeric_or_bool(b.dtype):
        return np.array_equal(a, b, equal_nan=True)
    # For non numeric types use pandas to compare
    # Reshapeing allows us to compare inputs with >2 dimensions
    return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
```

`else` is unnecessary here.

anndata/_core/merge.py
```diff
@@ -141,7 +141,21 @@ def equal_dask_array(a, b) -> bool:
...
+        """array_equal with equal_nan=True only works for these types."""
+        return np.issubdtype(x, np.number) or np.issubdtype(x, np.bool_)
+
+    b = asarray(b)
+
+    if a.shape != b.shape:
+        return False
+
+    if _is_numeric_or_bool(a.dtype) and _is_numeric_or_bool(b.dtype):
+        return np.array_equal(a, b, equal_nan=True)
```
ilan-gold, line 154: Why introduce `array_equal` at all?  I think `reshape` is generally very quick (returns a view of the same underlying in-memory data etc.), so is the array equality much faster here than pandas?

anndata/tests/test_concatenate.py
ivirshup: Which part?

`recarray` refers to a record array: https://numpy.org/doc/stable/reference/generated/numpy.recarray.html#numpy-recarray

n is a parameter passed to all the other functions here. Generally it changes the size of the result, but we ignore that in a couple cases. Here I ignore it because it made some test cases take over a minute.

anndata/_core/merge.py
ivirshup: I think this is more of a style thing. I think having the returns under their own conditions is a little more clear that the function could have returned higher up.

And broadly if the linter says it's fine, I think it's fine.

anndata/_core/merge.py
ivirshup: So my initial thinking was that sometimes `reshape` and creating the dataframe copies. But I tested a few things at the repl and found that pandas was generally faster here, so I've made the change.

anndata/_core/merge.py
```diff
@@ -141,7 +141,13 @@ def equal_dask_array(a, b) -> bool:
...
 @equal.register(np.ndarray)
 def equal_array(a, b) -> bool:
-    return equal(pd.DataFrame(a), pd.DataFrame(asarray(b)))
+    # Reshaping allows us to compare inputs with >2 dimensions
+    # We cast to pandas since it will still work with non-numeric types
+    b = asarray(b)
+    if a.shape != b.shape:
+        return False
+
+    return equal(pd.DataFrame(a.reshape(-1)), pd.DataFrame(b.reshape(-1)))
```
flying-sheep, line 150: What does `_.reshape(-1)` mean? The numpy docs say

> One shape dimension can be -1. In this case, the value is inferred from the length of the array and remaining dimensions.

So what you do basically creates 1D `pd.DataFrames`? Shouldn’t we make and compare `pd.Series` then?

flying-sheep (approved):

anndata/_core/merge.py
ivirshup: That would make sense, but in practice the Series constructor acts differently and throws errors for some of the non-numeric dtypes.

Issue #1203, opened 2023-10-19
Minimum dependencies test job


ivirshup:
### Please describe your wishes and possible alternatives to achieve the desired result.

We should be testing against minimum versions of our dependencies. This is motivated by several recent issues related to pandas versions.

I propose we add a test job that grabs the minimum bounds of our dependencies and tests against it.

* This should test against our minimum python version
* This should run on main and release branches
* This should include optional dependencies
* We should figure out what the minimum version required is for our dependencies. I would propose:
  * Recently added and optional dependencies can have high lower bounds
  * Core dependencies should try and follow [SPEC 0](https://scientific-python.org/specs/spec-0000/)
* Keeping the job up to date with the `pyproject.toml` must be automated.

Unfortunately, it doesn't look like there's a great way to automate the job:

* https://github.com/pypa/pip/issues/8085

We can probably just write a script to parse this from our `pyproject.toml`.

cc: @flying-sheep

flying-sheep:
> We can probably just write a script to parse this from our pyproject.toml.

Very easily. If we run that job in Python 3.11, we don’t even have to add a test dependency, we can just use the built-in `tomllib` module to parse `pyproject.toml` and grab the list of version specifier strings with it.

Then parse it using [packaging.specifiers.SpecifierSet](https://packaging.pypa.io/en/latest/specifiers.html) and transform it into a list of `==` specs.

Then begins the “fun” of finding out what the real minimum versions are that we support.

ivirshup:
This gets more complicated for cupy since:

* You gotta specify the version of CUDA if you install cupy from pip
* CUDA is installed for the system image, not via pip

So we may need to upgrade out GPU testing system to let us use other versions of CUDA, so that we can support older cupy to test against them.

ivirshup:
Another potential solution is to use https://pypi.org/project/pypi-timemachine/, which would help with transitive dependencies. However, I think this raises some new questions like: "what date do we set it to" and "what if a particular dependency should be from a newer date".

flying-sheep:
Yeah, I feel like a PyPI time machine is more for “I can’t figure out how to get a working environment for this old thing anymore”, not our use case.

PR #1314, opened 2024-01-16
Test against mindeps

ivirshup:
- [x] Closes https://github.com/scverse/anndata/issues/1203
- [x] Tests added

TODO:

- [x] Get all tests running

Won't do in this PR:

* GPU support (it's going to be more difficult with the self-hosted runner + gotta get cuda versions working right
* Fancier minimum version detection

Current problems:

* Warnings are different in earlier versions...
* Pandas has different results for `pd.value_counts` depending on version
* Should this be collecting coverage to check on version bounded things?


ci/scripts/min-deps.py
```diff
@@ -0,0 +1,72 @@
+#!python3
+from __future__ import annotations
+
+import argparse
+from pathlib import Path
+
+import tomli
```
flying-sheep, line 7:
```suggestion
if sys.version_info >= (3, 11):
    import tomllib
else:
	import tomli as tomllib
```

ci/scripts/min-deps.py
```diff
@@ -0,0 +1,72 @@
...
+        usage="pip install `python min-deps.py pyproject.toml`",
+    )
+    parser.add_argument(
+        "path", type=Path, help="pyproject.toml to parse minimum dependencies from"
+    )
+    parser.add_argument("--extras", type=str, nargs="*", help="extras to install")
+
+    args = parser.parse_args()
+
+    pyproject = tomli.loads(args.path.read_text())
```
flying-sheep, line 59:
```suggestion
    pyproject = tomllib.loads(args.path.read_text())
```

ivirshup:
This last doctest is a pain (`concat_on_disk`), because pandas actually changed the output for `value_counts` between v1 and v2.

I'd rather not just skip it, but unsure if there's a good solution here. Ideally we'd just ignore the lines that are different or skip this bit only when pandas is older.

anndata/tests/test_io_warnings.py
```diff
@@ -43,6 +45,13 @@ def test_old_format_warning_not_thrown(tmp_path):
 
     with warnings.catch_warnings(record=True) as record:
         warnings.simplefilter("always", ad.OldFormatWarning)
+        if Version(h5py.__version__) < Version("3.2"):
```
flying-sheep, line 48:
```suggestion
        if Version(version("h5py")) < Version("3.2"):
```

let’s use the standard API (`importlib.metadata:version`) and not a convention

anndata/tests/test_concatenate.py
```diff
@@ -1350,7 +1351,7 @@ def test_concat_size_0_dim(axis, join_type, merge_strategy, shape):
             FutureWarning,
             match=r"The behavior of DataFrame concatenation with empty or all-NA entries is deprecated",
         )
-        if shape[axis] == 0
+        if shape[axis] == 0 and Version(pd.__version__) >= Version("2.1")
```
flying-sheep, line 1354:
```suggestion
        if shape[axis] == 0 and Version(version("pandas")) >= Version("2.1")
```

same as above

flying-sheep:
> This last doctest is a pain (concat_on_disk), because pandas actually changed the output for value_counts between v1 and v2. I'd rather not just skip it, but unsure if there's a good solution here.

What’s the difference in the format? maybe we can modify it in a way that smooths over the differences, e.g. `dict(sorted(x.items()))` if the order is different or something.

anndata/tests/test_io_warnings.py
ivirshup: I don't think we should hit the disk for a version check of an imported package, as a rule.

I would agree that this could be better if this were asking the version of an dynamically selected package, but we are asking for the version of a package we know.

ivirshup:
The name of the series and the name of the index are different. For example:

```python
>>> pd.DataFrame({"a": [1, 1, 2, 3, 4, 4, 4]})["a"].value_counts()
4    3
1    2
2    1
3    1
Name: a, dtype: int64
>>> pd.__version__
'1.4.4'
```

```python
>>> pd.DataFrame({"a": [1, 1, 2, 3, 4, 4, 4]})["a"].value_counts()
a
4    3
1    2
2    1
3    1
Name: count, dtype: int64
>>> pd.__version__
'2.1.2'
```

I don't like the idea of modifying the command just to get it reproducible since it's meant to be an example for the user, and it's a weird thing to select as an example without the context.

If we had a different example to show, that could be useful. We could even just remove this line.

flying-sheep:
> I don't like the idea of modifying the command just to get it reproducible since it's meant to be an example for the user, and it's a weird thing to select as an example without the context.

Sure, only if it makes sense. a trailing `.to_dict()` might!

ivirshup:
I'm converting back to draft while I get this working for scanpy. I think there will be a few changes that it'd be good to get in here too.

anndata/tests/test_io_warnings.py
flying-sheep: As a rule, I ask for numbers before I consider “let’s use the hack instead of the dedicated API”. The numbers say: “Both add <1ms to function execution time, so don’t worry”

```console
$ python -m timeit -s "from importlib.metadata import version; version('pandas')" "version('pandas')"
1000 loops, best of 5: 225 usec per loop
$ python -m timeit -s "import pandas as pd" "pd.__version__"
10000000 loops, best of 5: 22.6 nsec per loop
```

ivirshup: Did you try on the HPC? File access is randomly incredibly slow. E.g. seconds.

ivirshup: Also, did the correctness issues on version numbers reported by importlib ever get fixed?

flying-sheep: > File access is randomly incredibly slow. E.g. seconds.

So Python importing is extremely slow too, which we do at run time in some cases. So this doesn‘t noticably exacerbate the problem.

> Also, did the correctness issues on version numbers reported by importlib ever get fixed?

Is there a bug I’m not aware of?

ivirshup: > Is there a bug I’m not aware of?

There are issues that we've discussed before: https://github.com/scverse/anndata/pull/1001#issuecomment-1593163740

As an example in an environment right now:

```sh
$ python3 -c "import anndata; from importlib.metadata import version; print(anndata.__version__, version('anndata'))"
0.10.4.dev0+g2ee11c4.d20240118 0.10.1
```

See also the thread on the python discourse I linked above.

> We can fix this specific symptom of the underlying problem by defining PANDAS_VERSION

As discussed in person, pandas provides this already as `pd.__version__`. Though this variable does not suffer from the correctness problems of `importlib.metadata.version` with editable or modified installs.

ivirshup: > and we should switch our stuff away from accessing .__version__ anywhere.

How about you open an issue for this and we use the convention we already use multiple places in this library (`.__version__`) here?

flying-sheep: Sure!

PR #1340, opened 2024-01-26
(feat): add setting to retain categories

### LINKED ISSUE #890: Option to retain categories in columns when subsetting anndata object.

ilan-gold:
- [x] Closes #890 
- [x] Tests added

### DIFF
--- a/anndata/_config.py
+++ b/anndata/_config.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
+import os
 import textwrap
+import warnings
 from collections.abc import Iterable
 from contextlib import contextmanager
 from inspect import Parameter, signature
-from typing import TYPE_CHECKING, NamedTuple, TypeVar
+from typing import TYPE_CHECKING, Any, NamedTuple, TypeVar
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Sequence
 
 T = TypeVar("T")
 
@@ -36,6 +38,7 @@ The following options are available:
 {options_description}
 
 For setting an option please use :func:`~anndata.settings.override` (local) or set the above attributes directly (global) i.e., `anndata.settings.my_setting = foo`.
+For assignment by environment variable, use the variable name in all caps with `ANNDATA_` as the prefix before import of :mod:`anndata`.
 """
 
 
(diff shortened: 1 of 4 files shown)


anndata/tests/test_config.py
```diff
@@ -110,3 +116,32 @@ def test_deprecation_no_message():
...
     assert settings._registered_options[option_3].type == type_3
     assert str(type_3) in settings.describe(option_3, print_description=False)
+
+
+def test_check_and_get_environ_var():
+    option_env_var = "ANNDATA_OPTION"
+    assert hash("foo") == check_and_get_environ_var(
+        option_env_var, "foo", ["foo", "bar"], lambda x: hash(x)
+    )
+    os.environ[option_env_var] = "bar"
```
flying-sheep, line 126: does Pytest reset os.environ after each test?

anndata/_config.py
```diff
@@ -288,11 +290,77 @@ def __doc__(self):
...
+        )
+        return cast(default_value)
+    return cast(environ_val)
+
+
+def check_and_get_bool(option, default_value):
+    return check_and_get_environ_var(
+        "ANNDATA_" + option.upper(),
+        str(int(default_value)),
+        ["0", "1"],
```
flying-sheep, line 330: There’s a lot of conventions for this. E.g.

1. In older versions of Python, there was a function for one: https://docs.python.org/3.8/distutils/apiref.html#distutils.util.strtobool
2. https://force-color.org/ defines “true” as “present and not an empty string”

So maybe we should skip `yes`/`no` and `on`/`off`, since `no` and `on` are so close, but support all capitalizations of `True`/`False` and the empty string. What do you think?

anndata/_config.py
```diff
@@ -288,11 +290,77 @@ def __doc__(self):
...
+
+settings.register(
+    categories_option,
+    categories_default_value,
+    categories_description,
+    validate_bool,
+)
+settings.remove_unused_categories = check_and_get_bool(
+    categories_option, categories_default_value
+)
```
flying-sheep, line 363: should this be part of the registration call?

docs/release-notes/0.11.0.md
```diff
@@ -3,6 +3,7 @@
 ```{rubric} Features
 ```
 * Add `settings` object with methods for altering internally-used options, like checking for uniqueness on `obs`' index {pr}`1270` {user}`ilan-gold`
+* Add `remove_unused_categories` option to `anndata.settings` to override current behavior.  Default is `True` (i.e., previous behavior).  Please refer to the documentation for usage.  {pr}`1340` {user}`ilan-gold`
```
ivirshup, line 6: Could you link to the documentation here?

anndata/_config.py
```diff
@@ -288,11 +290,77 @@ def __doc__(self):
         )
 
 
+def check_and_get_environ_var(
+    key: str,
+    default_value: str,
+    allowed_values: Sequence[str] | None = None,
+    cast: Callable[[Any], T] = lambda x: x,
```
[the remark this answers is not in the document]
flying-sheep: I typed `cast[val]` for a reason.

```py
class Test(Enum):
    a = 1
    b = 2

Test["a"] is Test.a
assert Test(1) is Test.a
```

flying-sheep (changes requested):
See https://github.com/scverse/anndata/pull/1340#discussion_r1476195720

anndata/_config.py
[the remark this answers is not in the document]
ilan-gold: Forgot to add second change, apologies.

PR #1343, opened 2024-01-26
(fix): move zarr import inside try-catch in test helpers

### LINKED ISSUE #1342: We broke scanpy’s tests

### Report

The tests were broken in https://github.com/scverse/anndata/pull/1266

```console
pip install anndata  # anndata[testing] doesn‘t exist
```

```python
import anndata.tests.helpers
```

Traceback:

```pytb
ModuleNotFoundError: No module named 'zarr'
```

It’s time to finally fix #646

### Versions

```python
anndata 0.10.5
```

ilan-gold:
Is there some sort of way to test this?  Open to suggestions.  `ruff` does not support custom rules, I think?

- [x] Towards #1342
- [ ] Tests added

### DIFF
--- a/anndata/tests/helpers.py
+++ b/anndata/tests/helpers.py
@@ -781,7 +781,7 @@ try:
 except ImportError:
 
     class AccessTrackingStore:
-        def __init__(self) -> None:
+        def __init__(self, *_args, **_kwargs) -> None:
             raise ImportError(
                 "zarr must be imported to create an `AccessTrackingStore` instance."
             )
--- a/anndata/tests/helpers.py
+++ b/anndata/tests/helpers.py
@@ -12,7 +12,6 @@ import h5py
 import numpy as np
 import pandas as pd
 import pytest
-import zarr
 from pandas.api.types import is_numeric_dtype
 from scipy import sparse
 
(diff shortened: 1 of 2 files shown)


anndata/tests/helpers.py
```diff
@@ -759,21 +758,30 @@ def shares_memory_sparse(x, y):
...
+        def set_key_trackers(self, keys_to_track):
+            for k in keys_to_track:
+                self._access_count[k] = 0
+except ImportError:
+
+    class AccessTrackingStore:
+        def __init__(self) -> None:
+            raise ImportError(
+                "zarr must be imported to create an `AccessTrackingStore` instance."
+            )
```
ilan-gold, line 787: Not a huge fan of this, but I think the other option is to create `AccessTrackingStore` based on some dummy base class, which will error out on construction with a less helpful error.  And if we want to raise an error, we're sort of back to this.

ivirshup:
I think an `anndata-mindeps` test job would be the way to test this, and then testing all the functionality used by `scanpy-mindeps` uses under that job. I don't think a static check would be enough here, since a something could always be imported dynamically

docs/release-notes/0.10.6.md
```diff
@@ -2,6 +2,7 @@
 
 ```{rubric} Bugfix
 ```
+* Do not raise `ImportError` when `zarr` is imported in `anndata.tests.helpers` due to `scanpy` relying on functionality there. {pr}`1343` {user}`ilan-gold`
```
ivirshup, line 5:
```suggestion

* Defer import of zarr in test helpers, as scanpy CI job relies on them {pr}`1343` {user}`ilan-gold`
```

I'd prefer not advertising the private module.

anndata/tests/helpers.py
```diff
@@ -759,21 +758,30 @@ def shares_memory_sparse(x, y):
...
-    def set_key_trackers(self, keys_to_track):
-        for k in keys_to_track:
-            self._access_count[k] = 0
+        def set_key_trackers(self, keys_to_track):
+            for k in keys_to_track:
+                self._access_count[k] = 0
+except ImportError:
+
+    class AccessTrackingStore:
+        def __init__(self) -> None:
```
flying-sheep, line 784:
```suggestion
        def __init__(self, *_args, **_kwargs) -> None:
```

flying-sheep (changes requested):

ivirshup (approved):

PR #1346, opened 2024-01-26
Switch from using rubric in release notes

ivirshup:
Just something that always bothered me. Plus if we want to put release notes on github it'd be nice to have less that needs reformatting.


flying-sheep:
Good call! We should also make sure that people can link to patch versions.

For some reason, anchors like `#version-0-10` exist, while 0.10.5 currently has the (unstable) anchor `#id5`

ivirshup:
> For some reason, anchors like #version-0-10 exist, while 0.10.5 currently has the (unstable) anchor #id5

I have a local commit for this

The problem is that sphinx can't do ids that start with numbers, so I added "Version". I would also like to keep the date out of it so it just looks like 

https://anndata.readthedocs.io/en/latest/release-notes/index.html#version-0-10-5

But that might require putting the date on a separate line, which I'd rather not do.

flying-sheep:
Yeah, that would be great! Mabye we can define the `smaller` role in a way that hides it from Sphinx’ slug generation mechanism. Now that we use scanpydoc again, centralizing that is easy.

docs/conf.py
```diff
@@ -96,8 +102,8 @@
 
 
 def setup(app: Sphinx):
-    # Don’t allow broken links. DO NOT CHANGE THIS LINE, fix problems instead.
-    app.warningiserror = True
```
flying-sheep, line 100: By now, this is enforced in .readthedocs.yml

Having it twice helped against the most egregious offenders, but by now I’m in a more active role and can prevent the other instance from being removed.

docs/release-notes/index.md
```diff
@@ -1,8 +1,45 @@
 # Release notes
 
-```{include} /release-notes/release-latest.md
+(v0.11)=
```
flying-sheep, line 3: the `v` is load-bearing, without it the generated HTML anchors are generic `id40` ones.

docs/release-notes/0.8.1.md
flying-sheep, line 1: this release doesn’t actually exist

PR #1359, opened 2024-02-07
added yaml for gpu_ci

Intron7:
Adds `Yaml` file for the GPU test environment.

### DIFF
--- a/.ci/gpu_ci.yml
+++ b/.ci/gpu_ci.yml
@@ -1,0 +1,12 @@
+name: cupy_env
+channels:
+ - nvidia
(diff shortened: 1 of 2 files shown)


ivirshup:
Seeing a different error (but still failures) if I bump the cuda version to 12.3 https://github.com/scverse/anndata/actions/runs/7815601152/job/21319267093?pr=1360

ivirshup:
I'm going to merge this as a temporary fix, but I think its bad that we are now only testing with a version of CUDA from 2022, so will keep tracking this for the next release.

PR #1366, opened 2024-02-09
(fix): calculate mean slice correctly for sparse boolean->slice optimization

ilan-gold:
- [x] Fixes a small bug I found in `subset_by_major_axis_mask`
- [x] Tests added


anndata/tests/test_backed_sparse.py
```diff
@@ -149,26 +156,40 @@ def make_one_elem_mask(size: int) -> np.ndarray:
...
 # test behavior from https://github.com/scverse/anndata/pull/1233
 @pytest.mark.parametrize(
-    "make_bool_mask",
+    "make_bool_mask,should_trigger_optimization",
     [
-        make_randomized_mask,
-        make_alternating_mask,
-        make_one_group_mask,
-        make_one_elem_mask,
+        (make_randomized_mask, None),
```
ilan-gold, line 161: Not sure what to do about randomized masks since their behavior can't be reasoned about without calculating the mean slice length, which is somewhat circular reasoning in a unit test context that relies on that calculation being correct in the function itself.

anndata/tests/test_backed_sparse.py
```diff
@@ -128,13 +128,20 @@ def make_randomized_mask(size: int) -> np.ndarray:
...
     for i in range(0, size, 10):  # 10 is enough to trigger new behavior
         mask_alternating[i] = False
     return mask_alternating
 
 
+def make_alternating_mask_5(size: int) -> np.ndarray:
+    mask_alternating = np.ones(size, dtype=bool)
+    for i in range(0, size, 5):  # 5 is too low to trigger new behavior
+        mask_alternating[i] = False
+    return mask_alternating
```
ivirshup, line 142:
```suggestion
from functools import partial

def make_alternating_mask(size: int, step: int) -> np.ndarray:
    mask_alternating = np.ones(size, dtype=bool)
    for i in range(0, size, step):  # 5 is too low to trigger new behavior
        mask_alternating[i] = False
    return mask_alternating
    
make_alternating_mask_5 = partial(make_alternating_mask, step=5)
make_alternating_mask_10 = partial(make_alternating_mask, step=10)
```

Could we do something more like this?

anndata/tests/test_backed_sparse.py
```diff
@@ -149,26 +156,40 @@ def make_one_elem_mask(size: int) -> np.ndarray:
...
+        (make_one_elem_mask, False),
     ],
-    ids=["randomized", "alternating", "one_group", "one_elem"],
+    ids=["randomized", "alternating_10", "alternating_5", "one_group", "one_elem"],
 )
 def test_consecutive_bool(
+    mocker,
     ondisk_equivalent_adata: tuple[AnnData, AnnData, AnnData, AnnData],
     make_bool_mask: Callable[[int], np.ndarray],
+    should_trigger_optimization: bool | None,
```
ivirshup, line 173: Could you document this value? I think saying what the expected optimization is should be fine.

anndata/tests/test_backed_sparse.py
ivirshup: Fwiw, the 5 consecutive case would really benefit from the optimization.

anndata/tests/test_backed_sparse.py
ivirshup: I think this is fine.

ivirshup (changes requested):

anndata/tests/test_backed_sparse.py
ilan-gold: This is probably laptop dependent.  I will check when I get home but I am fairly certain the cutoff point was 7.

PR #1395, opened 2024-03-03
Invalid type of X ValueError lists unclear accepted types

### LINKED ISSUE #1394: Invalid type of X ValueError lists unclear accepted type

**Description**
Invalid type of X `ValueError` lists unclear accepted types.

**Example**
```python
from dask import array as da
from dask import dataframe as dd
import scanpy as sc

Xda = da.random.randint(low=0, high=100, size=(100, 1000), chunks=(10, 100))
Xdd = dd.from_dask_array(Xda, columns=[f"gene{i}" for i in range(1000)])
adata = sc.AnnData(Xdd)
```

returns

ValueError: `X` needs to be of one of ndarray, MaskedArray, spmatrix, Array, ZappyArray, Array, CupyArray, CupySparseMatrix, BaseCompressedSparseDataset, not <class 'dask.dataframe.core.DataFrame'>

E.g. what "Array" means, and why it is listed twice is not immediately visible.

**Suggested Enhancement**
I wonder whether showing more information from the list below would be more clear than simply the `__name__`?

<class 'numpy.ndarray'>
<class 'numpy.ma.core.MaskedArray'>
<class 'scipy.sparse._matrix.spmatrix'>
<class 'zarr.core.Array'>
<class 'anndata.compat.ZappyArray'>
<class 'dask.array.core.Array'>
<class 'anndata.compat.CupyArray'>
<class 'anndata.compat.CupySparseMatrix'>
<class 'anndata._core.sparse_dataset.BaseCompressedSparseDataset'>

Code responsible [here](https://github.com/scverse/anndata/blob/f0d3a6e926dc3cc2bfbfcf987ee90abe0833e2b1/anndata/_core/anndata.py#L515)

eroell:
List module info as well

- [x] Closes #1394
- [x] Tests added

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -514,7 +514,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                     break
             else:
                 class_names = ", ".join(
-                    f"{c.__module__}.{c.__name__}" for c in StorageType.classes()
+                    repr(c) for c in StorageType.classes()
                 )
 
                 raise ValueError(
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -513,10 +513,8 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 if isinstance(X, s_type.value):
                     break
             else:
-                class_names = ", ".join(repr(c) for c in StorageType.classes())
-
                 raise ValueError(
-                    f"`X` needs to be of one of {class_names}, not {type(X)}."
+                    f"X needs to be of one of numpy.ndarray, numpy.ma.core.MaskedArray, scipy.sparse.spmatrix, h5py.Dataset, zarr.Array, anndata.experimental.[CSC,CSR]Dataset, dask.array.Array, cupy.ndarray, cupyx.scipy.sparse.spmatrix, not {type(X)}."
                 )
             if shape is not None:
                 raise ValueError("`shape` needs to be `None` if `X` is not `None`.")
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -513,9 +513,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                 if isinstance(X, s_type.value):
                     break
             else:
-                class_names = ", ".join(
-                    repr(c) for c in StorageType.classes()
-                )
+                class_names = ", ".join(repr(c) for c in StorageType.classes())
 
                 raise ValueError(
                     f"`X` needs to be of one of {class_names}, not {type(X)}."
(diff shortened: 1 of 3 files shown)


anndata/_core/anndata.py
```diff
@@ -512,7 +513,10 @@ def _init_as_actual(
                 if isinstance(X, s_type.value):
                     break
             else:
-                class_names = ", ".join(c.__name__ for c in StorageType.classes())
+                class_names = ", ".join(
+                    f"{c.__module__}.{c.__name__}" for c in StorageType.classes()
```
ivirshup, line 517:
```suggestion
                    repr(c) for c in StorageType.classes()
```

How about this? Then these classes will print in a similar style to `type(X)`

ivirshup (commented):
A test to make sure the improved behaviour doesn't change would be good!

anndata/_core/anndata.py
eroell: Yes, looks also good to me!

ivirshup (commented):
That test looks good, and I think the place is fine.

This should get a release note. I think we can do this for a bugfix release.

It would be nice if we didn't report types that come from `anndata.compat`, since that may be confusing to users (they're never meant to be used by an actual user). I'm a little unsure of what to do there except hard code some names or types... But I think this can be considered separate from this since we were already doing it.

eroell:
E.g. like that? (not mentioning zappy and BaseCompressedSparseDataset)

`X` needs to be of one of <class 'numpy.ndarray'>, <class 'numpy.ma.core.MaskedArray'>, <class 'scipy.sparse._matrix.spmatrix'>, <class 'zarr.core.Array'>, <class 'dask.array.core.Array'>, <class ‘cupy.ndarray'>, <class 'cupyx.scipy.sparse.spmatrix'>, not {type(X)}.

ivirshup:
Could be:

X needs to be of one of numpy.ndarray, numpy.ma.core.MaskedArray, scipy.sparse.spmatrix, h5py.Dataset, zarr.Array, anndata.experimental.{CSC,CSR}Dataset, dask.array.Array, cupy.ndarray, cupyx.scipy.sparse.spmatrix, not {type(X)}.

Simplified some names to the user importable version, added `h5py.Dataset` and `anndata.experimental.{CSC,CSR}Dataset`

eroell:
> Could be:
> 
> X needs to be of one of numpy.ndarray, numpy.ma.core.MaskedArray, scipy.sparse.spmatrix, h5py.Dataset, zarr.Array, anndata.experimental.{CSC,CSR}Dataset, dask.array.Array, cupy.ndarray, cupyx.scipy.sparse.spmatrix, not {type(X)}.

Done that (changed to [CSC,CSR] instead of {CSC,CSR} since using f-string)

ivirshup (approved):
FWIW: I think you can escape the `{` by doing `f"{{"`

But LGTM

PR #1400, opened 2024-03-06
Update `to_df` to include metadata

### LINKED ISSUE #1399: Include metadata when converting to Pandas DataFrame
## Description

`AnnData.to_df` only converts data from `.X` or a layer to a DataFrame. It'd be great to also add metadata column from `AnnData.obs`. This feature would make it easier to work with Pandas `groupby` on the generated DataFrame, for example.

WeilerP:
## Changes

* Adds argument `obs` to include columns from `.obs` in generated metadata.

## Related issues

Closes #1399.

### DIFF
--- a/anndata/_core/anndata.py
+++ b/anndata/_core/anndata.py
@@ -1284,7 +1284,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
 
     T = property(transpose)
 
-    def to_df(self, layer=None) -> pd.DataFrame:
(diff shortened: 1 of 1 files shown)


WeilerP:
I'll add unit tests once it's clear if this feature is desired and how to include it.

ivirshup:
Just to write down some edge cases:

* What if there are repeated:
    * `obs_names`
    * `var_names`
* What if `(key in adata.obs.columns) and (key in adata.var_names)`

Issue #1434, opened 2024-03-22
Upcoming Scipy 1.13 incompatible with file-backed sparse matrices


flying-sheep:
### Please make sure these conditions are met

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest version of anndata.
- [ ] (optional) I have confirmed this bug exists on the master branch of anndata.

### Report

Code:

```python
from anndata.experimental import sparse_dataset

sparse_dataset(some_group)[0,0]
```

or e.g.

```shell
pytest -k test_read_partial_adata
```

Traceback:

```pytb
________________________ test_read_partial_adata[zarr] _________________________
>           part = read_elem_partial(storage["obsp"][key], indices=(obs_idx, obs_idx))
anndata/tests/test_io_partial.py:86: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
anndata/_io/specs/registry.py:371: in read_elem_partial
    return _REGISTRY.get_partial_reader(
anndata/_io/specs/methods.py:623: in read_sparse_partial
    return sparse_dataset(elem)[indices]
anndata/_core/sparse_dataset.py:427: in __getitem__
    sub = mtx[row, col]
/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/scipy/sparse/_index.py:91: in __getitem__
    return self._get_columnXarray(row[:,0], col.ravel())
/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/scipy/sparse/_compressed.py:683: in _get_columnXarray
    return self._major_index_fancy(major)._minor_index_fancy(minor)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
        csr_row_index(
            M,
            indices,
            self.indptr.astype(idx_dtype, copy=False),
>           self.indices.astype(idx_dtype, copy=False),
            self.data,
            res_indices,
            res_data
        )
E       TypeError: Array.astype() got an unexpected keyword argument 'copy'

/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/scipy/sparse/_compressed.py:709: TypeError
```

### Versions

```python
[...]

PR #1435, opened 2024-03-22
Fix pre-release tests

flying-sheep:
@ilan-gold this should unblock you

See https://github.com/scverse/anndata/issues/1434 (doesn‘t fix that one though)

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -61,7 +61,7 @@ jobs:
         condition: eq(variables['DEPENDENCIES_VERSION'], 'minimum')
 
       - script: |
-          uv pip install --system --compile --pre "anndata[dev,test] @ ."
+          uv pip install --system --compile --pre "anndata[dev,test] @ ." "scanpy>=1.10.0rc1"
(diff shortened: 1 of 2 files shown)


ilan-gold (approved):
I'm fine with upper bounding packages.  I will take a look at the issue.  Is @ivirshup ok with upper bounding this?  I can't think of a good reason not to but he often has something up his sleeve!

flying-sheep:
Scipy is unlikely to change their implementation to not use `copy=True`, so it’s on us to make this work.

Since it’s not a temporary bug in scipy but a change that broke us, I introduced a temporary upper bound (as opposed to a `!=1.13.0.rc1` constraint) until we have a fix.

ivirshup:
> In an ideal world, all past anndata versions would have the same bound, but that’s impossible to do.

I believe conda-forge actually allows this

flying-sheep:
Yeah, I was thinking about PyPI. There it would also kinda maybe be possible by yanking everything and releasing .post1 releases but I don't think that's a reasonable standard to hold ourselves to and it would force people to change their lock files

PR #1459, opened 2024-04-09
Move tests out of package

### LINKED ISSUE #1451: Move tests out of the package
We currently have unit tests inside of the anndata package. We should separate them out following the recommended structure from Pytest (that we use in the scverse template).

Not having them separated also creates an issue with the `src` layout and `--pyargs`:

1. We excise `tests` from the package during packaging using a exclude config
2. this means that running tests on an editable install vs a regular install works differently:

   - In an editable install, `pytest --pyargs anndata` will run the unit tests inside of `anndata/tests`. ✅
   - In an editable install, `pytest --pyargs anndata ./anndata/tests` will run the unit tests twice
   - In a regular install, `pytest --pyargs anndata` will only run doctest since there are no unit tests in `anndata/tests`
   - In a regular install, `pytest --pyargs anndata ./anndata/tests` will run doctests and unit tests ✅

flying-sheep:
- [x] Closes #1451
- [x] Tests added


ivirshup:
What's up with the coverage?

flying-sheep:
`pytest --cov` instead of `coverage run`. Although because of https://github.com/nedbat/coveragepy/issues/1696, `coverage run` is also not optimal.

`coverage run` always works, whereas `pytest --cov` breaks if you somehow get pytest to import your module before it initializes the plugin.

pyproject.toml
```diff
@@ -125,16 +119,14 @@ exclude_also = [
 
 [tool.pytest.ini_options]
 addopts = [
-    # "--import-mode=importlib",  # TODO: enable: https://github.com/scverse/anndata/issues/1451
+    "--import-mode=importlib",
```
ivirshup, line 122:
```suggestion
```

I'm getting quite a few errors locally, but removing this seems to solve that. Does removing this break anything?

pyproject.toml
```diff
@@ -125,16 +119,14 @@ exclude_also = [
...
     "--doctest-modules",
     "--pyargs",
 ]
 filterwarnings = [
-    'ignore:Support for Awkward Arrays is currently experimental',
-    'ignore:Outer joins on awkward\.Arrays',
-    # TODO: replace both lines above with this one once we figured out how prevent ImportPathMismatchError
-    # 'ignore::anndata._warnings.ExperimentalFeatureWarning',
+    'ignore::anndata._warnings.ExperimentalFeatureWarning',
+    'ignore::anndata.OldFormatWarning',
```
ivirshup, line 129:
```suggestion
    'ignore:Support for Awkward Arrays is currently experimental',
    'ignore:Outer joins on awkward\.Arrays',
```

This seems to clear up issues with pytest-cov, and still lets us do coverage in parallel

ivirshup (commented):
More specifically, the coverage is different here due to the changes to `filterwarnings` in the `pyproject.toml`. I significantly prefer being able to run tests in parallel (which I believe isn't really possible with `coverage`) locally over specifying warnings to filter by class in the `pyproject.toml`.

I've made a suggestion which still allows parallel testing and coverage collection.

pyproject.toml
flying-sheep: That’s not right, you just deleted the `OldFormatWarning` line without replacement.

Unless the `warnings.filterwarnings` line in conftest.py was redundant.

Also if we go back to hacky message matching, we need to link the two issues that block this:

- https://github.com/nedbat/coveragepy/issues/1696
- https://github.com/pytest-dev/pytest-cov/issues/627

we need either of them to be fixed to have a satisfying solution (either keep using `pytest --cov` if that fix comes first or switch to `PYTHONSAFEPATH=1 coverage run -m pytest` if that fix comes first)

pyproject.toml
flying-sheep: cannot reproduce, could you please try investigating why that is?

pyproject.toml
ivirshup: Wait, why would we switch to `PYTHONSAFEPATH=1 coverage run -m pytest`?

I would really like to keep pytest-cov going so that I can collect coverage locally in ~30 seconds instead of ~5 min.

flying-sheep: Yeah, seems the alternatives don’t work as reliably: https://github.com/nedbat/coveragepy/issues/1341

flying-sheep:
OK, I think this looks good. It’s blocked on https://github.com/scverse/scanpy/issues/2993, which is in turn blocked on pytest’s 8.2 release.

ilan-gold (commented):
I still get `CoverageWarning: Module anndata was previously imported, but not measured (module-not-measured)` locally.  Is this intentional or expected?  Should we catch it?

Otherwise I was able to run the tests locally, in parallel, getting a codecov report that has non-zero `line-rate` for the various submodules.  Is there a tool for visualizing this xml?  I couldn't find anything

flying-sheep:
Yes, but you can also just check if the newest commit hash matches what the codecov comment says, and then check the link from the codecov comment: https://app.codecov.io/gh/scverse/anndata/pull/1459

Seems like somehow this PR increases coverage very very slightly?

ilan-gold:
> As said, https://github.com/scverse/anndata/commit/ba3c054355414ed4f622e8ef9d1e4eddac5607c8 fixed the issue that resulted in anndata being imported before pytest-cov, so now this PR doesn’t negatively affect coverage anymore.

I am seeing it in parallel, but can't find the issue about it which I saw earlier so am not 100% sure if that's expected.

My coverage looks good in parallel and similar to the one not in parallel, with the only coverage change being the test helpers.  So to me, this PR seems fine.

PR #1462, opened 2024-04-09
lower required pytest version

ivirshup:
So that we can have dev installations of scanpy and anndata in the same environment, as is often useful

See: https://github.com/scverse/scanpy/issues/2993

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ doc = [
 ]
 test = [
     "loompy>=3.0.5",
-    "pytest>=8",
+    "pytest>=8.0",
     "pytest-cov>=2.10",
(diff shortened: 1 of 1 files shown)


flying-sheep:
Let’s think about this a second, I’m still investigating

ivirshup:
Using `pytest==8.0.2` seems to work fine for the minimum dependency test job.

Are there other harms you are expecting from allowing that pytest version?

ivirshup:
Just had a quick call where we couldn't find a problem introduced by this, though both vaguely remember some problem related to pytest versions with the `src` pr. We can revert if we find problems here, but for right now this unblocks fixing scanpy dev environments.

PR #1468, opened 2024-04-10
update doc-string for sparse dataset

ivirshup:
- [ ] Closes #
- [x] Tests added the doctests

Curious to see how the randomness plays with CI *UPDATE* not well

### DIFF
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -603,16 +603,45 @@ def sparse_dataset(group: GroupStorageType) -> CSRDataset | CSCDataset:
     Example
     -------
 
-    >>> import zarr
-    >>> from anndata.experimental import sparse_dataset
-    >>> group = zarr.open_group("./my_test_store.zarr")
-    >>> group["data"] = [10, 20, 30, 40, 50, 60, 70, 80]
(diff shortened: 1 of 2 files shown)


src/anndata/_core/sparse_dataset.py
```diff
@@ -603,16 +603,42 @@ def sparse_dataset(group: GroupStorageType) -> CSRDataset | CSCDataset:
...
+    These can also be used inside of an AnnData object, no need for backed mode
+
+    >>> from anndata import AnnData
+    >>> adata = AnnData(
+    ...     layers={"backed": X}, obs=read_elem(f["obs"]), var=read_elem(f["var"])
+    ... )
+    >>> adata.layers["backed"]
+    CSRDataset: backend hdf5, shape (700, 765), data_dtype float32
+
+    Access pulls it into memory
```
ilan-gold, line 635: I think this is ambiguous because I would also say `adata.layers["backed"]` counts as accessing it but does not bring the data into memory.

Maybe "Indexing access i.e., from views, brings data into memory"

ivirshup:
```suggestion
    Indexing access (i.e., from views) brings selection into memory
```

ilan-gold: Looks good!

Issue #523, opened 2021-03-22
Pass Through `chunks` arg in `write_zarr` to `layers`


ilan-gold:
Would it be possible to replicate this functionality:

https://github.com/theislab/anndata/blob/2723c9c9e8bc0fe50a9513f292916616a603923b/anndata/_io/zarr.py#L48-L51

in the `layers` as well?

https://github.com/theislab/anndata/blob/2723c9c9e8bc0fe50a9513f292916616a603923b/anndata/_io/zarr.py#L58

Maybe there is a reason not to do this though

ivirshup:
So, I'm a little bit hesitant to define much about the behavior of `chunks` at the moment.  This is because we'd like to integrate `dask` with anndata, where chunks need to be aligned between arrays that are used together. In this instance, it may make more sense if sparse arrays aren't  chunked along the `indices` or `data` but the sparse arrays themselves would be chunked by their cartesian indices (e.g. you'd get chunks of sparse arrays instead of chunks of dense arrays).

I don't see any harm in adding the behavior you suggest right now, but it could change soon. I think there would be value in being able to specify these things separately.

ilan-gold:
@ivirshup I agree that they should be specified separately, especially since they `layers` and `X` might be different types.  Perhaps in order to wrap up #524 as well, we could just add a `sparse_chunks` argument to `write_zarr`?  Then these could both be passed on to the `layers`?

flying-sheep:
@ilan-gold is this issue still relevant?

PR #1469, opened 2024-04-11
(feat): `read_elem_as_dask` method

### LINKED ISSUE #1105: Distributed writing for H5ad format due to h5py objects being unserializable

### Report

This is the code that will fail.

```python
import anndata as ad
import dask.array as da
import dask.distributed as dd

with dd.LocalCluster(n_workers=1,threads_per_worker=1) as cluster:
    with dd.Client(cluster) as client:
        adata = ad.AnnData(da.random.random((100, 100), chunks=(10, 10)))
        adata.write_h5ad("test.h5ad")
```
Usually the same code used to fail for both zarr and h5ad, but this PR will fix the issue with zarr https://github.com/scverse/anndata/pull/1079. For h5ad serialization of h5py might be overcome by whatever Xarray does as mentioned in this issue https://github.com/pydata/xarray/issues/4242

Traceback:

```pytb
023-08-25 11:17:10,491 - distributed.protocol.pickle - ERROR - Failed to serialize <ToPickle: HighLevelGraph with 1 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f6ae131c700>
 0. 140097021523072
>.
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/distributed/protocol/pickle.py", line 63, in dumps
    result = pickle.dumps(x, **dump_kwargs)
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/h5py/_hl/base.py", line 368, in __getnewargs__
    raise TypeError("h5py objects cannot be pickled")
TypeError: h5py objects cannot be pickled

  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/distributed/protocol/pickle.py", line 68, in dumps
    pickler.dump(x)
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/distributed/protocol/pickle.py", line 29, in reducer_override
    return deserialize, serialize(obj)
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/distributed/protocol/h5py.py", line 24, in serialize_h5py_dataset
    header, _ = serialize_h5py_file(x.file)
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/distributed/protocol/h5py.py", line 11, in serialize_h5py_file
    raise ValueError("Can only serialize read-only h5py files")
ValueError: Can only serialize read-only h5py files

...
    return Pickler.dump(self, obj)
  File "/home/sel/mambaforge/envs/dask/lib/python3.9/site-packages/h5py/_hl/base.py", line 368, in __getnewargs__
    raise TypeError("h5py objects cannot be pickled")
TypeError: h5py objects cannot be pickled
```

### Versions
```
-----
anndata             0.10.0.dev198+ga61d5d4
dask                2023.7.1
distributed         2023.7.1
numpy               1.22.4
pandas              2.0.0
scipy               1.9.3
session_info        1.0.0
zarr                2.13.3
-----
PIL                 9.2.0
asciitree           NA
asttokens           NA
attr                23.1.0
awkward             2.1.0
awkward_cpp         NA
backcall            0.2.0
bokeh               2.4.3
cffi                1.15.1
click               8.1.3
cloudpickle         2.2.0
colorama            0.4.6
comm                0.1.1
cython_runtime      NA
cytoolz             0.12.0
...
Python 3.9.15 | packaged by conda-forge | (main, Nov 22 2022, 15:55:03) [GCC 10.4.0]
Linux-6.1.44-1-MANJARO-x86_64-with-glibc2.38
-----
```
Session information updated at 2023-08-25 11:18

### LINKED ISSUE #1430: `read_elem_as_dask`

We should probably add some `read_elem_as_dask` methods to make reading stored objects from as dask arrays easier.

This largely needs special casing due to:

* Sparse arrays
* Dask not handling `h5py` object in the process based mode: https://github.com/scverse/anndata/issues/1105

I think we can skip dataframe support for now. It may be worth thinking about how to read objects into GPU memory as well.

I'm not completley set on the API, so input welcome. My initial thought is that we could just have a smaller registry of methods which only supports reads for dense and sparse arrays.

ilan-gold:
My thinking in terms of the naming for this would be to use this for `xarray` as well but we don't have to.

- [x] Closes #1430 and closes #1105 via our careful handling of the file descriptor
- [x] Tests added

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,9 +101,14 @@ nitpick_ignore = [
     ("py:class", "anndata.compat.DaskArray"),
     ("py:class", "anndata.compat.CupyArray"),
     ("py:class", "anndata.compat.CupySparseMatrix"),
+    ("py:class", "numpy.ma.core.MaskedArray"),
+    ("py:class", "dask.array.core.Array"),
     ("py:class", "awkward.highlevel.Array"),
     ("py:class", "anndata._core.sparse_dataset.BaseCompressedSparseDataset"),
     ("py:obj", "numpy._typing._array_like._ScalarType_co"),
+    # Something is picking these up as classes despite being aliases so this just suppresses the warning, but doesn't affect the build
+    ("py:class", "RWAbleDict"),
+    ("py:class", "RWAbleList"),
 ]
 suppress_warnings = [
     "ref.citation",
@@ -138,12 +143,19 @@ qualname_overrides = {
     "anndata._types.WriteCallback": "anndata.experimental.WriteCallback",
     "anndata._types.Read": "anndata.experimental.Read",
     "anndata._types.Write": "anndata.experimental.Write",
+    "anndata._types.RWAble": "anndata.experimental.RWAble",
+    "anndata._types.RWAbleDict": "anndata.experimental.RWAbleDict",
+    "anndata._types.RWAbleList": "anndata.experimental.RWAbleList",
 }
 autodoc_type_aliases = dict(
     NDArray=":data:`~numpy.typing.NDArray`",
-    InMemoryReadElem=":data:`~anndata.experimental.InMemoryReadElem`",
-    InMemoryType=":data:`~anndata.experimental.InMemoryArrayOrScalarType`",
-    InMemoryArrayOrScalarType=":data:`~anndata.experimental.InMemoryArrayOrScalarType`",
+    RWAble=":data:`~anndata.experimental.RWAble`",
+    RWAbleDict=":data:`~anndata.experimental.RWAbleDict`",
+    RWAbleList=":data:`~anndata.experimental.RWAbleList`",
+    **{
+        f"{v}variantInMemoryType": ":data:`~anndata.experimental.InMemoryElem`"
+        for v in ["In", "Co", "Contra"]
+    },
 )
 
 # -- Social cards ---------------------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ release = version = metadata.version("anndata")
 # default settings
 templates_path = ["_templates"]
 html_static_path = ["_static"]
-source_suffix = [".rst", ".md"]
+source_suffix = {".rst": "restructuredtext", ".md": "markdown"}
 master_doc = "index"
 default_role = "literal"
 exclude_patterns = [
@@ -110,7 +110,6 @@ nitpick_ignore = [
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
-    "autosummary.import_cycle",  # https://github.com/sphinx-doc/sphinx/issues/12589
 ]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,7 +110,7 @@ nitpick_ignore = [
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
-    "autosummary.import_cycle",  # https://readthedocs.com/projects/icb-anndata/builds/2349021/
+    "autosummary.import_cycle",  # https://github.com/sphinx-doc/sphinx/issues/12589
 ]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,6 +110,7 @@ nitpick_ignore = [
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
+    "autosummary.import_cycle",  # https://github.com/sphinx-doc/sphinx/issues/12589
 ]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -128,8 +128,8 @@ intersphinx_mapping = dict(
     scipy=("https://docs.scipy.org/doc/scipy/", None),
     sklearn=("https://scikit-learn.org/stable/", None),
     zarr=("https://zarr.readthedocs.io/en/stable/", None),
-    dask=("https://docs.dask.org/en/stable/", None),
     xarray=("https://xarray.pydata.org/en/stable/", None),
+    dask=("https://docs.dask.org/en/stable/", None),
 )
 qualname_overrides = {
     "h5py._hl.group.Group": "h5py.Group",
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,7 +110,6 @@ nitpick_ignore = [
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
-    "autosummary.import_cycle",  # https://github.com/sphinx-doc/sphinx/issues/12589
 ]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -126,6 +126,7 @@ intersphinx_mapping = dict(
     scipy=("https://docs.scipy.org/doc/scipy/", None),
     sklearn=("https://scikit-learn.org/stable/", None),
     zarr=("https://zarr.readthedocs.io/en/stable/", None),
+    dask=("https://docs.dask.org/en/stable/", None),
     xarray=("https://xarray.pydata.org/en/stable/", None),
 )
 qualname_overrides = {
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -106,9 +106,6 @@ nitpick_ignore = [
     ("py:class", "awkward.highlevel.Array"),
     ("py:class", "anndata._core.sparse_dataset.BaseCompressedSparseDataset"),
     ("py:obj", "numpy._typing._array_like._ScalarType_co"),
-    # Something is picking these up as classes despite being aliases so this just suppresses the warning, but doesn't affect the build
-    ("py:class", "RWAbleDict"),
-    ("py:class", "RWAbleList"),
 ]
 suppress_warnings = [
     "ref.citation",
@@ -144,14 +141,10 @@ qualname_overrides = {
     "anndata._types.Read": "anndata.experimental.Read",
     "anndata._types.Write": "anndata.experimental.Write",
     "anndata._types.RWAble": "anndata.experimental.RWAble",
-    "anndata._types.RWAbleDict": "anndata.experimental.RWAbleDict",
-    "anndata._types.RWAbleList": "anndata.experimental.RWAbleList",
 }
 autodoc_type_aliases = dict(
     NDArray=":data:`~numpy.typing.NDArray`",
     RWAble=":data:`~anndata.experimental.RWAble`",
-    RWAbleDict=":data:`~anndata.experimental.RWAbleDict`",
-    RWAbleList=":data:`~anndata.experimental.RWAbleList`",
     **{
         f"{v}variantInMemoryType": ":data:`~anndata.experimental.InMemoryElem`"
         for v in ["In", "Co", "Contra"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -110,6 +110,7 @@ nitpick_ignore = [
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
+    "autosummary.import_cycle",  # https://readthedocs.com/projects/icb-anndata/builds/2349021/
 ]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -60,6 +60,7 @@ extensions = [
     "sphinx.ext.linkcode",
     "nbsphinx",
     "IPython.sphinxext.ipython_console_highlighting",
+    "sphinx_toolbox.more_autodoc.autoprotocol",
 ]
 myst_enable_extensions = [
     "html_image",  # So README.md can be used on github and sphinx docs
@@ -94,6 +95,15 @@ nitpick_ignore = [
     # TODO: sphinx’ builtin autodoc.typehints extension isn’t handled by `qualname_overrides` yet
     # https://github.com/theislab/scanpydoc/issues/140
     ("py:class", "h5py._hl.group.Group"),
+    ("py:class", "h5py._hl.dataset.Dataset"),
+    # for experimental callback exports
+    ("py:class", "anndata.compat.ZappyArray"),
+    ("py:class", "anndata.compat.DaskArray"),
+    ("py:class", "anndata.compat.CupyArray"),
+    ("py:class", "anndata.compat.CupySparseMatrix"),
+    ("py:class", "awkward.highlevel.Array"),
+    ("py:class", "anndata._core.sparse_dataset.BaseCompressedSparseDataset"),
+    ("py:obj", "numpy._typing._array_like._ScalarType_co"),
 ]
 suppress_warnings = [
     "ref.citation",
@@ -123,7 +133,17 @@ qualname_overrides = {
     "h5py._hl.files.File": "h5py.File",
     "h5py._hl.dataset.Dataset": "h5py.Dataset",
     "anndata._core.anndata.AnnData": "anndata.AnnData",
+    "anndata._types.ReadCallback": "anndata.experimental.ReadCallback",
+    "anndata._types.WriteCallback": "anndata.experimental.WriteCallback",
+    "anndata._types.Read": "anndata.experimental.Read",
+    "anndata._types.Write": "anndata.experimental.Write",
 }
+autodoc_type_aliases = dict(
+    NDArray=":data:`~numpy.typing.NDArray`",
+    InMemoryReadElem=":data:`~anndata.experimental.InMemoryReadElem`",
+    InMemoryType=":data:`~anndata.experimental.InMemoryArrayOrScalarType`",
+    InMemoryArrayOrScalarType=":data:`~anndata.experimental.InMemoryArrayOrScalarType`",
+)
 
 # -- Social cards ---------------------------------------------------------
 
--- a/docs/release-notes/0.11.0.md
+++ b/docs/release-notes/0.11.0.md
@@ -8,7 +8,8 @@
 * Add `should_remove_unused_categories` option to `anndata.settings` to override current behavior.  Default is `True` (i.e., previous behavior).  Please refer to the [documentation](https://anndata.readthedocs.io/en/latest/generated/anndata.settings.html) for usage.  {pr}`1340` {user}`ilan-gold`
 * `scipy.sparse.csr_array` and `scipy.sparse.csc_array` are now supported when constructing `AnnData` objects {pr}`1028` {user}`ilan-gold` {user}`isaac-virshup`
 * Add `should_check_uniqueness` option to `anndata.settings` to override current behavior.  Default is `True` (i.e., previous behavior).  Please refer to the [documentation](https://anndata.readthedocs.io/en/latest/generated/anndata.settings.html) for usage.  {pr}`1507` {user}`ilan-gold`
-* Add `~anndata.experimental.read_elem_as_dask` function to handle i/o with sparse and dense arrays {pr}`1469` {user}`ilan-gold`
+* Add :func:`~anndata.experimental.read_elem_as_dask` function to handle i/o with sparse and dense arrays {pr}`1469` {user}`ilan-gold`
+* Add functionality to write from GPU {class}`dask.array.Array` to disk {pr}`1550` {user}`ilan-gold`
 
 #### Bugfix
 
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -63,6 +63,16 @@ def maybe_open_h5(path_or_group: Path | ZarrGroup, elem_name: str):
 _DEFAULT_STRIDE = 1000
 
 
+def compute_chunk_layout_for_axis_shape(
+    chunk_axis_shape: int, full_axis_shape: int
+) -> tuple[int, ...]:
+    n_strides, rest = np.divmod(full_axis_shape, chunk_axis_shape)
+    chunk = (chunk_axis_shape,) * n_strides
+    if rest > 0:
+        chunk += (rest,)
+    return chunk
+
+
 @_LAZY_REGISTRY.register_read(H5Group, IOSpec("csc_matrix", "0.1.0"))
 @_LAZY_REGISTRY.register_read(H5Group, IOSpec("csr_matrix", "0.1.0"))
 @_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("csc_matrix", "0.1.0"))
@@ -103,10 +113,7 @@ def read_sparse_as_dask(
         return chunk
 
     shape_minor, shape_major = shape if is_csc else shape[::-1]
-    n_strides, rest = np.divmod(shape_major, stride)
-    chunks_major = (stride,) * n_strides
-    if rest > 0:
-        chunks_major += (rest,)
+    chunks_major = compute_chunk_layout_for_axis_shape(stride, shape_major)
     chunks_minor = (shape_minor,)
     chunk_layout = (
         (chunks_minor, chunks_major) if is_csc else (chunks_major, chunks_minor)
@@ -147,13 +154,10 @@ def read_h5_array(
                 idx += (slice(start, stop),)
             return f[idx]
 
-    chunk_layout = ()
-    for i in range(len(shape)):
-        n_strides, rest = np.divmod(shape[i], chunks[i])
-        chunk = (chunks[i],) * n_strides
-        if rest > 0:
-            chunk += (rest,)
-        chunk_layout += (chunk,)
+    chunk_layout = tuple(
+        compute_chunk_layout_for_axis_shape(chunks[i], shape[i])
+        for i in range(len(shape))
+    )
 
     return da.map_blocks(
         make_dask_chunk,
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -79,7 +79,7 @@ def get_elem_name(x):
     raise NotImplementedError(f"Not implemented for {type(x)}")
 
 
-@get_elem_name.register(h5py.Group)
+@get_elem_name.register(H5Group)
 def _(x):
     return x.name
 
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -4,7 +4,7 @@ from contextlib import contextmanager
 from functools import singledispatch
 from pathlib import Path, PurePosixPath
 from types import MappingProxyType
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING
 
 import h5py
 import numpy as np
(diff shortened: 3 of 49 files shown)


src/anndata/_io/specs/registry.py
```diff
@@ -44,7 +44,7 @@ def _from_read_parts(
     ) -> IORegistryError:
         # TODO: Improve error message if type exists, but version does not
         msg = (
-            f"No {method} method registered for {spec} from {src_typ}. "
+            f"No {method} method registered for {spec} from {src_typ} in registry {registry}. "
```
ilan-gold, line 47: Ok I will remove this, it's way too much

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,105 @@
...
+@_LAZY_REGISTRY.register_read(H5Group, IOSpec("csc_matrix", "0.1.0"))
+@_LAZY_REGISTRY.register_read(H5Group, IOSpec("csr_matrix", "0.1.0"))
+def read_sparse_as_dask_h5(elem, _reader):
+    filename = elem.file.filename
+    elem_name = elem.name
+    with h5py.File(filename, "r") as f:
+        e = f[elem_name]
+        shape = e.attrs["shape"]
+        encoding_type = e.attrs["encoding-type"]
+        dtype = e["data"].dtype
```
ivirshup: Can't we use the passed `elem` here?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,105 @@
...
+        return da.from_array(elem, chunks=(h5_chunks,) * len(elem.shape))
+    return da.from_array(elem)
+
+
+@_LAZY_REGISTRY.register_read(H5Array, IOSpec("string-array", "0.2.0"))
+def read_h5_string_array(elem, _reader):
+    from anndata._io.h5ad import read_dataset
+
+    elem = read_dataset(elem)
+    return read_h5_array(elem, _reader)
```
ivirshup, line 98: Ooh. I'm a little less sure about this. I was thinking this could just be sparse and dense arrays for now.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,103 @@
...
+from scipy import sparse
+
+import anndata as ad
+from anndata.compat import H5Array, H5Group, ZarrArray, ZarrGroup
+
+from .registry import _LAZY_REGISTRY, IOSpec
+
+# TODO: settings
+stride = 100
+h5_chunks = 1000
```
ilan-gold, line 17: Not sure what to do about this.  The code path is quite complicated for i/o since we're relying on the registry.  The `settings` object would be one way to go about this.  And then another would be updating the registry to allow for passing in `kwargs`...

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,103 @@
...
+        file = h5py.File(filename_or_elem, "r")
+        try:
+            yield file[elem_name]
+        finally:
+            file.close()
+    else:
+        try:
+            yield filename_or_elem
+        finally:
+            pass
```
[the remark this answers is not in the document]
ivirshup: I had an argument whose value was either `h5py.File` or `zarr.open` which was then called on the path.

src/anndata/_io/specs/lazy_methods.py
ivirshup: I think it would make sense for there to be a `chunks` argument to read elem as Dask. Something like: `chunks: None | int | tuple[int, ...] = None`.

tests/test_io_elementwise.py
```diff
@@ -195,7 +243,7 @@ def test_write_anndata_to_root(store):
     ["attribute", "value"],
     [
         ("encoding-type", "floob"),
-        ("encoding-version", "10000.0"),
+        ("encoding-version", "SIZE0.0"),
```
ivirshup, line 246: Unintentional find and replace?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,103 @@
...
+from anndata.compat import H5Array, H5Group, ZarrArray, ZarrGroup
+
+from .registry import _LAZY_REGISTRY, IOSpec
+
+# TODO: settings
+stride = 100
+h5_chunks = 1000
+
+
+def make_dask_array(is_csc, shape, make_dask_chunk, dtype):
```
ivirshup, line 20: I'm not sure I got a ton from this being its own function as it's not re-used anywhere. Personally wouldn't separate it out.

If it does stick around, I would definitely add type hints.

tests/test_io_elementwise.py
```diff
@@ -47,6 +54,39 @@ def store(request, tmp_path) -> H5Group | ZarrGroup:
...
+        SIZE,
+        format=sparse_format,
+        density=0.01,
+        random_state=np.random.default_rng(),
+    )
+    X_dask = da.from_array(X, chunks=(100, 100))
+
+    write_elem(store, "X", X)
+    write_elem(store, "X_dask", X_dask)
+    return store
```
ivirshup, line 87: If we're going to use these in more places, I would suggest not making them square. This has definitely caused problems for us in the past. Maybe even make the shape/ chunk size a parameter?

I would also write a short doc string about what is actually being returned. I think it's a little strange that one of these groups has two things in it, while other only has one.

ilan-gold: I think we can just do the chunking along the axis.  The shape shouldn't matter.  We can reoslve it from `sparse_format`

src/anndata/_io/specs/lazy_methods.py
ilan-gold: I see what's up - I didn't know kwargs got passed through.  Done now.  The chunks here is much less complicated and is really `chunk_size` so should be an `int` and I also made `stride` a kwarg

src/anndata/_io/specs/lazy_methods.py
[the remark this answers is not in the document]
ilan-gold: We have no way of resolving the zarr store's full file path from what I understand.  So we can't re-open the zarr store without something like

https://github.com/scverse/anndata/blob/e2df2d0d49f27558302b2238aab3bff571bb5afb/src/anndata/_core/sparse_dataset.py#L69-L73

which seems somehow more complicated.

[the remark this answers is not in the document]
ilan-gold: And then we'd need the check at the end anyway for `close` for the h5.  Maybe there's some slick way to do this.  I will spend a few more minutes fiddling but if you have something please share!

docs/release-notes/0.10.8.md
```diff
@@ -13,3 +13,5 @@
 
 ```{rubric} Performance
 ```
+
+* Add `~anndata.experimental.read_elem_as_dask` function to handle i/o with sparse and dense arrays {pr}`1469` {user}`ilan-gold`
```
ilan-gold, line 17: @ivirshup Is this right? `0.10.8` or are we waiting until `0.11.0`?

ivirshup: It's a feature, so I'd go `0.11`

tests/test_io_elementwise.py
```diff
@@ -146,30 +203,53 @@ def test_io_spec_cupy(store, value, encoding_type):
...
+    import dask.distributed as dd
+
+    file = h5py.File(tmp_path / "test.h5", "w")
+    store = file["/"]
+    arr_store = create_sparse_store(sparse_format, store)
+    X_dask_from_disk = read_elem_as_dask(arr_store["X"])
+    X_from_disk = read_elem(arr_store["X"])
+    file.close()
+    with dd.LocalCluster(n_workers=1, threads_per_worker=1) as cluster:
+        with dd.Client(cluster) as client:  # noqa: F841
```
flying-sheep, line 251:
```suggestion
    with (
        dd.LocalCluster(n_workers=1, threads_per_worker=1) as cluster,
        dd.Client(cluster) as _client,
    ):
```

flying-sheep (commented):
looks good, but needs a bit of work for maintainability.

One thought: should there be options for `read_elem_as_dask`? Like `chunk_size` or so?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,95 @@
...
+
+@_LAZY_REGISTRY.register_read(H5Group, IOSpec("csc_matrix", "0.1.0"))
+@_LAZY_REGISTRY.register_read(H5Group, IOSpec("csr_matrix", "0.1.0"))
+@_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("csc_matrix", "0.1.0"))
+@_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("csr_matrix", "0.1.0"))
+def read_sparse_as_dask(elem, _reader, stride: int = 100):
+    import dask.array as da
+
+    filename_or_elem = elem.file.filename if isinstance(elem, H5Group) else elem
+    elem_name = elem.name if isinstance(elem, H5Group) else Path(elem.path).name
```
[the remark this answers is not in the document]
ilan-gold: What should be a `PurePosixPath`?  We don't control the `elem`, so I think you need to cast it.

ilan-gold:
> One thought: should there be options for read_elem_as_dask? Like chunk_size or so?

That's a good point.  I will think on this a bit.  What we have now is fairly geared towards one type of chunking.  So maybe `stride` should be exported, but not sure the best way to do it, and I am not sure just allowing `kwargs` on `read_elem_as_dask` is the way to go

src/anndata/_io/specs/lazy_methods.py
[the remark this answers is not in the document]
flying-sheep: `elem.path` is the relative path inside of the zarr hierarchy, right? converting that to `Path` (a concrete path) makes no sense. It’s probably not too bad, but as said: Please read the intro here: https://docs.python.org/3/library/pathlib.html

flying-sheep:
>I am not sure just allowing kwargs on read_elem_as_dask is the way to go

yeah, I’ve gone away from that. Having a `<thing>_kwargs: SomeTypedDict` parameter works in some cases, but just doing things manually is often best.

src/anndata/_io/specs/lazy_methods.py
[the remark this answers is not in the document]
ilan-gold: @flying-sheep I was just going to ask - why a `PurePosixPath` instead of `PurePath`?  I guess since it is relative?

ilan-gold:
I think I have a clearer idea of how this should work.

src/anndata/_io/specs/lazy_methods.py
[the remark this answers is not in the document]
flying-sheep: No, both POSIX and windows paths can be relative. But virtual paths like this often use `/` as delimiter and are therefore POSIX paths.

ilan-gold:
I actually do not have a clear idea of how to do this.  I didn't realize `read_elem` doesn't take any kwargs - I was confusing that with `read_elem_partial`.  We'd have to change all of the reader functions to accept some sort of `kwargs`.  I could see us going down the path of `dataset_kwargs` for `write_elem` but I don't know if that's really a successful model: https://github.com/scverse/anndata/issues/523 is still an open issue.

ilan-gold:
Adding `dataset_kwargs` also constitutes a breaking change because of `callback` for the experimental `read_dispatched`...although not sure how strice we are about this

flying-sheep:
Doesn’t need to be breaking, we can use `inspect` to check if the callback supports the argument and only pass it in if it does.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,175 @@
...
+def make_index(
+    *,
+    is_csc: Literal[False],
+    stride: int,
+    shape: tuple[int, int],
+    block_id: tuple[int, int],
+) -> tuple[slice]: ...
+
+
+def make_index(
```
ivirshup, line 39: I would rename this. The name sounds like it is building and index as in an index for speeding up look ups. But what the function is doing is making an object that can be used to index in to another array. 

Maybe `make_block_indexer`? Or `_block_indexer`?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,175 @@
...
+    *, is_csc: bool, stride: int, shape: tuple[int, int], block_id: tuple[int, int]
+) -> tuple[slice, slice] | tuple[slice]:
+    index1d = slice(
+        block_id[is_csc] * stride,
+        min((block_id[is_csc] * stride) + stride, shape[0]),
+    )
+    if is_csc:
+        return (slice(None, None, None), index1d)
```
ivirshup, line 47:
```suggestion
        return (slice(None), index1d)
```

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,175 @@
...
+def read_sparse_as_dask(
+    elem: H5Group | ZarrGroup,
+    _reader,
+    dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
+):
+    import dask.array as da
+
+    path_or_group = Path(elem.file.filename) if isinstance(elem, H5Group) else elem
+    elem_name = (
+        elem.name if isinstance(elem, H5Group) else PurePosixPath(elem.path).name
```
ivirshup, line 89: This could be useful wrapped up in a helper. E.g. a function like `elemname(elem) -> str`. It's like `filename` in `file_backing.py`.

ivirshup (commented):
I don't love passing `dataset_kwargs` as the way to specify chunks. `dataset_kwargs` in `write_elem` is generally splatted right into `create_dataset`, while the argument here is not doing that. I think a `read_elem_as_dask` just having an `chunks` keyword argument could be nicer.

src/anndata/tests/helpers.py
```diff
@@ -494,7 +494,7 @@ def assert_equal_cupy(a, b, exact=False, elem_name=None):
 def assert_equal_ndarray(a, b, exact=False, elem_name=None):
     b = asarray(b)
     if not exact and is_numeric_dtype(a) and is_numeric_dtype(b):
-        assert a.shape == b.shape, format_msg(elem_name)
+        assert a.shape == b.shape, (a.shape, b.shape)
```
ivirshup, line 497: I would like to keep the functionality where the format_message is included. Otherwise it can be difficult to figure out which element the error is coming from.

docs/release-notes/0.11.0.md
```diff
@@ -8,6 +8,7 @@
 * Add `should_remove_unused_categories` option to `anndata.settings` to override current behavior.  Default is `True` (i.e., previous behavior).  Please refer to the [documentation](https://anndata.readthedocs.io/en/latest/generated/anndata.settings.html) for usage.  {pr}`1340` {user}`ilan-gold`
 * `scipy.sparse.csr_array` and `scipy.sparse.csc_array` are now supported when constructing `AnnData` objects {pr}`1028` {user}`ilan-gold` {user}`isaac-virshup`
 * Add `should_check_uniqueness` option to `anndata.settings` to override current behavior.  Default is `True` (i.e., previous behavior).  Please refer to the [documentation](https://anndata.readthedocs.io/en/latest/generated/anndata.settings.html) for usage.  {pr}`1507` {user}`ilan-gold`
+* Add `~anndata.experimental.read_elem_as_dask` function to handle i/o with sparse and dense arrays {pr}`1469` {user}`ilan-gold`
```
ivirshup, line 11: `read_elem_as_dask` is currently not exported from `anndata.experimental`

ilan-gold:
> I don't love passing dataset_kwargs as the way to specify chunks. dataset_kwargs in write_elem is generally splatted right into create_dataset, while the argument here is not doing that. I think a read_elem_as_dask just having an chunks keyword argument could be nicer.

The question for me then is (a) whether we are prepared to go forward with an API that might have increasing number of arguments and (b) whether it makes sense to special-case this function in the `Reader` code.  The argument has to go through there and there is currently no mechanism is `Reader.read_elem` for this.

ilan-gold:
The only thing I'm unsure of is the discovery of `dataset_kwargs` in the `callback` since the typing on `callback` suggests it should be all positional https://github.com/scverse/anndata/blob/main/src/anndata/experimental/_dispatch_io.py#L15 i.e., maybe we should just check for the length of the args instead of name?

ivirshup:
> (a) whether we are prepared to go forward with an API that might have increasing number of arguments

So long as it's exported from experimental, yes. I think a `meta` argument would also make sense here.

> So we could leave the dataset_kwargs as-is elsewhere and then make chunks only for read_elem_as_dask

That's what I was thinking

----

> The only thing I'm unsure of is the discovery of dataset_kwargs in the callback

Yeah. I think we can skip thinking about this a bit here since none of these calls are recursive, right?

flying-sheep:
> The only thing I'm unsure of is the discovery of dataset_kwargs in the callback since the typing on callback suggests it should be all positional

It’s simply that Python typing is sadly too primitive to reflect named arguments.

As a rule of thumb, most functions should have 0-2 positional arguments, some 3, and almost none 4.

ilan-gold:
> As a rule of thumb, most functions should have 0-2 positional arguments, some 3, and almost none 4. Ideally that’s enforced by using * (and /)

Right I agree, but my python was marginal when I wrote this.  So there is no real API except that purely positional type hint.  What do we think should be done?

ilan-gold:
I would be in favor of making everything positional except `dataset_kwargs` and then making this very clear in the docs.  I don't consider this breaking because (a) it's experimental and (b) we (I) never made this clear in the first place.

flying-sheep:
maybe define that call protocol and use it. Something like this, but I’m not sure about the `iospec` thing, I think that’s not part of the public API.

```py
from typing import Protocol, TypedDict

class DatasetKwargs(TypedDict, total=False):
    chunks: tuple[int, ...]

def WriterCallback(Protocol):
    def __call__(
        self,
        write_func: Callable[[StorageType, str, Any], None],
        store: GroupStorageType,
        key: str,
        elem: Any,
        *,
        dataset_kwargs: DatasetKwargs = ...,
        iospec: IOSpec = ...,
   ) -> None: ...
```

ilan-gold:
> maybe define that call protocol and use it

Yup

> Something like this, but I’m not sure about the iospec thing, I think that’s not part of the public API.

That's a good point.  We should probably export it from `experimental` then as well.  I can do this in a separate PR along with the protocol then

ilan-gold:
> I just checked, the iospec parameter is part of the public API of the callback.

I though you were referring to the object `IOSpec` itself so people could type-hint and/or inspect its contents to know what to do with it.

flying-sheep:
That too of course! If it‘s part of a public API, it needs to be public.

ilan-gold:
@flying-sheep If you know of a place where we have everything that can be stuck in an `AnnData` object listed, please share.  Typing `write_func` or `read_func` without `Any` would require this, but we can also just leave `Any`.  I'll proceed with `Any` but that might not be such a bad thing to have anyway lying around.

ilan-gold:
> @flying-sheep If you know of a place where we have everything that can be stuck in an AnnData object listed, please share. Typing write_func or read_func without Any would require this, but we can also just leave Any. I'll proceed with Any but that might not be such a bad thing to have anyway lying around.

Maybe not, could possibly get away with generics...

flying-sheep:
If it works, sure! If not, Any is acceptable when gradually adding type info. When you don’t know the type of an existing API, it’s too much to ask to go on a wild goose chase for it.

On the other hand when I’m writing completely new APIs I’m supposed to know the actual type, so if I added Any, it would be out of lazyness or because I wouldn’t know what I’m doing.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,194 @@
...
+        yield path_or_group
+        return
+    file = h5py.File(path_or_group, "r")
+    try:
+        yield file[elem_name]
+    finally:
+        file.close()
+
+
+_DEFAULT_STRIDE = 1000
```
ilan-gold, line 67: Ok I just spent a long time trying to debug why a zarr store was being overfetched, and it turns out it wasn't.  This number is just crazy high.  So I'm not entirely sure what the best way forward is.  This number is (probably) good for bulk tasks like filtering or PCA but absolutely useless for random access.  For example, on a CSC matrix, you end up fetching 5% of the dataset if there are 20,000 genes.  So we need to be VERY careful here.  Not sure what the answer is, actually.  I don't think exposing `stride` is great because it's specific to csr/csc and also a subset of the functionality of `chunks`, but the flipside is that resolving the size of a sparse matrix requires looking at `attrs`.

ilan-gold: Exposing a per-matrix (i.e., `layers`) chunking is tough with a monolithic API https://github.com/scverse/anndata/pull/1247 but a monolithic API (i.e., `read_backed`) makes it easier to deal with telling people how to do things, which is currently very difficult.  I'll need to think about that PR.

ilan-gold: But with that PR, it doesn't really make sense to expose a method called `read_elem_as_xarray` because that's way too specific - only our dataframes work there.  So I'm not sure what the answer is.  Perhaps changing this to `read_lazy`  or something is not a bad idea.

docs/conf.py
```diff
@@ -110,6 +110,7 @@
 suppress_warnings = [
     "ref.citation",
     "myst.header",  # https://github.com/executablebooks/MyST-Parser/issues/262
+    "autosummary.import_cycle",  # https://readthedocs.com/projects/icb-anndata/builds/2349021/
```
ilan-gold, line 113: No idea why this all of a sudden came up

tests/test_io_elementwise.py
```diff
@@ -156,30 +213,113 @@ def test_io_spec_cupy(store, value, encoding_type, as_dask):
...
+
+def test_read_lazy_h5_cluster(sparse_format, tmp_path):
+    import dask.distributed as dd
+
+    file = h5py.File(tmp_path / "test.h5", "w")
+    store = file["/"]
+    arr_store = create_sparse_store(sparse_format, store)
+    X_dask_from_disk = read_elem_as_dask(arr_store["X"])
+    X_from_disk = read_elem(arr_store["X"])
+    file.close()
```
ivirshup, line 280:
```suggestion
    with h5py.File(tmp_path / "test.h5", "w") as file:
        store = file["/"]
        arr_store = create_sparse_store(sparse_format, store)
        X_dask_from_disk = read_elem_as_dask(arr_store["X"])
        X_from_disk = read_elem(arr_store["X"])
```

Should be the same, and more common way to do this

src/anndata/_io/specs/lazy_methods.py
ivirshup: I think this is probably a fine value for now. And agree on not exposing `stride`. Also don't know what the answer is.

ivirshup (commented):
I don't think any `read_elem` methods should be given a `dataset_kwargs` argument. I think that argument has a clear convention for the `write_elem` methods, which is not similar to how it is being used here. It at the very least should have a different name.

I also don't think this PR should touch the callback signature, since you can't even use the callbacks alongside this feature.

What do you think about:

* just not using the dispatch system for `read_as_dask`? There's no need for it to support recursion or callbacks since it can only be used on the sparse and dense arrays.
* Subclassing the `Reader` with `DaskReader` which allows extra arguments to it's methods?

ilan-gold:
>  just not using the dispatch system for read_as_dask? There's no need for it to support recursion or callbacks since it can only be used on the sparse and dense arrays.
> Subclassing the Reader with DaskReader which allows extra arguments to it's methods?

Marvelous idea

tests/test_io_elementwise.py
```diff
@@ -53,6 +64,52 @@ def store(request, tmp_path) -> H5Group | ZarrGroup:
...
+
+    Returns
+    -------
+        A store with a key, `X` that is simply a sparse matrix, and `X_dask` where that same array is wrapped by dask
+    """
+    import dask.array as da
+
+    X = sparse.random(
+        SIZE,
+        SIZE,
```
ivirshup, line 99: I think I remember where we had problems with shape being square. I think we had some tests which were bugged because they were accessing the shape of the wrong dimension to check the expected shape of the result. It doesn't matter if you do it access the wrong dimension when the arrays are square.

Can we just pass a shape argument in here, or use a non-square size?

ivirshup (approved):
This looks great. I'm actually surprised how easily the subclassing seems to have worked out.

I've left one comment, which I think should be addressed but I hope is simple. One more minor nit which is simple, but can be left.

I think `get_elem_name` should live next to `anndata._core.file_backing.filename`. I don't know that `file_backing` is actually the best module for that, and maybe it would fit better in a `util` module. Also, `get_elem_name` seems like an actually implemented version of `AnnData._io.specs.registry.elem_key`. I think we should just delete `elem_key`.

tests/test_io_elementwise.py
ilan-gold: done with non-square shape

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+    path_or_group = Path(filename(elem)) if isinstance(elem, H5Group) else elem
+    elem_name = get_elem_name(elem)
+    shape: tuple[int, int] = tuple(elem.attrs["shape"])
+    dtype = elem["data"].dtype
+    is_csc: bool = elem.attrs["encoding-type"] == "csc_matrix"
+
+    stride: int = _DEFAULT_STRIDE
+    if chunks is not None:
+        if len(chunks) != 2:
+            raise ValueError("`chunks` must be a tuple of two integers")
```
flying-sheep, line 71: why is it `tuple[int, ...]` then and not `tuple[int, int]`?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+        if chunks[int(not is_csc)] != shape[int(not is_csc)]:
```
flying-sheep, line 72: how about

```
major_dim, minor_dim = (1, 0) if is_csc else (0, 1)
```

and using these instead of complex expression?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+            raise ValueError(
+                "Only the major axis can be chunked. "
+                f"Try setting chunks to {((-1, _DEFAULT_STRIDE) if is_csc else (_DEFAULT_STRIDE, -1))}"
+            )
+        stride = chunks[int(is_csc)]
+
+    def make_dask_chunk(
+        block_info: Union[  # noqa: UP007
```
flying-sheep, line 80: why the noqa?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+            None,
+        ] = None,
+    ):
+        # We need to open the file in each task since `dask` cannot share h5py objects when using `dask.distributed`
+        # https://github.com/scverse/anndata/issues/1105
+        if block_info is None:
+            raise ValueError("Block info is required")
+        with maybe_open_h5(path_or_group, elem_name) as f:
+            mtx = ad.experimental.sparse_dataset(f)
+            array_location = block_info[None]["array-location"]
```
flying-sheep, line 94: can’t this be something like

```py
(xxx_start, xxx_end), (yyy_start, yyy_end) = block_info[None]["array-location"]
```

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+        if len(chunks) != 2:
+            raise ValueError("`chunks` must be a tuple of two integers")
+        if chunks[int(not is_csc)] != shape[int(not is_csc)]:
+            raise ValueError(
+                "Only the major axis can be chunked. "
+                f"Try setting chunks to {((-1, _DEFAULT_STRIDE) if is_csc else (_DEFAULT_STRIDE, -1))}"
+            )
+        stride = chunks[int(is_csc)]
+
+    def make_dask_chunk(
```
flying-sheep, line 79: why is this inline? does it access any outer scope variables?

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -0,0 +1,169 @@
...
+
+    path = Path(elem.file.filename)
+    elem_name = elem.name
+    shape = tuple(elem.shape)
+    dtype = elem.dtype
+    chunks: tuple[int, ...] = (
+        chunks if chunks is not None else (_DEFAULT_STRIDE,) * len(shape)
+    )
+
+    def make_dask_chunk(
```
flying-sheep, line 132: same comments as above

flying-sheep (changes requested):

src/anndata/_io/specs/lazy_methods.py
flying-sheep: ah, I see now it does

ilan-gold: Yes `path_or_group`

flying-sheep (approved):

flying-sheep:
OK now I’m done lol

Issue #534, opened 2021-04-03
Calling write_h5ad without converting strings to categoricals


ljjh20:
Hi AnnData team,

It seems I am unable to write to h5ad without having `strings_to_categoricals` be called. 
https://github.com/theislab/anndata/blob/dad9a085afdfb8d23b30cbbbe890acffdae4d621/anndata/_io/h5ad.py#L85

I am using the output h5ad file in Java, not with scanPy, so having my string data encoded as categoricals causes me a lot of problems. Would it be possible to make this optional? 

Thanks!

ivirshup:
Related: https://github.com/theislab/scanpy/issues/1747

I think I'd be fine with a keyword argument for this.

Since it's significantly faster to use the categoricals so we'd probably stick with that for the default. There should be categorical style arrays available in java though, Arrow dictionary encoded arrays for example.

adamgayoso:
I don't see why this should be default, even if it makes things faster in scanpy it can happen at any time (not strictly at saving) and every subsequent call will be a no-op basically

falexwolf:
Suggestion: https://github.com/scverse/anndata/pull/1474

flying-sheep:
Fixed by #1474

PR #1474, opened 2024-04-22
Make `strings_to_categoricals` upon `.write()` optional

falexwolf:
To illustrate that it'd be a simple change, this PR adds the 3 lines for what @ivirshup suggested [here](https://github.com/scverse/anndata/issues/534#issuecomment-813901847) 3 years ago:

> I think I'd be fine with a keyword argument for this.

In the two issues below, @grst, @adamgayoso, @lazappi & @ljjh20 expressed that "forced" sanitization is problematic.

- https://github.com/scverse/anndata/issues/534
- https://github.com/scverse/scanpy/issues/1747

I found these issues only now but also agree that a write function should not mutate the object that's about to be written. @chaichontat made me aware.

Things turn out particularly bad for gene symbols, which should remain strings even for performance reasons and compatibility with JS. Having 30k categories in a 32k dimensional genes vector is detrimental, not beneficial.


lazappi:
Just chipping in to say that while I think this would be great, it doesn't really solve https://github.com/scverse/scanpy/issues/1747 as **scanpy** functions will still cause this conversion (unless something has changed). So this helps if you are just reading/writing `AnnData` but anything that goes through **scanpy** is still likely to be converted to categoricals.

You probably knew this already but wanted to make it clear for anyone reading this later.

falexwolf:
> Just chipping in to say that while I think this would be great

Great that you're also onboard, @lazappi!

> as scanpy functions will still cause this conversion (unless something has changed)

Yes, you're right about the Scanpy plotting functionality. But I think the issue is more pressing for anndata, given its widespread use outside of Scanpy.

ilan-gold (approved):

flying-sheep (approved):

ilan-gold:
@flying-sheep I am disabling auto-merge.  Isaac brought up a good point about whether this should be in `settings` on zulip.  What do you think?  I think it's fine to leave this here because you might have different behavior for different datasets.  If you agree we can merge (two against one haha).

ilan-gold:
@ivirshup your call then.

ilan-gold:
I am fine leaving this as-is @flying-sheep so will enable auto-merge

ilan-gold:
> I think we didn’t fully agree on if this should be an “and” or an “or” for things to go into settings.

I am comfortable with "or" here

flying-sheep:
weird. We actually don’t document `anndata.io.write_{zarr,h5ad}` yet, and this PR didn’t add the API to `anndata.AnnData.write_{zarr,h5ad}`. I think we should fix all that.

PR #1499, opened 2024-05-16
Numpy 2 compatibility

flying-sheep:
- [ ] Closes #
- [x] Tests added

### DIFF
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -115,7 +115,11 @@ def not_missing(v) -> bool:
 # TODO: Hopefully this will stop being an issue in the future and this code can be removed.
 @singledispatch
 def equal(a, b) -> bool:
-    return np.array_equal(a, asarray(b))
+    a = asarray(a)
+    b = asarray(b)
+    if a.ndim == b.ndim == 0:
+        return bool(a == b)
+    return np.array_equal(a, b)
 
 
 @equal.register(pd.DataFrame)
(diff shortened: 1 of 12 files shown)


src/anndata/_core/merge.py
```diff
@@ -115,7 +115,11 @@ def not_missing(v) -> bool:
 # TODO: Hopefully this will stop being an issue in the future and this code can be removed.
 @singledispatch
 def equal(a, b) -> bool:
-    return np.array_equal(a, asarray(b))
+    a = asarray(a)
+    b = asarray(b)
+    if a.ndim == b.ndim == 0:
+        return bool(a == b)
+    return np.array_equal(a, b)
```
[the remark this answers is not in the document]
flying-sheep: @ivirshup introduced this code in https://github.com/scverse/anndata/pull/378, so maybe he should comment

src/anndata/_core/sparse_dataset.py
```diff
@@ -403,7 +403,7 @@ def shape(self) -> tuple[int, int]:
         if shape is None:
             # TODO warn
             shape = self.group.attrs.get("h5sparse_shape")
-        return tuple(shape)
+        return tuple(map(int, shape))
```
ilan-gold, line 406: shape can not have ints?

flying-sheep: they’re `np.int64s`, which isn’t a problem in behavior, but affects the `repr` in numpy 2, which breaks our doctests.

Since we convert the array to a tuple, let’s convert all the way to Python

PR #1509, opened 2024-05-27
(chore): change boolean settings names to use `should`

ilan-gold:
Since this is coming in 0.11 I think we should change these to use `should` verbs at the front.  I like verbs in front of booleans because it makes clear what they do.  Otherwise, I would just close them and leave them both as the action without `should`

- [ ] Closes #
- [ ] Tests added


flying-sheep (approved):
Sure, sounds good to me! Similar idea as having booleans named `is_*` or `has_*`, which I also like.

PR #1514, opened 2024-06-04
(fix): fix for new `scipy.sparse` matrix strings in `sparse_dataset` docstring tests

ilan-gold:
See https://github.com/scverse/anndata/runs/25637727260

```
Expected:
    <100x765 sparse matrix of type '<class 'numpy.float32'>'
        with 25003 stored elements in Compressed Sparse Row format>
Got:
    <Compressed Sparse Row sparse matrix of dtype 'float32'
    	with 25003 stored elements and shape (100, 765)>
```

- [ ] Closes #
- [x] Tests added

### DIFF
--- a/docs/release-notes/0.10.8.md
+++ b/docs/release-notes/0.10.8.md
@@ -5,7 +5,7 @@
 
 * Write out `64bit` indptr when appropriate for {func}`~anndata.experimental.concat_on_disk` {pr}`1493` {user}`ilan-gold`
 * Support for Numpy 2 {pr}`1499` {user}`flying-sheep`
-
+* `xfail` `~anndata.experimental.sparse_dataset` docstring test on account of new `~scipy` version {pr}`1514` {user}`ilan-gold`
 ```{rubric} Documentation
 ```
 
--- a/docs/release-notes/0.10.8.md
+++ b/docs/release-notes/0.10.8.md
@@ -5,7 +5,7 @@
 
 * Write out `64bit` indptr when appropriate for {func}`~anndata.experimental.concat_on_disk` {pr}`1493` {user}`ilan-gold`
 * Support for Numpy 2 {pr}`1499` {user}`flying-sheep`
-* {func}`~pytest.mark.xfail` {func}`~anndata.experimental.sparse_dataset` docstring test on account of new {mod}`scipy` version {pr}`1514` {user}`ilan-gold`
+* Fix {func}`~anndata.experimental.sparse_dataset` docstring test on account of new {mod}`scipy` version {pr}`1514` {user}`ilan-gold`
 ```{rubric} Documentation
 ```
 
--- a/docs/release-notes/0.10.8.md
+++ b/docs/release-notes/0.10.8.md
@@ -5,7 +5,7 @@
 
 * Write out `64bit` indptr when appropriate for {func}`~anndata.experimental.concat_on_disk` {pr}`1493` {user}`ilan-gold`
 * Support for Numpy 2 {pr}`1499` {user}`flying-sheep`
-* `xfail` `~anndata.experimental.sparse_dataset` docstring test on account of new `~scipy` version {pr}`1514` {user}`ilan-gold`
+* {func}`~pytest.xfail` {func}`~anndata.experimental.sparse_dataset` docstring test on account of new {mod}`scipy` version {pr}`1514` {user}`ilan-gold`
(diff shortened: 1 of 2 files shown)


ilan-gold:
Hmm ok you conditionally skip certain tests by custom condition in doctests.  Will try this

docs/release-notes/0.10.8.md
```diff
@@ -5,7 +5,7 @@
 
 * Write out `64bit` indptr when appropriate for {func}`~anndata.experimental.concat_on_disk` {pr}`1493` {user}`ilan-gold`
 * Support for Numpy 2 {pr}`1499` {user}`flying-sheep`
-
+* `xfail` `~anndata.experimental.sparse_dataset` docstring test on account of new `~scipy` version {pr}`1514` {user}`ilan-gold`
```
flying-sheep, line 8:
```suggestion
* {func}`~pytest.xfail` {func}`~anndata.experimental.sparse_dataset` docstring test on account of new {mod}`scipy` version {pr}`1514` {user}`ilan-gold`
```

you need to add roles here to make links. `~` is there to remove leading elements so, ``{func}`~foo.bar` `` becomes a link with the text `bar` while ``{mod}`~scipy` `` is identical to ``{mod}`scipy` ``

src/anndata/_core/sparse_dataset.py
```diff
@@ -607,6 +607,11 @@ def sparse_dataset(group: GroupStorageType) -> CSRDataset | CSCDataset:
     >>> import scanpy as sc
     >>> import h5py
     >>> from anndata.experimental import sparse_dataset, read_elem
+    >>> import scipy
+    >>> import pytest
+    >>> from packaging.version import Version
+    >>> if Version(scipy.__version__) >= Version("1.14.0rc1"):
+    ...     pytest.xfail("scipy.sparse string formatting changed in 1.14.0rc1 and up")
```
flying-sheep, line 614: Isaac said that he doesn‘t like using pytest stuff visibly in docstrings.

You could e.g. use `# doctest: +ELLIPSIS` and keep the parts that stay the same.

flying-sheep (changes requested):

src/anndata/_core/sparse_dataset.py
ilan-gold: I was trying to fiddle around with this but I couldn't figure out how to get pytest to use a custom doctest runner: https://stackoverflow.com/a/55480919/8060591

flying-sheep: Why would that be necessary? you just do this instead of what you’re doing

```diff
->>> X[100:200]  # doctest: +NORMALIZE_WHITESPACE
-<100x765 sparse matrix of type '<class 'numpy.float32'>'
-    with 25003 stored elements in Compressed Sparse Row format>
+>>> X[100:200]  # doctest: +ELLIPSIS
+<...sparse matrix of...float32...with 25003 stored elements...>
```

ilan-gold: Well it would allow us to match exactly if I could get it work.  But this is also fine.

flying-sheep:
Huh, I don’t understand why that works. I thought the new format looks like this:

```
    <Compressed Sparse Row sparse matrix of dtype 'float32'
    	with 25003 stored elements and shape (100, 765)>
```

so why does this match? Is the test job no longer running with the affected scipy version?

```
<...100...765...sparse matrix of type...float32...Compressed Sparse Row...>
```

flying-sheep (changes requested):
Yeah, scipy 1.14.0rc1 isn’t used in the current run, so this probably doesn’t work.

ilan-gold:
They revoked the version due to mac wheels breaking, so I didn't get a chance to test.  And as for why it works/doesn't I understood ellipsis as _any_ substring match https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS but now that I'm re-reading it, that meant that the ellipsis itself matches and not the text content.  To some degree, this brings us back to the original problem.  How exactly should we test this if things need to be in order?

flying-sheep:
I suggested this pattern:

> `<...sparse matrix of...float32...with 25003 stored elements...>`

which only contains parts that are in the same order in both versions, i.e. the patterns should match both the old and the new repr.

It doesn’t contain shape or CSC/CSR format, since those flip sides

flying-sheep (approved):

PR #1515, opened 2024-06-06
Fix settings

flying-sheep:
This removes the class attributes and replaces them with instance attributes.

Previously, doing anything with a `SettingsManager` instance would modify the class.

I also modified the tests to use an instance, so `.deprecate` doesn’t persist after a test. Tests should be without side effects.

### How to review

without whitespace changes: https://github.com/scverse/anndata/pull/1515/files?w=1

I removed the superfluous `monkeypatch.scope`: Monkey patches get reset at the end of each test, so unless you need to reset them before the end of the test, no need to use scopes.

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -32,73 +32,68 @@ def validate_int_list(val) -> bool:
     return True
 
 
-settings = SettingsManager()
-settings.register(option, default_val, description, validate_bool)
-
-settings.register(option_2, default_val_2, description_2, validate_bool)
-
-settings.register(
-    option_3,
-    default_val_3,
-    description_3,
-    validate_int_list,
-    type_3,
-)
+@pytest.fixture
+def settings() -> SettingsManager:
+    settings = SettingsManager()
+    settings.register(option, default_val, description, validate_bool)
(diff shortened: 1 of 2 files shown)


tests/test_settings.py
```diff
@@ -32,73 +32,68 @@ def validate_int_list(val) -> bool:
...
-        option_env = "test_var_env"
-        default_val_env = False
-        description_env = "My doc string env!"
-        option_env_var = "ANNDATA_" + option_env.upper()
-        mp.setenv(option_env_var, "b")
-
-        class TestEnum(Enum):
-            a = False
-            b = True
+def test_register_with_env(settings: SettingsManager, monkeypatch: pytest.MonkeyPatch):
```
ilan-gold, line 49: Do you know if `monkeypatch` is correctly scoped to remove the patch once the test is torn down (i.e., no need for context)?  I did this to be safe because I couldn't find anything definite, although I suspect that it is correctly scoped (I just found this blog post: https://holgerkrekel.net/2009/03/03/monkeypatching-in-unit-tests-done-right/ but couldn't find anything in the docs, so I'm sure it's fine).  This is much cleaner

ilan-gold (approved):

PR #1523, opened 2024-06-11
Fix sparse dataset 2D slicing

### LINKED ISSUE #1520: 2D indexing (no “:”) in `sparse_dataset` is not lazy
https://github.com/scverse/anndata/blob/b78a8056008eac5732379f9024c3b46d2b7d5d3f/src/anndata/_core/sparse_dataset.py#L179

calls `self._get_submatrix(...)`, which calls [`scipy.sparse.sparsetools.get_csr_submatrix(...)`](https://github.com/scipy/scipy/blob/e1fecdfb0eb4f1bc01345b0a39930ab1cc94e1da/scipy/sparse/sparsetools/csr.h#L1181).

And that one is implemented in C++, so it converts all of `self.indices` and `self.data` into arrays on each slice access.

This eliminates the whole purpose of `sparse_dataset` …

flying-sheep:
- [x] Closes #1520
- [x] Tests added


tests/test_backed_sparse.py
```diff
@@ -113,9 +115,28 @@ def test_backed_indexing(
...
+
+    __array__orig = cls.__array__
+
+    def __array__(self, *args) -> np.ndarray:
+        arr = __array__orig(self, *args)
+        assert len(arr) < len(csr_mem.X.data)
+        return arr
+
+    monkeypatch.setattr(cls, "__array__", __array__)
+
```
ilan-gold, line 137: At the expense of losing the H5 test (which I personally think is fine), we have https://github.com/scverse/anndata/blob/97a17923f6c9207203e1bd13719cfd90ebd6e354/src/anndata/tests/helpers.py#L858-L883 for tracking what is accessed.  IMO it's a bit simpler than this since you can track chunks explicitly.  You could probably just add a case to `test_data_access` to check that this slice x slice works as expected.

src/anndata/_core/sparse_dataset.py
```diff
@@ -181,7 +181,11 @@ def _get_sliceXslice(self, row: slice, col: slice) -> ss.csr_matrix:
                     self._get_contiguous_compressed_slice(row), shape=out_shape
                 )
             return self._get_arrayXslice(np.arange(*row.indices(self.shape[0])), col)
-        return super()._get_sliceXslice(row, col)
+
+        # This is better than super()._get_sliceXslice,
+        # as that loads the complete .data and .indices arrays into memory
+        # TODO: Can probably still be more efficient
```
ilan-gold, line 187: I don't think this TODO is meaningful.  We're not going to be writing C code to do this, and I don't think it would be a huge improvement anyway.

src/anndata/_core/sparse_dataset.py
```diff
@@ -181,7 +181,11 @@ def _get_sliceXslice(self, row: slice, col: slice) -> ss.csr_matrix:
...
+        return self[row, :][:, col]
```
ilan-gold, line 188: I can't re-write the function in a suggestion but I would probably change the `elif` block above to be an `else` block and then `res = ss.csr_matrix(...` which could be indexed by `col` afterwards if `out_shape[1] < shape[1]`

tests/test_backed_sparse.py
flying-sheep: thanks for the pointer!

src/anndata/_core/sparse_dataset.py
flying-sheep: I think so. Let’s see if what I think you meant works

flying-sheep: Seems like it!

tests/test_backed_sparse.py
ilan-gold: Why keep this around?

tests/test_backed_sparse.py
```diff
@@ -318,12 +345,69 @@ def test_indptr_cache(
...
+        id_ = "-".join(map(idify, [idx_maj_raw, idx_min, maj_kind]))
+        yield pytest.param(idx_maj, idx_min, exp, id=id_)
+
+
 @pytest.mark.parametrize("sparse_format", [sparse.csr_matrix, sparse.csc_matrix])
+@pytest.mark.parametrize(
+    ("idx_maj", "idx_min", "exp"),
+    width_idx_kinds(
+        (
+            [0],
```
ilan-gold, line 391: Probably should add some cases where `len(idx) > 1` no for `mk_idx_kind` so that we test the behavior there as well?

tests/test_backed_sparse.py
```diff
@@ -318,12 +345,69 @@ def test_indptr_cache(
     assert store.get_access_count("X/indptr") == 2
 
 
+Kind = Literal["slice", "int", "array", "mask"]
+
+
+def mk_idx_kind(idx: Sequence[int], *, kind: Kind, l: int) -> Idx | None:
```
ilan-gold, line 351: Maybe a doc string here?  Just a description would be good to help reason about what is going on

ilan-gold (commented):
Just some cleanup stuff.  Maybe another test since you wrote all that code?

tests/test_backed_sparse.py
ilan-gold: Sorry, consecutive integers, not start-stop pair

flying-sheep:
You mean another test parameter set or something else utilizing the helpers?

ilan-gold:
> You mean another test parameter set or something else utilizing the helpers?

Yes, something besides `[0]` as the `idx`

flying-sheep:
like in fce1d678f36f8e9b30ca0c6549e17c9bc7a02aff?

ilan-gold (approved):

PR #1532, opened 2024-06-18
(fix): allow setting `X` as sparse when `X` is dense view

### LINKED ISSUE #768: Cannot set dense .X attribute with sparse matrix when adata is a view
It is not possible to set the `.X` attribute when `adata` is a view. The resulting error does not hint at this and I would expect anndata to handle this in the background or  give some sort of warining. It is related to this part of the code in [anndata/_core/anndata.py].(https://github.com/scverse/anndata/blob/master/anndata/_core/anndata.py#L683-L688)

## Minimal reproducible example
```python
import anndata
import numpy
from scipy.sparse import csr_matrix

x = np.zeros((100, 30))
adata = anndata.AnnData(x)[:30]
adata.X = csr_matrix(x[:30])
```
while `adata.X = x[:30]` works just fine. 

## Traceback

<details>

```pytb
TypeError: float() argument must be a string or a number, not 'csr_matrix'

/tmp/ipykernel_8909/1816563358.py in <module>
      5 x = np.zeros((100, 30))
      6 _adata = anndata.AnnData(x)[:30]
----> 7 _adata.X = csr_matrix(x[:30])

~/miniconda3/envs/scib-pipeline-R4/lib/python3.7/site-packages/anndata/_core/anndata.py in X(self, value)
    686                     ):
    687                         value = sparse.coo_matrix(value)
--> 688                     self._adata_ref._X[oidx, vidx] = value
    689                 else:
    690                     self._X = value

ValueError: setting an array element with a sequence.
```
</details>

_____

with @michalk8.

ilan-gold:
- [x] Closes #768
- [x] Tests added

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -662,6 +662,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):
                         else:
                             memory_class = sparse.coo_matrix
                         value = memory_class(value)
+                    elif sparse.issparse(value) and isinstance(
+                        self._adata_ref._X, np.ndarray
+                    ):
+                        value = value.toarray()
                     self._adata_ref._X[oidx, vidx] = value
                 else:
                     self._X = value
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -9,15 +9,10 @@ import warnings
 from collections import OrderedDict
 from collections.abc import Iterable, Mapping, MutableMapping, Sequence
 from copy import copy, deepcopy
-from enum import Enum
 from functools import partial
 from pathlib import Path
 from textwrap import dedent
-from typing import (  # Meta  # Generic ABCs  # Generic
-    TYPE_CHECKING,
-    Any,
-    Literal,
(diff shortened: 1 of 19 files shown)


src/anndata/_core/anndata.py
```diff
@@ -662,6 +662,10 @@ def X(self, value: np.ndarray | sparse.spmatrix | SpArray | None):
                         else:
                             memory_class = sparse.coo_matrix
                         value = memory_class(value)
+                    elif sparse.issparse(value) and isinstance(
+                        self._adata_ref._X, np.ndarray
+                    ):
+                        value = value.toarray()
```
ilan-gold, line 668: We definitely could warn here, but on the other hand, we don't warn in the above case when we sparsify from dense (which entails a >2x memory increase)

flying-sheep (approved):
maybe we should warn in both cases. Your call.

Did you mean to leave this as draft?

ilan-gold:
@flying-sheep I guess I just wasn't even sure this was the desired result but your approval says that we are on the right track.  I will add a warning.

PR #1516, opened 2024-06-06
Coerce np.matrix instances to np.ndarray

flying-sheep:
- [x] Closes #1431
- [x] Tests added
- [x] Release note added (or unnecessary)

This unifies the array type coercion, which used to work differently betwen `X` and `AlignedMapping`s.

@ilan-gold, does it make sense to add AwkwardArray there or should we special-case it so it’s only valid in {obs,var}m, not in X and layers?

PR #1537, opened 2024-06-21
Fix assigning and initializing with h5py datasets

flying-sheep:
Fixes an unreleased bug introduced in #1516

AnnData’s `StorageType` enum doesn’t contain `h5py.Dataset`, but in scanpy’s tests we set `adata.obsm["thing"] = h5py_file["some_dataset"]`
My path unified handling of `AlignedArray`s and `X` so now that “anndata dev” test job in Scanpy throws an error.

How should we do this with backed?
- Should we just allow putting `Dataset`s everywhere without making `AnnData` formally `isbacked`?
- Should we make it `isbacked` if `X` happens to be a `Dataset` with the correct path for backed mode?
- Should we disallow setting `X` specifically to `Dataset` and throw an error that one should use `adata.filename = ...` for that?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.9
+    rev: v0.4.10
     hooks:
       - id: ruff
         types_or: [python, pyi, jupyter]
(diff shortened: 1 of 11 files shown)


ilan-gold (approved):
To me, a layman when it comes to `backed` mode, this seems reasonable.

I don't see anything in `AnnData` that would warrant a test (everything looks like it is sectioned off via `self.isbacked`) but just want to be sure other operations works.  I think a verbal confirmation of some testing in a shell would even be fine.  Things like setting, deleting, and writing `X`.

ilan-gold:
I'm also going to do a fix for that test.

PR #1557, opened 2024-07-05
(feat): add `callback` typing for `{read,write}_dispatched`

ilan-gold:
- [ ] See https://github.com/scverse/anndata/pull/1469#issuecomment-2208412624
- [ ] Tests added

### DIFF
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -48,8 +48,7 @@ if TYPE_CHECKING:
         ArrayStorageType,
         GroupStorageType,
         InMemoryArrayOrScalarType,
-        RWAbleDict,
-        RWAbleList,
+        RWAble,
     )
     from anndata.compat import SpArray
 
@@ -332,7 +331,7 @@ def write_raw(
 
 @_REGISTRY.register_read(H5Group, IOSpec("dict", "0.1.0"))
 @_REGISTRY.register_read(ZarrGroup, IOSpec("dict", "0.1.0"))
-def read_mapping(elem: GroupStorageType, *, _reader: Reader) -> RWAbleDict:
+def read_mapping(elem: GroupStorageType, *, _reader: Reader) -> dict[str, RWAble]:
     return {k: _reader.read_elem(v) for k, v in elem.items()}
 
 
@@ -341,7 +340,7 @@ def read_mapping(elem: GroupStorageType, *, _reader: Reader) -> RWAbleDict:
 def write_mapping(
     f: GroupStorageType,
     k: str,
-    v: RWAbleDict,
+    v: dict[str, RWAble],
     *,
     _writer: Writer,
     dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
@@ -361,7 +360,7 @@ def write_mapping(
 def write_list(
     f: GroupStorageType,
     k: str,
-    elem: RWAbleList,
+    elem: list[RWAble],
     *,
     _writer: Writer,
     dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -102,7 +102,7 @@ def _to_cpu_mem_wrapper(write_func):
         cupy_val: CupyArray | CupyCSCMatrix | CupyCSRMatrix,
         *,
         _writer: Writer,
-        dataset_kwargs=MappingProxyType,
+        dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
     ):
         return write_func(
             f, k, cupy_val.get(), _writer=_writer, dataset_kwargs=dataset_kwargs
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -4,7 +4,7 @@ from collections.abc import Mapping
 from functools import partial
 from itertools import product
 from types import MappingProxyType
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
 from warnings import warn
 
 import h5py
@@ -26,6 +26,7 @@ from anndata.compat import (
     CupyCSCMatrix,
     CupyCSRMatrix,
     DaskArray,
+    SpArray,
     ZarrArray,
     ZarrGroup,
     _decode_structured_array,
@@ -33,10 +34,12 @@ from anndata.compat import (
     _read_attr,
 )
 
-from .registry import _REGISTRY, IOSpec, read_elem, read_elem_partial
+from .registry import _REGISTRY, IOSpec, Reader, read_elem, read_elem_partial
 
 if TYPE_CHECKING:
     from os import PathLike
+
+    from anndata._core.storage import StorageType
 
 H5Array = h5py.Dataset
 H5Group = h5py.Group
@@ -109,7 +112,9 @@ def _to_cpu_mem_wrapper(write_func):
 @_REGISTRY.register_read(H5File, IOSpec("", ""))
 @_REGISTRY.register_read(H5Group, IOSpec("", ""))
 @_REGISTRY.register_read(H5Array, IOSpec("", ""))
-def read_basic(elem, _reader):
+def read_basic(
+    elem: StorageType, _reader: Reader
+) -> dict | np.ndarray[Any, Any] | np.ndarray | sparse.spmatrix | SpArray:
     from anndata._io import h5ad
 
     warn(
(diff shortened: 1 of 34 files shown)


ilan-gold:
This is a first-pass, and if it looks good, I would do `write` as well

src/anndata/_io/specs/methods.py
```diff
@@ -109,7 +112,9 @@ def wrapper(
 @_REGISTRY.register_read(H5File, IOSpec("", ""))
 @_REGISTRY.register_read(H5Group, IOSpec("", ""))
 @_REGISTRY.register_read(H5Array, IOSpec("", ""))
-def read_basic(elem, _reader):
+def read_basic(
+    elem: StorageType, _reader: Reader
+) -> dict | np.ndarray[Any, Any] | np.ndarray | sparse.spmatrix | SpArray:
```
ilan-gold, line 117: Not sure about the meaningfulness of `np.ndarray[Any, Any] | np.ndarray` and the like - need to look into it

flying-sheep:
really cool! At the moment, having all of `reader`, `Reader`, `read_callback` and `read_func: Callable[[StorageType, Reader], InMemoryType]` is really confusing, but that can probably only be mitigated a little bit by changing the names: most likely our abstractions are just confusing.

ilan-gold:
> really cool! At the moment, having all of reader, Reader, read_callback and read_func: Callable[[StorageType, Reader], InMemoryType] is really confusing, but that can probably only be mitigated a little bit by changing the names: most likely our abstractions are just confusing.

I totally agree.

ilan-gold:
Made a best effort to remove some the `Any` but it's tough to keep track of both what is read in by our functions right now and _could_ be read in because `AnnData` accepts it as an `elem` i.e., via direct setting.  So, `cupy.ndarray` comes to mind, for example.

flying-sheep:
Protocols are usually `PascalCase`-named, right? Or is the `snake_case` a convention I don’t know about?

flying-sheep:
Two thoughts:

- maybe the types should go into their own table? they’re crowding the actual list of API functions
- the “The callback has the following signature: […]” text is now redundant with the callback type

ilan-gold:
> maybe the types should go into their own table? they’re crowding the actual list of API functions

Was just about to comment the same

ilan-gold:
> maybe the types should go into their own table? they’re crowding the actual list of API functions

This could cause a circular import though because the `Protocols` use the classes as types...

ilan-gold:
Ah, no, should be good in a type checking block

flying-sheep:
hm, maybe `scanpydoc`’s `qualname_overrides` doesn’t work with generics syntax: https://github.com/theislab/scanpydoc/blob/b0fe50f1c42ea1e79f9744fa5e79ebf3b4d53474/src/scanpydoc/elegant_typehints/_formatting.py#L50

ilan-gold:
Hmm, not sure.  Let me push and you can have a look.  It seems the generics are indeed breaking the docs but I'm not sure how.

ilan-gold:
To be clear, the problem seems to be that generics break the link to `Read`/`Write` - without generics, no issue

flying-sheep:
OK, I have a fix in scanpydoc: https://github.com/theislab/scanpydoc/pull/158

flying-sheep:
One more thought: with this in place, we can also separate the internal callback API (with `_reader`/`_writer` parameters) from the public API.

ilan-gold:
> One more thought: with this in place, we can also separate the internal callback API (with _reader/_writer parameters) from the public API.

I am not sure I see it.  The `read_func` and `write_func` callbacks that `{Write,Read}Callback` take are based on those "internal callback API" functions if I understand you correctly.

src/anndata/_io/specs/methods.py
```diff
@@ -110,7 +121,9 @@ def wrapper(
 @_REGISTRY.register_read(H5File, IOSpec("", ""))
 @_REGISTRY.register_read(H5Group, IOSpec("", ""))
 @_REGISTRY.register_read(H5Array, IOSpec("", ""))
-def read_basic(elem, _reader):
+def read_basic(
+    elem: H5File | H5Group | H5Array, *, _reader: Reader
```
flying-sheep, line 125: I think `h5py.File` is a subclass of `h5py.Group` so technically there’s no need to specify it in any read signature, but we can just leave it in of course.

src/anndata/_io/specs/methods.py
```diff
@@ -301,13 +330,22 @@ def write_raw(f, k, raw, _writer, dataset_kwargs=MappingProxyType({})):
 
 @_REGISTRY.register_read(H5Group, IOSpec("dict", "0.1.0"))
 @_REGISTRY.register_read(ZarrGroup, IOSpec("dict", "0.1.0"))
-def read_mapping(elem, _reader):
+def read_mapping(
+    elem: GroupStorageType, *, _reader: Reader
+) -> dict[str, InMemoryArrayOrScalarType]:
```
flying-sheep, line 335: could also be recursive, right?

Maybe something like

```py
RWAble = InMemoryArrayOrScalarType | "RWAbleDict" | "RWAbleList"
RWAbleDict = dict[str, RWAble]
RWAbleList = list[RWAble]
```

src/anndata/_io/specs/methods.py
```diff
@@ -301,13 +330,22 @@ def write_raw(f, k, raw, _writer, dataset_kwargs=MappingProxyType({})):
...
     return {k: _reader.read_elem(v) for k, v in elem.items()}
 
 
 @_REGISTRY.register_write(H5Group, dict, IOSpec("dict", "0.1.0"))
 @_REGISTRY.register_write(ZarrGroup, dict, IOSpec("dict", "0.1.0"))
-def write_mapping(f, k, v, _writer, dataset_kwargs=MappingProxyType({})):
+def write_mapping(
+    f: GroupStorageType,
+    k: str,
+    v: dict,
```
flying-sheep, line 344: this should be the same type as above: dict from str to everything we can read/write including itself.

src/anndata/_io/specs/methods.py
```diff
@@ -320,7 +358,14 @@ def write_mapping(f, k, v, _writer, dataset_kwargs=MappingProxyType({})):
 
 @_REGISTRY.register_write(H5Group, list, IOSpec("array", "0.2.0"))
 @_REGISTRY.register_write(ZarrGroup, list, IOSpec("array", "0.2.0"))
-def write_list(f, k, elem, _writer, dataset_kwargs=MappingProxyType({})):
+def write_list(
+    f: GroupStorageType,
+    k: str,
+    elem: list,
```
flying-sheep, line 364: similar situation

flying-sheep (approved):
Looks great! Some optional things I noticed, but nothing major.

The “Classes that implement this protocol” line is confusing here (callbacks are usually not defined as classes), and maybe we should configure `more_autodoc` to not show it.

One thing that I fixed: `MappingProxyType` is the type we use if we want an immutable dict, i.e. an instance of `Mapping` type that complains if we accidentally try to mutate it. So just like `dict` is almost never used as a parameter signature¹ in favor of `MutableMapping` (or more often just `Mapping`), `MappingProxyType` is a concrete type that almost never appears in a parameter signature.

¹except in cases like `write_dict` because we literally only support serializing `dict` instances, no other mappings

flying-sheep:
Oof, the `Covariant` and so on variants broke the TypeAlias links. I almost missed that, but fixed it in 3a332ade227b5d945125b60d69dd34e15a0f7bd9

src/anndata/_io/specs/methods.py
ilan-gold: I actually don't know.  Can something like `obsm` be recursive?

ilan-gold: Screenshot 2024-07-11 at 11 33 58

ilan-gold: I guess this can be used for `uns` as well...

ilan-gold:
@flying-sheep I suspect the recursion is breaking the docs - would you agree https://readthedocs.com/projects/icb-anndata/builds/2342598/?

ilan-gold:
That being said, this is much simpler! Super clean!

flying-sheep:
> I suspect the recursion is breaking the docs

yeah, I think so! Maybe we should just revert that commit and deal with that in a subsequent PR

ilan-gold:
Ok this is a `scanpydoc` issue again, I think: https://readthedocs.com/projects/icb-anndata/builds/2342598/

```python
File "/home/docs/checkouts/readthedocs.org/user_builds/icb-anndata/envs/1557/lib/python3.12/site-packages/scanpydoc/rtd_github_links/__init__.py", line 187, in _module_path
    file = Path(inspect.getabsfile(obj))
...
TypeError: module, class, method, function, traceback, frame, or code object was expected, got GenericAlias
```

flying-sheep:
I can deal with that!

flying-sheep:
This should do the trick: https://pypi.org/project/scanpydoc/0.13.6/

ilan-gold:
The current error seems to be caused by the docstring of `builtins.dict` itself. We could just change to `MutableMapping` or `Mapping` to avoid this but that feels somewhat odd

flying-sheep:
I figured out how to actually fix it!

flying-sheep (approved):
OK, there we go! Maybe check how the summary and the individual docs render, but I like it!

ilan-gold:
Weird that `write_dispatched` has the link to `RWAble` but `InMemoryElem` does not

Screenshot 2024-07-11 at 16 46 03
Screenshot 2024-07-11 at 16 46 17

ilan-gold:
Hmm seems like this has been this way for a few commits.  I would be fine leaving this TBH, we have gone very far with this.

flying-sheep:
Ah, it’s all so many messy systems interacting. InMemoryElem is a TypeAlias, so I think Sphinx does that one, while sphinx_autodoc_typehints does parameter types.

I think we spent long enough with this, thank you for tackling it! I think this is a very valuable step towards properly typechecking anndata!

PR #1558, opened 2024-07-05
IO for nullable string arrays

### LINKED ISSUE #679: Nullable string columns
Split off from #504

It would be nice to have support for nullable string arrays. It would be good to have a consistent in-memory representation for these so we can reason about performance. However, this does not currently exist in our dependency stack. I currently think this feature will be dependent on upstream developments in pandas StringArray type.

This is less urgent than nullable integers and booleans since we already have nullable categorical arrays, and currently aggressively cast strings to categorical for performance reasons.

flying-sheep:
- [x] Closes #679
- [x] Tests added

What to use as backend?

- Numcodecs talks about dataframe stuff here: https://github.com/zarr-developers/numcodecs/issues/452


src/anndata/_io/specs/methods.py
```diff
@@ -395,11 +405,23 @@ def read_string_array(d, _reader):
...
+def read_string_array_nullable(d, _reader):
+    return pd.array(read_array(d.asstr(), _reader=_reader), dtype="string")
+
+
 @_REGISTRY.register_read_partial(H5Array, IOSpec("string-array", "0.2.0"))
 def read_string_array_partial(d, items=None, indices=slice(None)):
     return read_array_partial(d.asstr(), items=items, indices=indices)
 
 
+@_REGISTRY.register_read_partial(H5Array, IOSpec("string-array-nullable", "0.1.0"))
```
ilan-gold, line 418: Probably best to stop adding things to `read_partial`

flying-sheep: I removed them again

ilan-gold:
@flying-sheep Didn't we want this to be opt-in? (see https://github.com/scverse/anndata/issues/679#issuecomment-2276177688)

ilan-gold (changes requested):
Didn't we want this as opt-in? https://github.com/scverse/anndata/issues/679#issuecomment-2276177688

flying-sheep:
yup, you’re right, I should have requested a review after adding that part

src/anndata/_io/specs/methods.py
```diff
@@ -1028,30 +1034,82 @@ def write_nullable_integer(
...
+)
+def write_nullable_string(
+    f: GroupStorageType,
+    k: str,
+    v: pd.arrays.StringArray,
+    *,
+    _writer: Writer,
+    dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
+):
+    from ..._settings import settings
```
ilan-gold, line 1051: Why import here instead of at the top?

src/anndata/_io/specs/methods.py
```diff
@@ -1028,30 +1034,82 @@ def write_nullable_integer(
...
+        msg = (
+            "`anndata.settings.allow_write_nullable_strings` is False, "
+            "because writing of `pd.arrays.StringArray` is new "
+            "and not supported in anndata < 0.11, still use by many people. "
+            "Opt-in to writing these arrays by toggling the setting to True."
+        )
+        raise RuntimeError(msg)
+    g = f.require_group(k)
+    if (mask := v.isna()).any():
+        _writer.write_elem(g, "mask", mask, dataset_kwargs=dataset_kwargs)
```
ilan-gold, line 1063: If there are no `NA` values, maybe we should give an `ImplicitModificationWarning` here and write out the whole thing as a normal string array?  Not sure.  For `zarr` at least, the semantics around missing arrays for object stores are not amazing.  I have seen 403s, 404s and now this:

```python
In [3]: zarr.open("https://storage.googleapis.com/vitessce-demo-data/maynard-2021/151673.sdata.zarr/images/visium_151673_annotations")
Out[3]: <zarr.hierarchy.Group '/'>

In [4]: zarr.open("https://storage.googleapis.com/vitessce-demo-data/maynard-2021/151673.sdata.zarr/images/visium_151673_annotations/0")
Out[4]: <zarr.core.Array (7, 600, 600) float64>

In [5]: zarr.open("https://storage.googleapis.com/vitessce-demo-data/maynard-2021/151673.sdata.zarr/images/visium_151673_annotations/foo")
NotImplementedError                       Traceback (most recent call last)
Cell In[5], line 1
----> 1 zarr.open("https://storage.googleapis.com/vitessce-demo-data/maynard-2021/151673.sdata.zarr/images/visium_151673_annotations/foo")

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/zarr/convenience.py:129, in open(store, mode, zarr_version, path, **kwargs)
    127         return open_array(_store, mode=mode, **kwargs)
    128     else:
--> 129         return open_group(_store, mode=mode, **kwargs)
    131 else:
    132     if contains_array(_store, path):

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/zarr/hierarchy.py:1587, in open_group(store, mode, cache_attrs, synchronizer, path, chunk_store, storage_options, zarr_version, meta_array)
   1585         if contains_array(store, path=path):
   1586             raise ContainsArrayError(path)
-> 1587         init_group(store, path=path, chunk_store=chunk_store)
   1589 elif mode in ["w-", "x"]:
   1590     if contains_array(store, path=path):

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/zarr/storage.py:682, in init_group(store, overwrite, path, chunk_store)
    679     store["zarr.json"] = store._metadata_class.encode_hierarchy_metadata(None)  # type: ignore
    681 # initialise metadata
--> 682 _init_group_metadata(store=store, overwrite=overwrite, path=path, chunk_store=chunk_store)
    684 if store_version == 3:
    685     # TODO: Should initializing a v3 group also create a corresponding
    686     #       empty folder under data/root/? I think probably not until there
    687     #       is actual data written there.
    688     pass

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/zarr/storage.py:743, in _init_group_metadata(store, overwrite, path, chunk_store)
    741 key = _prefix_to_group_key(store, _path_to_prefix(path))
    742 if hasattr(store, "_metadata_class"):
--> 743     store[key] = store._metadata_class.encode_group_metadata(meta)
    744 else:
    745     store[key] = encode_group_metadata(meta)

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/zarr/storage.py:1470, in FSStore.__setitem__(self, key, value)
   1468     if self.fs.isdir(path):
   1469         self.fs.rm(path, recursive=True)
-> 1470     self.map[key] = value
   1471     self.fs.invalidate_cache(self.fs._parent(path))
   1472 except self.exceptions as e:

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/fsspec/mapping.py:175, in FSMap.__setitem__(self, key, value)
    173 key = self._key_to_str(key)
    174 self.fs.mkdirs(self.fs._parent(key), exist_ok=True)
--> 175 self.fs.pipe_file(key, maybe_convert(value))

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/fsspec/asyn.py:118, in sync_wrapper.<locals>.wrapper(*args, **kwargs)
    115 @functools.wraps(func)
    116 def wrapper(*args, **kwargs):
    117     self = obj or args[0]
--> 118     return sync(self.loop, func, *args, **kwargs)

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/fsspec/asyn.py:103, in sync(loop, func, timeout, *args, **kwargs)
    101     raise FSTimeoutError from return_result
    102 elif isinstance(return_result, BaseException):
--> 103     raise return_result
    104 else:
    105     return return_result

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/fsspec/asyn.py:56, in _runner(event, coro, result, timeout)
     54     coro = asyncio.wait_for(coro, timeout=timeout)
     55 try:
---> 56     result[0] = await coro
     57 except Exception as ex:
     58     result[0] = ex

File ~/Projects/Theis/anndata/venv/lib/python3.11/site-packages/fsspec/asyn.py:408, in AsyncFileSystem._pipe_file(self, path, value, **kwargs)
    407 async def _pipe_file(self, path, value, **kwargs):
--> 408     raise NotImplementedError

NotImplementedError:
```

ilan-gold: I see that this is similar to some business about `mask` being present or not for `nullable-{boolean,integer}` (so maybe we would never run into this because we don't open `mask` directly) but this feels different than other cases of masked values since we are checking for any NAs here.  Could you explain a bit maybe what is going on there and how it is similar/different to this?

src/anndata/_io/specs/methods.py
```diff
@@ -1028,30 +1034,82 @@ def write_nullable_integer(
...
 def read_nullable_boolean(
     elem: GroupStorageType, *, _reader: Reader
 ) -> pd.api.extensions.ExtensionArray:
-    if "mask" in elem:
-        return pd.arrays.BooleanArray(
-            _reader.read_elem(elem["values"]), mask=_reader.read_elem(elem["mask"])
-        )
-    else:
-        return pd.array(_reader.read_elem(elem["values"]))
+    return _read_nullable(elem, _reader=_reader, array_type=pd.arrays.BooleanArray)
```
ilan-gold, line 1082: I think we have semantics for this sort of repeated implementation

https://github.com/scverse/anndata/blob/main/src/anndata/_io/specs/methods.py#L1120-L1121

So maybe just adding a `partial` here to specify the `array_type` would do it?

ilan-gold: this would prevent us from re-implementing the same function signature a bunch

src/anndata/_io/specs/methods.py
```diff
@@ -1028,30 +1034,82 @@ def write_nullable_integer(
...
+    _reader: Reader,
+    # BaseMaskedArray
+    array_type: Callable[
+        [NDArray[np.number], NDArray[np.bool_]], pd.api.extensions.ExtensionArray
+    ],
+) -> pd.api.extensions.ExtensionArray:
+    values = _reader.read_elem(elem["values"])
+    if "mask" not in elem:
+        return pd.array(values)
+    return array_type(values, mask=_reader.read_elem(elem["mask"]))
```
ilan-gold, line 1112: Would really like to understand what is going on when you have no `mask` i.e., how this could occur

src/anndata/tests/helpers.py
```diff
@@ -82,30 +98,77 @@ def gen_vstr_recarray(m, n, dtype=None):
...
+        return issubclass(a, b)
+    if isinstance(a, pd.api.extensions.ExtensionDtype):
+        return isinstance(a, b)
+    try:
+        return np.issubdtype(a, b)
+    except TypeError:
+        pytest.fail(f"issubdtype can’t handle everything yet: {a} {b}")
+
+
+def gen_random_column(
```
ilan-gold, line 114: Can we change the order of this function so that it goes

1. check the non-`n_bits` dtypes first
2. get the bits without an `AttributeError` (since presumably that exception was handled above in 1.)
3. Check the rest of the dtypes

ilan-gold (changes requested):

src/anndata/_io/specs/methods.py
flying-sheep: > I see that this is similar to some business about mask being present or not

it’s exactly identical, I reuse that code. We could just leave out this check and always write the mask, even if it’s all-unmasked

src/anndata/_io/specs/methods.py
flying-sheep: When the `pd.*.BaseMaskedArray`’s `_mask` is `None`:

https://github.com/scverse/anndata/blob/3afc34b948a6c16752dbf0c34a135951fb0664ea/src/anndata/_io/specs/methods.py#L1033-L1034

src/anndata/tests/helpers.py
flying-sheep: good idea!

src/anndata/_io/specs/methods.py
```diff
@@ -1023,35 +1030,83 @@ def write_nullable_integer(
...
+            "and not supported in anndata < 0.11, still use by many people. "
+            "Opt-in to writing these arrays by toggling the setting to True."
         )
-    else:
-        return pd.array(_reader.read_elem(elem["values"]))
+        raise RuntimeError(msg)
+    g = f.require_group(k)
+    _writer.write_elem(g, "mask", v.isna(), dataset_kwargs=dataset_kwargs)
+    _writer.write_elem(
+        g, "values", v.fillna("")._ndarray, dataset_kwargs=dataset_kwargs
```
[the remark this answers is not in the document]
ilan-gold: Or
```suggestion
        g, "values", np.array(v.fillna("")), dataset_kwargs=dataset_kwargs
```

ilan-gold (changes requested):

src/anndata/_io/specs/methods.py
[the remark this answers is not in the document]
ilan-gold: `np.array` calls `to_numpy` in any case

src/anndata/_io/specs/methods.py
```diff
@@ -1023,35 +1030,83 @@ def write_nullable_integer(
     dataset_kwargs: Mapping[str, Any] = MappingProxyType({}),
 ):
     g = f.require_group(k)
-    if v._mask is not None:
-        _writer.write_elem(g, "mask", v._mask, dataset_kwargs=dataset_kwargs)
+    _writer.write_elem(g, "mask", v._mask, dataset_kwargs=dataset_kwargs)
```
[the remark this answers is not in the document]
flying-sheep: > Can anyone think of any reason that always writing the mask might break some things?

Not only I can’t think of a reason, but `self._mask` is typed as `np.ndarray` from the first commit of the extension array types. No idea why anyone thought it could be `None`.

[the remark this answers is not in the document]
flying-sheep: > While we're here, we should stop relying on the private API IMO.

I’d rather do that in another PR. `_data` is also a private API.

src/anndata/_io/specs/methods.py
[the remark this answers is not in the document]
flying-sheep: good call!

src/anndata/_io/specs/methods.py
```diff
@@ -1014,44 +1021,85 @@ def read_partial_categorical(elem, *, items=None, indices=(slice(None),)):
 @_REGISTRY.register_write(
     ZarrGroup, pd.arrays.BooleanArray, IOSpec("nullable-boolean", "0.1.0")
 )
+@_REGISTRY.register_write(
+    H5Group, pd.arrays.StringArray, IOSpec("nullable-string-array", "0.1.0")
+)
+@_REGISTRY.register_write(
+    ZarrGroup, pd.arrays.StringArray, IOSpec("nullable-string-array", "0.1.0")
+)
 def write_nullable_integer(
```
ilan-gold, line 1030:
```suggestion
def write_nullable(
```

src/testing/anndata/_pytest.py
```diff
@@ -32,10 +32,14 @@ def pytest_configure(config: pytest.Config) -> None:
...
 @pytest.fixture(autouse=True)
-def _suppress_env_for_doctests(request: pytest.FixtureRequest) -> None:
+def _anndata_test_env(request: pytest.FixtureRequest) -> None:
+    import anndata
+
     if isinstance(request.node, pytest.DoctestItem):
         request.getfixturevalue("_doctest_env")
 
+    anndata.settings.reset(anndata.settings._registered_options.keys())
+
```
ilan-gold, line 42: Not sure how I feel about a behavioral change like this being used on every test....took me a second to find this after seeing the lack of context manager.  So I'm not sure how easy this will be for the next person.   I won't block on it though.  I see the argument to be made that `settings` has a sort of implicit `function` scope i.e., changing in one test shouldn't change in another.  Could be nice to start a document listing all the custom things we do since they can be hard to find for people new to the repo

ilan-gold (approved):

src/testing/anndata/_pytest.py
flying-sheep: That would indeed be nice to have.

I think it’s very important to do all these resets in the tests (scanpy has many more) since they
1. document what global state your package has (and usually you don’t want it to have much of that)
2. make sure the tests are actually correct and no leakage from other tests causes them to be order-dependent. Another way to help with that would be to use something like https://pypi.org/project/pytest-randomly/ to randomize test order

src/anndata/_io/specs/methods.py
```diff
@@ -1014,44 +1021,85 @@ def read_partial_categorical(elem, *, items=None, indices=(slice(None),)):
...
+def _string_array(
+    values: np.ndarray, mask: np.ndarray
 ) -> pd.api.extensions.ExtensionArray:
-    if "mask" in elem:
-        return pd.arrays.BooleanArray(
-            _reader.read_elem(elem["values"]), mask=_reader.read_elem(elem["mask"])
-        )
-    else:
-        return pd.array(_reader.read_elem(elem["values"]))
+    """Construct a string array with pandas’ BaseMaskedArray API."""
```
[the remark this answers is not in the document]
flying-sheep: What I mean is the API `(values: npt.NDArray[Any], mask: npt.NDArray[np.bool_]) → pd.Array`

```suggestion
    """Construct a string array from values and mask."""
```

src/anndata/_io/specs/methods.py
```diff
@@ -506,10 +511,12 @@ def write_vlen_string_array_zarr(
...
-    # Workaround for https://github.com/zarr-developers/numcodecs/issues/514
-    # TODO: Warn to upgrade numcodecs if fixed
-    if not elem.flags.writeable:
-        elem = elem.copy()
+    if Version(numcodecs.__version__) < Version("0.13"):
+        msg = "Old numcodecs version detected. Please update for improved performance and stability."
+        warnings.warn(msg)
+        # Workaround for https://github.com/zarr-developers/numcodecs/issues/514
+        if hasattr(elem, "flags") and not elem.flags.writeable:
+            elem = elem.copy()
```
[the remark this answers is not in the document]
flying-sheep: Huh. Seems like coverage has been broken for a while, why are we getting anything?

src/testing/anndata/_pytest.py
ilan-gold: Sounds very reasonable to me.  Good call then.

src/anndata/_io/specs/methods.py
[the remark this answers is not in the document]
ilan-gold: It looks like 3.9 is publishing?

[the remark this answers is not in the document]
flying-sheep: For potential future people seeing this: only the mindeps and 3.12 jobs are publishing.

The reason it works is because we’re publishing twice: once to azure, once to codecov.io. The second one worked all along.

PR #1586, opened 2024-08-09
(fix): setting `index` on `DataFrameView`

### LINKED ISSUE #1201: A subset of a view can get a different .obs than the view itself

### Report

When a view object is created and its .obs property is modified, my understanding is that the .obs of the view and of its _adata_ref can diverge. A concrete manifestation of this can occur when subsetting the view, in which case the .obs of the subset seems to be extracted from the _adata_ref, hence being different than that of the view. Here is a minimal example.

Code:

```python
import anndata
import numpy as np

a1 = anndata.AnnData(X=np.array([[1, 2, 3], [4, 5, 6]]), obs={'obs_names': ['aa', 'bb'], 'property': [True, True]}, var={'var_names': ['c', 'd', 'e']})
a1 = a1[:, ['c', 'd']]
a1.obs.index = a1.obs.index.map(lambda x: x[-1])
selector = np.array([True, True])

# These should be the same but they are not
print(a1[selector].obs)
print(a1.obs)
```

Output:
```
    property
aa      True
bb      True
   property
a      True
b      True
```

### Versions

```python
-----
anndata             0.9.1
numpy               1.22.4
session_info        1.0.0
-----
bottleneck          1.3.5
cloudpickle         2.2.1
cython_runtime      NA
dateutil            2.8.2
debugpy             1.8.0
dill                0.3.6
h5py                3.8.0
importlib_metadata  NA
mpl_toolkits        NA
mpmath              1.3.0
natsort             8.3.1
numexpr             2.8.3
opt_einsum          v3.3.0
packaging           21.3
pandas              1.4.3
pyarrow             10.0.1
pydev_ipython       NA
pydevconsole        NA
pydevd              2.9.5
pydevd_file_utils   NA
pydevd_plugins      NA
pydevd_tracing      NA
pytz                2023.3
ruamel              NA
scipy               1.7.3
six                 1.16.0
sympy               1.11.1
torch               2.0.1
tqdm                4.64.1
typing_extensions   NA
zipp                NA
-----
Python 3.9.12 (main, Apr  5 2022, 01:53:17) [Clang 12.0.0 ]
macOS-10.16-x86_64-i386-64bit
-----
Session information updated at 2023-10-12 17:09
```

ilan-gold:
- [x] Closes #1201
- [x] Tests added

### DIFF
--- a/src/anndata/_core/views.py
+++ b/src/anndata/_core/views.py
@@ -274,6 +274,19 @@ class DataFrameView(_ViewMixin, pd.DataFrame):
         with view_update(*self._view_args) as df:
             df.drop(*args, inplace=True, **kw)
 
+    def __setattr__(self, key: str, value: Any):
+        if key == "index":
+            warnings.warn(
+                f"Trying to modify {key} from attribute `.{self._view_args.attrname}` of view, "
+                "initializing view as actual.",
+                ImplicitModificationWarning,
+                stacklevel=2,
+            )
+            with view_update(*self._view_args) as container:
+                setattr(container, key, value)
+        else:
+            super().__setattr__(key, value)
+
 
 @singledispatch
 def as_view(obj, view_args):
--- a/src/anndata/_core/views.py
+++ b/src/anndata/_core/views.py
@@ -277,7 +277,7 @@ class DataFrameView(_ViewMixin, pd.DataFrame):
     def __setattr__(self, key: str, value: Any):
         if key == "index":
             warnings.warn(
-                f"Trying to modify {key} from attribute `.{self._view_args.attrname}` of view, "
+                f"Trying to modify {key} of attribute `.{self._view_args.attrname}` of view, "
                 "initializing view as actual.",
                 ImplicitModificationWarning,
                 stacklevel=2,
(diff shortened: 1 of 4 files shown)


ilan-gold:
This is backwards - we _don't_ want the orginal object updated.

src/anndata/_core/views.py
```diff
@@ -274,6 +274,19 @@ def drop(self, *args, inplace: bool = False, **kw):
         with view_update(*self._view_args) as df:
             df.drop(*args, inplace=True, **kw)
 
+    def __setattr__(self, key: str, value: Any):
+        if key == "index":
+            warnings.warn(
+                f"Trying to modify {key} from attribute `.{self._view_args.attrname}` of view, "
```
flying-sheep, line 280:
```suggestion
                f"Trying to modify {key} of attribute `.{self._view_args.attrname}` of view, "
```

I think this is the right grammar? But you’re the native speaker, so idk!

flying-sheep (approved):
Looks good! Did you check if this warning gets raised anywhere in our tests or even some utility function?

If so, that should probably be fixed in this PR. Otherwise LGTM!

ilan-gold:
> Looks good! Did you check if this warning gets raised anywhere in our tests or even some utility function?

We do `--strict-warnings` so tests would fail if it did, no?

src/anndata/_core/views.py
ilan-gold: I think both are ok, but yours is probably more fluid yea

flying-sheep:
We specifically don’t for this one yet:

https://github.com/scverse/anndata/blob/daed86fa09f03c895a35dad059c39602de75ad88/pyproject.toml#L138-L140

ilan-gold:
@flying-sheep No additional warnings there so we're good.  So will merge.  But why don't we just add `pytest.warns` to the two places where those warnings are raised?

flying-sheep:
> But why don't we just add pytest.warns to the two or three places where those warnings are raised?

we should definitely do that!

PR #1589, opened 2024-08-09
(fix): disallow using dataframes with multi index columns

### LINKED ISSUE #1202: Compatibility with pd.MultiIndex dimension labels
The following line breaks in pandas>2 for MultiIndex X columns. https://github.com/scverse/anndata/blob/ff64448c44836b8134b67f828fb91f70dc59e347/anndata/_core/anndata.py#L120.

Due to change in behavior of `is_string_dtype`.

```py
import pandas as pd
from pandas.api.types import is_string_dtype

is_string_dtype(pd.MultiIndex.from_tuples([('a','b')])) # False in pandas>=2

is_string_dtype(pd.MultiIndex.from_tuples([('a','b')])) # True in pandas<2
```

`anndata==0.9.2`

Also found in
`anndata==0.10.2` (latest release)

ilan-gold:
- [x] Closes #1202
- [x] Tests added

### DIFF
--- a/src/anndata/_core/aligned_mapping.py
+++ b/src/anndata/_core/aligned_mapping.py
@@ -258,16 +258,22 @@ class AxisArraysBase(AlignedMappingBase):
         return df
 
     def _validate_value(self, val: Value, key: str) -> Value:
-        if isinstance(val, pd.DataFrame) and not val.index.equals(self.dim_names):
-            # Could probably also re-order index if it’s contained
-            try:
-                pd.testing.assert_index_equal(val.index, self.dim_names)
-            except AssertionError as e:
-                msg = f"value.index does not match parent’s {self.dim} names:\n{e}"
-                raise ValueError(msg) from None
-            else:
-                msg = "Index.equals and pd.testing.assert_index_equal disagree"
-                raise AssertionError(msg)
+        if isinstance(val, pd.DataFrame):
+            if isinstance(val.columns, pd.MultiIndex):
+                raise ValueError(
+                    "MultiIndex columns are not supported in AnnData. "
+                    "Please use a single-level index."
+                )
+            if not val.index.equals(self.dim_names):
+                # Could probably also re-order index if it’s contained
+                try:
+                    pd.testing.assert_index_equal(val.index, self.dim_names)
+                except AssertionError as e:
+                    msg = f"value.index does not match parent’s {self.dim} names:\n{e}"
+                    raise ValueError(msg) from None
+                else:
+                    msg = "Index.equals and pd.testing.assert_index_equal disagree"
+                    raise AssertionError(msg)
         return super()._validate_value(val, key)
 
     @property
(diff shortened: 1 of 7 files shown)


src/anndata/_core/storage.py
```diff
@@ -84,6 +84,11 @@ def coerce_array(
             value = value.A
         return value
     if isinstance(value, pd.DataFrame):
+        if isinstance(value.columns, pd.MultiIndex) and allow_df:
```
ilan-gold, line 87: Could remove, but not bad for exhaustiveness sake

src/anndata/_core/storage.py
```diff
@@ -84,6 +88,8 @@ def coerce_array(
             value = value.A
         return value
     if isinstance(value, pd.DataFrame):
+        if allow_df:
+            raise_value_error_if_multiindex_columns(value)
```
ilan-gold, line 92: could remove due to no coverage, but not bad for exhaustiveness' sake

flying-sheep: I went the opposite route and removed the `xxxm` checking to rely on this instead.

flying-sheep (approved):

flying-sheep (approved):

src/anndata/utils.py
```diff
@@ -400,3 +400,12 @@ def is_hidden(attr) -> bool:
             for item in type.__dir__(cls)
             if not is_hidden(getattr(cls, item, None))
         ]
+
+
+def raise_value_error_if_multiindex_columns(df: pd.DataFrame):
+    if isinstance(df.columns, pd.MultiIndex):
+        msg = (
+            "MultiIndex columns are not supported in AnnData. "
+            "Please use a single-level index."
```
flying-sheep, line 409: Please add (and use) an attr name parameter here. This should make it easier if the error gets raised during initialization.

```suggestion
            f"Please use a single-level index for {attr}."
```

flying-sheep (changes requested):

src/anndata/_core/aligned_mapping.py
```diff
@@ -265,7 +265,7 @@ def to_df(self) -> pd.DataFrame:
 
     def _validate_value(self, val: Value, key: str) -> Value:
         if isinstance(val, pd.DataFrame):
-            raise_value_error_if_multiindex_columns(val)
+            raise_value_error_if_multiindex_columns(val, f"{self.attrname}.{key}")
```
flying-sheep, line 268: Isn’t this more like

```suggestion
            raise_value_error_if_multiindex_columns(val, f"{self.attrname}[{key!r}]")
```

flying-sheep (changes requested):

src/anndata/_core/aligned_mapping.py
ilan-gold: Sure, I'll quotes since it is more like what people write.

ilan-gold: Ah, the repr does that ok.

flying-sheep (approved):

PR #1593, opened 2024-08-12
(fix): use `typing_extensions` for `NamedTuple`

ilan-gold:
I was off on the version in my comment or we just missed I meant "greater than": https://github.com/python/cpython/pull/92027

In any case, we can just use `typing_extensions`: https://github.com/python/typing_extensions/pull/44

Previously, this was causing issues with 3.10 as this was the case not covered (this feature broke in 3.9, but is fixed in 3.11, and we special cased <3.10).

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_settings.py
+++ b/src/anndata/_settings.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import inspect
 import os
-import sys
 import textwrap
 import warnings
 from collections.abc import Iterable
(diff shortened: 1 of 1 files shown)


ilan-gold:
Hmmm, I thought `typing_extensions` was either already a dependent or included in python.  Better not to add a dependency just for this.

PR #1596, opened 2024-08-13
GPU tests: install with uv

Intron7:
This would allow testing with cuda-12. It would also fully use uv to install packages outside of python and uv


Intron7:
Ok so for the cupy array I ran into a bug with the default value ([see older gpu tests](https://github.com/scverse/anndata/actions/runs/10369935722/job/28706763100)). I could fix it with switching from `cupy.full` to `cp.zeros`. Also newer versions of cupy allow for numpy 2.0 so I disabled that.

Intron7:
I dont know why codecov didnt work for the gpu run

.github/workflows/test-gpu.yml
```diff
@@ -57,8 +57,11 @@ jobs:
           environment-file: ci/gpu_ci.yml
           post-cleanup: "all"
 
+      - name: Install cupy
+        run: uv pip install cupy-cuda12x numba pytest pytest-cov pytest-xdist
+
       - name: Install AnnData
-        run: uv pip install -e .[dev,test,gpu]
+        run: uv pip install -e .[dev,test]
```
ilan-gold, line 64: 1. Why do we need these `pytest` deps if they come from `test` (Except for `xdist`)?
2. Why are we using `cupy` as the dep in `gpu` instead of a pre-built wheel?  
3. Since it seems like we need `numba` why not make a `test-gpu` section in the `pypoject.toml`?

pyproject.toml
```diff
@@ -102,7 +102,6 @@ test = [
 ]
 gpu = [
     "cupy",
-    "numpy<2.0.0",
```
ilan-gold, line 105: So I think this is needed (see below)

ilan-gold (changes requested):

src/anndata/_core/merge.py
```diff
@@ -592,7 +592,7 @@ def _apply_to_cupy_array(self, el, *, axis, fill_value=None):
         out_shape = list(el.shape)
         out_shape[axis] = len(self.new_idx)
 
-        out = cp.full(tuple(out_shape), fill_value)
```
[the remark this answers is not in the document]
ilan-gold: Ok maybe not something I changed.  But this is why I upper bounded `numpy` https://github.com/scverse/anndata/pull/1540/files

.github/workflows/test-gpu.yml
flying-sheep: also please never mutate an environment after it has been created. that means

1. only use one command to create the environment, then use it.

   you can use `-e` in the same call as specifying other dependencies!

3. no `pip: ...` entry in an environment.yaml that’s installed by conda, as that translates to `conda install ... && pip install ...`. If you need to install conda packages and pip packages, use something that supports that, e.g. conda-lock or pixi

flying-sheep (changes requested):

ci/gpu_ci.yml
```diff
@@ -1,13 +1,6 @@
 name: cupy_env
```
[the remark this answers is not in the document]
ilan-gold: Can delete this file now?

pyproject.toml
```diff
@@ -102,6 +102,16 @@ test = [
...
+    "numpy <2.0.0",
+]
+
+gpucu12 = [
+    "cupy-cuda12x",
+    "numpy<2.0.0",
+]
+
+gpucu11 = [
+    "cupy-cuda11x",
```
ilan-gold, line 114:
```suggestion
    "numpy <2.0.0",
]

cu12 = [
    "cupy-cuda12x",
    "numpy<2.0.0",
]

cu11 = [
    "cupy-cuda11x",
```

ilan-gold: or `cuda11`/`cuda12`?

ilan-gold: I think `gpu` is redundant

.github/workflows/test-gpu.yml
```diff
@@ -51,19 +51,21 @@ jobs:
...
         run: nvidia-smi
 
-      - uses: mamba-org/setup-micromamba@v1
+      - name: Install Python
+        uses: actions/setup-python@v5
         with:
-          micromamba-version: "1.5.6-0"
-          environment-file: ci/gpu_ci.yml
-          post-cleanup: "all"
+          python-version: "3.x"
```
ilan-gold, line 57: Probably want to fix this to a version?

ilan-gold: Or multiple?

ilan-gold (changes requested):

.github/workflows/test-gpu.yml
Intron7: I think this uses the newest stable release thats what we do for the release aswell

flying-sheep: In the release workflow, we don’t actually import the package AFAIK, so it’s fine if we build it with a version that isn’t supported by e.g. numba yet.

I agree with the sentiment: I don’t want multiple workflows that all hardcode the newest release that numba (and by extension we) support, I’d like to outsource or at least centralize this release specifier.

It’s not super important right now though, as we’ll notice when the next stable release breaks numba …

.github/workflows/test-gpu.yml
```diff
@@ -51,19 +51,19 @@ jobs:
...
         run: nvidia-smi
 
-      - uses: mamba-org/setup-micromamba@v1
+      - name: Install Python
+        uses: actions/setup-python@v5
         with:
-          micromamba-version: "1.5.6-0"
-          environment-file: ci/gpu_ci.yml
-          post-cleanup: "all"
+          python-version: "3.11"
```
flying-sheep, line 57: why not 3.12?

pyproject.toml
```diff
@@ -102,7 +102,14 @@ test = [
...
-    "numpy<2.0.0",
+]
+
+cu12 = [
+    "cupy-cuda12x",
+]
+
+cu11 = [
+    "cupy-cuda11x",
 ]
```
flying-sheep, line 113: maybe we shouldn’t have an extra here at all and just document how to get GPU support?

Intron7: we could also throw out the complete cupy setup and just add docs for it

ilan-gold: Well, I was opposed to having both that _and_ a `gpu` extra.  But I would be fine with @flying-sheep's solution if we get rid of `gpu`.  My only thing would be if we add another library in the future for GPU.  @ivirshup might have insight as to what could come.  Maybe `cudf`?

flying-sheep (approved):
I moved this to the 0.10.9 milestone: all dev process PRs that can should be backported

ilan-gold (changes requested):
@Intron7 Please add a fragment for this.  You can create a new towncrier header for "Development Process" and add it to the `pyproject.toml` at the bottom,

ilan-gold:
Sorry if it wasn't clear, I meant re: the pyproject.toml.  People don't need to know about our internal use of `uv`

ilan-gold:
@Intron7 Also a fragment with the news? You can find examples in https://github.com/scverse/anndata/tree/main/docs/release-notes so the file for this PR would be `1596.dev.md`

ilan-gold:
Sorry if that wasn't clear

docs/release-notes/1596.dev.md
```diff
@@ -0,0 +1,2 @@
+update the CI to use `uv` {user}`Intron7`
+create new `cupy` install options for cuda 11 & 12 {user}`Intron7`
```
ilan-gold, line 2:
```suggestion
create new `cupy` installation options for cuda 11 & 12 called `cu11` and `cu12` {user}`Intron7`
```

ilan-gold (changes requested):

docs/release-notes/1596.dev.md
ilan-gold: We shouldn't be putting CI changes in fragments and should only have one note per fragment.

ilan-gold (approved):

PR #1598, opened 2024-08-14
(chore): export `read_elem` and `write_elem` from the main package

### LINKED ISSUE #1158: Move `read_elem`, `write_elem` from experimental into main API

`read_elem`, `write_elem` are stable enough to moved out of experimental.

Possible end locations include:

* Main namespace: `ad.read_elem`
* `io` submodule: `ad.io.read_elem`
* `storage` submodule `ad.storage.read_elem`

I'd like to make an `io` module for all the read and write functions, but maybe a new module that would also house other on disk access (e.g. `sparse_dataset`) would also be good.

ilan-gold:
To do this properly, needed to export `RWAble` and the sparse backed classes

- [x] Closes #1158
- [x] Tests added


src/anndata/_types.py
```diff
@@ -49,7 +49,8 @@
     H5Array,
     ZarrArray,
     ZappyArray,
-    BaseCompressedSparseDataset,
+    CSRDataset,
+    CSCDataset,
```
ilan-gold, line 53: This wasn't really correct to begin with since `BaseCompressedSparseDataset` isn't documented

flying-sheep: They look very sparsely documented though.

We should decide how much of that API we want to commit to and hide all attributes/methods we don’t want to support (especially that deprecated one)
 
image

flying-sheep (changes requested):
see above: define public API of CS{R,C}DataSet

src/anndata/_types.py
ilan-gold: See https://github.com/scverse/anndata/pull/1608

flying-sheep (approved):

ilan-gold:
@flying-sheep After seeing I have been thinking about this a bit.  Is it bad practice to keep the `CS{C,R}Dataset` classes experimental while exporting something from core that can deal with them? I don't think so.  Even though [`X`](https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.X.html#anndata.AnnData.X) has the wrong type annotation, it should include `CS{R,C}Dataset`, and this would be an example of a public API referencing an experimental one.

src/anndata/__init__.py
```diff
@@ -97,3 +104,15 @@ def read(*args, **kwargs):
     "experimental",
     "settings",
 ]
+
+
+class CustomExperimental(types.ModuleType):
+    def __getattribute__(self, key: str) -> Any:
```
flying-sheep, line 110: why not just define `def __getattr__(key: str) -> Any: ...` in the experimental module directly?

https://peps.python.org/pep-0562/ exists since 3.7

flying-sheep (changes requested):

src/anndata/__init__.py
ilan-gold: This has the advantage of never having to thinking about this again.  As we move things out of experimental into the public API, we'll never have to update this.

ilan-gold: Maybe you could do `__getattr__` without a circular import? I guess I can try but I just kind of assumed not

ilan-gold: Although now that I think about it, maybe a blanket warning isn't such a good idea...would want to be explicit about deprecations

flying-sheep: You can just import inside of `__getattr__`, no?

`experimental/__init__.py`:
```py
[…]

def __getattr__(key: str) -> Any:
    import anndata

    if key in anndata.__all__: ...
```

src/anndata/experimental/__init__.py
```diff
@@ -1,21 +1,29 @@
...
+    from typing import Any
+
+deprecated = ["CSRDataset", "CSCDataset", "read_elem", "write_elem"]
+
+
+def __getattr__(key: str) -> Any:
+    if key in deprecated:
+        msg = f"Importing {key} from `anndata.experimental` is deprecated. Import from `anndata` directly."
+        warnings.warn(msg, FutureWarning)
+    return globals()[key]
```
flying-sheep, line 25: `__getattr__` is only called if the name is *not* in `globals()`, so we need to remove the import and change this: 

```suggestion
        import anndata

        return getattr(anndata, key)
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

ilan-gold: That makes more sense! "in a module that would act like a normal __getattr__ method"

flying-sheep (approved):

PR #1604, opened 2024-08-22
(feat): use `towncrier` for docs automations

ilan-gold:
Blocked by https://github.com/twisted/towncrier/issues/661 and https://github.com/twisted/towncrier/issues/648#issuecomment-2304965344.

Also, need to figure out exactly where to inject this into the CI.   If not, this can be done manually, although I'd rather not.

- [ ] Closes #
- [ ] Tests added


pyproject.toml
```diff
@@ -64,6 +64,8 @@ dev = [
     "setuptools_scm",
     # test speedups
     "pytest-xdist",
+    # until https://github.com/twisted/towncrier/pull/651 is released
+    "towncrier @ git+https://github.com/twisted/towncrier",
```
flying-sheep, line 68: you could move this to `hatch.toml`, as having these URLs directly in `project.(optional-)dependencies` technically goes against spec.

that way we also don’t need to add `tool.hatch.metadata.allow-direct-references = true`

flying-sheep (commented):
Wow, if that’s it, that looks extremely clean, seems instantly nicer than what we have now.

maybe this can be relatively easily automated to a release-please like workflow

pyproject.toml
```diff
@@ -192,3 +196,19 @@ strict = true
...
 skip = ".git,*.pdf,*.svg"
 ignore-words-list = "theis,coo,homogenous"
+
+[tool.towncrier]
+package = "anndata"
+directory = "docs/release-notes"
+filename = "docs/release-notes/{version}.md"
+single_file = false
+package_dir = "src"
+issue_format = "[#{issue}](https://github.com/scverse/anndata/pull/{issue})"
```
flying-sheep, line 206:
```suggestion
issue_format = "{{issue}}`{issue}`"
```

pyproject.toml
```diff
@@ -192,3 +196,19 @@ strict = true
...
+title_format = "({version})=\n### {version} `{project_date}`"
```
flying-sheep, line 207: After merging #1346, I found that these don’t actually work: seems like MyST doesn’t support anchors that start with a number.

So in https://github.com/scverse/scanpy/pull/3172, I started prefixing them with a `v` like this to get anchors:

```suggestion
title_format = "(v{version})=\n### {version} `{project_date}`"
```

pyproject.toml
ilan-gold: Why? This is the current format that we have

flying-sheep: The syntax we use in our docs is ``{issue}`123` ``

Is this not used in our docs?

ilan-gold: Ah, I see what this would do.  Apologies

flying-sheep:
FYI: I pushed eefafad:

- https://github.com/scverse/anndata/blob/eefafad90d6ef3ee9656d8e5815af2f92496ffbd/hatch.toml#L1-L3
  allows to just `hatch run towncrier`
- https://github.com/scverse/anndata/blob/eefafad90d6ef3ee9656d8e5815af2f92496ffbd/pyproject.toml#L67
  allows to rely on standard metadata without `allow-direct-references`

ilan-gold:
> We can maybe simply use [sections](https://towncrier.readthedocs.io/en/stable/configuration.html#sections) for our minor version headers:

Is there something wrong with our current system?  I like the separate branches, like `v0.10` will simply contain fragments (after the the inital feature release) that aren't features so there's no need to stash or for separate directories etc.

flying-sheep:
Actually we can’t use sections for what I was thinking.

My comment was about this structure:

grafik

when we do a feature release, we’d want to insert

```md
(v0.11)=
## Version 0.11
```

before the new `0.11.0` version we’re creating.

and when we do a bugfix release, we want to insert our new section to be inserted below the “Version 0.10” header.

There is only one insertion point: [see here](https://towncrier.readthedocs.io/en/stable/customization/newsfile.html)

flying-sheep:
Ooof, now that I wrote all this code, I realized there’s an even better solution:

Replace the whole content of `release-notes/index.md` with

```md
# Release notes

```{release-notes} .
```

and implement the `release-notes` directive that collects all `*.*.*.{rst,md}` file names in the directory and constructs the whole doctree that file currently contains.

flying-sheep:
TODO:

- [ ] migrate 0.11.0.md to feature fragments
- [ ] backport everything except for feature fragments to `0.10.x` branch

docs/extensions/release_notes.py
```diff
@@ -0,0 +1,111 @@
...
+        for (major, minor), versions in version_groups:
+            self.render_version_group(major, minor, versions)
+        return []
+
+    def render_version_group(
+        self, major: int, minor: int, versions: Iterable[tuple[Version, Path]]
+    ) -> None:
+        target = nodes.target(
+            ids=[f"v{major}-{minor}"],
+            names=[f"v{major}.{minor}"],
```
ilan-gold, line 59: I was/have always been wondering about this - why use `v` some places and not in others?

ilan-gold: i.e., it is not used here https://github.com/scverse/anndata/pull/1604/files#diff-bdbc86648710a391d76972ef99a2b1cb8587ec09d9b4aea5d403497f5cf2d806R63-R65

ilan-gold: For me, I find it clearer to only do one or the other (`v` or no `v`)

ilan-gold: Aside from the two refs to this, nothing else uses these and we use `v` nowhere else to refer to releases from what I can tell

flying-sheep: >why use v some places and not in others

MyST didn’t turn the `(x.y)=` (_target_) syntax into a URL anchor unless you added the `v`.

So Isaac spelled the `x.y` headings “Version x.y” which results in `…/release-notes#version-x-y` slugs.
I didn’t want the word “version” on the page hundreds of times, so I experimented until I figured out that MyST *will* respect the target syntax if there’s at least one letter.

The changes in this PR are to stay consistent with the slugs we already have because that’s the whole point: that links to to `…/release-notes#version-0.10` or `…/release-notes#v0.10.8` are stable.

ilan-gold: > MyST doesn’t turn the (x.y)= (target) syntax into a URL anchor unless you add letters.

I didn't know you need the letters, ok.

That makes sense, I didn't know the history of this.

ilan-gold:
@flying-sheep It will be easier to backport this if we simply don't do the `0.11` fragments on this branch.  I will open up a follow-up PR.

pyproject.toml
```diff
@@ -192,3 +194,19 @@ strict = true
...
 skip = ".git,*.pdf,*.svg"
 ignore-words-list = "theis,coo,homogenous"
+
+[tool.towncrier]
+package = "anndata"
+directory = "docs/release-notes"
+filename = "docs/release-notes/{version}.md"
+single_file = false
+package_dir = "src"
+issue_format = "{{issue}}`{issue}`"
```
ilan-gold, line 204: This is a bit redundant since our current PR system has the `{pr}` or `{issue}` tag at the end.  I am going to update the fragments here.

flying-sheep, line 204: Since we actually link PRs and not issues:

```suggestion
issue_format = "{{pr}}`{issue}`"
```

flying-sheep (approved):

PR #1608, opened 2024-08-26
(fix): clarify public backed sparse docstring/api

ilan-gold:
For https://github.com/scverse/anndata/pull/1598#pullrequestreview-2260162414

Perhaps should go in separately to `main` as it is a cleaner git blame IMO
- [ ] Closes #
- [ ] Tests added


flying-sheep:
So you want us to commit to the whole API?

If not necessarily: There are ways to make users see only part of it, I can help with that.

ilan-gold:
> If not necessarily: There are ways to make users see only part of it, I can help with that.

Sorry went to lunch before getting to that.  I made private what I thought should be, and left what I thought should be public as such.  Feel free to comment otherwise

flying-sheep (approved):
Awesome! I don’t think we want people to manually initialize it, but I think not much harm is done if they try do do that.

src/anndata/_core/sparse_dataset.py
```diff
@@ -569,21 +572,29 @@ def _to_backed(self) -> BackedSparseMatrix:
...
 
-    def to_memory(self) -> ss.csr_matrix | ss.csc_matrix:
+    def to_memory(self) -> spmatrix | SpArray:
+        """Returns an in-memory representation of the sparse matrix.
+
+        Returns
+        -------
+            The in-memory representation of the sparse matrix.
+        """
         format_class = get_memory_class(self.format)
```
flying-sheep, line 585: This never returns `csr_array`/`csc_array`, should it?

ilan-gold: At the moment, it shouldn't.  But we need at add a setting in for that, thanks for the reminder.   Will PR now

flying-sheep: OK, then that doesn’t stop this PR

flying-sheep (approved):

src/anndata/_core/sparse_dataset.py
```diff
@@ -569,21 +572,29 @@ def _to_backed(self) -> BackedSparseMatrix:
...
+        mtx.indptr = self._indptr
         return mtx
 
-    def to_memory(self) -> ss.csr_matrix | ss.csc_matrix:
+    def to_memory(self) -> spmatrix | SpArray:
+        """Returns an in-memory representation of the sparse matrix.
+
+        Returns
+        -------
+            The in-memory representation of the sparse matrix.
```
flying-sheep, line 583:
```suggestion
        The in-memory representation of the sparse matrix.
```

Issue #1195, opened 2023-10-11
General array-api support


ivirshup:
### Please describe your wishes and possible alternatives to achieve the desired result.

With the [array-api](https://data-apis.org/array-api/latest/) it should be possible to have some level of support for a greater variety of array types, including `pytorch` and `jax`. E.g.:

* https://github.com/scverse/anndata/issues/697
* https://github.com/scverse/anndata/issues/1019

We have a proof of concept with using array-api compat on top of `cupy` arrays.

# Challenges

## Dispatch

We currently use type based dispatch to select appropriate methods for different array types. It's not immediately clear to me how we integrate the array-api here. Potentially with a `Protocol` or `ABC`?

## Specialized code for performance

We currently have a bunch of non-standard code for various operations for performance reasons. For instance, using pandas indexing code on numpy arrays.

## IO dispatch

IO dispatch is particularly difficult. It's likely we just need to use the `dlpack` based interchange here.

## Views

Current view infrastructure is very type based. Right now, I think we'd either need to rework this or have special code for each array type, which kinda defeats the purpose. Maybe an xarray dependency would be easier here.

## Sparse

The array API currently doesn't account for sparse types:

* https://github.com/scipy/scipy/issues/18915

A notable problem here is IO, since dlpack based interchange would densify the array.

## Downstream libraries

Finally, how do downstream libraries deal with this? It creates the possibility of passing these libraries array types they haven't seen before. What is the appropriate behavior on their end? Do you ask the user to convert? Does the library convert on its own (even if that could be extremely expensive)?

flying-sheep:
We can define a type if none exists using the same pattern that `abc.ABCMeta`/`abc.ABC`/`typing.Protocol` uses: [`__instancecheck__`](https://docs.python.org/3/reference/datamodel.html#class.__instancecheck__) (here combined with [`is_array_api_obj`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_array_api_obj)):

```py
import array_api_compat

class ArrayApiMeta(type):
    def __instancecheck__(cls, instance: object) -> bool:
        return array_api_compat.is_array_api_obj(instance)

class ArrayApi(meta=ArrayApiMeta):
    pass
```

PR #1616, opened 2024-08-28
(fix): correct typing of `AnnData.X`

ilan-gold:
- [ ] Re: https://github.com/scverse/anndata/pull/1598#issuecomment-2314890727
- [ ] Tests added


src/anndata/_core/anndata.py
```diff
@@ -541,7 +542,17 @@ def shape(self) -> tuple[int, int]:
...
 
     @property
-    def X(self) -> np.ndarray | sparse.spmatrix | SpArray | ArrayView | None:
+    def X(
+        self,
+    ) -> (
+        CSRDataset
+        | CSCDataset
+        | np.ndarray
+        | sparse.spmatrix
```
flying-sheep, line 551:
```suggestion
        | sparse.csr_matrix
        | sparse.csc_matrix
```

flying-sheep (commented):
Shouldn’t we unify this with `ArrayDataStructureType`?

I don’t quite know why that one is an enum, maybe we can replace it with

```py
ArrayDataStructureType = Union[All|Of|This|Here]
```

and then do

```
from typing import get_args

isinstance(get_args(ArrayDataStructureType))
```

ilan-gold:
> I don’t quite know why that one is an enum, maybe we can replace it with

I also didn't know but don't really question these sorts of things.  I imagine it served a purpose at some point, but like I said in the related PR, that seems to have changed recently.

flying-sheep:
I guess then we should unify the two. Either as part of this PR or later, as you wish

src/anndata/_core/anndata.py
ilan-gold: So somewhere, we are sticking a `dia_matrix` into an `AnnData` object.  For now I'm going to leave this, and file an issue.

ilan-gold: Nope! too quick! this caught a bug in a test! good!

flying-sheep:
We could maybe use [`__instancecheck__`](https://docs.python.org/3/reference/datamodel.html#class.__instancecheck__) and [`is_array_api_obj`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_array_api_obj):

```py
import array_api_compat

class ArrayApiMeta(type):
    def __instancecheck__(cls, instance: object):
        return array_api_compat.is_array_api_obj(instance)

class ArrayApi(meta=ArrayApiMeta):
    pass
```

ilan-gold:
I'm going to push back on the array-api here because I think it's out of scope and testing for it will be non-trivial.  I think we should make a separate branch/PR for that specifically, and stick it in `0.12`.  We can test for jax, which now has a working CPU implementation (like numpy).  This would be a very cool feature.

ilan-gold:
There's even an issue for this: https://github.com/scverse/anndata/issues/1195

src/anndata/_core/storage.py
```diff
@@ -28,45 +27,23 @@
...
+    np.ndarray,
+    ma.MaskedArray,
+    sparse.csr_matrix,
+    sparse.csc_matrix,
+    SpArray,
+    AwkArray,
+    H5Array,
+    ZarrArray,
+    ZappyArray,
+    BaseCompressedSparseDataset,
```
flying-sheep, line 42: Similar to above with `csr_matrix` and `csc_matrix`, we should probably do this, what do you think?

```suggestion
    CSRDataset,
    CSCDataset,
```

flying-sheep: self-answer: yes, we should, this way we get links.

src/anndata/_core/storage.py
```diff
@@ -100,7 +87,7 @@ def coerce_array(
         except (ValueError, TypeError) as _e:
             e = _e
     # if value isn’t the right type or convertible, raise an error
-    msg = f"{name} needs to be of one of {join_english(ArrayDataStructureType.qualnames())}, not {type(value)}."
+    msg = f"{name} needs to be of one of {join_english([str(x) for x in array_data_structure_types])}, not {type(value)}."
```
flying-sheep, line 90:
```suggestion
    msg = f"{name} needs to be of one of {join_english(map(str, array_data_structure_types))}, not {type(value)}."
```

flying-sheep (approved):

PR #1625, opened 2024-08-29
Run towncrier when building docs

### LINKED ISSUE #1622: (dev): Validate `towncrier` fragments.

We should probably have some sort of test that builds the docs _after_ running `towncrier` to ensure fragments are valid.  Another option would be some sort of `myst` script to check that the fragment "compiles" without warning (although I don't know enough to say whether or not that is a valid approach).

flying-sheep:
- [x] Closes #1622
- [x] Tests added

TODO:

- [x] only run if necessary (it will create a file if not)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,6 +4,9 @@ build:
   tools:
     python: "3.12"
   jobs:
+    post_checkout:
+      # unshallow so version can be derived from tag
+      - git fetch --unshallow || true
     pre_build:
       - "( find docs/release-notes -regex '[^.]+[.][^.]+.md' | grep -q . ) && towncrier build --keep"
       - ls docs/release-notes
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,7 +9,7 @@ build:
       - git fetch --unshallow || true
     pre_build:
       # run towncrier to preview the next version’s release notes
-      - "( find docs/release-notes -regex '[^.]+[.][^.]+.md' | grep -q . ) && towncrier build --keep"
+      - ( find docs/release-notes -regex '[^.]+[.][^.]+.md' | grep -q . ) && towncrier build --keep
 sphinx:
   configuration: docs/conf.py
   fail_on_warning: true # do not change or you will be fired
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -82,7 +77,10 @@ doc = [
     "myst_parser",
     "sphinx_design>=0.5.0",
     "readthedocs-sphinx-search",
+    # for unreleased changes
+    "anndata[dev-base,dev-doc]",
```
ilan-gold, line 81: Why `dev-base` as well?

ilan-gold (approved):

.readthedocs.yml
```diff
@@ -3,6 +3,13 @@ build:
   os: ubuntu-20.04
   tools:
     python: "3.12"
+  jobs:
+    post_checkout:
+      # unshallow so version can be derived from tag
+      - git fetch --unshallow || true
+    pre_build:
+      # run towncrier to preview the next version’s release notes
+      - "( find docs/release-notes -regex '[^.]+[.][^.]+.md' | grep -q . ) && towncrier build --keep"
```
ilan-gold, line 12: Sorry wait, what is the effect of this on our releases?

ilan-gold (changes requested):

.readthedocs.yml
ilan-gold: It seems to change the file of the current version.  So if I run `towncrier build --version=0.10.9` with no files, it overrides the `0.10.9.md` file.  Similarly, this will start adding our `dev` notes to our `latest` docs, which I guess is ok?

flying-sheep: nothing. this line means “if there are fragments when RTD builds the docs, we create an ephemeral release notes file that doesn’t get committed anywhere, and then build the docs containing that pre-version’s release notes”

pyproject.toml
flying-sheep: good point, I thought we need setuptools-scm, but that’s just for runtime `__version__`

.readthedocs.yml
ilan-gold: I guess this really solidifies the fact that there should be no stray fragments after _any_ release on the release base branch.  I like it.

ilan-gold (approved):

PR #1626, opened 2024-08-29
(chore): automation of towncrier release process

### LINKED ISSUE #1620: (dev): Automation for cherry-picking release notes to `main` from `0.X.X` branch

Right now, if we make a release notes fix on a minor branch e.g., `0.10.x` branch, we need to manually cherry pick it to `main`, although this should be doable with an automated PR from the release process.

ilan-gold:
- [x] Closes #1620
- [ ] Tests added


hatch.toml
```diff
@@ -9,3 +9,11 @@ dependencies = ["setuptools"]  # https://bitbucket.org/pybtex-devs/pybtex/issues
 [envs.docs.scripts]
 build = "sphinx-build -M html docs docs/_build -W --keep-going {args}"
 clean = "git clean -fX -- docs"
+
+[envs.towncrier]
+features = ["doc"]
+dependencies = ["setuptools"]  # https://bitbucket.org/pybtex-devs/pybtex/issues/169
```
ilan-gold, line 15: Subject to change upon #1625

ilan-gold:
A fun thing about this is that I just realized that it would allow for even more automation i.e., if we wanted to have automatic releases from closed out milestones: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#milestone

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+import argparse
+import subprocess
+
+from packaging.version import parse
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        prog="towncrier-automation",
+        description="This script runs townncrier for a given version, creates a branch off of the current one, and then creates a PR into the original branch with the changes.  The PR will be backported to main if the current branch is not main.",
```
flying-sheep, line 13:
```suggestion
        description=(
            "This script runs towncrier for a given version, "
            "creates a branch off of the current one, "
            "and then creates a PR into the original branch with the changes. "
            "The PR will be backported to main if the current branch is not main."
        ),
```

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+        "--dry-run",
+        help="Whether or not to dry-run the actual creation of the pull request",
+        action="store_true",
+    )
+    args = parser.parse_args()
+    version = args.version
+    parse(args.version)
+
+    # Run towncrier
+    if subprocess.call(["towncrier", "build", f"--version={version}", "--yes"]):
```
flying-sheep, line 27: isn’t this inverse? if the command succeeds, the status code is `0`, but this branch is entered

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+    # Check if we are on the main branch to know if we need to backport
+    base_branch = subprocess.run(
+        ["git", "rev-parse", "--abbrev-ref", "HEAD"],
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.strip()
+    pr_description = ""
+    if base_branch != "main":
+        pr_description = "on-merge: backport to main"
```
flying-sheep, line 39: please use a ternary in super simple cases like this

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+    branch_name = f"release_notes_{args.version}"
+
+    # Create a new branch + commit
+    subprocess.call(["git", "checkout", "-b", branch_name])
```
flying-sheep, line 43:
```suggestion
    subprocess.call(["git", "switch", "-c", branch_name])
```

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+            "pr",
+            "create",
+            "--base",
+            base_branch,
+            "--head",
+            branch_name,
+            "--title",
+            pr_title,
+            "--body",
+            pr_description,
```
flying-sheep, line 61:
```suggestion
            f"--base={base_branch}",
            f"--head={branch_name}",
            f"--title={pr_title}",
            f"--body={pr_description}",
```

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,74 @@
...
+            "--dry-run" if args.dry_run else "",
```
flying-sheep, line 62: Please don’t add an empty positional argument here. You can use `*([x] if cond else [])` instead.

flying-sheep (changes requested):
Pretty straightforward, but you’re not checking the return codes. My Ruff updates should fix that.

hatch.toml
flying-sheep: what are you referring to? the `doc` feature will do what we need, no?

ci/scripts/towncrier_automation.py
ilan-gold: Ah this must have gotten lost in the sauce...seems like I actually lost one or two things.  I had `retcode` here

ilan-gold: Ah I must have committed to the wrong branch.  I also got rid of `call` right

hatch.toml
ilan-gold: Using https://github.com/scverse/anndata/pull/1625/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R86 instead i.e., `dev-doc`.  We don't need all of `doc`, no?

flying-sheep:
> Closed milestone opens an auto-merge PR that maybe has a branch title to trigger a release? Just some thoughts

yeah, that would work. or the other way around, whatever is easier.

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,75 @@
...
+        check=True,
+    )
+
+    # Enable auto-merge
+    if not args.dry_run:
+        subprocess.run(
+            ["gh", "pr", "merge", branch_name, "--auto", "--squash"], check=True
+        )
+    else:
+        print("Dry run, not merging")
```
flying-sheep, line 71: hm, but all the other stuff is being done, maybe “dry run” isn’t the right term here.

ilan-gold: I just use it because it matches `gh` terminology.  No idea why there isn't a `dry-run` option here.  Apparently getting one (looking at old issues) was non-trivial for `gh pr`

ilan-gold:
> that’s just a pair of parentheses actually :D

Ah true, it's just a ternary in there.  Was working on the bus haha

hatch.toml
flying-sheep: true, and that’s merged now.

you can just use the default env, already has towncrier in it.

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,89 @@
...
+
+class Args(argparse.Namespace):
+    version: str
+    dry_run: bool
+
+
+def parse_args(argv: Sequence[str] | None = None) -> Args:
+    parser = argparse.ArgumentParser(
+        prog="towncrier-automation",
+        description=(
```
flying-sheep, line 22: I removed `usage` here, it’s autogenerated, and can be accessed via `-h`/`--help`

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,89 @@
...
+            "This script runs towncrier for a given version, "
+            "creates a branch off of the current one, "
+            "and then creates a PR into the original branch with the changes. "
+            "The PR will be backported to main if the current branch is not main."
+        ),
+    )
+    parser.add_argument("version", type=str, help="What the new version is")
```
flying-sheep, line 29: I changed this to a positional parameter. I’d generally do positional parameters for the single mandatory thing the command does. In this case releasing a version.

Before my change, if you didn’t specify `--version`, you’d get “TypeError: Expected str, not None” on the `version.parse(args.version)` line. Now you get “Missing mandatory parameter `version`”

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument(
+        "--dry-run",
+        help="Whether or not to dry-run the actual creation of the pull request",
+        action="store_true",
+    )
+    args = parser.parse_args(argv, Args())
+    version.parse(args.version)  # validate
```
flying-sheep, line 36: What happens when people specify `1.11` here? will it all still work, e.g. will the built file be `1.11.0.md`?

ilan-gold: Good call! Validated!

ci/scripts/towncrier_automation.py
```diff
@@ -0,0 +1,100 @@
...
+            "creates a branch off of the current one, "
+            "and then creates a PR into the original branch with the changes. "
+            "The PR will be backported to main if the current branch is not main."
+        ),
+    )
+    parser.add_argument(
+        "version",
+        type=str,
+        help=(
+            "The new version for the release must have at least three parts, like `major.minor.patch` and no `major.minor`."
```
flying-sheep, line 33:
```suggestion
            "The new version for the release must have at least three parts, like `major.minor.patch` and no `major.minor`. "
```

flying-sheep (approved):

PR #1627, opened 2024-08-29
Add hatch-test env customization

flying-sheep:
This allows us to locally test with an approximation of our online test envs:

```console
$ hatch test # regular, stable resolution. Also `hatch test -i deps=stable`
$ hatch test -i deps=pre  # pre-releases
$ hatch test -i deps=min  # Python 3.0 and minimum possible dep versions.
$ hatch test --all
```

Approximation because `UV_RESOLUTION=lowest-direct` doesn’t directly match our min-deps ci script. Maybe we should get rid of the custom version number semantics in that one and just switch to `UV_RESOLUTION=lowest-direct`

---

I want this because I had a custom `hatch.toml` that got nuked when we added one to the branch, oops!

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -9,3 +9,19 @@ dependencies = ["setuptools"]  # https://bitbucket.org/pybtex-devs/pybtex/issues
 [envs.docs.scripts]
 build = "sphinx-build -M html docs docs/_build -W --keep-going {args}"
 clean = "git clean -fX -- docs"
+
+[envs.hatch-test]
+default-args = []
+extra-dependencies = ["ipykernel"]
+features = ["dev", "test"]
+overrides.matrix.deps.env-vars = [
+    { key = "UV_PRERELEASE", value = "allow", if = ["pre"] },
+    { key = "UV_RESOLUTION", value = "lowest-direct", if = ["min"] },
+]
+overrides.matrix.deps.python = [
+    { if = ["min"], value = "3.9" },
+    { if = ["stable", "pre"], value = "3.12" },
+]
+
+[[envs.hatch-test.matrix]]
+deps = ["stable", "pre", "min"]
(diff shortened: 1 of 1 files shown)


hatch.toml
```diff
@@ -9,3 +9,19 @@ dependencies = ["setuptools"]  # https://bitbucket.org/pybtex-devs/pybtex/issues
 [envs.docs.scripts]
 build = "sphinx-build -M html docs docs/_build -W --keep-going {args}"
 clean = "git clean -fX -- docs"
+
+[envs.hatch-test]
+default-args = []
+extra-dependencies = ["ipykernel"]
```
ilan-gold, line 15: Why `ipykernel`?

hatch.toml
```diff
@@ -9,3 +9,19 @@ dependencies = ["setuptools"]  # https://bitbucket.org/pybtex-devs/pybtex/issues
 [envs.docs.scripts]
 build = "sphinx-build -M html docs docs/_build -W --keep-going {args}"
 clean = "git clean -fX -- docs"
+
+[envs.hatch-test]
```
ilan-gold, line 13: Why not use `uv` by default? Is this inherited from `default`?

hatch.toml
flying-sheep: Allows you to use this env as kernel for notebooks in VS Code.

hatch.toml
flying-sheep: - all envs *except* the `hatch-*` ones inherit from default, that’s why this works without `envs.towncrier.features` being defined: https://github.com/scverse/anndata/blob/94b2304cb2c08dd2c5f8d3cb7ee4da39d874975a/hatch.toml#L13-L15
- the `hatch-*` envs have `uv` by default

ilan-gold (approved):

PR #1028, opened 2023-06-27
(feat): Support for scipy.sparse.{csr,csc}_array

ivirshup:
Starting support for scipy.sparse csc_array and csr_array. These are the new array-like interfaces for sparse matrices from scipy.

Fixes #746, #859

Still very WIP

TODO:

 - [x] Figure out compat with scipy<1.11
 - [ ] Demo of how to load an array from disk
 - [x] Support for `.X`
 - [x] Tests (figure out which ones need adding)
 - [x] Fix concatenation tests
 - [x] Fix view tests
 - [x] Test for indices dtype after concatenation (e.g. #859)
 - [x] Release note
 - [x] Remove support for `dask.array[sparse.sparray]`

PR #1633, opened 2024-08-30
(feat): allow reading of `scipy.sparse.cs{c,r}_array`

ilan-gold:
- [x] Discovered in https://github.com/scverse/anndata/pull/1608#discussion_r1738686391 and following up on https://github.com/scverse/anndata/pull/1028#discussion_r1559858337
- [x] Tests added


src/anndata/_core/sparse_dataset.py
```diff
@@ -358,10 +361,12 @@ class BaseCompressedSparseDataset(ABC):
     format: Literal["csr", "csc"]
     """The format of the sparse matrix."""
     _group: GroupStorageType
+    _shall_use_sparse_array_on_read: bool
 
     def __init__(self, group: GroupStorageType):
         type(self)._check_group_format(group)
         self._group = group
+        self._shall_use_sparse_array_on_read = settings.shall_use_sparse_array_on_read
```
ilan-gold, line 369: Up for debate whether this should be dynamic or not.  I think once the class is initiated, so too should this property

flying-sheep: I think it would be pretty surprising if the setting “does nothing”

And if someone changes this setting in the middle of a loop while they’re operating on an AnnData object, I think they’re kinda asking for weird things to happen?

ilan-gold: Sorry, not sure I follow.  Those seem like like they are opposites.  If someone changes the setting while holding on to the object still, you think the behavior of the object (`CS{R,C}Dataset`) should change?  Or the opposite, that it should not because that is asking for trouble?

flying-sheep: I mean code like this: what would you expect `m` to be?

```py
adata = ad.read_backed(...)  # this line would lock in `adata.X._shall_use_sparse_array_on_read`

# several notebook cells later
subset = adata[adata.obs["sample"] == 1]
with ad.settings.override(shall_use_sparse_array_on_read=True):
    m = subset.X.to_memory()
```

ilan-gold: Ok, this makes sense.  Should be dynamic then!

flying-sheep (approved):

src/anndata/_core/sparse_dataset.py
```diff
@@ -313,6 +316,13 @@ def mean_slice_length(slices):
 def get_memory_class(
     format: Literal["csr", "csc"], *, use_sparray_in_io: bool = False
 ) -> type[_cs_matrix]:
+    if not CAN_USE_SPARSE_ARRAY and use_sparray_in_io:
+        msg = (
+            "scipy.sparse.cs{r,c}array is not available in current scipy version."
+            "falling back to scipy.sparse.spmatrix for reading"
+        )
+        warnings.warn(msg, UserWarning, stacklevel=2)
+        use_sparray_in_io = False
```
flying-sheep, line 325: why not just put that into the setting.

if someone tries to set the setting to `True` when sparse arrays aren’t available, just throw an error instead.

src/anndata/_core/sparse_dataset.py
```diff
@@ -317,7 +317,10 @@ def get_memory_class(
     format: Literal["csr", "csc"], *, use_sparray_in_io: bool = False
 ) -> type[_cs_matrix]:
     if not CAN_USE_SPARSE_ARRAY and use_sparray_in_io:
-        msg = "scipy.sparse.cs{r,c}array is not available in current scipy version, falling back to scipy.sparse.spmatrix for reading"
+        msg = (
+            "scipy.sparse.cs{r,c}array is not available in current scipy version."
```
flying-sheep, line 321: don’t forget that you have to add a space after the first `.` in multiline messages

flying-sheep (approved):

PR #1635, opened 2024-09-02
(fix): ensure all indexing methods return correct format on backed sparse

ilan-gold:
- [x] Follow up #1633 
- [x] Tests added

### DIFF
--- a/tests/test_backed_sparse.py
+++ b/tests/test_backed_sparse.py
@@ -270,13 +270,41 @@ def test_dataset_append_memory(
 
 
 @pytest.mark.parametrize("sparse_format", [sparse.csr_matrix, sparse.csc_matrix])
+@pytest.mark.parametrize(
+    ("subset_func", "subset_func2"),
+    map(
+        lambda x: pytest.param(*x),
+        product(
+            *(
+                [
+                    [
+                        ad.tests.helpers.array_subset,
+                        ad.tests.helpers.slice_subset,
+                        ad.tests.helpers.array_int_subset,
+                        ad.tests.helpers.array_bool_subset,
+                        lambda x: make_randomized_mask(len(x)),
+                        lambda x: make_alternating_mask_15(len(x)),
+                        lambda x: make_alternating_mask_5(len(x)),
+                        lambda x: make_one_group_mask(len(x)),
+                        lambda x: make_one_elem_mask(len(x)),
+                    ]
+                ]
+                * 2
+            )
+        ),
+    ),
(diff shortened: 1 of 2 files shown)


tests/test_backed_sparse.py
```diff
@@ -270,13 +270,41 @@ def test_dataset_append_memory(
 
 
 @pytest.mark.parametrize("sparse_format", [sparse.csr_matrix, sparse.csc_matrix])
+@pytest.mark.parametrize(
+    ("subset_func", "subset_func2"),
+    map(
+        lambda x: pytest.param(*x),
+        product(
```
flying-sheep, line 277: - you don’t need `pytest.param` if you don’t use its keyword arguments, you can just use tuples for that. And `product` already emits tuples
- (no longer relevant if you just use plain `product`, but it’s more pythonic to use comprehensions instead of map with a lambda)
- `product` has a `repeat` parameter that you can use for much less confusing brace syntax.

tests/test_backed_sparse.py
```diff
@@ -270,13 +270,34 @@ def test_dataset_append_memory(
...
+            ad.tests.helpers.array_subset,
+            ad.tests.helpers.slice_subset,
+            ad.tests.helpers.array_int_subset,
+            ad.tests.helpers.array_bool_subset,
+            lambda x: make_randomized_mask(len(x)),
+            lambda x: make_alternating_mask_15(len(x)),
+            lambda x: make_alternating_mask_5(len(x)),
+            lambda x: make_one_group_mask(len(x)),
+            lambda x: make_one_elem_mask(len(x)),
+        ],
```
flying-sheep, line 286: Should `anndata.tests.helpers.subset_func` be changed to have these parameters?

If no, why are there two different parameter lists of subset functions? Can we put them next to each other and give them speaking names explaining why there are two lists?

flying-sheep (commented):
looks good, I‘m just confused why this file contains more ways to make masks that aren’t covered in `subset_func`

tests/test_backed_sparse.py
ilan-gold: So these functions were meant to hit (or not) the optimized path within `subset_by_major_axis_mask`, but we don't need to go that deep here so I will these (the coverage in `__getitem__` is the same without these).  They are used in the `test_consecutive_bool` specifically for this purpose where we a mocker/spy to check for optimized path.

flying-sheep (approved):

PR #1637, opened 2024-09-02
(chore): ensure views of anndata produce distinguishable view classes

ilan-gold:
- [ ] Closes #
- [ ] Tests added


ilan-gold:
@flying-sheep Happy to add a protocol here for having `base` if you think that's the way to go.  I am a bit ambivalent - this solution is perhaps less robust but requires exposing less API surface area.

flying-sheep:
Hmm, good question.

I think *if possible* I’d prefer exporting `_SetItemMixin` under a different name. Maybe something containing `COW` (copy-on-write) to clarify semantics?

But it might not be possible to do that since https://github.com/python/typing/issues/213 isn’t implemented yet. We might want to add our use case to https://github.com/CarliJoy/intersection_examples/issues

Issue #1634, opened 2024-09-02
0.11 API


flying-sheep:
# new top-level

- [x] #1644
- [x] #1641

# .experimental
 
- [x] #1642
- [x] `experimental.read_elem_as_dask` → `experimental.read_elem_lazy`? Not yet: https://github.com/scverse/anndata/issues/1634#issuecomment-2325894399
- [x] #1643

# new modules

- [x] #1645

ilan-gold:
> idea: move read_zarr in together with .h5ad as “native file format”

So one `read` function?  We then need a dependency on `zarr` then probably

> experimental.sparse_dataset needs to come to the same place as CS{RC}Dataset or be removed.

Exported then, it is fairly stable.

> Maybe .arrays and .types submodule?

Hmmm, I like `types` but `arrays` is not great if we're also going to export views, maybe some xarray dataset stuff etc.  Better name, I'm all for it.  But even there, I don't want to actually export anything beyond what's needed for type checking and the bare minimum needed for basic usage (only `CS{R,C}dataset` ATM)

> [InMemoryElem](https://anndata.readthedocs.io/en/latest/generated/anndata.InMemoryElem.html#anndata.InMemoryElem) needs a docstring for a nicer list entry on the API page (like RWAble)

Yes ok!

flying-sheep:
> So one read function?

no, just move it up in the docs:

![grafik](https://github.com/user-attachments/assets/85c606a6-13ab-4237-9458-98638b28c49b)

> Exported then, it is fairly stable.

It’s 3 lines, I don’t know if we need to have it at all. People could trivially re-implement it.

But I’m also OK with exporting it.

> Better name, I'm all for it. 

ah yes, the famous second-hardest thing in computer science: naming things. We have the following categories of types that we export or might want to export:

- Containers (owning data): AnnData, AlignedMapping
- Views (referencing data in a container): AnnData(!), CupyArrayView, CSRArrayView, … (all(?) identical or inheriting from the parent array)
- Datasets (referencing data on disk): CSRDataset, …

Am I missing something? E.g. we have actual in-memory array containers?

`np.ndarray` and `AnnData` don’t have dedicated view classes, so maybe we should just export a protocol named `View` with the [np.ndarray.base](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.base.html) API? If there is just one `View` protocol / abstract base class for all array types we support, it could go wherever.

ilan-gold:
[...]

PR #1642, opened 2024-09-03
(chore): export `sparse_dataset` from the core

ilan-gold:
- [ ] Towards #1634
- [ ] Tests added

### DIFF
--- a/benchmarks/benchmarks/sparse_dataset.py
+++ b/benchmarks/benchmarks/sparse_dataset.py
@@ -4,9 +4,8 @@ import numpy as np
 import zarr
 from scipy import sparse
 
-from anndata import AnnData
+from anndata import AnnData, sparse_dataset
 from anndata._io.specs import write_elem
-from anndata.experimental import sparse_dataset
 
 
 def make_alternating_mask(n):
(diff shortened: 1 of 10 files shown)


benchmarks/benchmarks/sparse_dataset.py
```diff
@@ -5,8 +5,8 @@
 from scipy import sparse
 
 from anndata import AnnData
+from anndata._core.sparse_dataset import sparse_dataset
 from anndata._io.specs import write_elem
```
flying-sheep, line 9:
```suggestion
from anndata import AnnData, sparse_dataset, write_elem
```

flying-sheep (approved):

benchmarks/benchmarks/sparse_dataset.py
ilan-gold: I _think_ the benchmark needs this file to be stable? https://github.com/scverse/anndata/actions/runs/10680419101/job/29601783819

flying-sheep: ah of course, if we want to run it with old commits. OK.

PR #1645, opened 2024-09-03
add `.abc` and `.typing` modules

### LINKED ISSUE #1634: 0.11 API

flying-sheep:
Fixes #1634

TODO:

- [x] Add actual API we want to commit to to `abc.*`

### DIFF
--- a/src/anndata/abc.py
+++ b/src/anndata/abc.py
@@ -12,6 +12,9 @@ if TYPE_CHECKING:
     from .compat import SpArray
 
 
+__all__ = ["CSRDataset", "CSCDataset"]
+
+
 class _AbstractCSDataset(ABC):
     """Base for the public API for CSRDataset/CSCDataset."""
 
@@ -57,9 +60,3 @@ class CSRDataset(_AbstractCSDataset, ABC):
 class CSCDataset(_AbstractCSDataset, ABC):
     __doc__ = _sparse_dataset_doc.format(format="CSC")
     format = "csc"
-
-
-__all__ = [
-    "CSRDataset",
-    "CSCDataset",
-]
--- a/src/anndata/abc.py
+++ b/src/anndata/abc.py
@@ -1,0 +1,9 @@
+from __future__ import annotations
+
+from ._core.sparse_dataset import AbstractCSCDataset as CSCDataset
+from ._core.sparse_dataset import AbstractCSRDataset as CSRDataset
+
+__all__ = [
+    "CSRDataset",
+    "CSCDataset",
+]
--- a/src/anndata/abc.py
+++ b/src/anndata/abc.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
-from abc import ABC
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from typing import ClassVar, Literal
 
     import numpy as np
+    from scipy.sparse import csc_matrix, csr_matrix
+
+    from .compat import SpArray
 
 
 class _AbstractCSDataset(ABC):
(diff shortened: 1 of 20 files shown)


ilan-gold:
> Add actual API we want to commit to to abc.*

I made an attempt at what I thought was reasonable via `BaseCompressedSparseDataset`'s private/`_` methods but I would also be ok with a first pass at this rolling that back further to not include `append`, `__setitem__`, or `group` i.e., people should be able to "get" (either `__getitem__` or `to_memory()`),  ask for the backing type (`zarr` or `hdf5`), know the `dtype`, and get the shape.

ilan-gold:
After all, easier to add than remove...

flying-sheep:
We could export the `Index` type for `__getitem__` or some simplification.

ilan-gold:
> We could export the Index type for __getitem__ or some simplification.

Absolutely.  Maybe a `types` module then here as well?

ilan-gold:
Either that, or punt on `types` until 0.12.  I don't have a preference there.

src/anndata/abc.py
```diff
@@ -0,0 +1,62 @@
...
+    shape: tuple[int, int]
+    """Shape of the matrix."""
+
+    dtype: np.dtype
+    """The :class:`numpy.dtype` of the `data` attribute of the sparse matrix."""
+
+    backend: Literal["zarr", "hdf5"]
+    """Which file type is used on-disk."""
+
+    # TODO: index type
```
ilan-gold, line 33: Export that in `types`? I noticed it in a few spots in the github search for `_code`

src/anndata/typing.py
```diff
@@ -0,0 +1,64 @@
...
+    AwkArray,
+    H5Array,
+    ZarrArray,
+    ZappyArray,
+    abc.CSRDataset,
+    abc.CSCDataset,
+    DaskArray,
+    CupyArray,
+    CupySparseMatrix,
+]
```
ilan-gold, line 45: What's supposed to be the difference between this file and `_types` if this isn't public?

ilan-gold (changes requested):
Please look at #1647 or fix CI some other way!

src/anndata/abc.py
flying-sheep: Do you mean `_types` and `_core`?

flying-sheep:
> Please look at https://github.com/scverse/anndata/pull/1647 or fix CI some other way!

whoop, seems like we did the same thing!

ilan-gold:
> whoop, seems like we did the same thing!

Haha, but yours is much cleaner!

src/anndata/typing.py
flying-sheep: Everything in here is public, just not all names.

There are a few internal APIs that use subsets of the public types, and these are included in here too.

I’d have put these into `_types` if Python was able to deal with this kind of circularity.

So I guess the logic is “all public types and the types they depend on?

flying-sheep:
I wouldn’t say so, I just simplified the error message again (after making it more complex in another PR :see_no_evil:)

src/anndata/abc.py
ilan-gold: Sorry! `typing`!

ilan-gold: So we should export `Index` in `typing`

src/anndata/typing.py
ilan-gold: Sounds good!

ilan-gold (approved):

PR #1648, opened 2024-09-04
(chore): add `Index` type to `__getitem__` for sparse dataset

ilan-gold:
Starting point.  I think this should go in the next 0.11 release candidate (maybe along with your doc rendering fix @flying-sheep but doesn't have to).  I would also be in favor of exporting this type since it is used elsewhere in quite important places.
- [ ] Closes #
- [ ] Tests added


flying-sheep:
With “doc rendering fix”, you mean making it so “TypeVar” gets rendered more nicely or … ?

ilan-gold:
> With “doc rendering fix”, you mean making it so “TypeVar” gets rendered more nicely or … ?

YEs you had mentioned wanting it, and it maybe being a blocker?

src/anndata/abc.py
```diff
@@ -30,10 +30,19 @@ class _AbstractCSDataset(ABC):
...
+        """Load a slice or an element from the sparse dataset into memory.
+
+        Parameters
+        ----------
+        index
+            Index to load.
+
+        Returns
+        -------
+            The desired data read off disk.
```
flying-sheep, line 44:
```suggestion
        The desired data read off disk.
```

flying-sheep (approved):

PR #1654, opened 2024-09-04
(fix): towncrier automation fixes

ilan-gold:
- [ ] Closes #
- [ ] Tests added


flying-sheep:
Why is is necessary to uglify this?

Does Git have such a bad commandline parser that it doesn’t support this?

Issue #634, opened 2021-10-29
HDF5 Cloud Storage


jreadey:
Is there interest in storing data in the cloud?  E.g. using AWS S3.  With HDF5 this is problematic since h5py requires a posix-based filesystem.  I maintain the h5pyd project (https://github.com/HDFGroup/h5pyd) which gets around this by providing a h5py compatible api to a sharded data store (similar to zarr).  I think it should be possible to have h5ad support either h5py or h5pyd, but first wanted to gauge interest in this approach.
Thanks!

ivirshup:
Hey! We definitely have interest in cloud based storage, but so far have been largely eyeing Zarr and possibly arrow for this. This looks interesting. My initial thoughts compared to the other formats:

Strengths:

* hdf5 is our most common backend, most common way files are distributed
* Good cross language support compared to other cloud friendly formats

Weaknesses:

* Requires HDF5 service running on the server side, not sure how hard/ common this will be

Needs investigation:

* Drop in support?
* How does performance look?

Could be of interest to @ryan-williams, @ilan-gold, @joshua-gould

ilan-gold:
@ivirshup I haven't had time to dig into this on the AnnData side yet, but a good starting point here for remote data (wihtout switching to a new hdf5 reader, which is what what this issue seems to be about) would be to literally just try passing in a URL to a zarr-backed AnnData store to the `AnnData`constructor since (I think) `zarr` supports URLs natively.  For example, this line makes me think that this should just work out of the box:
https://github.com/theislab/anndata/blob/0ec97410702f61b29acef80b6377d1232699fa94/anndata/_io/zarr.py#L252

So for example, if you have an AnnData store you could do
```python
adata.write_zarr('path/to/local/my_store.zarr')
```
Then in a shell (since I don't know offhand in `python`)
```shell
gsutil cp -r path/to/local/my_store.zarr gs://my_bucket/
```
And then
```python
import anndata as ad
import aiohttp, requests, zarr, fsspec
adata_cloud = ad.read_zarr(my_google_url)
```
This doesn't seem to work for my examples (I get an empty `AnnData` store from https://storage.googleapis.com/vitessce-demo-data/anndata-test/pbmc3k_processed.zarr although there is no error) but this would be the idea if it worked.  Maybe I missed something though.  I'll do some more digging.  Despite this apparently not working with remote zarr stores (I really think 
[...]

PR #1665, opened 2024-09-04
(feat): custom `reopen` with `read_elem_as_dask` for remote `h5ad`

ilan-gold:
Not sure about the typing on `reopen` (with `Generator[...]` as before, it was complaining, but a normal `Callable` with ` return` felt wrong); however, the following seems to work:

```python
import dask.distributed as dd
import remfile
import h5py
from anndata.experimental import read_elem_as_dask

# optionally log all requests to see roughly that this is lazy https://stackoverflow.com/a/24588289/8060591
import logging
import contextlib
from http.client import HTTPConnection

def debug_all_requests():

    HTTPConnection.debuglevel = 1
    logging.basicConfig()
    logging.getLogger().setLevel(logging.DEBUG)
    requests_log = logging.getLogger("requests.packages.urllib3")
    requests_log.setLevel(logging.DEBUG)
    requests_log.propagate = True

def debug_no_requests():

    HTTPConnection.debuglevel = 0
    root_logger = logging.getLogger()
    root_logger.setLevel(logging.WARNING)
    root_logger.handlers = []
    requests_log = logging.getLogger("requests.packages.urllib3")
    requests_log.setLevel(logging.WARNING)
    requests_log.propagate = False

ADATA_URI = "https://allen-brain-cell-atlas.s3.us-west-2.amazonaws.com/expression_matrices/WMB-10Xv2/20230630/WMB-10Xv2-TH-log2.h5ad"

cluster = dd.LocalCluster(n_workers=1, threads_per_worker=1)
client = dd.Client(cluster)

file_h5 = h5py.File(remfile.File(ADATA_URI), "r")
def reopen():
   yield h5py.File(remfile.File(ADATA_URI), "r")["X"]

elem = read_elem_as_dask(file_h5["X"], reopen=reopen)
elem[:20, :].compute()
```

The question is whether or not we'd want to internalize this - my guess is "no."  I think we should only promise support for core zarr/hdf5.  The flip side then is how do we test this....good questions to grapple with!

- [x] Towards https://github.com/scverse/anndata/issues/634#issuecomment-1906214981 with `dask`
- [ ] Tests added


ilan-gold:
I should note that I could not figure out how to extract the `remfile.RemFile.RemFile` object out of the `h5py.File` object. If I could, maybe no need for a `reopen` param...

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -36,7 +36,7 @@
 @contextmanager
 def maybe_open_h5(
     path_or_group: Path | ZarrGroup, elem_name: str
-) -> Generator[StorageType, None, None]:
+) -> Callable[[], Iterator[StorageType]]:
```
flying-sheep, line 39: The previous type was correct, the new one isn’t.

The way decorators interact with Python is that you normally type the decorated function (e.g. if it contains `yield`, the return type is `Generator`). The decorator than transforms the function from whatever it is to whatever the decorator wants.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -67,13 +67,18 @@ def make_dask_chunk(
     *,
     wrap: Callable[[ArrayStorageType], ArrayStorageType]
     | Callable[[H5Group | ZarrGroup], _CSRDataset | _CSCDataset] = lambda g: g,
+    reopen: None | Callable[[], Iterator[StorageType]] = None,
```
flying-sheep, line 70: so the idea is “reopen is a callable that can be transformed into a context manager using `contextlib.contextmanager`.

Why not just “reopen is a callable that returns an [`contextlib.AbstractContextManager[StorageType]`](https://docs.python.org/3/library/contextlib.html#contextlib.AbstractContextManager)?

ivirshup:
Do you know what the difference is between what `remfile` and `fsspec` are doing?

ilan-gold:
@ivirshup My understanding is that `fsspec` + `kerchunk` requires writing a separate JSON index file: https://fsspec.github.io/kerchunk/reference.html#kerchunk.hdf.SingleHdf5ToZarr.  `remfile` reads directly.

ilan-gold:
As an aside here: even if we use `fsspec` + `hfpy` i.e., `fsspec`'s file object, this sort of thing is still necessary.  So the need for the PR remains

PR #1668, opened 2024-09-05
Revert uglification of CLI options

flying-sheep:
@ilan-gold in #1654, this was made uglier.

My local version of `gh` has no problem with the cleaner version of CLI option specification, so why not like this?


ilan-gold (approved):
Apologies! was very frustrated and trying many different things but the dry run was working so there was no need to try this!

flying-sheep:
OK. Please also check out e844f186dc8241df1d4dcfd711009877fd07040b, the validation was broken too!

(`len(Version(...).release) == 3` is always True, we must match the raw string)

ilan-gold:
> (len(Version(...).release) == 3 is always True, we must match the raw string)

Maybe I'm missing something:

```python
>>> from packaging import version
>>> version.Version("1.0")
<Version('1.0')>
>>> version.Version("1.0").release
(1, 0)
```

ci/scripts/towncrier_automation.py
```diff
@@ -40,11 +44,12 @@ def parse_args(argv: Sequence[str] | None = None) -> Args:
...
-        raise ValueError(
-            f"Version argument {args.version} must contain major, minor, and patch version."
-        )
-    version.parse(args.version)  # validate
+    if (match := VERSION.fullmatch(args.version)) is None:
+        msg = f"Version argument {args.version} is not a valid version."
+        raise ValueError(msg)
+    if len(match["release"].split(".")) < 3:
+        msg = f"Version argument {args.version} must contain major, minor, and patch version."
+        raise ValueError(msg)
```
ilan-gold, line 52: Am I doing something wrong here?


```python
>>> import re
>>> from packaging import version
>>> from packaging.version import VERSION_PATTERN

>>> pattern = re.compile(VERSION_PATTERN, re.VERBOSE | re.IGNORECASE)
>>> pattern.fullmatch("1.0.0.0") # not None?
<re.Match object; span=(0, 7), match='1.0.0.0'>
>>> version.Version("1.0.0.0").release
(1, 0, 0, 0)
```

ilan-gold: So it seems like what I had before was right, no?

flying-sheep: huh, indeed! I think I confused that with the fact that `Version.patch` will be `0` no matter if it was supplied or not.

Will revert, sorry!

PR #1677, opened 2024-09-17
(chore): add `dask` optional dep

ilan-gold:
- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,8 @@ dev-test = ["pytest-xdist"] # local test speedups
 gpu = ["cupy"]
 cu12 = ["cupy-cuda12x"]
 cu11 = ["cupy-cuda11x"]
+# https://github.com/dask/dask/issues/11290
+dask = ["dask[array]>=2022.09.2,<2024.8.0"]
 
 [tool.hatch.version]
 source = "vcs"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,11 +96,12 @@ test = [
     "boltons",
     "scanpy",
     "httpx", # For data downloading
-    "dask[array,distributed]>=2022.09.2,<2024.8.0",
+    "dask[distributed]>=2022.09.2",
     "awkward>=2.3",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -96,16 +96,19 @@ test = [
     "boltons",
     "scanpy",
     "httpx", # For data downloading
-    "dask[array,distributed]>=2022.09.2,<2024.8.0",
+    "dask[distributed]>=2022.09.2",
```
flying-sheep, line 99: The constraint is already there through `anndata[dask]`, no need to add it again.

```suggestion
    "dask[distributed]",
```

pyproject.toml
```diff
@@ -96,16 +96,19 @@ test = [
...
     "awkward>=2.3",
     "pyarrow",
     "pytest_memray",
-    "pytest-mock"
+    "pytest-mock",
+    "anndata[dask]"
```
flying-sheep, line 104: why do we still miss some trailing commas? Time to format our TOML I think.

```suggestion
    "anndata[dask]",
```

flying-sheep (changes requested):

pyproject.toml
ilan-gold: Are `distributed` and `array` linked?

ilan-gold: i.e.I think you can install different versions of each?  Just tested this.

ilan-gold: Hmm, no `dask[array]` is _not_ a separate package, ok.

ilan-gold: Or rather, `distributed` is a separate package, but hooks back to `dask`

ilan-gold: Hence the confusion

flying-sheep: `dask[distributed]>=2022.09.2` means “install dask >=2022.09.2 with the extra `distributed` active”

flying-sheep (approved):

PR #1682, opened 2024-09-20
(feat): `io` submodule

### LINKED ISSUE #1681: `io` submodule

We should have a public submodule `io` since we're starting to grow our IO api.

This should include all public methods related to `io`, including:

* `read_elem`, `write_elem`
* `sparse_dataset`

Eventually I would also like whatever public API we have for the registries to be included here.

I think it would make sense to put some of the more esoteric reader functions in here, like `read_text`, `read_loom`, etc. However, I'm not so sure about `read_h5ad`, `read_zarr`. I think these could be exported from both the main namespace and `io`, but am not 100% sure on this.

ilan-gold:
A brief summary of how this was handled and why:

1. I moved everything into `io` related to i/o
2. I [left `read_zarr` and `read_h5ad` at the top-level](https://github.com/scverse/anndata/pull/1682/files#diff-222501f69879bbc6ce8adc36d9b9835e636af244de21e24e964671a574434f9eR84-R85) for now, although I think that should be deprecated in favor of everything going through `io` _in the future_ despite their "high" status above the others. They are now documented as being from `io` to begin to encourage that behavior although we aren't throwing warnings for e.g., `anndata.read_h5ad` yet
3. I also made a backwards compat for people who use `from anndata._io import blah` [since uhhhhhhh yikes](https://github.com/search?q=%22anndata._io%22&type=code) by leaving that module alone.  However, at least if anyone tries to import from the top-level `from anndata._io import read_h5ad`, it will warn.  Also, I added a public note about this issue since it seems pervasive.
4. I added tests for the deprecations.  For now, we need to check the scanpy imports for warnings but we can open the PR for scanpy before its release to adapt.  I think the imports are cached or something so `pytest.warns` can't run reliably every time because the warning doesn't come every time....not sure what's up with that

- [x] Closes #1681
- [x] Tests added


tests/test_readwrite.py
```diff
@@ -753,7 +756,10 @@ def test_scanpy_pbmc68k(tmp_path, diskfmt, diskfmt2):
 @pytest.mark.skipif(not find_spec("scanpy"), reason="Scanpy is not installed")
 def test_scanpy_krumsiek11(tmp_path, diskfmt):
     filepth = tmp_path / f"test.{diskfmt}"
-    import scanpy as sc
+    with pytest.warns(
+        FutureWarning, match=r"Importing read_.* from `anndata` is deprecated"
+    ):
+        import scanpy as sc
```
flying-sheep, line 762: just importing scanpy shouldn’t trigger a warning.

src/anndata/_core/anndata.py
```diff
@@ -995,8 +995,8 @@ def filename(self, filename: PathLike | None):
             # as the data is stored on disk, we can safely set self._X to None
             self._X = None
 
-    def _set_backed(self, attr, value):
-        from .._io.utils import write_attribute
+    def _set_baiod(self, attr, value):
```
flying-sheep, line 998: ?

tests/test_readwrite.py
ilan-gold: It imports `read_csv` etc. from top-level instead of `io` because the feature doesn't exist yet as far as `scanpy` is concerned

ilan-gold: Issue created with a to-do list

flying-sheep: Then these blocks should be changed to ignore the warnings instead of testing that they’re raised.

Otherwise these tests would start failing once things are fixed on the scanpy side.

docs/api.md
```diff
@@ -33,8 +33,8 @@ Reading anndata’s native formats `.h5ad` and `zarr`.
 .. autosummary::
    :toctree: generated/
 
-   read_h5ad
-   read_zarr
+   io.read_h5ad
+   io.read_zarr
```
flying-sheep, line 37: maybe we should document that those are also available as `anndata.read_*`?

flying-sheep (changes requested):
Very nice and clean! We should add a release note fragment for this change.

We do turn `FutureWarnings` into errors right? So it’s impossible that you missed one in here, right?

tests/test_readwrite.py
flying-sheep: Ah, you ended up doing that here anyway: 1174453a4cc7fb5be135548f531b6cf895a3cae9

docs/api.md
ilan-gold: I think we shouldn't.   I think we should be encouraging people to use `io`.  We won't break their code for a long time if they continue to use `anndata.read_{zarr,h5ad}` (currently these are the only two from the top level _not_ deprecated).  But I still think we should be encouraging this.  Maybe a note in their docstrings then?

tests/test_readwrite.py
ilan-gold: Done!

ilan-gold:
> We do turn FutureWarnings into errors right? So it’s impossible that you missed one in here, right?

Could you elaborate? When I search `FutureWarning` in our code base, I don't see anything that might signal they are errors, and a lot of `pytest.warns`.

flying-sheep:
One of our test job runs with `--strict-warnings`, which turns all warnings into errors except the ones from an allowlist (that exists because we haven’t gotten around to dealing with them).

These are the default filters:

https://github.com/scverse/anndata/blob/34e9783f955a1121850e9e36dd35ad9adbecbb2a/src/testing/anndata/_pytest.py#L26-L31

This is the allowlist:

https://github.com/scverse/anndata/blob/34e9783f955a1121850e9e36dd35ad9adbecbb2a/pyproject.toml#L148-L154

This is the code that turns all warnings into errors except the above.

https://github.com/scverse/anndata/blob/34e9783f955a1121850e9e36dd35ad9adbecbb2a/src/testing/anndata/_pytest.py#L97-L108

README.md
```diff
@@ -36,6 +36,10 @@ Please consider making a tax-deductible [donation](https://numfocus.org/donate-t
   >
 </a>
 
+## Public API
+
+Our public API is documented [in the API section of these docs](https://anndata.readthedocs.io/en/stable/api.html).  We cannot guarantee the stability of our internal APIs, whether it's the location of a function, its arguments, or something else.  In other words, we do not officially support (or encourage users to do) something like `from anndata._core import AnnData` as `_core` is both not documented and [contains a leading underscore](https://peps.python.org/pep-0008/#public-and-internal-interfaces).  However, we are aware that [many users do use these internal APIs](https://github.com/search?q=%22anndata._io%22&type=code) and thus encourage them to open an [issue](https://github.com/scverse/anndata/issues/new/choose) or migrate to the public API.  That is, if something is missing from out public API as documented, for example a feature you wish to be exported publicly, please open an issue.
```
flying-sheep, line 41: please

- maximally one sentence per source line
- no inline links
- PEP links are written ``{pep}`number` ``

src/anndata/__init__.py
```diff
@@ -71,23 +86,15 @@ def read(*args, **kwargs):
     "abc",
     "experimental",
     "typing",
+    "_io",
```
flying-sheep, line 89:
```suggestion
```

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -120,7 +120,7 @@ def read_sparse_as_dask(
     )
     memory_format = sparse.csc_matrix if is_csc else sparse.csr_matrix
     make_chunk = partial(
-        make_dask_chunk, path_or_group, elem_name, wrap=ad.sparse_dataset
+        make_dask_chunk, path_or_group, elem_name, wrap=ad._io.sparse_dataset
```
flying-sheep, line 123: please either access them from their public location or `from .. import sparse_dataset`

src/anndata/_io/__init__.py
flying-sheep, line 1: where did the deprecation `def __getattr__` go?

docs/release-notes/1682.feat.md
```diff
@@ -0,0 +1 @@
+A new {mod}`anndata.io` module contains all `read_*` and `write_*` functions, and all imports of such functions should go through this module.  Old ways of importing these functions i.e., `from anndata import read_csv` or `from anndata._io.specs import read_elem` will still work, but are now considered deprecated and give a warning on import. with the exception of {func}`anndata.io.read_zarr` and {func}`anndata.io.read_h5ad`, which will remain at the top-level `anndata` without warning.  However, they will be eventually deprecated, and will start to a throw a warning in a future release. `user`{ilan-gold}
```
flying-sheep, line 1: Please also max. one line per sentence.

> However, they will be eventually deprecated, and will start to a throw a warning in a future release.

is that so?

ilan-gold:
Sorry at @flying-sheep yes I knew that about `stricto-warnings`, but I meant about `FutureWarning` specifically, which is what confused me.   In any case I misread "impossible" as "possible" hence my confusion.  So yes, it's impossible we missed

src/anndata/_io/__init__.py
ilan-gold: Coming back....

docs/release-notes/1682.feat.md
ilan-gold: I would like to do so but no need to state it here since it isn't even happening yet.  Conversation for another time, good point.

README.md
ilan-gold: > PEP links are written {pep}`number` 

Any way to get the specific section?  I want people to go right there.  Also any docs about this stuff you could link to would be helpful.  I have trouble navigating the myst (I think that's where this comes from?) docs.

README.md
```diff
@@ -36,6 +36,20 @@ Please consider making a tax-deductible [donation](https://numfocus.org/donate-t
   >
 </a>
 
+## Public API
+
+Our public API is documented in the [API section][] of these docs.
```
flying-sheep, line 41: please use roles for internal links: `func` for functions and so on.

Here, `/docs/api.md` becomes a `doc` with identifier `api`.

And again: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-pep

README.md
flying-sheep: Since this is the README, if we want this to be readable on the github page, we can’t use that stuff anyway sorry for the misdirection (we could use it if we used rST instead of MyST …)

For the record, you can use anchors: ``{pep}`number#anchor` ``

ilan-gold: > For the record, you can use anchors: {pep}`number#anchor` 

Cool, I suspected as much but wasn't sure.  Tough that I don't see much in the docs about it... https://myst-parser.readthedocs.io/en/latest/search.html?q=pep

ilan-gold: Maybe it's an rst thing? Would be great to understand more deeply

docs/api.md
flying-sheep: I guess since we import it by default, it’s just 3 more characters to type and no other change.

src/anndata/_io/__init__.py
```diff
@@ -1,40 +1,58 @@
...
+    from .specs import read_elem, write_elem
+    from .write import write_csvs, write_loom
+
+    def write_zarr(*args, **kw):
+        from .zarr import write_zarr
+
+        return write_zarr(*args, **kw)
+
+    # We use this in test by attribute access
+    from . import specs  # noqa: F401, E402
```
flying-sheep, line 30: Any reason why we shouldn’t just stop doing that?

src/anndata/io/__init__.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import annotations
+
+import warnings
+
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore", message=r"Importing read_.* from `anndata` is deprecated"
+    )
```
flying-sheep, line 8: This seems a bit backwards. I’m OK with this approach, but I’m curious:

Is there anything that prevented doing this?

`io/__init__.py`
```py
from ._io.foo import read_x
from ._io.bar import read_y
...

__all__ = ["read_x", "read_y", ...]
```

`_io.__init__.py`
```py
def __getattr__(name: str) -> Any:
    from . import io
    attr = getattr(io, name)
    warnings.warn(...)
    return attr

__all__: list[str] = []
```

flying-sheep (approved):
Looks great! Two questions

src/anndata/io/__init__.py
ilan-gold: Hmmm, hadn't thought of that.  I think that approach should be a bit cleaner...let me take a whack at it.  I'll re-request review if I land on something...

src/anndata/_io/__init__.py
ilan-gold: Would need to investigate what it's doing there or what that comment even means.  Maybe an issue?

ilan-gold: Well, I deleted it and nothing happened so....

src/anndata/io/__init__.py
ilan-gold: Done, seems to work.

src/anndata/_io/__init__.py
flying-sheep: I put that import and comment there because Isaac made the tests do

```py
import anndata as ad

do_stuff_with(ad._io.specs.foobar)
```

instead of importing that module. It happened to work because some other code somewhere else caused `anndata._io.specs` to be imported, but “happens to work because of side effects” is not what I want to build upon.

Now that the tests no longer access stuff that way, this import is no longer necessary.

flying-sheep (approved):

ilan-gold:
> Why breaking change ‼️ ?

Hmm I guess I was just thinking "begins process for breaking change" but yea, it's not.

ilan-gold:
> I got rid of a few more obsolete import hacks: https://github.com/scverse/anndata/commit/a947d3ec279f78a5e26fd5fc2db94c02c217dc59

I don't think they were so obsolete: https://dev.azure.com/scverse/anndata/_build/results?buildId=8163&view=logs&j=60c6783d-556f-53a9-3d94-3b9ccc51ad8d&t=3522b173-13e1-56b6-93d8-ed20bad6f9bb&l=60!  I cleaned it up a bit to be more uniform.

flying-sheep:
Ah weird, I’m sure I ran the tests without zarr locally, but looking at the code I guess I probably didn’t.

I don’t like this:

image

I think we should instead just do

```py
try:
    from .zarr import read_zarr, write_zarr
except ImportError as e:
    def read_zarr(*_, *__) -> NoReturn:
        raise e from None
    def write_zarr(*_, *__) -> NoReturn:
        raise e from None
```

docs/release-notes/0.10.2.md
```diff
@@ -4,7 +4,7 @@
 #### Bug fixes
 
 * Added compatibility layer for packages relying on `anndata._core.sparse_dataset.SparseDataset`.
-  Note that this API is *deprecated* and new code should use {class}`~anndata.abc.CSRDataset`, {class}`~anndata.abc.CSCDataset`, and {func}`~anndata.sparse_dataset` instead.
+  Note that this API is *deprecated* and new code should use {class}`~anndata.abc.CSRDataset`, {class}`~anndata.abc.CSCDataset`, and {func}`~anndata.io.sparse_dataset` instead.
```
ivirshup, line 7: I don't think we should change the text of old release notes. But I do think it's fine if the links don't work anymore. So instead I'd just do:

```suggestion
  Note that this API is *deprecated* and new code should use `anndata.CSRDataset`, `~anndata.CSCDataset`, and `anndata.sparse_dataset` instead.
```

src/anndata/io.py
```diff
@@ -0,0 +1,53 @@
...
+    read_umi_tools,
+)
+from ._io.specs import read_elem, write_elem
+from ._io.write import write_csvs, write_loom
+
+if "zarr" in sys.modules or TYPE_CHECKING:
+    from ._io.zarr import read_zarr, write_zarr
+else:
+    # Importing zarr is slow, so unless it’s already imported,
+    # we wrap these functions into shims.
```
ivirshup, line 24: Don't we import zarr at import time to define the `ZarrArray`/ `ZarrGroup` classes?

docs/release-notes/0.10.2.md
flying-sheep: For the future, that’s less work. But Ilan already did the work, so I don’t think it needs to be undone.

src/anndata/io.py
flying-sheep: Doesn’t look like it.

`PYTHONPROFILEIMPORTTIME=1 python -c 'import anndata' | complete | get stderr | pipx run tuna /dev/stdin`
grafik

`PYTHONPROFILEIMPORTTIME=1 python -c 'import zarr; import anndata' | complete | get stderr | pipx run tuna /dev/stdin`
grafik

flying-sheep: Maybe that was some weird version resolution, I noticed `uv` acting up in recent times.

Maybe I just got some old zarr version that took one second to import :thinking:

ilan-gold: So just need to update the comment?

flying-sheep:
```suggestion
    # In case zarr is not imported (and maybe not installed),
    # wrap these functions into shims.
```

PR #1688, opened 2024-09-23
document zarr functions properly

flying-sheep:


ilan-gold (approved):

flying-sheep:
I just noticed that the reson we don’t do that is probably zarr’s slow import.

flying-sheep:
OK, I pushed an alternative that works. Another one would be

```py
def __getattr__(attr_name: Literal["read_zarr", "write_zarr"]):
    # This makes the lazy import above into a direct one. Importing zarr is slow.
    if attr_name in {"read_zarr", "write_zarr"}:
        from ._io import zarr

        return getattr(zarr, attr_name)

    raise AttributeError(attr_name)
```

which is less hacky, but makes type checking sadder.

src/anndata/io.py
```diff
@@ -27,6 +29,11 @@ def write_zarr(*args, **kw):
     return write_zarr(*args, **kw)
 
 
+if os.environ.get("READTHEDOCS", "") != "True":
```
ilan-gold, line 32: Is this boolean right?

ilan-gold:
> OK, I pushed an alternative that works. Another one would be

In any case, I think this would probably be better.  In vscode, this would also give the docstring as well, no?

src/anndata/io.py
flying-sheep: This is correct: https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS

ilan-gold: But now the build is failing because this READTHEDOCS not equal to True? https://dev.azure.com/scverse/anndata/_build/results?buildId=8179&view=logs&j=60c6783d-556f-53a9-3d94-3b9ccc51ad8d&t=3522b173-13e1-56b6-93d8-ed20bad6f9bb&l=61

flying-sheep: Ah I misunderstood your question. The `!=` is *not* correct. Spicy autocomplete fails again, this should be a `==`

flying-sheep:
The stuff we have for overriding locations is only for links. In this case we want `autosummary` to see the real function, so we need to replace the shim with it.

An altenative I can think of: I could replace the condition with `if 'zarr' in sys.modules`, then if people import `zarr` first, the shims get removed.
We could then do just that in `conf.py` and add a comment documenting this.

flying-sheep:
OK, there we go. This should work nicely.

ilan-gold (approved):

PR #1692, opened 2024-09-24
Fix typing for optional imports

flying-sheep:
Inspired by #1682

The coverage hasn’t changed from before this PR I think. We could definitely add `# pragma: no cover` to the `else` branches for zarr, as we test only with it, but idk what the best approach for the others is.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,7 +7,6 @@ from importlib import metadata
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
 
-import zarr as _  # noqa: F401  # Makes {read,write}_zarr show up in docs.
 from docutils import nodes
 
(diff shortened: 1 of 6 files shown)


ilan-gold (commented):
The coverage on the `AccessTrackingStore` is weird....do you see it touched locally?

flying-sheep:
Currently, I see codecov.io reporting that all of its lines are covered. The only uncovered code in the diff is

```py
     class DirectoryStore:
         def __init__(self, *_args, **_kwargs) -> None:
             cls_name = type(self).__name__
             msg = f"zarr must be imported to create a {cls_name} instance."
             raise ImportError(msg)
```

ilan-gold (approved):
Must have been erroneous on the rendering then!

PR #1702, opened 2024-10-01
Simplify roundtrip io tests

flying-sheep:


ivirshup:
@ilan-gold @flying-sheep, can we revert and get a review? I personally find this harder to follow

PR #1707, opened 2024-10-09
feat!: Unify X and layers

### LINKED ISSUE #244: Unification of X and layers
@gtca’s proposal in  #706:

> This is an implementation proposal draft for unifying `.X` and layers as mentioned e.g. in #244. Expanding on previous issues, this one tracks concrete implementation details for achieving the end goal.
> 
> At this stage, feedback on the topic is appreciated — either here or in #244.
> 
> ## `.X` attribute as a reference
> 
> The proposal is to use `adata.X` as a reference to one of the layers. In the text of the proposal below, `*.X` or `*adata.X` will denote the data in the layer that `adata.X` points to (akin to pointer dereferencing).
> 
> ### Benefits
> 
> Some benefits include:
> 
> - Reduce the implementation complexity as `.X` and `.layers` items store conceptually identical entities.
> - Reduce memory footprint in common workflows. E.g. when having layers with spliced and unspliced counts, `adata.X` will just point to one of them instead of containing its copy.
> - Make it easier to implement idempotent APIs (to be expanded in other issues).
> 
> ### Options
> 
> #### 1. Explicit layer name
> 
> `.X` links to an explicit layer. This reference can be changed by a function that is run on an AnnData object. E.g. a normalisation function can create a new layer and update the reference instead of modifying the values in-place. In-place APIs will continue to modify the values in the `*.X`.
> 
> When `*.X` is deleted, `.X` can be either `None` or, less favourably, point to the layer that was defined previously in case there's a memory of which layers were linked to in `.X` before.
> 
> #### 2. Last layer
> 
> `.X` can link to `adata.layers[list(adata.layers.keys())[-1]]`. This is easier to reason about but might be an issue when there's no explicit order in layers (e.g. raw counts -> normalised counts -> scaled counts) and new layers are added (e.g. unspliced counts).
> 
> ### Additional methods
> 
> There should be a way to explicitly change which layer `.X` points to.
> 
> ### Alternatives
> 
> An alternative, as mentioned in #244, would be to have `.X` as a specific layer. While this is simpler in the short-term, that might not provide some of the benefits listed above.
> 
> ## Effect for the end user
> 
> Taking the current scanpy API as an example where the matrix is modified in-place, nothing changes for the end user:
> 
> ```py
> sc.pp.scale(adata)
> # modifies values stored in the layer
> # that .X links to
> ```
> 
> ## On disk
> 
> For HDF5, a non-breaking change should be achievable with the HDF5 references functionality. It is to be checked that R (`rhdf5`, `hdf5r`) and Julia HDF5 libraries are able to work with references.
> 
> For Zarr, feedback is needed if this is possible. It is probably not an issue if it's not as Zarr storage for AnnData is less mature/common.
> 
> ## Potential
> 
> This is to be expanded in other issues to come.
> 
> ### Idempotent analysis API
> 
> In practice, the scaling example above typically looks like this together with its context:
> 
> ```py
> adata.layers["lognorm"] = adata.X.copy()
> sc.pp.scale(adata)
> ```
> 
> This feature would allow to implement idempotent APIs, i.e.
> 
> ```py
> sc.pp.scale(adata, in_layer="lognorm", out_layer="scaled")
> # .X now links to .layers["scaled"]
> ```
> 
> that would not modify counts, and, importantly, will not affect the end user API provided that the appropriate default arguments for `in_layer` and `out_layer` are chosen.

prior considerations by me:

> 
> X should become a layer in `AnnData.layers`. `AnnData.X` should just be a shortcut for that. Answering the questions in https://github.com/theislab/anndata/pull/242#issuecomment-550174042:
> 
> - The key needs to be determined. I think `None` fits the idea “default” best, and I’ll use it as a strawman, but `"X"` and other options can be discussed.
> 
>     Scanpy functions allowing to specify the layers they operate on should probably have the signature `layers: Union[Literal['all'], Sequence[str]] = (None,)` or `... = 'all'`, indicating operation on `X` or all layers.
> 
> - Iteration over `layers.items()` should yield `X` as the first key-value pair (Accordingly for `keys`/`values` of course)
> - Disk representation: X should probably be stored in the layers group, if we can stomach the backwards compat code
> - Raw should get layers, see #96. I think sharing a base class between Raw and AnnData makes a lot of sense.
> - @ivirshup said “obs is to obsm as X is to layers”, but I don’t think so: `obsm` will e.g. also contain dimension reductions like `X_pca`, so I don’t think `obsm[None]` should be `obs`.

### LINKED ISSUE #678: concat when some values of `X` are `None`
Subpart of #467

What should we do when we `concat` multiple anndata objects where some of them have `None` for their `X` value?

While I think we should broadly do the same thing that we do for values in `layers`, I think this could have some weird behaviour.

Here is my current proposal for the behavior:

* All `AnnData`'s have `X` – current behavior
* No `AnnData`'s have `X` – same as if no `AnnData` had this key. It remains `None`.
* Some `AnnData`'s have `X` and it's an outer join. We use the `fill_value` and create `.X` arrays for the objects missing them
* Some `AnnData`'s have `X` and it's an inner join. Result has `X = None`.

The final case feels off to me, but it's the least weird behavior I could think of. I personally would expect this to result in an inner join on the variables with an outer join on the values for `X`. That is, we still take the intersection of the alternative axis, but we fill the `X`s which were missing.

But this would inconsistent with the behavior for `layers` (#244), which I think is worse.

I think this leads to the broader issue of the `join` argument controlling both combining the dimensions and which elements are kept (#591).

flying-sheep:
- [x] Closes #244, closes #678
- [x] Tests added

TODO:

- [x] decide on `dtype` arg: just remove? or ignore+warning? (unless we already do that, then remove)
- [x] backed mode: replace with Dask? *not yet*
- [x] #678


ilan-gold:
So re: the backed question, I think the answer is not to get rid of it, but to clarify its behavior.  To do that, I think we should merge #1927 first, which may include a breaking change.  If we stop allowing overriding of on-disk stores (which currently relies on inheritance form scipy.sparse private methods), then this question becomes much more simplified.

The next step after #1927 would probably be to simplify it.  From what I can tell, the reaosn we need to track if the file is open or not is because if writing.  If everything were purely read-only, I think it would be less of an issue.

My vision there would be in "backed" mode (after the two above points are implemented), we would simply take the `X` attribute on-disk, and read it in using `ad.io.sparse_dataset` with `mode="r"` and call it a day.  That's how `layers` and other elements handle `sparse_dataset` so `X` shouldn't be different

src/anndata/_core/anndata.py
```diff
@@ -561,125 +537,18 @@ def shape(self) -> tuple[int, int]:
...
-        # elif self.n_obs == 1 or self.n_vars == 1:
-        #     if issparse(X): X = X.toarray()
-        #     return X.flatten()
-        # else:
-        #     return X
+        return self.layers.get(None)
 
     @X.setter
-    def X(self, value: XDataType | None):  # noqa: PLR0912
-        if value is None:
```
ilan-gold: I think a lot of this deleted logic contains things that `layers` can't currently do.  For example, overriding the backing `X`; currently, if you try to override a part of a view of a layers, it will just actualize and not override the old data.  But if you do `adata_view.X = 1` it will fill ones in to the parent `AnnData` object.  I'm not sure what the history is there, but it accounts for a pretty big discrepancy in functionality

ilan-gold:
My rough plan here:

1. https://github.com/scverse/anndata/pull/2327
2. Turn off the above behavior by simply removing the setting on `main` (i.e., `0.13`) - hopefully this doesn't make merges into `0.12.x` impossible
3. https://github.com/scverse/anndata/pull/2071 is in-flight and relevant to setting `X` so likely would want this to go in first
4. https://github.com/scverse/anndata/issues/2147 in parallel to 3. will force us to start thinking about managing open-or-closed hdf5 files outside of `X` which is what will be needed here for `backed` mode (and `backed` mode would also probably be superseded by #2147 but not until `0.14` at least)
5. Merge this PR, which is in a mostly complete state except for the above

ilan-gold:
Now it seems like we've hit a point where the blocking factor is the interaction between `layers` and `backed` mode.  This PR at this point got rid of `_mutated_copy` which previously handled detaching a CoW-ed view of `AnnData` from its file handle so `my_backed[subset].X = None` would not error out but it does as of the current state of things because it passes through to `layers` which then calls `copy`.  So we likely need to bring this function back in some capacity but this gets at the intricacies of backed mode + layers.

ilan-gold:
The main open question here (raised in #2357 and asked [here on zulip](https://scverse.zulipchat.com/#narrow/channel/315789-data-structures/topic/Why.20Does.20.60backed.60.20mode.20not.20open.20the.20file.20and.20leave.20it.20open.3F/with/576321492)) is what even backed mode is.  Already was the case that backed mode was this weird "if it's set, open the file on access to `.X` thing" but now [that is turbo apparent](https://github.com/scverse/anndata/pull/1707/changes#diff-93807bfddb872b9cf4ea7586b7ee8feb73bb9cab28969a82b4b6974001ff3462R569-R581) since nothing really happens in `X` anymore because there is no `_X` anymore.

In general, I think figuring out what backed mode should be now that X is copy-on-write (i.e., `adata[subset].X` doesn't write through) is the open question (asked in the description of #2357).

src/anndata/_core/aligned_mapping.py
```diff
@@ -314,20 +337,57 @@ class LayersBase(AlignedMappingBase):
...
-class Layers(AlignedActual, LayersBase):
-    pass
+
+class Layers[K: str | None](AlignedActual[K], LayersBase):
+    def __init__(self, parent: AnnData | Raw, *, store: MutableMapping[str, Value]):
+        super().__init__(parent, store=store)
+        if None not in self._data and self.parent.filename is not None:
+            self.is_none_backed = True
+        else:
+            self.is_none_backed = False
```
flying-sheep:
```suggestion
        self.is_none_backed = None not in self._data and self.parent.filename is not None
```

src/anndata/_core/anndata.py
```diff
@@ -633,13 +618,16 @@ def X(self, value: _XDataType | None):
...
-            new = self._mutated_copy(X=value)
-            self._init_as_actual(new)
-            return
-        self._X = value
+            self._init_as_actual(self._copy(X=value))
+            return None
+        if value is not None:
+            self.layers[None] = value
+        else:
+            self.layers.pop(None)
```
flying-sheep: I think this is clearer, `pop` is usually used when its return value is used or to ignore when the key is missing using `.pop(key, None)`

```suggestion
            del self.layers[None]
```

or we do

```suggestion
            self.layers.pop(None, None)
```

src/anndata/_core/anndata.py
```diff
@@ -974,8 +962,8 @@ def isbacked(self) -> bool:
         """`True` if object is backed on disk, `False` otherwise."""
         is_filename_none = self.filename is not None
         is_x_none = (
-            getattr(self._adata_ref if self._is_view else self, "_X", None) is None
-        )
+            self._adata_ref.layers if self._is_view else self.layers
+        ).is_none_backed
         return is_filename_none and is_x_none
```
flying-sheep: Hm, `is_filename_none`’s name is inverted. Anyway:

- is it possible that `self.filename is not None` but `self.layers.is_none_backed`, or is this check redundant?
- shouldn’t this work? I feel like `is_none_backed` should also be true, if we’re a view of a backed AnnData, right?

```suggestion
        return self.filename is not None and self.layers.is_none_backed
```

src/anndata/_core/anndata.py
```diff
@@ -1024,8 +1012,10 @@ def filename(self, filename: PathLike[str] | str | None):
                 self.write(filename, as_dense=as_dense)
             # open new file for accessing
             self.file.open(filename, "r+")
-            # as the data is stored on disk, we can safely set self._X to None
-            self._X = None
+            # As the data is stored on disk, we can safely set remove if it previously was
+            # in layers. Setting `X` to `None` now would raise an error because `self.isbacked`.
+            if None in self.layers:
+                self.layers.pop(None)
```
flying-sheep:
```suggestion
            self.layers.pop(None, None)
```

or if you think the above is too confusing:

```suggestion
            if None in self.layers:
                del self.layers[None]
```

src/anndata/_io/h5ad.py
```diff
@@ -86,19 +86,22 @@ def write_h5ad(
...
-                    is_backed=adata.isbacked and adata.filename == filepath,
-                    as_dense=as_dense,
-                    dataset_kwargs=dataset_kwargs,
-                )
-            elif k == "raw":
+            if k == "raw":
                 _write_raw(
                     f, adata.raw, as_dense=as_dense, dataset_kwargs=dataset_kwargs
                 )
             else:
```
flying-sheep: could just do `continue` here instead of an enormous `else` branch

src/anndata/experimental/multi_files/_anncollection.py
```diff
@@ -56,7 +56,7 @@ def check_type(attr, key=None):
         arrs = []
         for a in adatas:
             attr_arr = getattr(a, attr)
-            if key is not None:
+            if (key is None and attr == "layers") or key is not None:
```
flying-sheep:
```suggestion
            if key is not None or attr == "layers":
```

src/anndata/tests/helpers.py
```diff
@@ -767,7 +767,7 @@ def assert_equal_mapping(
     a: Mapping, b: object, *, exact: bool = False, elem_name: str | None = None
 ):
     assert isinstance(b, Mapping)
-    assert set(a) == set(b), format_msg(elem_name)
+    assert set(a) == set(b), format_msg(elem_name) + f" {a.keys()} != {b.keys()}"
```
flying-sheep:
```suggestion
    assert set(a) == set(b), f"{format_msg(elem_name)} {a.keys()} != {b.keys()}"
```

tests/test_layers.py
```diff
@@ -21,7 +21,8 @@ def X(request):
 def test_creation(X: np.ndarray | None):
     adata = AnnData(X=X, layers=dict(L=L.copy()))
 
-    assert list(adata.layers.keys()) == ["L"]
+    # TODO: phil, when reviewing this, you had { "L", None } in both cases before - but if `X` is `None` should the key be there?
```
flying-sheep: it shouldn’t, you’re right! maybe if anndata is backed, because then X exists, it’s just backed.

```suggestion
```

tests/test_repr.py
```diff
@@ -59,4 +59,7 @@ def test_removal(adata, adata_attr):
     attr = adata_attr
     assert re.search(rf"^\s+{attr}:.*$", repr(adata), flags=re.MULTILINE)
     delattr(adata, attr)
-    assert re.search(rf"^\s+{attr}:.*$", repr(adata), flags=re.MULTILINE) is None
+    if attr != "layers":
+        assert re.search(rf"^\s+{attr}:.*$", repr(adata), flags=re.MULTILINE) is None
+    else:
```
flying-sheep:
```suggestion
    else:  # `del adata.layers` doesn’t delete `X` for backwards compat reasons
```

tests/test_x.py
```diff
@@ -88,8 +88,9 @@ def test_del_set_equiv_X():
     assert_equal(orig, copy)
 
     # Check that deleting again is still fine
-    del orig.X
-    assert orig.X is None
+    # TODO: Do we even want to keep supporting this operation i.e., del adata.X if X is None?
+    # del orig.X
+    # assert orig.X is None
```
flying-sheep: normally `del`ing a thing means it’s gone, but we want to keep the attributes on anndata and just delete the contents, but I think it’s OK that repeated del stops working.

ilan-gold:
For `concat` (i.e., #678), the default `inner` will now have a warning when some X's are `None` while `outer` will start to properly do the outer join i.e., add a fill value.

Put that in the release note.

ilan-gold (approved):

PR #1714, opened 2024-10-14
(fix): gpu ci uses max version from `pyproject.toml`

ilan-gold:
- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -51,13 +51,10 @@ jobs:
       - name: Nvidia SMI sanity check
         run: nvidia-smi
 
-      - name: Install toml-cli from crates.io
-        uses: baptiste0928/cargo-install@v3
-        with:
-          crate: toml-cli
-
       - name: Extract max Python version from classifiers
         run: |
+          sudo apt-get update && sudo apt-get install -y cargo && cargo install toml-cli
+          echo "export PATH=$HOME/.cargo/bin:\$PATH" >> $GITHUB_ENV
           classifiers=$(toml get pyproject.toml project.classifiers | grep -oP 'Python :: \K[0-9.]+')
           max_version=$(echo "$classifiers" | sort -V | tail -1)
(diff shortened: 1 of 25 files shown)


.github/workflows/test-gpu.yml
```diff
@@ -51,10 +51,17 @@ jobs:
       - name: Nvidia SMI sanity check
         run: nvidia-smi
 
+      - name: Extract max Python version from classifiers
+        run: |
+          sudo apt-get update && sudo apt-get install -y cargo && cargo install toml-cli
```
flying-sheep, line 56: You could also use this: https://github.com/marketplace/actions/cargo-install

ilan-gold: ah was looking for something like this just now.  I also thought cargo came on the ubuntu images by default but evidently not `install`? 

Also, do you think we should be doing this for all of our CI i.e., extracting from `pyproject.toml`?

ilan-gold: Ok I think that is only for rust projects: https://github.com/scverse/anndata/actions/runs/11330444061/job/31508293550?pr=1714

flying-sheep: I still think that using that action is better, because it sets up PATH and uses caching.

ilan-gold: https://github.com/scverse/anndata/actions/runs/11330444061/job/31508293550 Not sure

ilan-gold: I read this as `cargo` not being on the path

flying-sheep (approved):
Awesome! I’m not a huge fan of almost all[^1] short-form CLI parameters, so if you could switch to long-form I’d even be happier.

Also `\K` means something regarding capture right? If I’m right, maybe use a capture group instead, I think people are more likely to know that one. Otherwise please add a comment explaining what it is.

[^1]: except super-well known ones like `something -vv` or `grep -eP`

flying-sheep (approved):
Ah great, now I understand what’s going on!

PR #1719, opened 2024-10-17
(fix): correct default fill values for dask-sparse

ilan-gold:
- [x] Noticed in #1247
- [x] Tests added


flying-sheep (approved):
Doesn’t look like we use the test “is a sparse matrix/array or a Dask array containing a sparse matrix/array” anywhere else so I’m not suggesting to make this a helper in this PR yet

flying-sheep:
Dask sparse is only in 0.11 right? I modified the milestones under that assumption. If I’m wrong and you want to backport this, please change it back

ilan-gold:
I don't think so, just the API to do i/o is

flying-sheep:
OK, needs manual backport then

PR #1722, opened 2024-10-21
(chore): add docstring example for `read_elem_as_dask`

ilan-gold:
Noticed during my lab presentation, would be super helpful
- [ ] Closes #
- [x] Tests added
- [x] Release note added (or unnecessary): making a docstring

### DIFF
--- a/src/anndata/_io/specs/registry.py
+++ b/src/anndata/_io/specs/registry.py
@@ -395,10 +395,60 @@ def read_elem_as_dask(
     chunks, optional
        length `n`, the same `n` as the size of the underlying array.
        Note that the minor axis dimension must match the shape for sparse.
+       Defaults to `(1000, adata.shape[1])`.
 
     Returns
     -------
         DaskArray
+
+    Examples
+    --------
+
+    Setting up our example:
+
+    >>> from scanpy.datasets import pbmc3k
+    >>> import tempfile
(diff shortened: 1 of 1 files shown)


src/anndata/_io/specs/registry.py
```diff
@@ -395,10 +395,60 @@ def read_elem_as_dask(
...
+    ...     g["X"], chunks=(500, adata.shape[1])
+    ... )
+    >>> adata.X
+    dask.array<make_dask_chunk, shape=(2700, 32738), dtype=float32, chunksize=(500, 32738), chunktype=scipy.csr_matrix>
+
+    Reading a dense matrix from a zarr store lazily:
+
+    >>> adata.layers["dense"] = ad.experimental.read_elem_as_dask(g["layers/dense"])
+    >>> adata.layers["dense"]
+    dask.array<from-zarr, shape=(2700, 32738), dtype=float32, chunksize=(169, 2047), chunktype=numpy.ndarray>
```
flying-sheep, line 437: this doesn‘t seem to match the docs above:

> Defaults to `(1000, adata.shape[1])`.

ilan-gold: Ah, good point.  That is only for CSR sparse!

flying-sheep (approved):

PR #1724, opened 2024-10-21
Keep .X when setting filename to None

### LINKED ISSUE #1723: .X is lost when setting anndata.filename to None

### Report

When setting the `filename` of a backed anndata object to `None`, it's `.X` gets lost.

Reproducible example of how `to_memory()` works as expected:

```python
import anndata as ad
adata = ad.AnnData(
    X = np.array([[1, 2], [3, 4]]),
    obs=pd.DataFrame({"cell_type": ["A", "B"]}),
)
adata.write_h5ad("sample_data.h5ad")

adata = ad.read_h5ad("sample_data.h5ad", backed="r")

adata_mem = adata.to_memory()

adata_mem.X
```
```
array([[1, 2],
       [3, 4]])
```

Example of how using the `.filename` setter looses X: 
```python
import anndata as ad
adata = ad.AnnData(
    X = np.array([[1, 2], [3, 4]]),
    obs=pd.DataFrame({"cell_type": ["A", "B"]}),
)
adata.write_h5ad("sample_data.h5ad")

adata = ad.read_h5ad("sample_data.h5ad", backed="r")

# This should just put the anndata object to memory mode I suppose
adata.filename = None

adata.X is None
```
```
True
```

Running this on the main branch.

might become irrelevant in the future, but the fix seems to be a simple bug in `_to_memory_mode`.

### Versions

anndata             0.11.0rc4.dev4+g3260222
awkward             2.6.8
numpy               2.0.0
pandas              2.2.2
scanpy              1.10.0rc2.dev126+g48706caa
session_info        1.0.0

eroell:
- [X] Closes #1723 
- [ ] Tests added

Does [_to_memory_mode](https://github.com/scverse/anndata/blob/664b629d8617e2f6cafeb8013b8edfc80742b804/src/anndata/_core/file_backing.py#L104) only get called by the filename setter anyways? And are there reasons to not just call to_memory from this setter?

If not worth to go in depths here anymore, the removed `_` would fix a bug for now I guess


ilan-gold:
@eroell Please add a test to check the behavior is correct of this so we can verify the fix actually works

eroell:
added, @ilan-gold

ilan-gold (approved):

PR #1725, opened 2024-10-21
remove label in ci

Intron7:
This will remove the label to run the CI after one run. This will safe cost in the long run.


flying-sheep:
ah, I somehow missed that you already tested it, that’s why I added and removed the label.

PR #1726, opened 2024-10-21
(feat): support for `zarr-python>=3` 

### LINKED ISSUE #1817: Support zarr v3

### Report

See
Image

### Versions

```python
latest
```

ilan-gold:
- [x] Closes #1817
- [x] Tests added


ilan-gold:
blockers:

- [ ] https://github.com/zarr-developers/zarr-python/issues/2500

flying-sheep:
This should fix the docs build: https://github.com/theislab/scanpydoc/pull/191

flying-sheep:
why 3ac48cb681b51d067610c9ad394b177c954b4223? doesn’t the “minimum versions” job already check zarr v2?

ilan-gold:
Ah yea, you're totally right.

pyproject.toml
```diff
@@ -148,6 +147,7 @@ filterwarnings_when_strict = [
     "default:(Observation|Variable) names are not unique. To make them unique:UserWarning",
     "default::scipy.sparse.SparseEfficiencyWarning",
     "default::dask.array.core.PerformanceWarning",
+    "default:anndata will no longer support zarr v2:FutureWarning"
```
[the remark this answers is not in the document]
ilan-gold: Wait, no it should be the reverse? Or should we not do anything? Sorry, I forgot that FutureWarning is the one that is louder than DeprecationWarning.  I meant to use the quieter of the two.  But maybe also not so bad to add this?  I think zarr-python v2 is getting patches only for six months or something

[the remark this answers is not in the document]
flying-sheep: there are three:
- `PendingDeprecationWarning`: super silent, not even active during testing by default, no clue why it exists
- `DeprecationWarning`: quiet, enabled in test frameworks and `__main__` (`python -m module` and `python script.py`)
- `FutureWarning`: visible to users

src/anndata/_core/sparse_dataset.py
```diff
@@ -588,10 +588,11 @@ def append(self, sparse_matrix: ss.csr_matrix | ss.csc_matrix | CSArray) -> None
         append_data = sparse_matrix.data
         append_indices = sparse_matrix.indices
         if isinstance(sparse_matrix.data, ZarrArray) and not is_zarr_v2():
-            append_data = append_data[...]
-            append_indices = append_indices[...]
-        data[orig_data_size:] = append_data
-
+            append_data_in_memory = append_data[...]
+            data[orig_data_size:] = append_data_in_memory
+            del append_data_in_memory
```
flying-sheep, line 593: why not like this? if the rhs loads into memory, that happens before the `__setitem__`.

```suggestion
            data[orig_data_size:] = append_data[...]
```

src/anndata/_core/sparse_dataset.py
```diff
@@ -534,9 +551,9 @@ def append(self, sparse_matrix: ss.csr_matrix | ss.csc_matrix | CSArray) -> None
                 f"{self.format!r} and {sparse_matrix.format!r}"
             )
             raise ValueError(msg)
-        indptr_offset = len(self.group["indices"])
+        [indptr_offset] = self.group["indices"].shape
```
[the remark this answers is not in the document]
flying-sheep: that’s even cleaner, as it encodes our expectation that this is 1D

src/anndata/_io/specs/methods.py
```diff
@@ -532,23 +547,38 @@ def write_vlen_string_array_zarr(
...
+        from numcodecs import VLenUTF8
+
+        dataset_kwargs = dataset_kwargs.copy()
+        if "compressor" in dataset_kwargs:
+            dataset_kwargs.pop("compressor")
+        f.create_array(
+            k,
+            shape=elem.shape,
+            dtype=str if ad.settings.zarr_write_format == 3 else object,
+            filters=[VLenUTF8()] if ad.settings.zarr_write_format == 2 else None,
```
flying-sheep, line 578: This is a bit confusing. why `== 3` and `== 2`? Are there more valid values?

if not, maybe

```py
dtype, filters = ... if ad.settings.zarr_write_format == 3 else ...
f.create_array(k, shape=elem.shape, dtype=dtype, filters=filters, **dataset_kwargs)
```

src/anndata/_io/specs/methods.py
```diff
@@ -603,7 +633,13 @@ def write_recarray_zarr(
...
     from anndata.compat import _to_fixed_length_strings
 
-    f.create_dataset(k, data=_to_fixed_length_strings(elem), **dataset_kwargs)
+    elem = _to_fixed_length_strings(elem)
+    if isinstance(f, H5Group) or is_zarr_v2():
+        f.create_dataset(k, data=elem, shape=elem.shape, **dataset_kwargs)
+    else:
+        # TODO: zarr on-disk v3 doesn’t support this dtype
+        f.create_array(k, shape=elem.shape, dtype=elem.dtype, **dataset_kwargs)
+        f[k][...] = elem
```
flying-sheep, line 642: this is a repeated pattern, right? maybe make a helper?

src/anndata/_io/specs/methods.py
```diff
@@ -1134,18 +1180,56 @@ def _string_array(
...
+            shape=(),
+            dtype=str
+            if ad.settings.zarr_write_format == 3
+            else (object if isinstance(value, str) else dtype),
+            filters=[VLenUTF8()]
+            if ad.settings.zarr_write_format == 2 and isinstance(value, str)
+            else None,
+            **dataset_kwargs,
+        )
+        a[...] = np.array(value)
```
flying-sheep, line 1232: this looks very similar to the above code using `VLenUTF8`, maybe create a helper?

src/anndata/_io/specs/methods.py
```diff
@@ -1134,18 +1180,56 @@ def _string_array(
...
+        "scaleoffset",
+        "compressor",
+    ):
+        dataset_kwargs.pop(arg, None)
+    return dataset_kwargs
 
 
-def write_scalar(
+def write_scalar_zarr(
     f: GroupStorageType,
```
flying-sheep, line 1205: should be of type `zarr.Group` then

src/anndata/_io/specs/methods.py
```diff
@@ -603,7 +633,13 @@ def write_recarray_zarr(
 ):
     from anndata.compat import _to_fixed_length_strings
 
-    f.create_dataset(k, data=_to_fixed_length_strings(elem), **dataset_kwargs)
+    elem = _to_fixed_length_strings(elem)
+    if isinstance(f, H5Group) or is_zarr_v2():
+        f.create_dataset(k, data=elem, shape=elem.shape, **dataset_kwargs)
+    else:
+        # TODO: zarr on-disk v3 doesn’t support this dtype
```
flying-sheep, line 640: this seems like a rather large TODO!

src/anndata/experimental/merge.py
```diff
@@ -100,27 +100,34 @@ def _gen_slice_to_append(
...
+
+    from anndata._io.zarr import open_write_group
+
+    return open_write_group(store, mode=mode)
 
 
 @as_group.register(ZarrGroup)
 @as_group.register(H5Group)
-def _(store, *args, **kwargs):
+def _(store, *, mode: str) -> ZarrGroup | H5Group:
```
flying-sheep, line 129: codecov says this is never hit. seems weird!

flying-sheep (commented):
Cool, this seems to be much less hacky than I feared!

Anything missing in this check-list? https://github.com/scverse/anndata/pull/1726#issuecomment-2483443061

if so: did the blockers become non-blockers?

src/anndata/_core/sparse_dataset.py
ilan-gold: Just a brain fart!

src/anndata/_io/specs/methods.py
flying-sheep: Ah, so the issue is that I don’t know what that is.

I assumed this means “on-disk as opposed to in-memory”, i.e. all file-based stores and not some specific feature.

src/anndata/_io/specs/methods.py
ilan-gold: The differences between all of these blocks was small but signficant enough that an abstraction didn't seem worth it - some have the condition as including a check for hdf5, others dont etc. The way I see it it would have to be something like

```python
def create_array(key: str, dtype: np.dtype, shape: tuple[int], filters: list | None, value: Any)
```
with all these conditions done before calling `polymorphic_create_array` i.e., this one looks similar but is actaully a bit different.  So the abstraction would basically just be around the condition, `create_array`/`create_dataset`, and then write, but I think it's fine to leave these in since `is_zarr_v2` will disappear at some point at which point, some but not all of these look-alike blocks will disappear.

If you have suggestions, I'm all ears

src/anndata/_io/specs/methods.py
ilan-gold: Right, so there is zarr-python the package, which had a major version upgrade to 3.  There is also the on-disk v3 format, which is different than v2.  v3 does not support recarray.

ilan-gold:
> if so: did the blockers become non-blockers?

The actual blockers have been resolved.  Everything else is handled in this PR and should be marked with an issue if it appears in the codebase

src/anndata/_io/specs/methods.py
flying-sheep: I see! Since this is only 2 spots which are sufficiently different, I agree.

However, for this block, I’d not nest `if else`-expressions. How about something like this? I think this is a correct translation. I’m also not quite sure why `dtype is str` when the write format is `3` independently of `value`’s type, can you explain that?

```py
match ad.settings.zarr_write_format, value:
    case 2, str():
        filters, dtype = [VLenUTF8()], object
    case 2, _:
        filters, dtype = None, np.array(value).dtype
    case 3, _:
        filters, dtype = None, str
```

src/anndata/_io/specs/methods.py
flying-sheep: got it! could you rephrase a bit then, please?

```suggestion
        # TODO: zarr’s on-disk format v3 doesn’t support this dtype
```

src/anndata/_io/specs/methods.py
ilan-gold: This was a parenthesis issue - I certainly meant not to map everything to `str`

ilan-gold:
One thing I'm thinking about now -  a lot of code is not run because it is behind the flag of the format being 3, which is untested.  I'm not sure how to proceed here.  I had added the setting when I thought it was important/possible to support v3 on disk, but the recarray stuff prevents that.

So the way I see it, there are two options:

1. Continue as-is, potentially with some added tests, allowing people to write as much of the anndata store as possible, modulo the fact that they cannot write recarrays
2. Disallow the `3` option in the setting, with the option to remove or keep the conditional-on-v3-disk code
3. Remove the setting altogether including the v3 code

I am tempted to go for option 2, and leave the code so that we can easily test things out and potentially iterate on this.  WDYT @flying-sheep ?

src/anndata/experimental/merge.py
ilan-gold: Was that way before then, it would seem.  It's possible no one was testing `concat_on_disk` with an actual store

src/anndata/_io/specs/methods.py
```diff
@@ -1225,10 +1225,10 @@ def write_scalar_zarr(
         match ad.settings.zarr_write_format, value:
             case 2, str():
                 filters, dtype = [VLenUTF8()], object
+            case 3, str():
+                filters, dtype = None, str
             case 2, _:
                 filters, dtype = None, np.array(value).dtype
-            case 3, _:
-                filters, dtype = None, str
```
flying-sheep: And what about the `case 3, _:` you just created?

tests/test_readwrite.py
```diff
@@ -344,13 +345,14 @@ def test_zarr_compression(tmp_path):
...
+    def check_compressed(value, key):
+        if value.shape != ():
             if value.compressor != compressor:
                 not_compressed.append(key)
 
-    with zarr.open(str(pth), "r") as f:
-        f.visititems(check_compressed)
+    f = zarr.open(str(pth), mode="r")
+    for key in f.array_keys():
+        check_compressed(f[key], key)
```
flying-sheep, line 355: doesn’t visititems go all the way down? Is that not necessary for this test?

flying-sheep:
Looking good apart from the `case 3, _`

tests/test_readwrite.py
ilan-gold: `visititems` doesn't exist on zarr v3 :/

src/anndata/_io/specs/methods.py
ilan-gold: just ran into this! fixing now

tests/test_readwrite.py
flying-sheep: My question was if the structure in the test can be deeper than one level. If so, the code you replaced `visititems` with doesn’t do the same thing.

ilan-gold: apologies, shouldn't have assumed that it would go deeper!

flying-sheep (approved):

jan-engelmann:
🙌

PR #1729, opened 2024-10-22
(feat): support ellipsis indexing

### LINKED ISSUE #1696: Support ellipsis indexing

### Report

In 1.5.1 sklearn changed their _array_indexing method to use ellipsis indexing/slicing: 
https://github.com/scikit-learn/scikit-learn/blob/1.5.1/sklearn/utils/_indexing.py#L33

Running train_test_split on AnnData now results in an IndexError.

Code:

```python
import anndata
import session_info
from sklearn.model_selection import train_test_split
import scanpy as sc

session_info.show(html=False, dependencies=True)

adata = sc.datasets.ebi_expression_atlas("E-MTAB-4888")
adata_val, adata_test = train_test_split(adata, test_size=0.5, stratify=adata.obs['Sample Characteristic[cell type]'])
```

Traceback:

```pytb
  File "/home/user/project/src/data/test.py", line 6, in <module>
    adata_val, adata_test = train_test_split(adata, test_size=0.5, stratify=adata.obs['Sample Characteristic[cell type]'])
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/sklearn/utils/_param_validation.py", line 213, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/sklearn/model_selection/_split.py", line 2805, in train_test_split
    return list(
           ^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/sklearn/model_selection/_split.py", line 2807, in <genexpr>
    (_safe_indexing(a, train), _safe_indexing(a, test)) for a in arrays
     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/sklearn/utils/_indexing.py", line 267, in _safe_indexing
    return _array_indexing(X, indices, indices_dtype, axis=axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/sklearn/utils/_indexing.py", line 33, in _array_indexing
    return array[key, ...] if axis == 0 else array[:, key]
           ~~~~~^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/anndata/_core/anndata.py", line 1021, in __getitem__
    oidx, vidx = self._normalize_indices(index)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/anndata/_core/anndata.py", line 1002, in _normalize_indices
    return _normalize_indices(index, self.obs_names, self.var_names)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/anndata/_core/index.py", line 39, in _normalize_indices
    ax1 = _normalize_index(ax1, names1)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/env/lib/python3.12/site-packages/anndata/_core/index.py", line 109, in _normalize_index
    raise IndexError(f"Unknown indexer {indexer!r} of type {type(indexer)}")
IndexError: Unknown indexer Ellipsis of type <class 'ellipsis'>
```

I'm not sure if this behavior is intended.

### Versions

```python
-----
anndata             0.10.9
scanpy              1.10.2
session_info        1.0.0
sklearn             1.5.1
-----
PIL                 10.3.0
cffi                1.17.1
colorama            0.4.6
cycler              0.12.1
cython_runtime      NA
dateutil            2.9.0
h5py                3.11.0
igraph              0.11.6
joblib              1.4.2
kiwisolver          1.4.7
legacy_api_wrap     NA
leidenalg           0.10.2
llvmlite            0.43.0
matplotlib          3.9.2
mpl_toolkits        NA
natsort             8.4.0
numba               0.60.0
numpy               1.26.4
packaging           24.1
pandas              2.2.2
pyparsing           3.1.4
pytz                2024.1
scipy               1.14.1
six                 1.16.0
texttable           1.7.0
threadpoolctl       3.5.0
torch               2.4.1
torchgen            NA
tqdm                4.66.5
typing_extensions   NA
yaml                6.0.2
-----
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0]
Linux-6.1.0-25-amd64-x86_64-with-glibc2.36
```

ilan-gold:
Not sure if this is a feature or a fix TBH but going with feature and putting it in 0.10 because it addresses a pretty big use-case that broke without it (so in a not-so-strict sense, is a bug)
- [x] Closes #1696
- [x] Tests added

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,6 +112,7 @@ nitpick_ignore = [
     ("py:obj", "numpy._typing._array_like._ScalarType_co"),
     # https://github.com/sphinx-doc/sphinx/issues/10974
     ("py:class", "numpy.int64"),
+    # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/498
     ("py:class", "types.EllipsisType"),
 ]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,17 +121,17 @@ def setup(app: Sphinx):
 
 
 intersphinx_mapping = dict(
-    h5py=("https://docs.h5py.org/en/latest/", None),
-    hdf5plugin=("https://hdf5plugin.readthedocs.io/en/latest/", None),
-    loompy=("https://linnarssonlab.org/loompy/", None),
-    numpy=("https://numpy.org/doc/stable/", None),
-    pandas=("https://pandas.pydata.org/pandas-docs/stable/", None),
+    h5py=("https://docs.h5py.org/en/latest", None),
+    hdf5plugin=("https://hdf5plugin.readthedocs.io/en/latest", None),
+    loompy=("https://linnarssonlab.org/loompy", None),
+    numpy=("https://numpy.org/doc/stable", None),
+    pandas=("https://pandas.pydata.org/pandas-docs/stable", None),
     python=("https://docs.python.org/3", None),
-    scipy=("https://docs.scipy.org/doc/scipy/", None),
-    sklearn=("https://scikit-learn.org/stable/", None),
-    zarr=("https://zarr.readthedocs.io/en/stable/", None),
-    xarray=("https://docs.xarray.dev/en/stable/", None),
-    dask=("https://docs.dask.org/en/stable/", None),
+    scipy=("https://docs.scipy.org/doc/scipy", None),
+    sklearn=("https://scikit-learn.org/stable", None),
+    zarr=("https://zarr.readthedocs.io/en/stable", None),
+    xarray=("https://docs.xarray.dev/en/stable", None),
+    dask=("https://docs.dask.org/en/stable", None),
 )
 qualname_overrides = {
     "h5py._hl.group.Group": "h5py.Group",
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,6 +112,7 @@ nitpick_ignore = [
     ("py:obj", "numpy._typing._array_like._ScalarType_co"),
     # https://github.com/sphinx-doc/sphinx/issues/10974
     ("py:class", "numpy.int64"),
+    ("py:class", "types.EllipsisType"),
 ]
 
 
--- a/src/anndata/_core/index.py
+++ b/src/anndata/_core/index.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from collections.abc import Iterable, Sequence
 from functools import singledispatch
 from itertools import repeat
-from types import EllipsisType
 from typing import TYPE_CHECKING
 
 import h5py
@@ -14,7 +13,7 @@ from scipy.sparse import issparse, spmatrix
 from ..compat import AwkArray, DaskArray, SpArray
 
 if TYPE_CHECKING:
-    from ..compat import Index, IndexRest
+    from ..compat import Index, Index1D
 
 
 def _normalize_indices(
@@ -30,11 +29,9 @@ def _normalize_indices(
         if len(index) > 2:
             # only one remaining ellipsis
             num_ellipsis = sum(i is Ellipsis for i in index)
-            if num_ellipsis == 1:
-                index = tuple(i for i in index if i is not Ellipsis)
-            elif num_ellipsis > 1:
+            if num_ellipsis > 1:
                 raise IndexError("an index can only have a single ellipsis ('...')")
-            if len(index) > 2:
+            if num_ellipsis == 0:
                 raise ValueError("AnnData can only be sliced in rows and columns.")
         # deal with pd.Series
         # TODO: The series should probably be aligned first
@@ -55,8 +52,7 @@ def _normalize_index(
     | str
     | Sequence[bool | int | np.integer]
     | np.ndarray
-    | pd.Index
-    | EllipsisType,
+    | pd.Index,
     index: pd.Index,
 ) -> slice | int | np.ndarray:  # ndarray of int or bool
     if not isinstance(index, pd.RangeIndex):
@@ -81,8 +77,6 @@ def _normalize_index(
         return slice(start, stop, step)
     elif isinstance(indexer, np.integer | int):
         return indexer
-    elif isinstance(indexer, EllipsisType):
-        return slice(None)
     elif isinstance(indexer, str):
         return index.get_loc(indexer)  # int
     elif isinstance(
@@ -140,15 +134,30 @@ def _fix_slice_bounds(s: slice, length: int) -> slice:
     return slice(start, stop, step)
 
 
-def unpack_index(index: Index) -> tuple[IndexRest, IndexRest]:
+def unpack_index(index: Index) -> tuple[Index1D, Index1D]:
     if not isinstance(index, tuple):
+        if index is Ellipsis:
+            index = slice(None)
         return index, slice(None)
-    elif len(index) == 2:
+    num_ellipsis = sum(i is Ellipsis for i in index)
+    if num_ellipsis > 1:
+        raise IndexError("an index can only have a single ellipsis ('...')")
+    # If index has Ellipsis, filter it out (and if not, error)
+    if len(index) > 2:
+        if not num_ellipsis:
+            raise IndexError("Received a length 3 index without an ellipsis")
+        index = tuple(i for i in index if i is not Ellipsis)
         return index
-    elif len(index) == 1:
-        return index[0], slice(None)
-    else:
-        raise IndexError("invalid number of indices")
+    # If index has Ellipsis, replace it with slice
+    if len(index) == 2:
+        index = tuple(slice(None) if i is Ellipsis else i for i in index)
+        return index
+    if len(index) == 1:
+        index = index[0]
+        if index is Ellipsis:
+            index = slice(None)
+        return index, slice(None)
+    raise IndexError("invalid number of indices")
 
 
 @singledispatch
--- a/src/anndata/_core/index.py
+++ b/src/anndata/_core/index.py
@@ -29,9 +29,9 @@ def _normalize_indices(
     if isinstance(index, tuple):
         if len(index) > 2:
             # only one remaining ellipsis
-            num_ellipsis = sum(isinstance(i, EllipsisType) for i in index)
+            num_ellipsis = sum(i is Ellipsis for i in index)
             if num_ellipsis == 1:
-                index = tuple(i for i in index if not isinstance(i, EllipsisType))
+                index = tuple(i for i in index if i is not Ellipsis)
             elif num_ellipsis > 1:
                 raise IndexError("an index can only have a single ellipsis ('...')")
             if len(index) > 2:
--- a/src/anndata/_core/index.py
+++ b/src/anndata/_core/index.py
@@ -26,13 +26,8 @@ def _normalize_indices(
     if isinstance(index, pd.Series):
         index: Index = index.values
     if isinstance(index, tuple):
-        if len(index) == 2:
-            # deal with pd.Series
-            # TODO: The series should probably be aligned first
-            if isinstance(index[1], pd.Series):
-                index = index[0], index[1].values
-            if isinstance(index[0], pd.Series):
-                index = index[0].values, index[1]
+        # TODO: The series should probably be aligned first
+        index = tuple(i.values if isinstance(i, pd.Series) else i for i in index)
     ax0, ax1 = unpack_index(index)
     ax0 = _normalize_index(ax0, names0)
     ax1 = _normalize_index(ax1, names1)
--- a/src/anndata/_core/index.py
+++ b/src/anndata/_core/index.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Iterable, Sequence
 from functools import singledispatch
 from itertools import repeat
+from types import EllipsisType
 from typing import TYPE_CHECKING
 
 import h5py
@@ -47,7 +48,8 @@ def _normalize_index(
     | str
     | Sequence[bool | int | np.integer]
     | np.ndarray
-    | pd.Index,
+    | pd.Index
+    | EllipsisType,
     index: pd.Index,
 ) -> slice | int | np.ndarray:  # ndarray of int or bool
     if not isinstance(index, pd.RangeIndex):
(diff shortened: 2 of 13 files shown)


src/anndata/compat/__init__.py
```diff
@@ -46,7 +47,7 @@ class Empty:
     pass
 
 
-Index1D = slice | int | str | np.int64 | np.ndarray
+Index1D = slice | int | str | np.int64 | np.ndarray | EllipsisType
 Index = Index1D | tuple[Index1D, Index1D] | scipy.sparse.spmatrix | SpArray
```
flying-sheep, line 51: Since we use `Index1D` elsewhere, I think we should use more correct types:

`...` doesn‘t stand for one dimension, it stands for “the other dimensions”, so it only makes sense in the context of a multidimensional array

```suggestion
Index1D = slice | int | str | np.int64 | np.ndarray
IndexRest = Index1D | Ellipsis
Index = IndexRest | tuple[Index1D, IndexRest] | tuple [IndexRest, Index1D] | scipy.sparse.spmatrix | SpArray
```

tests/test_views.py
```diff
@@ -786,6 +786,14 @@ def test_dataframe_view_index_setting():
     assert a2.obs.index.values.tolist() == ["a", "b"]
 
 
+def test_ellipsis_index(adata, subset_func, matrix_type):
```
flying-sheep, line 789: Should also test `adata[...]` and `adata[..., subset_var_names]`

ilan-gold: Good call!

src/anndata/compat/__init__.py
flying-sheep: ah, sorry, I mean [types.EllipsisType](https://docs.python.org/3/library/types.html#types.EllipsisType)

I thought it would work like `None`, but I guess since `... is Ellipsis` and `...` has special meaning for typing, they needed to distinguish `tuple[int, ...]` (int tuple of unspecified length) from `tuple[int, EllipsisType]` (duple with an int and an `Ellipsis`/`...`)

tests/test_views.py
flying-sheep: Oh! And even more tricky:
- `adata[obs, var, ...]`
- `adata[obs, ..., var]`
- `adata[..., obs, var]`

in these cases it’ll expand to nothing actually.

ilan-gold: Is this in the scope of this issue though?  I would think this is more along the lines of ND AnnData?

src/anndata/compat/__init__.py
ilan-gold: Similarly, the two-elem tuple conflicts with https://github.com/scverse/anndata/pull/1729#discussion_r1810772405

tests/test_views.py
flying-sheep: If we want to support ellipsis indexing, we should support its full semantics, and those are

> like `:` but for each remaining axis.

So if 2 axes are remaining, it expands to `:, :`, and if 0 axes are remaining, it vanishes.

ilan-gold: Done.  We error on more than one ellipsis as well now with a more helpful message than before

docs/conf.py
```diff
@@ -112,6 +112,7 @@
     ("py:obj", "numpy._typing._array_like._ScalarType_co"),
     # https://github.com/sphinx-doc/sphinx/issues/10974
     ("py:class", "numpy.int64"),
+    ("py:class", "types.EllipsisType"),
```
ilan-gold, line 115: No idea what to do about this

flying-sheep: I’ll take a look.

By the way, the new `--exception-on-warning` flag is awesome to debug Sphinx stuff, finally this has become as easy as regular Python code.

src/anndata/_core/index.py
```diff
@@ -130,7 +140,7 @@ def _fix_slice_bounds(s: slice, length: int) -> slice:
     return slice(start, stop, step)
 
 
-def unpack_index(index: Index) -> tuple[Index1D, Index1D]:
+def unpack_index(index: Index) -> tuple[IndexRest, IndexRest]:
```
flying-sheep, line 143: Shouldn’t this be `(Index without Ellipsis) -> tuple[Index1D, Index1D]`?

docs/conf.py
flying-sheep:
```suggestion
    # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/498
    ("py:class", "types.EllipsisType"),
```

src/anndata/_core/index.py
ilan-gold: No, because this can return `EllipsisType` on either of the entries

src/anndata/_core/index.py
```diff
@@ -27,7 +28,14 @@ def _normalize_indices(
         index: Index = index.values
     if isinstance(index, tuple):
         if len(index) > 2:
-            raise ValueError("AnnData can only be sliced in rows and columns.")
+            # only one remaining ellipsis
+            num_ellipsis = sum(isinstance(i, EllipsisType) for i in index)
+            if num_ellipsis == 1:
+                index = tuple(i for i in index if not isinstance(i, EllipsisType))
```
flying-sheep, line 34: checks should be `i is Ellipsis` or `i is ...`

src/anndata/_core/index.py
```diff
@@ -72,6 +81,8 @@ def name_idx(i):
         return slice(start, stop, step)
     elif isinstance(indexer, np.integer | int):
         return indexer
+    elif isinstance(indexer, EllipsisType):
```
flying-sheep, line 84: ditto

src/anndata/_core/index.py
flying-sheep: my point is that it probably shouldn’t. see below

ilan-gold: In other words, `Index` can contain an ellipsis

flying-sheep (changes requested):
Our index handling isn’t very consistent, and while parts of that make sense (`BaseCompressedSparseDataset` doesn‘t need to handle string indexing), we should decide what we want. Currently:

- in `anndata._core.index._normalize_indices`, you remove the ellipses first, so `unpack_index` never gets an ellipsis passed from there.
- `Raw` doesn’t use `_normalize_indices`, it seems to have some hacky slimmed down version of it that `raise`s when an index tuple isn’t len() 2. That might be OK since we want to phase out Raw.
- `BaseCompressedSparseDataset` has an even simpler version that just passes in the index unchanged to `unpack_index`. This means that trying to index it with something that one doesn’t support breaks with a hard-to-decipher error

so we need to decide:
1. should `unpack_index` handle all cases? or does it only handle preprocessed indices?
2. how do we reuse code here smartly? We should optimally only remove ellipses in one place of our code base, but still support both classes that allow string indexing and classes that don’t

src/anndata/_core/index.py
ilan-gold: Ok, but @flying-sheep then if we want to use `unoack_index` everywhere, we need to let it accept tuples of length > 2

ilan-gold:
@flying-sheep took a shot - only one ellipsis removal, and `Index` type was updated to be a bit more accurate (I think)

ilan-gold:
With the exception of `raw` - I could fix there, but we don't test for it and want to get rid of it anyway.  someone would have to do `adata.raw[ellipsis_index]` and I don't think they should, quite frankly.  `adata[ellipsis_index].raw` is fine

src/anndata/_core/index.py
```diff
@@ -30,11 +29,9 @@ def _normalize_indices(
...
             # only one remaining ellipsis
             num_ellipsis = sum(i is Ellipsis for i in index)
-            if num_ellipsis == 1:
-                index = tuple(i for i in index if i is not Ellipsis)
-            elif num_ellipsis > 1:
+            if num_ellipsis > 1:
                 raise IndexError("an index can only have a single ellipsis ('...')")
-            if len(index) > 2:
+            if num_ellipsis == 0:
                 raise ValueError("AnnData can only be sliced in rows and columns.")
```
flying-sheep, line 35: seems like you do this check a second time in `unpack_index`.

if there is a reason, maybe extract it as a helper and add a comment explaining why you need to do it before calling `unpack_index`

tests/test_backed_sparse.py
```diff
@@ -127,6 +127,28 @@ def test_backed_indexing(
     assert_equal(csr_mem[:, var_idx].X, dense_disk[:, var_idx].X)
 
 
+@pytest.mark.parametrize(
+    "indexing_func",
+    [
+        lambda x: (Ellipsis, slice(0, 10)),
+        lambda x: (slice(0, 10), Ellipsis),
+        lambda x: (slice(0, 10), slice(0, 10), Ellipsis),
+        lambda x: (Ellipsis, slice(0, 10), slice(0, 10)),
```
flying-sheep, line 136: you don’t seem to use `x` in any of the cases here, also I’m adding the “ellipsis in the middle” case:

```suggestion
        (..., slice(0, 10)),
        (slice(0, 10), ...),
        (slice(0, 10), slice(0, 10), ...),
        (..., slice(0, 10), slice(0, 10)),
        (slice(0, 10), ..., slice(0, 10)),
```

or something like

```py
class Idx:
    def __class_getitem__(idx):
        return idx
```


```suggestion
        Idx[..., :10],
        Idx[:10, ...],
        Idx[:10, :10, ...],
        Idx[..., :10, :10],
        Idx[:10, ..., :10],
```

src/anndata/compat/__init__.py
```diff
@@ -53,6 +53,8 @@ class Empty:
     IndexRest
     | tuple[Index1D, IndexRest]
     | tuple[IndexRest, Index1D]
+    | tuple[Index1D, Index1D, EllipsisType]
+    | tuple[EllipsisType, Index1D, Index1D]
```
flying-sheep, line 57:
```suggestion
    | tuple[Index1D, Index1D, EllipsisType]
    | tuple[EllipsisType, Index1D, Index1D]
    | tuple[Index1D, EllipsisType, Index1D]
```

flying-sheep (commented):
awesome! does this give `BaseCompressedSparseDataset` the ability to be indexed with `...` as well?

ilan-gold:
>  awesome! does this give BaseCompressedSparseDataset the ability to be indexed with ... as well?


It should, I added tests for it

tests/test_views.py
```diff
@@ -786,6 +790,49 @@ def test_dataframe_view_index_setting():
...
+    subset_with_ellipsis = tuple(
+        Ellipsis if i == ellipsis_index_loc else next(generator) for i in range(3)
+    )
+    subset_with_slice = (obs_subset, var_subset)
+    subset_ellipsis = adata[subset_with_ellipsis]
+    subset = adata[subset_with_slice]
+    assert_equal(subset_ellipsis, subset)
+
+
+def test_ellipsis_index_3d_2_ellipses_errors(adata: ad.AnnData):
```
flying-sheep, line 831: this parameter is unused

src/anndata/_core/index.py
flying-sheep: I checked, all this can just be

```suggestion
        elif isinstance(index, tuple):
            index = tuple(i.array if isinstance(i, pd.Series) else i for i in index)
```

tests/conftest.py
```diff
@@ -10,12 +11,37 @@
...
+        ((slice(0, 10), ..., slice(0, 10)), (slice(0, 10), slice(0, 10))),
+    ],
+    ids=[
+        "ellipsis",
+        "obs-ellipsis",
+        "var-ellipsis",
+        "obs-var-ellipsis",
+        "ellipsis-obs-var",
+        "obs-ellipsis-var",
+    ],
```
flying-sheep, line 39: for so many params, do `pytest.param(id=...)` instead of `ids`

src/anndata/_core/index.py
```diff
@@ -26,14 +26,13 @@ def _normalize_indices(
...
-            index = index[0], index[1].values
-        if isinstance(index[0], pd.Series):
-            index = index[0].values, index[1]
+        if len(index) == 2:
+            # deal with pd.Series
+            # TODO: The series should probably be aligned first
+            if isinstance(index[1], pd.Series):
+                index = index[0], index[1].values
+            if isinstance(index[0], pd.Series):
+                index = index[0].values, index[1]
```
flying-sheep, line 35: as said, why not just this?

```suggestion
        # TODO: The series should probably be aligned first
        index = tuple(i.values if isinstance(i, pd.Series) else i for i in index)
```

tests/conftest.py
```diff
@@ -10,12 +11,62 @@
...
+
 
 @pytest.fixture
 def backing_h5ad(tmp_path):
     return tmp_path / "test.h5ad"
 
 
+@pytest.fixture(
+    params=[
+        pytest.param(((...), (slice(None), slice(None))), id="ellipsis"),
```
flying-sheep, line 25: `(...)` isn’t a tuple, maybe you wanted it to be one?

flying-sheep (approved):
looking great! there are a few minor issues with the types, otherwise perfect!

tests/conftest.py
ilan-gold: hmmm true, yes, the return type should just include this then in a union.  thanks!

flying-sheep:
I changed the milestone since the relnote is `feature` and we soon release 0.11

PR #1744, opened 2024-11-08
(fix): cache arrays in `BaseCompressedSparseDataset`

ilan-gold:
- [x] Noticed while creating the screenshot tutorial for the announcement
- [x] Tests added


src/anndata/tests/helpers.py
```diff
@@ -1057,7 +1057,7 @@ class AccessTrackingStore(DirectoryStore):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._access_count = Counter()
-        self._accessed_keys = {}
+        self._accessed_keys = dict()
```
flying-sheep, line 1060: ?

flying-sheep (commented):
Does what it promises and has nice tests.

However I think Isaac only cached the longest array intentionally. I forgot the reason though.

Sure, we now have much bigger data so maybe if there was a tradeoff, things are now different. But maybe worth checking.

ilan-gold:
> However I think Isaac only cached the longest array intentionally. I forgot the reason though.

We only cached `indptr` previously as a `np.ndarray` i.e., we read it into memory because it needs to be read in anywya ever time you make an access.

As for the other arrays, I don't think there was a good reason - I think we wanted to just start with that but I posted on zulip.

ilan-gold:
(I was the one who started doing the caching)

flying-sheep:
Hm, since these arrays are storage classes anyway, and probably don’t cost any noticable amount of memory, I don‘t foresee a problem.

ilan-gold:
Let's review this in the absence of an answer

ilan-gold:
(Or rather, give approval since you already seem to have reviewed)

src/anndata/_core/sparse_dataset.py
```diff
@@ -545,16 +546,18 @@ def append(self, sparse_matrix: ss.csr_matrix | ss.csc_matrix | SpArray) -> None
...
         # indices
         indices = self.group["indices"]
         orig_data_size = indices.shape[0]
         indices.resize((orig_data_size + sparse_matrix.indices.shape[0],))
         indices[orig_data_size:] = sparse_matrix.indices
 
+        # Clear cached property
+        for attr in ["_indptr", "_indices", "_data"]:
+            if hasattr(self, attr):
+                delattr(self, attr)
```
flying-sheep, line 559: before it was

```py
if hasattr(self, "indptr"):  # no underscore
    del self._indptr  # underscore
```

now it checks and `del`s the same name (with underscore).

is the new test testing that this works as expected or could that change be a problem?

ilan-gold: Right, I think this was a bug.  So without this change, the test fails:

```python
>           assert old_array_shapes[name] != getattr(diskmtx, f"_{name}").shape
E           AssertionError: assert (101,) != (101,)
E            +  where (101,) = array([  0,   1,   1,   2,   2,   4,   4,   4,   5,   5,   5,   5,   8,\n         8,   9,  11,  12,  12,  14,  16,  18,...76,  76,  78,  78,  79,  81,  82,  83,  85,  86,\n        87,  90,  92,  94,  97,  97,  97,  97,  97, 100], dtype=int32).shape
E            +    where array([  0,   1,   1,   2,   2,   4,   4,   4,   5,   5,   5,   5,   8,\n         8,   9,  11,  12,  12,  14,  16,  18,...76,  76,  78,  78,  79,  81,  82,  83,  85,  86,\n        87,  90,  92,  94,  97,  97,  97,  97,  97, 100], dtype=int32) = getattr(CSRDataset: backend zarr, shape (200, 100), data_dtype float64, '_indptr')```

ilan-gold: Ancillary, but direct relevant to the PR

flying-sheep (approved):

PR #1752, opened 2024-11-11
Merge 0.11 docs with rcs

flying-sheep:
This would be an idea. It’s pretty manual again and we wanted less manual process. IDK if we can automate this better.

https://scverse.zulipchat.com/#narrow/channel/315185-general/topic/AnnData.200.2E11.2E0/near/481456047


ilan-gold (approved):
Seems doable @flying-sheep but maybe not a huge priority since it only needs to be done on >minor releases.  Maybe a python markdown parser + script that runs in the release job?

Zethson (commented):
I'm trying to figure out how other big projects seem to do it and if I see it right, they:

1. Have full release notes per RC -> RC 3 also has all changes of RC2 and RC1 and not just the diff between RCs
2. The final release has ALL changes and does not mention the RCs again.
3. If I see this right, packages like numpy, pandas or Pytorch only render non RC releases on their docs but announce the RC changelogs via Github releases.

This PR is fine with me but I just wanted to mention that for discussion.

Zethson (approved):

PR #1758, opened 2024-11-12
Harmonize min-deps script with scanpy

flying-sheep:
Before this PR, a dependency that only had `!=` constraints broke the script

### DIFF
--- a/ci/scripts/min-deps.py
+++ b/ci/scripts/min-deps.py
@@ -1,4 +1,11 @@
 #!/usr/bin/env python3
+# /// script
+# dependencies = [
+#   "tomli; python_version < '3.11'",
+#   "packaging",
+# ]
+# ///
+
 from __future__ import annotations
 
 import argparse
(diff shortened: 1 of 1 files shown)


ilan-gold (commented):
This looks the same as before, no? The names just changed?

flying-sheep:
No, the logic of the `if` statement is different. Before it checked if there were no conditions, now it checks if there are no conditions that we handle below.

ci/scripts/min-deps.py
```diff
@@ -33,12 +40,14 @@ def min_dep(req: Requirement) -> Requirement:
     if req.extras:
         req_name = f"{req_name}[{','.join(req.extras)}]"
 
-    specs = [spec for spec in req.specifier if spec.operator in {">", ">=", "~=", "=="}]
-    if not specs:
+    filter_specs = [
+        spec for spec in req.specifier if spec.operator in {"==", "~=", ">=", ">"}
+    ]
+    if not filter_specs:
```
ilan-gold, line 46: I guess what I was saying was that this condition looks identical but renamed

flying-sheep: I don‘t see it :shrug: what do you suggest to fix that?

ilan-gold (approved):

PR #1767, opened 2024-11-14
(chore): tighten `spmatrix` usage

### LINKED ISSUE #1649: (chore): clean up use of `spmatrix` and `sparray`

At the moment, we use _a lot_ of `spmatrix` and `sparray` as types where they shouldn't be used.   We should clean up the types and do the promised deprecation on instance checking: https://github.com/scverse/anndata/pull/1616/files#diff-3b4874efa9bc9a522fa0cc5a88c69e233d7b46f2993331162bab12c4a6371bdcR70-R74

We should also look for other faulty instance checks

ilan-gold:
- [x] Closes #1649
- [x] Tests added

### DIFF
--- a/docs/release-notes/1767.breaking.md
+++ b/docs/release-notes/1767.breaking.md
@@ -1,0 +1,1 @@
+Tighten usage of {class}`scipy.sparse.spmatrix` for describing sparse matrices in types and instance checks to only {class}`scipy.sparse.csr_matrix` and {class}`scipy.sparse.csc_matrix` {user}`ilan-gold`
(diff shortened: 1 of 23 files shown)


ilan-gold:
Problematically, older scanpy versions have the `read` import and we probably want to support older versions for a while since https://github.com/scverse/scanpy/tree/6d234a7ec8c33348b7778e9ccb8fe7f226e4723d is only three weeks old.  So fixing this probably requires bumping the minimum supported version of scanpy with anndata, which I don't think we are ready to do.

ilan-gold:
Since this isn't until `0.12` I would like to wait to see if `numba` comes out with `python 3.13` support and then drop 3.10 and do all of the scientific python ecosystem all at once: https://github.com/scverse/anndata/pull/1768 

Then we don't have to worry about `register`

ilan-gold:
Since this isn't until `0.12` I would like to wait to see if `numba` comes out with `python 3.13` support and then drop 3.10 and do all of the scientific python ecosystem all at once: https://github.com/scverse/anndata/pull/1768 

Then we don't have to worry about `register`

PR #1768, opened 2024-11-15
(chore): adopt scientific python deprecation schedule

### LINKED ISSUE #1649: (chore): clean up use of `spmatrix` and `sparray`

ilan-gold:
This PR forces us to adopt the scientific python deprecation schedule as much as possible.  It also cleans up our use of spmatrix/sparray because of the bumping of `scipy.sparse` so that handling `sparray` is always done via a concrete class

- [x] Closes #1649
- [x] Tests added


.azure-pipelines.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           set -e
           uv pip install --system --compile tomli packaging
           deps=`python3 ci/scripts/min-deps.py pyproject.toml --extra dev test`
-          uv pip install --system --compile $deps pytest-cov "anndata @ ."
+          uv pip install --system --compile $deps pytest-cov "anndata @ .[test,dev]"
```
ilan-gold, line 60: @flying-sheep I think this is a bug, no? The min-deps are created without respect for upper bounds so the constraints can end up not including the upper bounds because the `min-deps` script doesn't care about it? I think this is the correct fix given the fact that the script is called "min-deps" and not "min-deps-with-upper-bound" but maybe I'm off

.azure-pipelines.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           set -e
           uv pip install --system --compile tomli packaging
           deps=`python3 ci/scripts/min-deps.py pyproject.toml --extra dev test`
-          uv pip install --system --compile $deps pytest-cov "anndata @ ."
+          uv pip install --system --compile $deps pytest-cov "anndata[test,dev] @ ."
```
ilan-gold, line 60: @flying-sheep I think this is a bug, no? The min-deps are created without respect for upper bounds so the constraints can end up not including the upper bounds because the min-deps script doesn't care about it? I think this is the correct fix given the fact that the script is called "min-deps" and not "min-deps-with-upper-bound" but maybe I'm off

flying-sheep: Good point! However I think the better fix would be to change the script so it leaves upper bounds (and things like environment markers) alone.

The script does already do more than just minimal deps, as it only pins deps with minimum version boundaries. Deps without that (only upper bounds, no bounds, …) are passed through without any minimizing!

So I’d say we could do two things:
1. change the script to either leave upper bounds alone or just skip deps without a minimum boundary entirely (just returning pinned dependencies)
2. depending on the decision for 1., the workflow would stay unchanged from `main` or we should change the CLI so it takes a comma separated list of extras and reuse it for both locations (`--extra=$extras` and `"anndata[$extras] @ ."`)

flying-sheep (commented):
Looks really good! So many simplifications. I have a few minor questions and nitpicks apart from https://github.com/scverse/anndata/pull/1768#discussion_r1958130072, that’s all.

pyproject.toml
```diff
@@ -29,20 +29,20 @@ classifiers = [
...
     # pandas 2.1.0rc0 has pandas/issues/54622
-    "pandas >=1.4, !=2.1.0rc0, !=2.1.2",
-    "numpy>=1.23",
+    "pandas >=2.0.0, !=2.1.0rc0, !=2.1.2",
+    "numpy>=1.25",
     # https://github.com/scverse/anndata/issues/1434
-    "scipy >1.8",
-    "h5py>=3.7",
+    "scipy >1.11",
+    "h5py>=3.8",
```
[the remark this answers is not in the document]
ilan-gold: Yes, our min-deps script does not handle python version so it will try to install e.g., 3.7 h5py against python 3.11 even though it does not have a python3.11 qualifier, whihc came in 3.8 https://docs.h5py.org/en/stable/whatsnew/3.8.html

Other than that, numpy, scipy, and pandas are all compliant.

pyproject.toml
```diff
@@ -108,13 +108,15 @@ gpu = ["cupy"]
...
-dask = ["dask[array]>=2022.09.2,!=2024.8.*,!=2024.9.*"]
+dask = ["dask[array]>=2023.1.0,!=2024.8.*,!=2024.9.*"]
 
 [tool.hatch.version]
 source = "vcs"
 raw-options.version_scheme = "release-branch-semver"
 [tool.hatch.build.targets.wheel]
 packages = ["src/anndata", "src/testing"]
+[tool.hatch.metadata]
+allow-direct-references = true
```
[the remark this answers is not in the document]
ilan-gold: I use branches pointing at git sometimes for testing e.g., zarr v3 sometimes (somewhat often tbh).

src/anndata/compat/__init__.py
```diff
@@ -31,16 +23,8 @@
...
-if not CAN_USE_SPARSE_ARRAY:
-
-    class SpArray:
-        @staticmethod
-        def __repr__():
-            return "mock scipy.sparse.sparray"
-else:
-    SpArray = scipy.sparse.sparray
+SpMatrix = scipy.sparse.csr_matrix | scipy.sparse.csc_matrix
+SpArray = scipy.sparse.csr_array | scipy.sparse.csc_array
```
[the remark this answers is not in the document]
ilan-gold: Yes you're right, meant to do that after seeing you scanpy PR

src/anndata/_core/anndata.py
```diff
@@ -1189,7 +1189,7 @@ def _inplace_subset_obs(self, index: Index1D):
         self._init_as_actual(adata_subset)
 
     # TODO: Update, possibly remove
-    def __setitem__(self, index: Index, val: float | np.ndarray | sparse.spmatrix):
+    def __setitem__(self, index: Index, val: ArrayDataStructureType):
```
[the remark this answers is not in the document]
ilan-gold: Just missed it

src/anndata/_core/aligned_mapping.py
```diff
@@ -36,7 +35,7 @@
 OneDIdx = Sequence[int] | Sequence[bool] | slice
 TwoDIdx = tuple[OneDIdx, OneDIdx]
 # TODO: pd.DataFrame only allowed in AxisArrays?
-Value = pd.DataFrame | spmatrix | np.ndarray
+Value = pd.DataFrame | SpMatrix | np.ndarray
```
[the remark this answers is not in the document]
ilan-gold: Missed it

flying-sheep (approved):

src/anndata/_core/merge.py
```diff
@@ -844,6 +841,13 @@ def concat_arrays(arrays, reindexers, axis=0, index=None, fill_value=None):
             ],
             format="csr",
         )
+        scipy_version = Version(scipy.__version__)
+        # Bug where xstack produces a matrix not an array in 1.11.*
+        if use_sparse_array and (scipy_version.major, scipy_version.minor) == (1, 11):
+            if mat.format == "csc":
+                return sparse.csc_array(mat)
+            return sparse.csr_array(mat)
+        return mat
```
flying-sheep, line 850: should this be backported?

src/anndata/_core/storage.py
```diff
@@ -45,7 +45,7 @@ def coerce_array(
         (isinstance(value, base) and not isinstance(value, csr_c_format))
         for base, csr_c_format in [
             (sparse.spmatrix, sparse.csr_matrix | sparse.csc_matrix),
-            (SpArray, sparse.csr_array | sparse.csc_array),
+            (CSArray, sparse.csr_array | sparse.csc_array),
```
flying-sheep, line 48: bug!

src/anndata/_core/sparse_dataset.py
```diff
@@ -494,7 +494,7 @@ def __setitem__(self, index: Index | tuple[()], value) -> None:
         mock_matrix[row, col] = value
 
     # TODO: split to other classes?
-    def append(self, sparse_matrix: ss.csr_matrix | ss.csc_matrix | SpArray) -> None:
+    def append(self, sparse_matrix: ss.csr_matrix | ss.csc_matrix | CSArray) -> None:
```
flying-sheep, line 497: some lines like this could be prettified by using `CSMatrix`

src/anndata/abc.py
```diff
@@ -31,7 +31,7 @@ class _AbstractCSDataset(ABC):
     """Which file type is used on-disk."""
 
     @abstractmethod
-    def __getitem__(self, index: Index) -> float | csr_matrix | csc_matrix | SpArray:
+    def __getitem__(self, index: Index) -> float | csr_matrix | csc_matrix | CSArray:
```
flying-sheep, line 34: can also be prettified

PR #1774, opened 2024-11-20
(fix): python debugger dask h5 meta array

ilan-gold:
Noticed while using the python debugger - dask sets `block_info` to `None` when run this way, which causes the function to error but setting `meta` (probably a good idea anyway) prevents this.

This also drive-by fixes an "internal" bug which did not have an effect downstream, but still noticed.  And I added a test (which passes anyway without this change, but still a sensible test).

- [ ] Closes #
- [x] Tests added
- [x] Release note added (or unnecessary): does not change runtime behavior

### DIFF
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -150,7 +150,7 @@ def read_h5_array(
             c if c not in {None, -1} else s for c, s in zip(chunks, shape, strict=True)
         )
         if chunks is not None
-        else (_DEFAULT_STRIDE,) * len(shape)
+        else tuple(min(_DEFAULT_STRIDE, s) for s in shape)
     )
 
     chunk_layout = tuple(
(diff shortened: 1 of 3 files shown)


tests/test_io_elementwise.py
```diff
@@ -66,22 +66,23 @@ def store(request, tmp_path) -> H5Group | ZarrGroup:
...
 
 
 @pytest.fixture(params=sparse_formats)
 def sparse_format(request):
     return request.param
 
 
-def create_dense_store(store, n_dims: int = 2):
-    X = np.random.randn(*[SIZE * (i + 1) for i in range(n_dims)])
+def create_dense_store(store: str, *, shape=DEFAULT_SHAPE) -> H5Group | ZarrGroup:
```
flying-sheep, line 77: types

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -150,7 +150,7 @@ def read_h5_array(
             c if c not in {None, -1} else s for c, s in zip(chunks, shape, strict=True)
         )
         if chunks is not None
-        else (_DEFAULT_STRIDE,) * len(shape)
+        else tuple(min(_DEFAULT_STRIDE, s) for s in shape)
```
flying-sheep, line 153: can you explain this change?

flying-sheep (approved):

PR #1790, opened 2024-12-08
(performance): speedup for `zarr`-based sparse indexing

ilan-gold:
```python
import scipy.sparse as sp
import numpy as np
import anndata as ad
import zarr
from pathlib import Path

data_path = Path('data/foo.zarr')
if not data_path.exists():
    z = zarr.open_group('data/foo.zarr', mode='w')
    arr = sp.random(1_000_000, 10_000, format="csr", random_state=np.random.default_rng())
    ad.io.write_elem(z, "X", arr)
z = zarr.open_group('data/foo.zarr', mode='r')

adata = ad.AnnData(X=ad.io.sparse_dataset(z["X"]))
idx = np.random.randint(0, adata.shape[0], 1024)
idx.sort()
%time adata[idx].X
```
On my computer, this is a 3-5X speedup.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -72,7 +72,9 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Install AnnData
-        run: uv pip install --system -e ".[dev,test,cu12]" llvmlite>=0.43
+        run:
+          printf "llvmlite>=0.43\nscanpy>=1.10.0rc1" | tee /tmp/constraints.txt
+          uv pip install --system -e ".[dev,test,cu12]" -c /tmp/constraints.txt
 
       - name: Env list
         run: pip list
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -72,7 +72,7 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Install AnnData
-        run:
+        run: |
           printf "llvmlite>=0.43\nscanpy>=1.10.0rc1" | tee /tmp/constraints.txt
           uv pip install --system -e ".[dev,test,cu12]" -c /tmp/constraints.txt
 
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -67,25 +67,14 @@ class BackedSparseMatrix(_cs_matrix):
             return sparse_dataset(self.data.parent).to_memory()
         if isinstance(self.data, ZarrArray):
             import zarr
-            from packaging.version import Version
 
-            is_zarr_v2 = Version(zarr.__version__) < Version("3.0.0b0")
-            if is_zarr_v2:
-                sparse_group = zarr.open(
+            return sparse_dataset(
+                zarr.open(
                     store=self.data.store,
                     mode="r",
                     chunk_store=self.data.chunk_store,  # chunk_store is needed, not clear why
                 )[Path(self.data.path).parent]
-            else:
-                anndata_group = zarr.open_group(store=self.data.store, mode="r")
-                sparse_group = anndata_group[
-                    str(
-                        Path(str(self.data.store_path))
-                        .relative_to(str(anndata_group.store_path))
-                        .parent
-                    )
-                ]
-            return sparse_dataset(sparse_group).to_memory()
+            ).to_memory()
         return super().copy()
 
     def _set_many(self, i: Iterable[int], j: Iterable[int], x):
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -67,14 +67,25 @@ class BackedSparseMatrix(_cs_matrix):
             return sparse_dataset(self.data.parent).to_memory()
         if isinstance(self.data, ZarrArray):
             import zarr
+            from packaging.version import Version
 
-            return sparse_dataset(
-                zarr.open(
+            is_zarr_v2 = Version(zarr.__version__) < Version("3.0.0b0")
+            if is_zarr_v2:
+                sparse_group = zarr.open(
                     store=self.data.store,
                     mode="r",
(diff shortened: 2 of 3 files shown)


ilan-gold:
Very good :)

.github/workflows/test-gpu.yml
```diff
@@ -72,7 +72,9 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Install AnnData
-        run: uv pip install --system -e ".[dev,test,cu12]" llvmlite>=0.43
+        run:
+          printf "llvmlite>=0.43\nscanpy>=1.10.0rc1" | tee /tmp/constraints.txt
+          uv pip install --system -e ".[dev,test,cu12]" -c /tmp/constraints.txt
```
ilan-gold, line 77: drive-by fix....

src/anndata/_core/sparse_dataset.py
```diff
@@ -250,30 +250,44 @@ def slice_as_int(s: slice, l: int) -> int:
...
-    indices = np.concatenate([x.indices[s[0] : s[-1]] for s in indptr_sels])
+    indptr_list = [x.indptr[slice(s.start, s.stop + 1)] for s in slices]
+    # HDF5 cannot handle out-of-order integer indexing
+    if isinstance(x.data, ZarrArray):
+        indptr_int = np.concatenate([np.arange(s[0], s[-1]) for s in indptr_list])
+        data = x.data[indptr_int]
+        indices = x.indices[indptr_int]
+    else:
+        data = np.concatenate([x.data[s[0] : s[-1]] for s in indptr_list])
+        indices = np.concatenate([x.indices[s[0] : s[-1]] for s in indptr_list])
```
flying-sheep, line 279:
```suggestion
    if isinstance(x.data, ZarrArray):
        indptr_int = np.concatenate([np.arange(start, end) for start, end in indptr_list])
        data = x.data[indptr_int]
        indices = x.indices[indptr_int]
    else:
        data = np.concatenate([x.data[start:end] for start, end in indptr_list])
        indices = np.concatenate([x.indices[start:end] for start, end in indptr_list])
```

src/anndata/_core/sparse_dataset.py
```diff
@@ -250,30 +250,44 @@ def slice_as_int(s: slice, l: int) -> int:
...
     # Need to track the size of the gaps in the slices to each indptr subselection
-    total = indptr_sels[0][0]
+    total = indptr_list[0][0]
     offsets = [total]
-    for i, sel in enumerate(indptr_sels[1:]):
-        total = (sel[0] - indptr_sels[i][-1]) + total
+    for i, sel in enumerate(indptr_list[1:]):
+        total = (sel[0] - indptr_list[i][-1]) + total
```
flying-sheep, line 284: Since you changed the name, `sel` is confusing here.

Also the whole loop is confusing. Please rewrite this to use `itertools.pairwise` instead of `enumerate`. (e.g. something like `for (s0, e0), (s1, e1) in pairwise(indptr_list): ...`)

And then use `sum(<generator comprehension>)` instead of a loop

flying-sheep (changes requested):
great! I think for clarity, wherever possible you should unpack an entry of `indptr_list` into `start, end` instead of using `s[0]` and `s[1]`.

you could either do `indptr_limits = [(x.indptr[s.start], x.indptr[s.stop + 1]) for s in slices]` or so, or just define `indptr_slices` as above.

src/anndata/_core/sparse_dataset.py
ilan-gold: This doesn't work because each array in `indptr_limits` is not actually 2 elements, it could be more than that

src/anndata/_core/sparse_dataset.py
ilan-gold: Again, we run into the issue that `indptr_list` contains >2 element arrays.  I guess we just need two data structures here then.

src/anndata/_core/sparse_dataset.py
flying-sheep: please read what I wrote above, this refers to another change I asked for

ilan-gold: Yup, I think I understood that you were suggesting two data structures.   I thought it was one in place of the other

src/anndata/_core/sparse_dataset.py
```diff
@@ -250,30 +250,47 @@ def slice_as_int(s: slice, l: int) -> int:
...
     # Need to track the size of the gaps in the slices to each indptr subselection
-    total = indptr_sels[0][0]
-    offsets = [total]
-    for i, sel in enumerate(indptr_sels[1:]):
-        total = (sel[0] - indptr_sels[i][-1]) + total
-        offsets.append(total)
-    start_indptr = indptr_sels[0] - offsets[0]
+    offsets = [indptr_indices[0][0]]
+    for (_, stop0), (start1, _) in pairwise(indptr_limits):
+        offsets.append((start1 - stop0) + offsets[-1])
```
flying-sheep, line 288: Should be something like this, right?

```suggestion
    gaps = (start1 - stop0 for (_, stop0), (start1, _) in pairwise(indptr_limits))
    offsets = accumulate(chain([indptr_limits[0][0]] gaps))
```

As said above, you could also do `indptr_slices` instead of `indptr_limits`, then we’d get

```diff
-indptr_int = np.concatenate([np.arange(start, stop) for start, stop in indptr_limits])
+indptr_int = np.concatenate([np.arange(s.start, s.stop) for s in indptr_slices])
```

and

```diff
-offsets = accumulate(chain([indptr_limits[0][0]] gaps))
+offsets = accumulate(chain([indptr_slices[0].start] gaps))
```

which I think is nicer

flying-sheep (approved):

src/anndata/_core/sparse_dataset.py
ilan-gold: The accumulation iterator is great.  It cleans up some code below as well.

flying-sheep (approved):

PR #1818, opened 2025-01-09
(chore): pin `zarr`

ilan-gold:
This represents a rather large change for us, which I had not realized.  We are pinning `zarr` as a test dependency but not as a normal one (not the first time I have misread that).  This could lead to users downloading the wrong version (in fact, it will, if they are not aware of out implicit constraint).

At this point it probably makes sense to add it in...

- [ ] Closes #
- [ ] Tests added


ilan-gold:
If there is agreement here, I would go ahead and remove all the dependency checking code for zarr.

ilan-gold:
No we won't do this.

PR #1819, opened 2025-01-09
(fix): upper bound zarr at runtime (and in docs + benchmark)

ilan-gold:
I would make a release just for this.
- [ ] Closes #
- [ ] Tests added


ilan-gold:
@flying-sheep Is there a way to cross reference old versions of docs? Or are we just going to have to take it on the chin here until we upgrade?

pyproject.toml
flying-sheep, line 1: maybe we should add a `zarr` extra that people can rely on instead of having to maintain their own pin.

then they can just specify `anndata[zarr]` and get the newest supported version

ilan-gold: This struck me as a minor version change.  I think this solution also would be necessary anyway to give a helpful message.

flying-sheep: I think the `raise RuntimeError` you added is the most helpful thing for users here. Although maybe it should be an ImportError, it’s not like users can make a change to the runtime parameters to fix it.

ilan-gold: The thought crossed my mind, I can update

josh-gree:
Is this runtime check not supposed to be for <3 and not <2?

flying-sheep:
`if v.major > 2` is the same as `if v.major >= 3`

josh-gree:
@flying-sheep sorry my bad - I think I am maybe running into some issues with versions and mudata perhaps - something quite strange was happening - was pretty sure I had zarr==2.18.4 and latest annadata - and was getting the `ImportError: zarr-python major version > 2 is not supported` error - I do apologise! I have checked in an isolated env and 2.18.4 with 0.11.3 does work fine!

PR #1829, opened 2025-01-16
(chore): disallow `AnnData(...)` with non-csr{r,c} matrices/arrays

ilan-gold:
- [x] Towards https://github.com/scverse/anndata/issues/1649, removing the ability to instantiate the class externally is a great start (thanks to @flying-sheep for the reminder: https://github.com/scverse/scanpy/pull/3431#discussion_r1918116620)
- [x] Tests added

### DIFF
--- a/tests/test_x.py
+++ b/tests/test_x.py
@@ -184,10 +184,10 @@ def test_set_dense_x_view_from_sparse():
     assert_equal(orig.X[30:], x[30:])  # change propagates through
 
 
-def test_warn_on_non_csr_csc_matrix():
-    X = sparse.eye(100)
-    with pytest.warns(
-        FutureWarning,
-        match=rf"AnnData previously had undefined behavior around matrices of type {type(X)}.*",
+def test_fail_on_non_csr_csc_matrix():
+    X = sparse.eye(100, format="coo")
+    with pytest.raises(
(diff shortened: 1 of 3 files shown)


flying-sheep:
You think we should support [`bsr_array`](https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.sparse.bsr_array.html)?

ilan-gold:
I think it could be worth opening an issue but we should probably fully close the linked issue first and then proceed with adding more.

tests/test_x.py
```diff
@@ -184,10 +184,10 @@ def test_set_dense_x_view_from_sparse():
     assert_equal(orig.X[30:], x[30:])  # change propagates through
 
 
-def test_warn_on_non_csr_csc_matrix():
-    X = sparse.eye(100)
```
flying-sheep, line 188: which `format` does `eye` return and why are you switching to `coo`?

flying-sheep (approved):

tests/test_x.py
ilan-gold: It was returning `DIA` before but I figured I'd just make it explicit here that it's not `cs{r,c}` (exactly because of your first question).

PR #1830, opened 2025-01-16
(docs): Qualname-override everything we can

flying-sheep:
Finally no more unlinked stuff!


Zethson:
Great! Is there a big effect on the build time of the docs? I noticed in other packages that intersphinx can make things substantially slower.

In any case, I think that this is much better.

flying-sheep:
Each intersphinx inventory adds maybe 0.2 seconds to the first build.

So locally with caching, there is no difference. On RTD, it’s not noticable, since the docs take much less time than the tests.

PR #1831, opened 2025-01-17
(chore): Update version number inference in dev environments

flying-sheep:
this PR has two closely related bugfixes:

## actually configure version inference correctly

Turns out I messed up in #1680 and put the config into the wrong section, disabling it. So this PR fixes that:

```diff
 [tool.hatch.version]
 ...
+raw-options.version_scheme = "release-branch-semver"
 [tool.hatch.build.hooks.vcs]
 ...
-raw-options.version_scheme = "release-branch-semver"
```

`main` branch before this PR:

```console
❯ hatch version
0.11.1.dev38+gfaec0f8
```

(this is wrong as it’s <0.11.3 which is already released)

`main` branch after this PR:

```console
❯ hatch version
0.12.0.dev40+gffbacdea
❯ hatch build
──── sdist ────
dist/anndata-0.12.0.dev40+gffbacdea.tar.gz
──── wheel ────
dist/anndata-0.12.0.dev40+gffbacdea-py3-none-any.whl
```

moving that config value to the `tool.hatch.version` table makes build time versions work as expected (e.g. `pip install -e .../anndata` in CI), but this PR also fixes a second problem:

## make the runtime version use our config

We follow “hatch-vcs-footgun-example”, which updated: https://github.com/maresb/hatch-vcs-footgun-example/pull/4

By using `hatchling` APIs instead of `setuptools_vcs` directly, we can make sure the config used by the runtime code is the same used at build time, i.e. in a dev environment, `anndata.__version__` gets calculated in the same way as the version embedded in the metadata.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ __pycache__/
 # Distribution / packaging
 /dist/
 /ci/min-deps.txt
-/src/anndata/_version.py
 /requirements*.lock
 /.python-version
 
--- a/docs/release-notes/1831.dev.md
+++ b/docs/release-notes/1831.dev.md
@@ -1,0 +1,1 @@
+Fix version number inference in development environments (CI and local) {user}`flying-sheep`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,6 @@ dask = ["dask[array]>=2022.09.2,!=2024.8.*,!=2024.9.*"]
 
 [tool.hatch.version]
 source = "vcs"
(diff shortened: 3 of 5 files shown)


ilan-gold (approved):
Works well now, and I think it's a bit cleaner that the example you linked to as well.  As long as CI passes, should be good.

flying-sheep:
The example repo is intended to capture best practices around when you do that (i.e. you want to infer the version at runtime). I’ll send a PR that way to capture the plugin error like we do here.

PR #1841, opened 2025-01-27
(fix): allow init without `X` and correct shape inferred

### LINKED ISSUE #1816: Initializing `AnnData` without `X` but with `layers` errors

### Report

See https://github.com/COMBINE-lab/alevin-fry/issues/151#issuecomment-2579079579 and below for discussion.  I think this problem could in theory be responsible, but also it might be unlikely.  TBD

Code:

```python
import anndata as ad
import numpy as np

ad.AnnData(X=None, layers={"foo": np.arange(4).reshape((2, 2)), "bar": np.arange(4).reshape((2, 2))}) # errors, probably erroneously
```
vs.
```python
import anndata as ad
import numpy as np

adata = ad.AnnData(X=np.arange(4).reshape((2, 2)), layers={"foo": np.arange(4).reshape((2, 2)), "bar": np.arange(4).reshape((2, 2))})
del adata.X
adata.write_h5ad("foo.h5ad")
ad.read_h5ad("foo.h5ad").X == None
```

So I think it's something in `__init__`.

Traceback:

```pytb
Cell In[4], line 1
----> 1 ad.AnnData(X=None, layers={"foo": np.arange(4).reshape((2, 2)), "bar": np.arange(4).reshape((2, 2))})

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/anndata.py:252, in AnnData.__init__(self, X, obs, var, uns, obsm, varm, layers, raw, dtype, shape, filename, filemode, asview, obsp, varp, oidx, vidx)
    250     self._init_as_view(X, oidx, vidx)
    251 else:
--> 252     self._init_as_actual(
    253         X=X,
    254         obs=obs,
    255         var=var,
    256         uns=uns,
    257         obsm=obsm,
    258         varm=varm,
    259         raw=raw,
    260         layers=layers,
    261         dtype=dtype,
    262         shape=shape,
    263         obsp=obsp,
    264         varp=varp,
    265         filename=filename,
    266         filemode=filemode,
    267     )

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/anndata.py:471, in AnnData._init_as_actual(self, X, obs, var, uns, obsm, varm, varp, obsp, raw, layers, dtype, shape, filename, filemode)
    468 self._clean_up_old_format(uns)
    470 # layers
--> 471 self.layers = layers

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py:435, in AlignedMappingProperty.__set__(self, obj, value)
    431 def __set__(
    432     self, obj: AnnData, value: Mapping[str, Value] | Iterable[tuple[str, Value]]
    433 ) -> None:
    434     value = convert_to_dict(value)
--> 435     _ = self.construct(obj, store=value)  # Validate
    436     if obj.is_view:
    437         obj._init_as_actual(obj.copy())

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py:407, in AlignedMappingProperty.construct(self, obj, store)
    405 def construct(self, obj: AnnData, *, store: MutableMapping[str, Value]) -> T:
    406     if self.axis is None:
--> 407         return self.cls(obj, store=store)
    408     return self.cls(obj, axis=self.axis, store=store)

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py:210, in AlignedActual.__init__(self, parent, store)
    208 self._data = store
    209 for k, v in self._data.items():
--> 210     self._data[k] = self._validate_value(v, k)

File ~/Projects/Theis/scanpy/venv/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py:97, in AlignedMappingBase._validate_value(self, val, key)
     91         dims = tuple(("obs", "var")[ax] for ax in self.axes)
     92         msg = (
     93             f"Value passed for key {key!r} is of incorrect shape. "
     94             f"Values of {self.attrname} must match dimensions {dims} of parent. "
     95             f"Value had shape {actual_shape} while it should have had {right_shape}."
     96         )
---> 97     raise ValueError(msg)
     99 name = f"{self.attrname.title().rstrip('s')} {key!r}"
    100 return coerce_array(val, name=name, allow_df=self._allow_df)

ValueError: Value passed for key 'foo' is of incorrect shape. Values of layers must match dimensions ('obs', 'var') of parent. Value had shape (2, 2) while it should have had (0, 0).
```

### Versions

```python
| Package | Version |
| ------- | ------- |
| numpy   | 2.1.3   |

| Dependency                    | Version     |
| ----------------------------- | ----------- |
| parso                         | 0.8.4       |
| more-itertools                | 10.3.0      |
| jaraco.collections            | 5.1.0       |
| packaging                     | 24.2        |
| sphinxcontrib-devhelp         | 2.0.0       |
| jaraco.functools              | 4.0.1       |
| psutil                        | 6.1.0       |
| prompt_toolkit                | 3.0.48      |
| python-dateutil               | 2.9.0.post0 |
| jaraco.context                | 5.3.0       |
| decorator                     | 5.1.1       |
| stack-data                    | 0.6.3       |
| charset-normalizer            | 3.4.1       |
| rich                          | 13.9.4      |
| awkward                       | 2.7.2       |
| numcodecs                     | 0.14.1      |
| traitlets                     | 5.14.3      |
| sparse                        | 0.16.0a9    |
| pyarrow                       | 18.1.0      |
| wcwidth                       | 0.2.13      |
| sphinxcontrib-htmlhelp        | 2.1.0       |
| jaraco.text                   | 3.12.1      |
| numba                         | 0.61.0rc1   |
| sphinxcontrib-serializinghtml | 2.0.0       |
| six                           | 1.17.0      |
| legacy-api-wrap               | 1.4.1       |
| Jinja2                        | 3.1.4       |
| fsspec                        | 2024.10.0   |
| attrs                         | 24.2.0      |
| sphinxcontrib-qthelp          | 2.0.0       |
| Pygments                      | 2.18.0      |
| llvmlite                      | 0.44.0rc1   |
| zarr                          | 2.18.3      |
| zappy                         | 0.2.0       |
| dask                          | 2024.7.1    |
| pandas                        | 2.2.3       |
| toolz                         | 1.0.0       |
| pytz                          | 2024.2      |
| msgpack                       | 1.1.0       |
| sphinxcontrib-jsmath          | 1.0.1       |
| cloudpickle                   | 3.1.0       |
| tblib                         | 3.0.0       |
| asttokens                     | 3.0.0       |
| MarkupSafe                    | 3.0.2       |
| asciitree                     | 0.3.3       |
| sphinxcontrib-bibtex          | 2.6.3       |
| ipython                       | 8.31.0      |
| setuptools-scm                | 8.1.0       |
| PyYAML                        | 6.0.2       |
| scipy                         | 1.14.1      |
| coverage                      | 7.6.8       |
| session-info2                 | 0.1.2       |
| sphinxcontrib-applehelp       | 2.0.0       |
| natsort                       | 8.4.0       |
| h5py                          | 3.12.1      |
| pure_eval                     | 0.2.3       |
| jedi                          | 0.19.2      |
| awkward_cpp                   | 43          |
| executing                     | 2.1.0       |
| setuptools                    | 75.6.0      |

| Component | Info                                                                                |
| --------- | ----------------------------------------------------------------------------------- |
| Python    | 3.12.3 | packaged by Anaconda, Inc. | (main, May  6 2024, 14:46:42) [Clang 14.0.6 ] |
| OS        | macOS-15.1-arm64-arm-64bit                                                          |
| Updated   | 2025-01-09 15:41                                                                    |
```

ilan-gold:
- [x] Closes #1816
- [x] Tests added

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -90,14 +90,17 @@ def _infer_shape_for_axis(
     xxxm: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
     layers: Mapping[str, np.ndarray | sparse.spmatrix] | None = None,
     xxxp: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
-):
+) -> int | None:
     for elem in [xxx, xxxm, xxxp]:
         if elem is not None and hasattr(elem, "shape"):
             return elem.shape[0]
     for elem in [layers, xxxm, xxxp]:
         if elem is not None:
             elem = cast(Mapping, elem)
-            return next(iter(elem.values())).shape[0]
+            next_elem = next(iter(elem.values()))
+            if hasattr(next_elem, "shape"):
+                size = cast(int, next_elem.shape[0])
+                return size
     return None
 
 
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -58,68 +58,6 @@ if TYPE_CHECKING:
     from .index import Index
 
 
-# for backwards compat
-def _find_corresponding_multicol_key(key, keys_multicol):
-    """Find the corresponding multicolumn key."""
-    for mk in keys_multicol:
-        if key.startswith(mk) and "of" in key:
-            return mk
-    return None
-
-
-# for backwards compat
-def _gen_keys_from_multicol_key(key_multicol, n_keys):
-    """Generates single-column keys from multicolumn key."""
-    keys = [f"{key_multicol}{i + 1:03}of{n_keys:03}" for i in range(n_keys)]
-    return keys
-
-
-def _check_2d_shape(X):
-    """\
-    Check shape of array or sparse matrix.
-
-    Assure that X is always 2D: Unlike numpy we always deal with 2D arrays.
-    """
-    if X.dtype.names is None and len(X.shape) != 2:
-        msg = f"X needs to be 2-dimensional, not {len(X.shape)}-dimensional."
-        raise ValueError(msg)
-
-
-def _infer_shape_for_axis(
-    xxx: pd.DataFrame | Mapping[str, Iterable[Any]] | None,
-    xxxm: np.ndarray | Mapping[str, Sequence[Any]] | None,
-    layers: Mapping[str, np.ndarray | sparse.spmatrix] | None,
-    xxxp: np.ndarray | Mapping[str, Sequence[Any]] | None,
-    axis: Literal[0, 1],
-) -> int | None:
-    for elem in [xxx, xxxm, xxxp]:
-        if elem is not None and hasattr(elem, "shape"):
-            return elem.shape[0]
-    for elem, id in zip([layers, xxxm, xxxp], ["layers", "xxxm", "xxxp"]):
-        if elem is not None:
-            elem = cast(Mapping, elem)
-            for sub_elem in elem.values():
-                if hasattr(sub_elem, "shape"):
-                    size = cast(int, sub_elem.shape[axis if id == "layers" else 0])
-                    return size
-    return None
-
-
-def _infer_shape(
-    obs: pd.DataFrame | Mapping[str, Iterable[Any]] | None = None,
-    var: pd.DataFrame | Mapping[str, Iterable[Any]] | None = None,
-    obsm: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
-    varm: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
-    layers: Mapping[str, np.ndarray | sparse.spmatrix] | None = None,
-    obsp: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
-    varp: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
-):
-    return (
-        _infer_shape_for_axis(obs, obsm, layers, obsp, 0),
-        _infer_shape_for_axis(var, varm, layers, varp, 1),
-    )
-
-
 class AnnData(metaclass=utils.DeprecationMixinMeta):
     """\
     An annotated data matrix.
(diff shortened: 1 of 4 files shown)


src/anndata/_core/anndata.py
```diff
@@ -85,6 +85,40 @@ def _check_2d_shape(X):
         raise ValueError(msg)
 
 
+def _infer_shape_for_axis(
```
ilan-gold, line 88: 2 things about these functions.

1. Do they belong in `utils`? I see we have a bunch at the top of the file here so not sure
2. Should we do some sort of validation that all the shapes match? I think this happens anyway elsewhere in the codebase e.g., https://github.com/scverse/anndata/blob/main/src/anndata/_core/aligned_mapping.py#L70-L100

flying-sheep: 1. I think the logic for where things are is mostly “the deepest module level shared by all spots using it”, so something used all throughout would be in `src/anndata/_utils.py`, something only used by IO routines would be in `src/anndata/io/_utils.py` and something only used in one spot would be in the same file.
2. We allow inferring the shape from whatever is specified right? So figuring out the shape is no longer strictly before setting aligned mappings.

   Technically the optimal way to do things would then be to kinda coerce everything together, infer the shape, then at the end set everything. But I think it’s fine to do things a bit less elegantly if that saves work (the result is the same: An error if the shape can’t be unambiguously inferred)

src/anndata/_core/anndata.py
```diff
@@ -85,6 +85,40 @@ def _check_2d_shape(X):
...
+    var: pd.DataFrame | Mapping[str, Iterable[Any]] | None = None,
+    obsm: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
+    varm: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
+    layers: Mapping[str, np.ndarray | sparse.spmatrix] | None = None,
+    obsp: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
+    varp: np.ndarray | Mapping[str, Sequence[Any]] | None = None,
+):
+    return (
+        _infer_shape_for_axis(obs, obsm, layers, obsp),
+        _infer_shape_for_axis(var, varm, layers, varp),
```
flying-sheep, line 118: both `n_obs` and `n_var` are inferred from `layers.shape[0]`, that’s wrong!

flying-sheep (changes requested):
There’s a bug with `layers`. Please fix and add a test inferring from non-square `layers` to catch that.

PR #1874, opened 2025-02-25
(chore): convert `DeprecationWarnings` to `FutureWarnings` where possible

ilan-gold:
I'm not sure about `PendingDeprecationWarning` (which remain untouched - if you look at the codebase, there are some strange uses of them which I think is a separate issue) but otherwise this PR removes `DeprecationWarning`.  I'm marking as 0.12 so as not to break people's CI.  In that case I would probably put concrete timelines on this like "Next minor release i.e., 0.13" we will remove these.

- [ ] Closes #
- [x] Tests added


src/anndata/utils.py
```diff
@@ -359,14 +359,14 @@ def warn_once(msg: str, category: type[Warning], stacklevel: int = 1):
 
 def deprecated(
     new_name: str,
-    category: type[Warning] = DeprecationWarning,
+    category: type[Warning] = FutureWarning,
```
ilan-gold, line 362: Maybe we remove this altogether and always force `FutureWarning`?

flying-sheep: works for me!

ilan-gold:
> PendingDeprecationWarnings are even weaker than DeprecationWarnings: They are completely ignored by default. So I say we should upgrade these too.

Yes I guess my point is that this is basically misused.  One is a parent class for `OldFormatWarning` where we will not remove the ability to read old files, so there is no intention to deprecate.  The other place is `__setitem__` for `SparseDataset` for which I don't have enough context right now to upgrade IMO.  I would need to investigate why we are removing and what the implications are.  So I want to deal with this separately.

PR #1880, opened 2025-02-26
initial support for Dask DataFrames in obsm/varm

ilia-kats:
This PR adds support for Dask DataFrames in `.obsm`/`.varm`.

- [ ] Closes #
- [x] Tests added

### DIFF
--- a/docs/release-notes/1880.feature.md
+++ b/docs/release-notes/1880.feature.md
@@ -1,0 +1,1 @@
+Add support for Dask DataFrames in `.obsm` and `.varm` ({user}`ilia-kats`)
--- a/src/anndata/_core/aligned_mapping.py
+++ b/src/anndata/_core/aligned_mapping.py
@@ -264,7 +264,16 @@ class AxisArraysBase(AlignedMappingBase):
     def _validate_value(self, val: Value, key: str) -> Value:
         if isinstance(val, pd.DataFrame):
             raise_value_error_if_multiindex_columns(val, f"{self.attrname}[{key!r}]")
-            if not val.index.equals(self.dim_names):
+            if (
+                not val.index.equals(self.dim_names)
+                and (
+                    val.index.dtype != "string"
+                    and self.dim_names.dtype != "O"
+                    or val.index.dtype != "O"
+                    and self.dim_names.dtype != "string"
+                )
+                and (val.index != self.dim_names).any()
+            ):
                 # Could probably also re-order index if it’s contained
                 try:
                     pd.testing.assert_index_equal(val.index, self.dim_names)
--- a/src/anndata/_core/file_backing.py
+++ b/src/anndata/_core/file_backing.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 
 import h5py
 
-from ..compat import AwkArray, DaskArray, ZarrArray, ZarrGroup
+from ..compat import AwkArray, DaskArray, DaskDataFrame, ZarrArray, ZarrGroup
 from .sparse_dataset import BaseCompressedSparseDataset
 
 if TYPE_CHECKING:
@@ -143,7 +143,8 @@ def _(x: BaseCompressedSparseDataset, *, copy: bool = False):
 
 
 @to_memory.register(DaskArray)
-def _(x: DaskArray, *, copy: bool = False):
+@to_memory.register(DaskDataFrame)
+def _(x: DaskArray | DaskDataFrame, *, copy: bool = False):
     return x.compute()
 
 
--- a/src/anndata/_core/storage.py
+++ b/src/anndata/_core/storage.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 from scipy import sparse
 
-from anndata.compat import CSArray
+from anndata.compat import CSArray, DaskDataFrame
 
 from .._warnings import ImplicitModificationWarning
 from ..utils import (
@@ -51,7 +51,7 @@ def coerce_array(
     if any(is_non_csc_r_array_or_matrix):
         msg = f"Only CSR and CSC {'matrices' if isinstance(value, sparse.spmatrix) else 'arrays'} are supported."
         raise ValueError(msg)
-    if isinstance(value, pd.DataFrame):
+    if isinstance(value, pd.DataFrame | DaskDataFrame):
         if allow_df:
             raise_value_error_if_multiindex_columns(value, name)
         return value if allow_df else ensure_df_homogeneous(value, name)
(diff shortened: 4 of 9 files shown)


ilia-kats:
The minimum_versions test is failing due to [an incompatibility between the old dask version and Python 3.11.9 specifically](https://github.com/dask/dask/issues/11038), not sure what to do here.

ilan-gold:
@ilia-kats What is the use-case here that https://github.com/scverse/anndata/pull/1247 would not address?

Furthermore, we have reports from `SpatialData` that dask dataframes are rather difficult to use, as was our experience motivating the above PR to use `xarray` instead.

ilan-gold:
I like that the PR is basically write-only, but want to understand more.

ilan-gold:
> call compute on the shape

ilia-kats:
So I've looked at `read_lazy` and xarray, and I don't think that fits my usecase. I need to handle arbitrary, probably in-memory, AnnData objects, while conserving as much memory as possible. If I'm reading the xarray code correctly, `Dataset.from_dataframe` [copies all data passed to it](https://github.com/pydata/xarray/blob/dd446d7d9c5f208cedc18b4b02fcf380a5ba7217/xarray/core/dataset.py#L7190). Using Dask DataFrames directly would, as far as I understand, completely avoid copying, possibly at the cost of higher runtime due to having to call `compute`.

In the very simplest case, it looks like calling `compute` on the `shape` does not actually do any passes through the data, but simply adds the shapes of the individual data frames:

```python
datest = dd.from_pandas(test)
datest2 = dd.from_pandas(test2)

concat = dd.concat([datest, datest2], axis=0, ignore_index=True)
concat.shape[0].optimize().pprint()
Fused(25743):
| FloorDiv: right=1
|   Add:
      Literal: value=40
|     Add: left=0
        Literal: value=10
```

ilan-gold:
@ilia-kats I assumed you wanted the dask dataframes for its lazy-loading capabilities, in which case memory shouldn't be such an issue.  Perhaps we can talk offline: ilan.gold@helmholtz-munich.de - I am around all day today

ilan-gold:
> I'm reading the xarray code correctly, Dataset.from_dataframe [copies all data passed to it](https://github.com/pydata/xarray/blob/dd446d7d9c5f208cedc18b4b02fcf380a5ba7217/xarray/core/dataset.py#L7190)

But why keep the dataframes around once they are converted?

ilan-gold:
Also the code you posted is only applicable for non-extension arrays - pandas extension arrays should be 0-copy: https://github.com/pydata/xarray/blob/dd446d7d9c5f208cedc18b4b02fcf380a5ba7217/xarray/core/dataset.py#L7272-L7277 

This includes https://pandas.pydata.org/docs/reference/api/pandas.arrays.NumpyExtensionArray.html (i.e., numpy), https://pandas.pydata.org/docs/reference/api/pandas.arrays.ArrowExtensionArray.html (i.e., arrow), https://pandas.pydata.org/docs/reference/api/pandas.Categorical.html (categoricals), https://github.com/pandas-dev/pandas/blob/v2.2.3/pandas/core/arrays/string_.py#L275-L657 (string arrays), and likely others

ilan-gold:
```python
import xarray as xr
import pandas as pd

 iris = pd.read_csv('https://raw.githubusercontent.com/mwaskom/seaborn-data/master/iris.csv')
ds = xr.Dataset.from_dataframe(iris)
assert ds["sepal_length"].data is iris["sepal_length"].array.to_numpy()
```

ilia-kats:
Thanks, I should have read the xarray code more carefully. I guess I'll go with that then.

ilan-gold:
@ilia-kats No problem, please reach out if you feel you have more needs.  Zulip is the best place for longer discussions :)

ilia-kats:
Ah, I think I'm supposed to use `anndata.experimental.backed._xarray.Dataset2D` as a wrapper. However, that is not part of the public API.

ilan-gold:
That makes sense @ilia-kats - if you would like to contribute a feature that allow for the conversion (i.e., setting on an `AnnData` object an element that is a `Dataset`), we would take that.  It shouldn't be that hard

ilia-kats:
Sure, I can give it a shot. However, I think that this PR (simply allowing Dask DataFrames) is less invasive than the conversion would be.

xinaesthete:
I notice that some of the doc strings still say things like "One-dimensional annotation of variables/ features (`pd.DataFrame`)" which was a little confusing to me when I noticed some of my CI for scripts expecting that had broken with this update.

ilan-gold:
@xinaesthete Could you explain a bit? This PR is super old and dask does not have an amazing maintenance status at the moment.  If someone wants to do https://github.com/scverse/anndata/issues/2043 then we could support them just fine.  

The TL;DR on that issue is that we use a limited subset of the APIs dataframes offer, and formalizing this as a runtime protocol would allow you to create wrappers around the dataframe of interest (polars, cudf etc.) without us having to implement something new for every dataframe implementation.

xinaesthete:
Not a big deal, and I think we'd been on a very old version of AnnData until recently, and we had some things that had been expecting `pd.DataFrame` explicitly for `var`/`obs` etc, and then fell over (at least `pyright` started complaining) where those didn't match up... which I'm sure we can deal with, but it looks like some documentation strings aren't quite in sync with that. Maybe just [this one](https://github.com/scverse/anndata/blob/3c90d3cc2aef2973f67a51e605442b6f411e80d7/src/anndata/_core/anndata.py#L846), maybe worth when someone has a moment checking if there are others. I'm fairly busy but may open a small PR if it helps.

ilan-gold:
> Maybe just [this one](https://github.com/scverse/anndata/blob/3c90d3cc2aef2973f67a51e605442b6f411e80d7/src/anndata/_core/anndata.py#L846), maybe worth when someone has a moment checking if there are others.

Yeah, this could be clarified but I can also see why it is that way.  It matches e.g., `varm`:

```
Multi-dimensional annotation of variables/features
```

I would be open to changing these to read "key-value store of ... annotation of ....".  Would that make more sense?

ilan-gold:
> expecting pd.DataFrame explicitly for var/obs

Yes we have a [`Dataset2D`](https://anndata.readthedocs.io/en/stable/generated/anndata.experimental.backed.Dataset2D.html) class now that can be used for lazy io i.e., over an internet connection.  You can then use `to_memory` (on the `AnnData` or `Dataset2D`) to bring stuff e.g., only your columns of interest, into memory.

I would like to remove it though and put it in a separate package, and then allow general DF apis hence #2043

PR #1885, opened 2025-02-28
Fix link URL in docstring

sjfleming:
Just fixing a URL in docs

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,10 @@ repos:
     rev: v1.9.4
     hooks:
       - id: biome-format
+  - repo: https://github.com/ComPWA/taplo-pre-commit
+    rev: v0.9.3
+    hooks:
+      - id: taplo-format
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.7
+    rev: v0.9.9
     hooks:
       - id: ruff
         args: ["--fix"]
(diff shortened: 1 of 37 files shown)


flying-sheep:
Thanks! Unfortunately this is a bit more involved, as we need to add a label to that header, and the tutorials live in a different repository.

If we just wanted to link to the document, we’d do ``:doc:`here </tutorials/notebooks/getting-started>` ``.

So we need to make a PR into the tutorial repo to add a label like

```markdown
(read-partial)=

## Partial reading of large data
```

and then just do this here: ``For an example, see :ref:`read-partial` ``, which would render to

> For an example, see [Partial reading of large data](#)

flying-sheep (approved):

sjfleming:
Thanks @flying-sheep  !

mschilli87:
Just came here looking for where to fix what I think is the same mistake I encountered in [the scanpy docs](https://scanpy.readthedocs.io/en/stable/generated/scanpy.read_h5ad.html).
Will this just be auto-updated someday or do I still need to do anything to make sure this gets propagated?

Anyway, thanks for everybody working on this fix. Wouldn't have expected it to be that 'involved' myself. 😉

flying-sheep:
> Will this just be auto-updated someday or do I still need to do anything to make sure this gets propagated?

You don’t have to do anything! Scanpy just re-exports the `read_*` functions from anndata and renders their docstrings, so docs of the newest anndata version at the time of a scanpy release are visible, which will be 0.11.4 with @sjfleming’s fix. Once I’m no longer sick I’ll do a new scanpy release to fix the IPython 9 incompatibility, then you’ll see it.

Btw @ilan-gold the 0.11.4 milestone needs to be closed and the issues transferred to a new one!

> Anyway, thanks for everybody working on this fix. Wouldn't have expected it to be that 'involved' myself. 😉

Having the notebooks in a different repo is just very kludgy. Very soon we’ll undo that – pip already uses git’s `--filter=blob:none` when installing packages from git URLs, and uv will have it. Once uv caught up, there’s no reason not to just have everything in a big repo.

PR #1887, opened 2025-03-05
(chore): disallow `Dataset2D` writing

ilan-gold:
I think forcing people to load the data into memory is a good way to (a) prevent them from writing data with lost data types (since `concat` cannot preserve extension array types) and (b) save me from stalling this release longer on account of having to create a proper lazy write scheme (the alternative just being that we call `to_memory` internally anyway).

- [ ] Closes #
- [x] Tests added

### DIFF
--- a/tests/lazy/conftest.py
+++ b/tests/lazy/conftest.py
@@ -9,7 +9,7 @@ import pytest
 from scipy import sparse
 
 import anndata as ad
-from anndata import AnnData, settings
+from anndata import AnnData
 from anndata._types import AnnDataElem
 from anndata.experimental import read_lazy
 from anndata.tests.helpers import (
@@ -26,12 +26,6 @@ if TYPE_CHECKING:
     from typing import Literal
 
 ANNDATA_ELEMS = typing.get_args(AnnDataElem)
-
-
-@pytest.fixture(autouse=True, scope="session")
-def write_nullable():
-    settings.allow_write_nullable_strings = True
-    return settings.allow_write_nullable_strings
 
 
 @pytest.fixture(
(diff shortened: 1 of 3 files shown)


src/anndata/_io/utils.py
```diff
@@ -312,3 +312,20 @@ def func_wrapper(
...
     return func_wrapper
+
+
+def no_write_dataset_2d(write):
+    def raise_error_if_dataset_2d_present(store, adata, *args, **kwargs):
+        from anndata.experimental.backed._compat import has_dataset_2d
+
+        if has_dataset_2d(adata):
+            msg = (
+                "Writing AnnData objects with a Dataset2D not supported. "
```
flying-sheep, line 323: should there be a “yet”?

src/anndata/experimental/backed/_xarray.py
```diff
@@ -128,10 +144,5 @@ def _remove_unused_categories_xr(
...
 @to_memory.register(Dataset2D)
-def to_memory(ds: Dataset2D, *, copy=False):
-    df = ds.to_dataframe()
-    index_key = ds.attrs.get("indexing_key", None)
-    if df.index.name != index_key and index_key is not None:
-        df = df.set_index(index_key)
-    df.index.name = None  # matches old AnnData object
-    return df
+def to_memory(ds: Dataset2D, *, copy=False) -> pd.DataFrame:
+    return ds.to_memory(copy=copy)
```
flying-sheep, line 148: could just be

```suggestion
to_memory.register(Dataset2D, Dataset2D.to_memory)
```

I think.

flying-sheep (approved):
Looks good! Maybe add a “yet” if you plan to add it in the future.

PR #1891, opened 2025-03-06
actually make CSRDataset and CSCDataset usable

flying-sheep:
oops, that should have been there from the start!

### DIFF
--- a/tests/test_backed_sparse.py
+++ b/tests/test_backed_sparse.py
@@ -334,9 +334,7 @@ def test_read_array(
         f = h5py.File(path, "a")
     ad.io.write_elem(f, "mtx", a)
     diskmtx = sparse_dataset(f["mtx"])
-    assert isinstance(
-        diskmtx, CSCDataset if sparse_format.format == "csc" else CSRDataset
-    )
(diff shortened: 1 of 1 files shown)


src/anndata/_core/sparse_dataset.py
```diff
@@ -639,6 +639,10 @@ class _CSCDataset(BaseCompressedSparseDataset, abc.CSCDataset):
     """Internal concrete version of :class:`anndata.abc.CSRDataset`."""
 
 
+abc.CSRDataset.register(_CSRDataset)
+abc.CSCDataset.register(_CSCDataset)
```
ilan-gold, line 643: This is a deep cut for me - don't these both already inherit from the ABC? I'd like to understand this one.  Does it have to do with the multiple inheritance?

```python
from anndata._core.sparse_dataset import _CSRDataset
from anndata.abc import CSRDataset
assert issubclass(_CSRDataset, CSRDataset)
```

flying-sheep: huh, you’re right.

I must have run into something else then, no idea why I thought that check wasn’t already right.

PR #1892, opened 2025-03-06
(feat): allow zarr v3 writing

ilan-gold:
TODO:

- [x] Look into specifics of v3 string writing: https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/Guidance.20for.20non-spec.20v3.20on-disk and a private conversation with Joe, and the linked PR all lead me to believe the v3 writing is kosher.  Consolidated metadata may change but we can always backfill that - I'm not particularly concerned.
---------
- [x] following up with #1817
- [x] Tests added


src/anndata/_settings.py
```diff
@@ -432,7 +432,7 @@ def validate_type(val: V) -> None:
 
 def validate_zarr_write_format(format: int):
     validate_int(format)
-    if format != 2:
+    if format not in [2, 3]:
```
flying-sheep, line 435: always use set literals when checking for membership in a set of literals (strings, ints, …)

tests/conftest.py
```diff
@@ -30,6 +30,26 @@ def backing_h5ad(tmp_path):
...
+            yield fmt
+
+
+@pytest.fixture
+def diskfmt2(diskfmt):
+    if diskfmt == "h5ad":
+        with ad.settings.override(zarr_write_format=2):
+            yield "zarr"
+    else:
+        yield "h5ad"
```
flying-sheep, line 50: So diskfmt2 is supposed to be “the other one”?

I thought we tested the cartesian product for a reason, i.e. `test_thing[h5ad,h5ad]`, `test_thing[h5ad,zarr]`, `test_thing[zarr,h5ad]`, `test_thing[zarr,zarr]` and not just `test_thing[h5ad,zarr]`, `test_thing[zarr,h5ad]`.

But maybe we test that stuff elsewhere and don’t actually need to test it here, just want to make sure.

tests/test_readwrite.py
```diff
@@ -368,42 +360,60 @@ def check_compressed(key, value):
...
+            compressor = Blosc(cname="zstd", clevel=3, shuffle=Blosc.BITSHUFFLE)
+        else:
+            from zarr.codecs import BloscCodec
+
+            compressor = BloscCodec(cname="zstd", clevel=3, shuffle="bitshuffle")
+        not_compressed = []
+
+        ad.io.write_zarr(pth, adata, compressor=compressor)
+
+        def check_compressed(value, key):
```
flying-sheep, line 380: use some early returns here, I can’t see the code all the way over there in China! 😉

flying-sheep (commented):
some preliminary comments, will look properly next time

tests/conftest.py
flying-sheep: yeah, if that’s the case, doing the same one twice doesn‘t test for anything new. OK!

tests/test_readwrite.py
```diff
@@ -368,42 +360,59 @@ def check_compressed(key, value):
     assert_equal(adata, expected)
 
 
-def test_zarr_compression(tmp_path):
-    from numcodecs import Blosc
+@pytest.mark.parametrize("zarr_write_format", [2, 3])
+def test_zarr_compression(tmp_path, zarr_write_format):
+    with ad.settings.override(zarr_write_format=zarr_write_format):
```
flying-sheep, line 365: Instead of adding an indentation level to the whole test, just set the setting.

We should have an autouse fixture resetting it after (and in the unlikely case that we do not, we need to add that!)

tests/test_readwrite.py
```diff
@@ -368,42 +360,59 @@ def check_compressed(key, value):
...
                     not_compressed.append(key)
+                return None
+            if not isinstance(read_compressor, BloscCodec) or (
+                any(
+                    getattr(read_compressor, attr) != getattr(compressor, attr)
+                    for attr in ["clevel", "cname", "shuffle", "blocksize"]
+                )
+                and compressor.typesize is None
+                and isinstance(read_compressor.typesize, int)
+            ):
```
flying-sheep, line 395: this looks complicated, what does this set of properties mean?

tests/test_structured_arrays.py
```diff
@@ -46,17 +45,27 @@ def test_io(tmp_path, diskfmt, diskfmt2):
...
+    def get_context(fmt: Literal["zarr", "h5ad"]):
+        return (
+            pytest.raises(
+                NotImplementedError, match=r"zarr v3 does not support structured dtypes"
+            )
+            if fmt == "zarr" and ad.settings.zarr_write_format == 3
+            else nullcontext()
+        )
+
+    with get_context(diskfmt):
```
flying-sheep, line 57:
```suggestion
    with (
        pytest.raises(NotImplementedError, match=r"zarr v3 does not support structured dtypes")
        if diskfmt == "zarr" and ad.settings.zarr_write_format == 3
        else nullcontext()
    ):
```

flying-sheep (approved):
Looks great! Apart from the check in the compression test that I don’t understand, there are just a few formatting things!

tests/test_readwrite.py
ilan-gold: Ok I will add a comment but basically the blosc codec is attached to a `typesize` now which corresponds to the `dtype` it uses for decoding.  But of course, this value can change based on what it is decoding.  So ever `read_compressor` that is attached to an `Array` will have a value that varies based on the `array` while the original compressor will not have a proper `typesize`.  I guess to "make things work" I can force the `typesize` to be 1 and this will be cleaner (see https://github.com/zarr-developers/numcodecs/pull/713 for a related bug and perhaps more info).

flying-sheep: yeah, when the purpose isn’t self-evident, a comment with some link is a good idea!

flying-sheep (approved):

PR #1624, opened 2024-08-28
Chunked writing of h5py.Dataset and zarr.Array

ivirshup:
- [x] Closes #1623
- [ ] Tests added
  - This might already be good on tests, but I should check
- [ ] Release note added (or unnecessary)
- [ ] Add benchmarks

This PR fixes #1623 by writing backed dense arrays in chunks.

Very open to feedback on the logic of how chunking pattern of writes is selected. Maybe we should prioritize the chunking of the destination array over the chunking of the source array?

cc: @ebezzi

Some proof it works:

```python
%load_ext memory_profiler

import h5py
from anndata.experimental import write_elem
import numpy as np

f = h5py.File("tmp.h5", "w")
X = np.ones((10_000, 10_000))

%memit write_elem(f, "X", X)
# peak memory: 945.67 MiB, increment: 0.56 MiB

%memit write_elem(f, "X2", f["X"])
peak memory: 1047.00 MiB, increment: 101.12 MiB

%memit write_elem(f, "X3", f["X"], dataset_kwargs={"compression":"gzip"})
peak memory: 1068.03 MiB, increment: 6.14 MiB
```

PR #1894, opened 2025-03-07
(fix): undo writing of arrays as elem in sparse

ilan-gold:
I accidentally introduced a "bug" (or maybe more of a file specification discrepancy) in https://github.com/scverse/anndata/pull/1624/commits/5938d86f0130265fa126ff3fddaac7e99143e41e which I would like to reverse.  The subarrays of a sparse matrix should not be written as "elems" in the sense of getting metadata, I think.  Certainly that was not the previous behavior so I think the behavior (modulo an update for zarr v3) should be restored

- [ ] Closes #
- [x] Tests added
- [x] Release note added (or **unnecessary**)

### DIFF
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -724,12 +724,16 @@ def write_sparse_compressed(
     for attr_name in ["data", "indices", "indptr"]:
         attr = getattr(value, attr_name)
         dtype = indptr_dtype if attr_name == "indptr" else attr.dtype
-        _writer.write_elem(
-            g,
-            attr_name,
-            attr,
-            dataset_kwargs={"dtype": dtype, **dataset_kwargs},
-        )
+        if isinstance(f, H5Group) or is_zarr_v2():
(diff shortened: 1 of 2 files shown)


flying-sheep:
Hmm, we have multiple elements that have recursive metadata: dicts in .uns, aligned mappings and so on. And what is a sparse matrix other than a container of three arrays with very specific, linked properties?

So I’m not sure we need to revert this.

ilan-gold:
> Why shouldn’t someone e.g. be able to use read_elem to read just one of them.

I guess it was more that [nothing in the specification](https://anndata.readthedocs.io/en/stable/fileformat-prose.html#sparse-arrays) mandates it, so I don't think we should all of a sudden start doing it.  It also somehow feels wrong to make `data` an elem since it has little meaning without its context where other elems are at least aware of one of their axis' sizes relative to the parent (with the exception of `uns` but that's the point of `uns`).  But it's a bit pedantic, yes.

flying-sheep (approved):

PR #1898, opened 2025-03-07
Format TOML with taplo

flying-sheep:
Huh, why is that not a thing? Should have been part of #1864


.taplo.toml
```diff
@@ -0,0 +1,5 @@
+[formatting]
```
ilan-gold, line 1: why put this `toml` under a `.` prefix?

ilan-gold (approved):

PR #1902, opened 2025-03-07
(feat): add `async` versions of `{write,read}_{elem,dispatched}`

### LINKED ISSUE #1897: `async` internally for `read_elem` and `write_elem`

After https://github.com/scverse/anndata/pull/1726 people will be able to have zarr return async compatible arrays. We can update all of our internal `io` to pool requests together using `asyncio.gather` and give us more throughput, especially for cloud-based analysis.

ilan-gold:
- [ ] Closes #1897
- [ ] Tests added


src/anndata/experimental/_dispatch_io.py
```diff
@@ -42,6 +43,32 @@ def read_dispatched(
     return reader.read_elem(elem)
 
 
+async def read_dispatched_async(
```
flying-sheep, line 46: Maybe this doesn’t need to be a separate function.

We could just check if the callback returns something awaitable, right?

ilan-gold: So we would have just 

```python
def read_dispatched (...) -> Coroutine | AnnData
````

and then it would be up to users to deal with it?  Or you mean that all of the "duplicated" code should be made to just handle coroutines and not (which I kind of like)

Koncopd:
I haven't checked the code, but i saw `asyncio.run` calls. Please note that `asyncio.run` doesn't work in jupyter.

ilan-gold:
@Koncopd yes I am aware.  This was just a first pass - we are going a different route with this feature, I think. Message me on zulip if you want more information, happy to chat.

Koncopd:
Yeah, i think the correct way here is to create an event loop in a separate thread and execute all async functions using this event loop, like `fsspec` does.

ilan-gold:
> Yeah, i think the correct way here is to create an event loop in a separate thread and execute all async functions using this event loop, like fsspec does.

Yes, the plan is actually to use `zarr's` own `sync` when needed (because we actually need to use _that_ event loop specifically) but I also am redoing this where we don't async-ify the whole code base, which I find clearer

PR #1921, opened 2025-03-17
Switch to (testing.)fast-array-utils

flying-sheep:
- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from contextlib import ExitStack
 from copy import deepcopy
 from operator import mul
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 import joblib
 import numpy as np
@@ -32,7 +32,7 @@ from anndata.tests.helpers import (
     slice_subset,
     subset_func,
 )
-from testing.fast_array_utils import Flags
+from testing.fast_array_utils import SUPPORTED_TYPES, Flags
 
 if TYPE_CHECKING:
     from types import EllipsisType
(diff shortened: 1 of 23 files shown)


tests/test_views.py
```diff
@@ -358,15 +357,16 @@ def test_not_set_subset_X_dask(matrix_type_no_gpu, subset_func):
...
+    assert not np.any(to_dense(adata.X != orig_X_val))
 
     assert init_hash == tokenize(adata)
     assert isinstance(subset.X, type(adata.X))
 
 
 @IGNORE_SPARSE_EFFICIENCY_WARNING
-def test_set_scalar_subset_X(matrix_type, subset_func):
-    adata = ad.AnnData(matrix_type(np.zeros((10, 10))))
+@pytest.mark.array_type(skip={Flags.Disk, *CUPY_SPARSE, *DASK_SPARRAY})
```
ilan-gold, line 367: I think this sort of thing will now run these tests on gpu whereas they did not before?

tests/test_views.py
```diff
@@ -339,8 +337,9 @@ def test_not_set_subset_X(matrix_type_base, subset_func):
 # TODO: Determine if this is the intended behavior,
 #       or just the behaviour we’ve had for a while
 @IGNORE_SPARSE_EFFICIENCY_WARNING
-def test_not_set_subset_X_dask(matrix_type_no_gpu, subset_func):
-    adata = ad.AnnData(matrix_type_no_gpu(asarray(sparse.random(20, 20))))
+@pytest.mark.array_type(skip=Flags.Gpu | Flags.Disk)
+def test_not_set_subset_X_dask(array_type: ArrayType, subset_func) -> None:
```
ilan-gold, line 341: In this case, this runs on sparse etc. as well, no?

ilan-gold (commented):
In general the amount of 
```
@pytest.mark.array_type(skip=Flags.Gpu | Flags.Disk)
```
is a lot (both this one specifically, and others like it).  It seems like there is probably some sort of default we could have, no?

PR #1906, opened 2025-03-10
(fix): Accelerate tests: migrate to Github Actions, remove slow param, parallelize

ilan-gold:
Our python 3.13 CI is regularly timing out so this is a first step.  I am investigating the culprint (concatenation on disk, if I am not mistaken)

- [x] Closes #1907
- [ ] Tests added
- [x] Release note added (or unnecessary)

PR #1924, opened 2025-03-18
Use same local Dask cluster in tests

flying-sheep:
Since #1906, tests often fail because multiple workers want to spin up a local cluster.

This could be fixed using `processes=False` clusters or by reusing a cluster.

TODO:

- [x] Tell people here once my account is activated: https://dask.discourse.group/t/using-pytest-with-localclusters/1053/2


flying-sheep:
OK, I have to merge this, GitHub behaves weird in other jobs and this might help.

Issue #1154, opened 2023-10-04
"How to" example of direct GPU IO


ivirshup:
### Please describe your wishes and possible alternatives to achieve the desired result.

Create an example of loading from disk into GPU memory using [`kvikio`](https://github.com/rapidsai/kvikio) and zarr.

Intron7:
I'm currently testing `kvikio` with `zarr` and we need to very careful with the compression. Some compressions break kvikio. We can only use `numcodecs.LZ4()` or no compression. I would add a compression argument in `write_zarr`

ivirshup:
I would also not that @Intron7 and I actually saw a performance decrease when using GPU with `LZ4` compression. It seems like `LZ4` decompression was actually quite slow at the moment, so we may want to wait on performance improving before creating this demo.

Having some trouble reproducing those at the moment though, due to some vscode issues.

PR #1927, opened 2025-03-20
feat: allow gpu io in `sparse_dataset` by removing `scipy` inheritance

### LINKED ISSUE #1456: Refactor `BaseCompressedSparseDataset` and subclasses to remove `backed_{csr,csc}_matrix`

Currently `BaseCompressedSparseDataset` goes through the following steps for backed data:

1. Get a backed class, which inherits from `{csr_matrix,_cs_matrix}`
2. Pass in indices through `__getitiem__` in the non-special-cased consecutive-boolean situation to the backed class - if we are in this situation, use the `subset_by_major_axis_mask` function
3. Clean up the indexers using the matrix backed class `spmatrix._validate_indices` and detect if one of the `_get_{Type}X{OtherType}` methods has been overriden by our backed class.  The method we override are fairly easy to detect because they follow the same pattern.
4. If we have not overridden the method for the given indexer in `backed_{csr,csc}_matrix`, the data is read into memory completely and `scipy` handles the indexing/loading from the given indices.  If not, `backed_{csr,csc}_matrix` uses its inheritance to dispatching to the `_get_{Type}X{OtherType}` methods which it has overridden. 
5. Return an actual `{csr,csc}_matrix` object.

For compatibility with `scipy>=1.13` we had to start doing step 3, detecting which methods we override.  Therefore, there's no real reason anymore to use the classes as an intermediary, and can instead just turn the overridden methods into their own functions.

Possible paint points are basically anything we are not aware that `scipy.sparse` is doing under the hood that might make things go smoother, especially around indexers, so this could be thorny.

ilan-gold:
Aside from the removed ability to write, all functionality should be ported over.

- [x] Closes #1456, towards #1154 and https://github.com/scverse/anndata/issues/659#issuecomment-1033548060
- [x] Tests added


src/anndata/_core/anndata.py
```diff
@@ -649,7 +649,8 @@ def X(self, value: XDataType | None):  # noqa: PLR0912
                 if self.is_view:
                     X = self.file["X"]
                     if isinstance(X, h5py.Group):
-                        X = sparse_dataset(X)
+                        msg = "Cannot write to views of sparse backed files"
+                        raise NotImplementedError(msg)
```
flying-sheep, line 653: Was this just broken in the first place or is this removing a feature?

I assume it’s a “broken feature” in that it might work in some cases but not generically?

tests/test_backed_hdf5.py
```diff
@@ -354,22 +354,9 @@ def test_backed_modification_sparse(
...
-        adata.X[2, 1] = 13
-
-    assert adata.isbacked
-
-    assert np.all(adata.X[0, :] == np.array([10, 0, 10]))
-    assert np.all(adata.X[1, :] == np.array([11, 0, 12]))
-    assert np.all(adata.X[2, :] == np.array([7, 0, 9]))
+    pat = r"Cannot write to views of sparse backed files"
+    with pytest.raises(NotImplementedError, match=pat):
+        adata[0, [0, 2]].X = 10
```
flying-sheep, line 359: when did the deprecation hit users?

src/anndata/_core/sparse_dataset.py
```diff
@@ -44,39 +46,101 @@
...
-    from ..compat import H5Array, Index, Index1D, Index1DNorm
-else:
-    from scipy.sparse import spmatrix as _cs_matrix
 
+type DenseType = np.ndarray | CupyArray
+type SparseMatrixType = CupyCSMatrix | CSMatrix | CSArray
 
-SCIPY_1_15 = Version(version("scipy")) >= Version("1.15rc0")
 
+def get_np_module() -> ModuleType:
```
flying-sheep, line 65: Isn’t this called `xp` in array api lingo?

src/anndata/_core/sparse_dataset.py
```diff
@@ -44,39 +46,101 @@
...
-    indptr: np.ndarray
+    data: ArrayT
+    indices: ArrayT
+    indptr: DenseType | ArrayT
+    format: Literal["csr", "csc"]
+    shape: tuple[int, int]
+
+    @property
+    def memory_format(self) -> type[SparseMatrixType]:
+        is_gpu = hasattr(zarr, "config") and "gpu" in zarr.config.get("ndbuffer")
```
flying-sheep: maybe unify this with `get_np_module`? It’s the same logic.

Also this can just be one expression by setting `getattr`’s 3rd argument (the default) to `{}`.

src/anndata/_core/sparse_dataset.py
```diff
@@ -98,276 +162,166 @@ def copy(self) -> CSMatrix:
...
                 ]
             return sparse_dataset(sparse_group).to_memory()
-        return super().copy()
+        msg = f"Unsupported array types {type(self.data)}"
+        raise ValueError(msg)
 
-    def _set_many(self, i: Iterable[int], j: Iterable[int], x):
-        """\
-        Sets value at each (i, j) to x
+    def _get_contiguous_compressed_slice(self, s: slice) -> CompressedVectors:
```
flying-sheep: that would be a `slice[int, int, None|Literal[1]]`, right?

src/anndata/_core/sparse_dataset.py
```diff
@@ -98,276 +162,166 @@ def copy(self) -> CSMatrix:
...
-            (use_sparray_in_io and issubclass(backed_class, CSArray))
-            or (not use_sparray_in_io and issubclass(backed_class, CSMatrix))
-        ):
-            return backed_class
-    msg = f"Format string {format} is not supported."
-    raise ValueError(msg)
+            row = key
+            col = slice(None)
+        major_index, minor_index = (row, col) if self.format == "csr" else (col, row)
+        return self._get(major_index, minor_index)
```
flying-sheep, line 239: So `__getitem__` is basically the only thing you copied from the former superclass?

What happens if not slicing with “something × full slice” before and after? Do the prior superclasses support more cases than we do now?

src/anndata/_core/sparse_dataset.py
```diff
@@ -549,10 +473,12 @@ def append(self, sparse_matrix: CSMatrix | CSArray) -> None:  # noqa: PLR0912, P
             sparse_matrix = sparse_matrix._to_backed()
 
         # Check input
-        if not ss.issparse(sparse_matrix):
+        if not (
+            ss.issparse(sparse_matrix) or isinstance(sparse_matrix, BackedSparseMatrix)
```
flying-sheep, line 471: needing this kind of change worries me, did you run scanpy tests with this PR?

flying-sheep (commented):
Great stuff, nice simplifications with the major/minor axis stuff. Some questions

src/anndata/_core/anndata.py
flying-sheep: we talked about this in person

tests/test_backed_hdf5.py
flying-sheep: a year ago or so

src/anndata/_core/sparse_dataset.py
ilan-gold: yes, but we're not using the array api because we are still tied to two libraries that don't use it (`scipy` and `cupy`)

src/anndata/_core/sparse_dataset.py
ilan-gold: Why? Previously the output of `sparse_matrix._to_backed()` inherited from `scipy`'s sparse classes so this check is doing what it did before.

src/anndata/_core/sparse_dataset.py
ilan-gold: If I understand you correctly, yes, that's the case. There should be the same number of cases supported as before.  Previously we had the three "{slice,int,array} x full slice" and we have them again

src/anndata/_core/sparse_dataset.py
flying-sheep: I’m saying that if there is any code e.g. in scanpy that used to do `isinstance(adata.X, CSBase)` and worked before, it will now use the other branch (which is probably assuming a dense array)

src/anndata/_core/sparse_dataset.py
flying-sheep: so `csr_matrix(...)[:, np.ndarray([0, 1])]` raises an error?

ilan-gold: Ah no, there is the default case as there was before of "if not supported by one of these, read the whole thing into memory." Apologies, that was in the old implementation as well. So we only overrode those three cases previously but didn't disable the other cases - in those cases the entire dataset would be read into memory and then `csr_matrix(...)[:, np.ndarray([0, 1])]` would work fine

src/anndata/_core/sparse_dataset.py
ilan-gold: I think you're misunderstanding the separation - `BackedSparseMatrix` is only accessed via `_to_backed` and is not public. `isinstance(adata.X, CSBase)` never passed.

src/anndata/_core/sparse_dataset.py
flying-sheep: OK, so do you think we should help users who relied on that to maybe `add_note` to the exception if one is raised, which tells them to slice differently or load things into memory manually?

ilan-gold: > OK, so do you think we should help users who relied on that to maybe add_note to the exception if one is raised, which tells them to slice differently or load things into memory manually?

I think the point I'm trying to make is that we shouldn't raise an exception.  The default case is just "read into memory and then subset the result, which is a `scipy.sparse.cs{r,c}_{array,matrix}` in which case `scipy` raises the exception

flying-sheep: Ah I misunderstood then. I thought we lost that default behavior in this PR, but if not all’s good!

src/anndata/_core/sparse_dataset.py
```diff
@@ -44,39 +46,104 @@
...
+    @property
+    def memory_format(self) -> type[SparseMatrixType]:
+        # zarr v2 has no config
+        if self.format == "csr":
+            if is_gpu():
+                return CupyCSRMatrix
+            return ss.csr_array if settings.use_sparse_array_on_read else ss.csr_matrix
+        if is_gpu():
+            return CupyCSCMatrix
+        return ss.csc_array if settings.use_sparse_array_on_read else ss.csc_matrix
```
flying-sheep:
```suggestion
        if is_gpu():
            return CupyCSRMatrix if self.format == "csr" else CupyCSCMatrix
        if settings.use_sparse_array_on_read:
            return ss.csr_array if self.format == "csr" else ss.csc_array
        return ss.csr_matrix if self.format == "csr" else ss.csc_matrix
```

flying-sheep (approved):

PR #1929, opened 2025-03-21
(fix): manage dask cluster from controller

flying-sheep:
I tested this locally with and without `-n auto`.

Due to my 16 cores it failed pretty reliably in parallel mode and now doesn’t.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,7 @@ test = [
     "dask[distributed]",
     "awkward>=2.3",
     "pyarrow",
-    "anndata[dask,lazy]",
+    "anndata[dask]",
 ]
 dev-test = [ "pytest-xdist[psutil]" ] # local test speedups
 gpu = [ "cupy" ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -102,7 +101,7 @@ test = [
     "dask[distributed]",
     "awkward>=2.3",
     "pyarrow",
-    "anndata[dask]",
+    "anndata[dask,lazy]",
```
ilan-gold, line 104: if we do this we don't need `test-full`

flying-sheep: Does this make the tests noticably slower? Otherwise why have `tests-full`?

ilan-gold: I mean, yes, there are more tests for sure (the entire `lazy` folder).  I separated it to ensure that we don't test our min dependencies against whatever `xarray` forces us to install.

flying-sheep: Well, I now changed it so local matches CI.

It was more reliably reproducible with the min-deps job, but it does affect the stable job too!

ilan-gold (approved):

PR #1930, opened 2025-03-21
Revert "(fix): manage dask cluster from controller"

ilan-gold:
Reverts scverse/anndata#1929

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -29,9 +29,6 @@ overrides.matrix.deps.python = [
     { if = [ "min" ], value = "3.11" },
     { if = [ "stable", "pre" ], value = "3.13" },
 ]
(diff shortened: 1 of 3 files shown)


hatch.toml
```diff
@@ -29,9 +29,6 @@ overrides.matrix.deps.python = [
     { if = [ "min" ], value = "3.11" },
     { if = [ "stable", "pre" ], value = "3.13" },
 ]
-overrides.matrix.deps.features = [
-    { if = [ "stable", "pre" ], value = "test-full" },
-]
```
flying-sheep, line 34: why revert this?

ilan-gold: Simply because it was in the PR that was reverted.  I didn't think to check that it was its own "feature"

PR #1948, opened 2025-03-31
(feat): `zarr` v3 guide

ilan-gold:
- [ ] Closes #
- [ ] Tests added


docs/index.md
```diff
@@ -20,6 +20,7 @@ benchmarks
 contributing
 release-notes/index
 references
+zarr-v3
```
flying-sheep, line 23: Are you sure that top-level is the right place for that?

If you want to advertise it more prominently, there are other ways.

docs/zarr-v3.md
```diff
@@ -0,0 +1,74 @@
...
+
+We anticipate enabling officially supporting this functionality officially for dense data, sparse data, and possibly awkward arrays in the next minor release, 0.13.
+
+## Asynchronous i/o
+
+At the moment, `anndata` exports no `async` functions.  However, `zarr-python` has a fully `async` API and provides its own event-loop so that users like `anndata` can interact with a synchronous API while still beenfitting from `zarr-python`'s asynchronous functionality under that API.  We anticipate providing `async` versions of {func}`anndata.io.read_elem` and {func}`anndata.experimental.read_dispatched` so that users can download data asynchronously without using the `zarr-python` event loop.  We also would like to create an asynchronous partial reader to enable iterative streaming of a dataset.
+
+
+[v3 format]: https://anndata.readthedocs.io/en/stable/generated/anndata.settings.html#anndata.settings.zarr_write_format
```
flying-sheep, line 89: No, never link to anything with a sphinx inventory using regular links. This includes everything on readthedocs, and most other Python projects that have a docs page.

We want to automatically have links update when things move around and to be notified when links break, which is only possible when you use intersphinx for linking.

You can use this to find links: https://sphobjinv.readthedocs.io/en/latest/cli/suggest.html

src/anndata/_io/specs/methods.py
```diff
@@ -632,7 +632,7 @@ def write_vlen_string_array_zarr(
         filters, dtype = (
             ([VLenUTF8()], object)
             if ad.settings.zarr_write_format == 2
-            else (None, str)
+            else (None, np.dtypes.StringDType())
```
flying-sheep, line 635: so a compatible numpy version is guaranteed when zarr v3 is used?

or are we already relying on a compatible numpy version.

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+`anndata` now uses the much improved {mod}`zarr` v3 package and also allows writing of datasets in the [v3 format], with the exception of structured arrays.
+Users should notice a significant performance improvement, especially for cloud data, but also likely for local data as well.
+Here is a quick guide on some of our learnings so far:
+
+## Remote data
+
+We now provide the {func}`anndata.experimental.read_lazy` feature for reading as much of the {class}`~anndata.AnnData` object as lazily as possible, using `dask` and {mod}`xarray`.
+Please note that this feature is experimental and subject to change.
+To enable this functionality in a performant and feature-complete way for remote data sources, we use [consolidated metadata] on the `zarr` store (written by default).
+Please note that this introduces consistency issues - if you update the structure of the underlying `zarr` store i.e., remove a column from `obs`, the consolidated metadata will no longer be valid.
```
flying-sheep, line 12: Please add some guidance here: how *can* one remove a column from `obs` after this? Also which operations update the structure? Can you e.g. add a column? I know that this is about remote data, which can currently not be written over https (right?) but this is still probably good to mention.

Also please use actual dashes (<kbd>Alt/option</kbd>+<kbd>-</kbd>) for dashes, not minus/hyphens.

```suggestion
Please note that this introduces consistency issues – if you update the structure of the underlying `zarr` store i.e., remove a column from `obs`, the consolidated metadata will no longer be valid.
```

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+
+There is currently a bug with `numcodecs` that prevents data written from other non-numcodecs `zstd` implementations from being read in by the default zarr pipeline (to which the above rust pipeline falls back if it cannot handle a datatype or indexing scheme, like `vlen-string`): https://github.com/zarr-developers/numcodecs/issues/424.
+Thus is may be advisable to use `BloscCodec` with `zarr` v3 file format data if you wish to use the rust-accelerated pipeline until this issue is resolved.
+
+The same issue with `zstd` applies to data that may eventually be written by the GPU `zstd` implementation (see below).
+
+## GPU i/o
+
+At the moment, it is unlikely your `anndata` i/o will work if you use [`zarr.enable_gpu`].
+It's *possible* dense data i/o i.e., using {func}`anndata.io.read_elem` will work as expected, but this functionality is untested - sparse data, awkward arrays, and dataframes will not.
```
flying-sheep, line 76:
```suggestion
It's *possible* dense data i/o i.e., using {func}`anndata.io.read_elem` will work as expected, but this functionality is untested – sparse data, awkward arrays, and dataframes will not.
```

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+Please note that this feature is experimental and subject to change.
+To enable this functionality in a performant and feature-complete way for remote data sources, we use [consolidated metadata] on the `zarr` store (written by default).
+Please note that this introduces consistency issues - if you update the structure of the underlying `zarr` store i.e., remove a column from `obs`, the consolidated metadata will no longer be valid.
+Further, note that without consolidated metadata, we cannot guarantee your stored `AnnData` object will be fully readable.
+And even if it is fully readable, it will almost certainly be much slower to read.
+
+There are two ways of opening remote [`zarr` stores] from the `zarr-python` package, `fsspec` and `obstore`, and both can be used with `read_lazy`.
+[`obstore` claims] to be more performant out-of-the-box, but notes that this claim has not been benchmarked with the `uvloop` event loop, which itself claims to be 2X more performant than the default event loop for `python`.
```
flying-sheep, line 17: This paragraph is a a bit nested, maybe more clearly itemize the options: Are they `fsspec`, `fsspec`+`uvloop`, `obstore`, or … ?

Also:

```suggestion
[`obstore` claims] to be more performant out-of-the-box, but notes that this claim has not been benchmarked with the `uvloop` event loop, which itself claims to be 2× more performant than the default event loop for `python`.
```

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Local data
+
+Local data generally poses a different set of challenges.
+First, write speeds can be somewhat slow and second, the creation of many small files on a file system can slow down a filesystem.
+For the "many small files" problem, `zarr` has introduced [sharding] in the v3 file format.
+Sharding requires knowledge of the array element you are writing, though, and therefore you will need to use {func}`anndata.experimental.write_dispatched` to use sharding.
```
flying-sheep, line 24: > Sharding requires knowledge of the array element you are writing, though, and therefore you will need to use {func}`anndata.experimental.write_dispatched` to use sharding.

why? please elaborate a tiny bit on how that knowledge comes in.

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+```python
+import anndata as ad
+from collections.abc import Mapping
+from typing import Any
+import zarr
+
+ad.settings.zarr_write_format = 3 # Absolutely crucial! Sharding is only for the v3 file format!
+
+def write_sharded(group: zarr.Group, adata: ad.AnnData):
+    def callback(func: ad.experimental.Write, g: zarr.Group, k: str, elem: ad.typing.RWAble, dataset_kwargs: Mapping[str, Any], iospec: ad.experimental.IOSpec):
```
flying-sheep, line 36: this is very wide and pokes out of the code box in the rendered doc.

Please reformat this (e.g. by pasting into https://play.ruff.rs/) so it looks good on the [rendered page](https://icb-anndata--1948.com.readthedocs.build/en/1948/zarr-v3.html). 

For later: maybe we should try out https://github.com/Freed-Wu/mdformat-ruff? I was against including it in the scverse template since mdformat doesn’t really support MyST, but we should be alert enough to spot when it breaks things :hand_with_index_finger_and_thumb_crossed:

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+            dataset_kwargs = { "shards": tuple(int(2 ** (16 / len(elem.shape))) for _ in elem.shape), **dataset_kwargs}
+            dataset_kwargs["chunks"] = tuple(i // 2 for i in dataset_kwargs["shards"])
+        elif iospec.encoding_type in { "csr_matrix", "csc_matrix" }:
+            dataset_kwargs = { "shards": (2**16,), "chunks": (2**8, ), **dataset_kwargs }
+        func(g, k, elem, dataset_kwargs=dataset_kwargs)
+    return ad.experimental.write_dispatched(group, "/", adata, callback=callback)
+```
+
+However, `zarr-python` can be slow with sharding throughput as well as writing throughput.
+Thus if you wish to speed up either writing, sharding, or both (or receive a modest speed-boost for reading), a bridge to the `zarr` implementation in Rust: https://zarrs-python.readthedocs.io/en/latest/ can help with that (see https://github.com/LDeakin/zarr_benchmarks for benchmarks):
```
flying-sheep, line 47: oof, please absolutely no blank links.

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+However, this pipeline is not compatible with all types of zarr store, especially remote stores and there are limitations on where rust can give a performance boost for indexing.
+We therefore recommend this pipeline for writing full datasets and reading contiguous regions of said written data.
+
+## Codecs
+
+The default `zarr-python` v3 codec for the [v3 format] is no longer `blosc` but `zstd`.
+While `zstd` is more widespread, you may find its performance to not meet your old expectations.
+Therefore, we recommend passing in the [`BloscCodec`] to `compressor` on {func}`~anndata.AnnData.write_zarr` if you wish to return to the old behavior.
+
+There is currently a bug with `numcodecs` that prevents data written from other non-numcodecs `zstd` implementations from being read in by the default zarr pipeline (to which the above rust pipeline falls back if it cannot handle a datatype or indexing scheme, like `vlen-string`): https://github.com/zarr-developers/numcodecs/issues/424.
```
flying-sheep, line 68: again: no blank links. We have a role for github issues, please use that.

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+However, this pipeline is not compatible with all types of zarr store, especially remote stores and there are limitations on where rust can give a performance boost for indexing.
+We therefore recommend this pipeline for writing full datasets and reading contiguous regions of said written data.
+
+## Codecs
+
+The default `zarr-python` v3 codec for the [v3 format] is no longer `blosc` but `zstd`.
+While `zstd` is more widespread, you may find its performance to not meet your old expectations.
+Therefore, we recommend passing in the [`BloscCodec`] to `compressor` on {func}`~anndata.AnnData.write_zarr` if you wish to return to the old behavior.
```
flying-sheep, line 66: does this play nice with `ZarrsCodecPipeline`?

docs/zarr-v3.md
```diff
@@ -0,0 +1,96 @@
...
+
+At the moment, it is unlikely your `anndata` i/o will work if you use [`zarr.enable_gpu`].
+It's *possible* dense data i/o i.e., using {func}`anndata.io.read_elem` will work as expected, but this functionality is untested - sparse data, awkward arrays, and dataframes will not.
+`kvikio` currently provides a [`GDS`-enabled store] although there are no working compressors at the moment exported from the `zarr-python` package (work is underway for `Zstd`: https://github.com/zarr-developers/zarr-python/pull/2863).
+
+We anticipate enabling officially supporting this functionality officially for dense data, sparse data, and possibly awkward arrays in the next minor release, 0.13.
+
+## Asynchronous i/o
+
+At the moment, `anndata` exports no `async` functions.
```
flying-sheep, line 83: That’ll change soon with #1902, right? So maybe we should merge this first, then make sure this line is changed in that PR.

flying-sheep (changes requested):
Hi, the text is mostly great, nicely structured! I commented on the few cases where I got confused.

First of all it’s always nice when a PR is about 1-2 docs pages to provide links in your initial PR comment (“PR body”? I never know how to refer to that text box).

Something like

````markdown
### [Rendered](https://icb-anndata--1948.com.readthedocs.build/en/1948/zarr-v3.html)
````

Please start using one sentence per line in markup documents, that way I can actually comment on what I want to comment on instead of one giant paragraph. I did it myself this time so I could comment.

Please also use typography:
- En-dashes: “–” (very easy to type on your MacBook: <kbd>Alt/option</kbd>+<kbd>-</kbd>)
- Times symbol: “×” (avoids confusion between `AnnData.X` and factors like “2×”)

File: docs/index.md (hunk 1 above)
Comment (ilan-gold) [reply docs/index.md:23]: I don't know where else to put it, open to ideas!

File: docs/zarr-v3.md (hunk 2 above)
Comment (ilan-gold) [reply docs/zarr-v3.md:89]: What should we do here because this attribute doesn't exist yet? So my solution was to use the link here and then we could change it once the feature is on `stable`

Comment (ilan-gold) [reply docs/zarr-v3.md:89]: Maybe this should be `latest` not `stable` but the point still stands (there's no inventory link)

File: src/anndata/_io/specs/methods.py (hunk 3 above)
Comment (ilan-gold) [reply src/anndata/_io/specs/methods.py:635]: Great catch, right, they even have code for this and I have seen it before: https://github.com/zarr-developers/zarr-python/blob/54b3d44b36e1876eab93239bc0de43c3d70e1e92/src/zarr/core/strings.py#L15

File: docs/zarr-v3.md (hunk 2 above)
Comment (flying-sheep) [reply docs/zarr-v3.md:101]: I think I need to explain this to you in person, I can’t really respond to this without explaining how readthedocs works on a fundamental level.

File: docs/zarr-v3.md [hunk 13]
```diff
@@ -1,6 +1,6 @@
 # zarr-v3 Guide/Roadmap
 
-`anndata` now uses the much improved {mod}`zarr` v3 package and also allows writing of datasets in the [v3 format], with the exception of structured arrays.
+`anndata` now uses the much improved {mod}`zarr` v3 package and also allows writing of datasets in the v3 format via {attr}`anndata.settings.remove_unused_category`, with the exception of structured arrays.
```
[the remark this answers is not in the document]
ilan-gold: brain is not working today...

docs/index.md
flying-sheep: It’s a guide, so maybe into `tutorials/`?

ilan-gold: Makes sense, I don't mind it being under tutorials.  We'll tweet about it/advertise lots :)

src/anndata/_io/specs/methods.py
```diff
@@ -629,11 +629,16 @@ def write_vlen_string_array_zarr(
...
+        match (
+            ad.settings.zarr_write_format,
+            Version(np.__version__) >= Version("2.0.0"),
+        ):
+            case 2, _:
+                filters, dtype = [VLenUTF8()], object
+            case 3, True:
+                filters, dtype = None, np.dtypes.StringDType()
+            case 3, False:
+                filters, dtype = None, np.dtypes.ObjectDType()
```
ilan-gold, line 641: Not sure how to force coverage here aside from using a new test job altogether

ilan-gold:
@flying-sheep I update the notebook a bit to use the new public data url and also added a note about `obstore` among some other minor cleanups.

src/anndata/_io/specs/methods.py
ilan-gold: We could add a no-cover but that also feels wrong somehow, but maybe it's the best way forward.  It's not that the code is unreachable (which is usually what I've seen no-cover for) but rather that we don't test it

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,106 @@
...
+We anticipate providing `async` versions of {func}`anndata.io.read_elem` and {func}`anndata.experimental.read_dispatched` so that users can download data asynchronously without using the `zarr-python` event loop.
+We also would like to create an asynchronous partial reader to enable iterative streaming of a dataset.
+
+[consolidated metadata]: https://zarr.readthedocs.io/en/stable/user-guide/consolidated_metadata.html
+[`zarr` stores]: https://zarr.readthedocs.io/en/stable/api/zarr/storage/index.html
+[`obstore` claims]: https://developmentseed.org/obstore/latest/performance
+[sharding]: https://zarr.readthedocs.io/en/stable/user-guide/performance.html#sharding
+[`BloscCodec`]: https://zarr.readthedocs.io/en/stable/api/zarr/codecs/index.html#zarr.codecs.BloscCodec
+[`zarr.enable_gpu`]: https://zarr.readthedocs.io/en/stable/user-guide/gpu.html#reading-data-into-device-memory
+[`GDS`-enabled store]: https://docs.rapids.ai/api/kvikio/nightly/api/#kvikio.zarr.GDSStore
```
flying-sheep, line 106: Link all of these using roles, not markdown links, as

- https://docs.rapids.ai/api/kvikio/nightly/objects.inv exists
- https://zarr.readthedocs.io/en/stable/objects.inv exists
- https://developmentseed.org/obstore/latest/objects.inv exists

So please update the intersphinx config as appropriate, find all these objects using e.g. the `sphobjinv` tool, and link them without URLs.

```suggestion
```

src/anndata/_io/specs/methods.py
flying-sheep: and we can’t/shouldn’t/don’t test it because … ?

“it’s a bunch of work for a tiny benefit” is a totally acceptable answer.

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,106 @@
...
+            dataset_kwargs["chunks"] = tuple(i // 2 for i in dataset_kwargs["shards"])
+        elif iospec.encoding_type in {"csr_matrix", "csc_matrix"}:
+            dataset_kwargs = {"shards": (2**16,), "chunks": (2**8,), **dataset_kwargs}
+        func(g, k, elem, dataset_kwargs=dataset_kwargs)
+
+    return ad.experimental.write_dispatched(group, "/", adata, callback=callback)
+```
+
+However, `zarr-python` can be slow with sharding throughput as well as writing throughput.
+Thus if you wish to speed up either writing, sharding, or both (or receive a modest speed-boost for reading), a bridge to the `zarr` implementation in Rust: https://zarrs-python.readthedocs.io/en/latest/ can help with that (see https://github.com/LDeakin/zarr_benchmarks for benchmarks):
```
flying-sheep, line 59: these are also still blank URLs. The first one should be linked via intersphinx, the second one should probably be the only regular markdown link in the document:

```md
… [`LDeakin/zarr_benchmarks` on GitHub][zarr-benchmarks] …

[zarr-benchmarks]: https://github.com/LDeakin/zarr_benchmarks
```

flying-sheep (changes requested):

src/anndata/_io/specs/methods.py
ilan-gold: It's actually minimal work.  We would just need to special case a test job for this where we do "numpy minimum" + "zarr maximum"

ilan-gold: But do we want to add an entire CI job and/or infrastructure just for that? Not sure.

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,101 @@
...
+`kvikio` currently provides a {class}`kvikio.zarr.GDSStore` although there are no working compressors at the moment exported from the `zarr-python` package (work is underway for `Zstd`: {pr}`zarr-developers/zarr-python#2863`.
+
+We anticipate enabling officially supporting this functionality officially for dense data, sparse data, and possibly awkward arrays in the next minor release, 0.13.
+
+## Asynchronous i/o
+
+At the moment, `anndata` exports no `async` functions.
+However, `zarr-python` has a fully `async` API and provides its own event-loop so that users like `anndata` can interact with a synchronous API while still beenfitting from `zarr-python`'s asynchronous functionality under that API.
+We anticipate providing `async` versions of {func}`anndata.io.read_elem` and {func}`anndata.experimental.read_dispatched` so that users can download data asynchronously without using the `zarr-python` event loop.
+We also would like to create an asynchronous partial reader to enable iterative streaming of a dataset. {doc}`zarr:user-guide/consolidated_metadata`
```
flying-sheep, line 98:
```suggestion
We also would like to create an asynchronous partial reader to enable iterative streaming of a dataset.
```

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,101 @@
...
+
+`anndata` now uses the much improved {mod}`zarr` v3 package and also allows writing of datasets in the v3 format via {attr}`anndata.settings.zarr_write_format`, with the exception of structured arrays.
+Users should notice a significant performance improvement, especially for cloud data, but also likely for local data as well.
+Here is a quick guide on some of our learnings so far:
+
+## Remote data
+
+We now provide the {func}`anndata.experimental.read_lazy` feature for reading as much of the {class}`~anndata.AnnData` object as lazily as possible, using `dask` and {mod}`xarray`.
+Please note that this feature is experimental and subject to change.
+To enable this functionality in a performant and feature-complete way for remote data sources, we use {doc}`conslidated metadata <zarr:user-guide/consolidated_metadata>` on the `zarr` store (written by default).
```
flying-sheep, line 11:
```suggestion
To enable this functionality in a performant and feature-complete way for remote data sources, we use {doc}`zarr:user-guide/consolidated_metadata` on the `zarr` store (written by default).
```

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,101 @@
...
+And even if it is fully readable, it will almost certainly be much slower to read.
+
+There are two ways of opening remote `zarr` stores from the `zarr-python` package, {class}`zarr.storage.FsspecStore` and {class}`zarr.storage.ObjectStore`, and both can be used with `read_lazy`.
+[`obstore` claims] to be more performant out-of-the-box, but notes that this claim has not been benchmarked with the `uvloop` event loop, which itself claims to be 2× more performant than the default event loop for `python`.
+
+## Local data
+
+Local data generally poses a different set of challenges.
+First, write speeds can be somewhat slow and second, the creation of many small files on a file system can slow down a filesystem.
+For the "many small files" problem, `zarr` has introduced `{ref} sharding <zarr:user-guide/performance#sharding>` in the v3 file format.
```
flying-sheep, line 23: Why not link to the `user-guide-sharding` label? It’s a different “Sharding” headline, but I think it contains the same information.

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,101 @@
...
+Therefore, we recommend passing in the {class}`zarr.codecs.BloscCodec` to `compressor` on {func}`~anndata.AnnData.write_zarr` if you wish to return to the old behavior.
+
+There is currently a bug with `numcodecs` that prevents data written from other non-numcodecs `zstd` implementations from being read in by the default zarr pipeline (to which the above rust pipeline falls back if it cannot handle a datatype or indexing scheme, like `vlen-string`): {issue}`zarr-developers/numcodecs#424`.
+Thus is may be advisable to use `BloscCodec` with `zarr` v3 file format data if you wish to use the rust-accelerated pipeline until this issue is resolved.
+
+The same issue with `zstd` applies to data that may eventually be written by the GPU `zstd` implementation (see below).
+
+## GPU i/o
+
+At the moment, it is unlikely your `anndata` i/o will work if you use `zarr.enable_gpu <zarr:user-guide-gpu>`.
```
flying-sheep, line 87: Isn’t it `zarr.config.enable_gpu`? Maybe just link to the document? The header is “Using GPUs with Zarr”, which would fit into the text if you want.

flying-sheep (changes requested):

docs/tutorials/zarr-v3.md
ilan-gold: I did not see that in the `objects.inv` when I searched it, but maybe I missed it

ilan-gold: Yes, sorry, but not sure what you mean by "just link to the document"

docs/tutorials/zarr-v3.md
flying-sheep: > If you do, please don’t customize the link text, it already reads “Sharding”.

ilan-gold: > If you do, please don’t customize the link text, it already reads “Sharding”.

not a huge fan of the capitalization...

src/anndata/_io/specs/methods.py
flying-sheep: got it. OK, no test it is!

docs/tutorials/zarr-v3.md
```diff
@@ -0,0 +1,105 @@
...
+And even if it is fully readable, it will almost certainly be much slower to read.
+
+There are two ways of opening remote `zarr` stores from the `zarr-python` package, {class}`zarr.storage.FsspecStore` and {class}`zarr.storage.ObjectStore`, and both can be used with `read_lazy`.
+[`obstore` claims] to be more performant out-of-the-box, but notes that this claim has not been benchmarked with the `uvloop` event loop, which itself claims to be 2× more performant than the default event loop for `python`.
+
+## Local data
+
+Local data generally poses a different set of challenges.
+First, write speeds can be somewhat slow and second, the creation of many small files on a file system can slow down a filesystem.
+For the "many small files" problem, `zarr` has introduced `{ref} sharding <zarr:user-guide-sharding>` in the v3 file format.
```
flying-sheep, line 23: this is still broken

flying-sheep (approved):
looks good apart from the broken syntax, please fix that before merging

PR #1959, opened 2025-04-10
Load AnnLoader lazily to avoid torch imports when not requested

### LINKED ISSUE #1950: Import torch lazily

https://github.com/scverse/anndata/blob/ccfb6e32806626765725ea9d1cec6eeede3176e3/src/anndata/experimental/pytorch/_annloader.py#L19

Torch has the tendency to be difficult to import because it has the tendency to randomly break on complex cluster systems where CUDA versions are not properly maintained and updated.

I suggest that we only import torch when its functionality is actually requested.

Zethson:
- [x] Closes #1950 
- [ ] Tests added: No because we don't have have a matrix with torch. Happy to add something if desired.


docs/release-notes/0.12.0rc1.md
```diff
@@ -24,3 +24,4 @@
 - Adopt the Scientific Python [deprecation schedule](https://scientific-python.org/specs/spec-0000/) {user}`ilan-gold` ({pr}`1768`)
 - Allow {mod}`zarr` v3 writing of data {user}`ilan-gold` ({pr}`1892`)
 - {func}`anndata.register_anndata_namespace` functionality for adding custom functionality to an {class}`~anndata.AnnData` object {user}`srivarra` ({pr}`1870`)
+- Load AnnLoader lazily to prevent expensive unnecessary `torch` imports when its available on the system. {user}`Zethson` & {user}`flying-sheep` ({pr}`1959`)
```
flying-sheep, line 27: this (without the `-`) belongs into a file called `1959.performance.md` that you create using `hatch towncrier:create`

0.12.0rc1 is already released and its release notes are complete.

flying-sheep (changes requested):

docs/release-notes/0.12.0rc1.md
Zethson: Sorry, I don't know what I was thinking. Not even my first time using towncrier. Done.

ilan-gold (approved):

PR #1961, opened 2025-04-11
Add Pylint lints

flying-sheep:


src/anndata/_core/index.py
```diff
@@ -47,9 +47,7 @@ def _normalize_index(
     from ..experimental.backed._compat import DataArray
 
     # TODO: why is this here? All tests pass without it and it seems at the minimum not strict enough.
-    if not isinstance(index, pd.RangeIndex) and (
-        index.dtype == float or index.dtype == int
-    ):
+    if not isinstance(index, pd.RangeIndex) and index.dtype in (np.float64, np.int64):
```
flying-sheep, line 50: I really feel the TODO here. I’m quite sure my change here is correct: `some_dtype == int` (and therefore `some_dtype in (int,)`) only evaluates to `True` when `some_dtype` is `np.dtype(np.int64)` (same for `float` and `np.float64`)

So why is that check?

Issue #1241, opened 2023-11-23
Extend linter config


flying-sheep:
### Please describe your wishes and possible alternatives to achieve the desired result.

See #1117.

flying-sheep:
Should also use preview lints like [PLR6201](https://docs.astral.sh/ruff/rules/literal-membership/#literal-membership-plr6201)

PR #1963, opened 2025-04-11
(chore): Add SIM lints

flying-sheep:
See https://github.com/scverse/anndata/issues/1241


src/anndata/_io/specs/lazy_methods.py
```diff
@@ -332,7 +332,7 @@ def read_nullable(
     elem_name = get_elem_name(elem)
     return MaskedArray(
         values=elem["values"],
-        mask=elem["mask"] if "mask" in elem else None,
+        mask=elem.get("mask", None),
```
flying-sheep, line 335: That’s the only one I wasn’t sure about, but this one’s covered by tests

PR #1966, opened 2025-04-11
Allow xarray Datasets to be used for obs/var/obsm/varm

ilia-kats:
- [ ] Closes #
- [x] Tests added

As [discussed](https://github.com/scverse/anndata/pull/1880#issuecomment-2789626806). This PR also moves the `Dataset2D` class from experimental into core. IMHO this makes more sense, since it as already used all over core and is now used even more.

Note that this is not a full-featured replacement for DataFrames: Pandas Extension Arrays are lost durint concatenation, since they are not supported by Dask Arrays. This affects categoricals and nullables. I would thus still like to have Dask DataFrame support.


ilia-kats:
I'm not sure what to do about the documentation warnings. I think the problem is that it's building the documentation for `Dataset2D`, which has type hints from `compat`, but it's not building any docs for `compat`. One could either remove `Dataset2D` from `api.md` or change the type hints to `xarray.DataArray`, which would mean that one needs to have xarray installed for type checking and building the docs.

ilan-gold:
> Note that this is not a full-featured replacement for DataFrames: Pandas Extension Arrays are lost durint concatenation, since they are not supported by Dask Arrays. This affects categoricals and nullables. I would thus still like to have Dask DataFrame support.

@ilia-kats For your purposes, you could concatenate the in-memory arrays with no issue.  The dask arrays are meant specifically for the "backed" extension arrays which are custom xarray classes i.e., they cannot be concatenated.  But for your use-case, you can concatenate the the arrays in-memory.

Re: allowing `DaskDataframe`, I would rather pin down what subset of the pandas API we are relying on (I think it's relatively small), and create a runtime-checkable `Protocol` which we export to say to users (like yourself): "These requirements enable you to use a dataframe-like object in places where a pandas dataframe is currently used."  Going this route would allow us to (a) allow users to create a polars-compatible wrapper (b) allow users to use `cuDF` in some capacity (c) you to create a dask dataframe wrapper (d) allow myself to potentially remove all the xarray code from this package altogether and create a new one where I can iterate faster (e) understand exactly what it is we are doing internally with pandas.  

We would then make this part of the next release.  I am not 100% sure this route is feasible, but I have a feeling it should be.

ilan-gold:
> I take it we're not moving forward with this PR then?

Not at all! We definitely should.  I was just adding that you can concatenate in-memory categoricals and nullable-booleans without losing their data types.

ci/scripts/min-deps.py
```diff
@@ -34,7 +34,7 @@ def min_dep(req: Requirement) -> Requirement:
     -------
 
     >>> min_dep(Requirement("numpy>=1.0"))
-    <Requirement('numpy==1.0.*')>
+    <Requirement('numpy~=1.0.0')>
```
ilan-gold, line 37: What changed here?

src/anndata/_core/merge.py
```diff
@@ -718,6 +728,36 @@ def _apply_to_awkward(self, el: AwkArray, *, axis, fill_value=None):
                 el = ak.pad_none(el, 1, axis=axis)  # axis == 0
             return el[self.idx]
 
+    def _apply_to_dataset2d(self, el: Dataset2D, *, axis, fill_value=None):
+        if fill_value is None:
+            fill_value = np.nan
+        index_dim = el.index_dim
+        if axis == 0:
+            # Dataset.reindex() can't handle ExtensionArrays
```
ilan-gold, line 736: Ok I will investigate and potentially submit a fix to them

src/anndata/_core/merge.py
```diff
@@ -1147,50 +1191,57 @@ def make_dask_col_from_extension_dtype(
...
+        else:
+            dtype = col.dtype.numpy_dtype
+        return da.map_blocks(
+            get_chunk,
+            chunks=chunk_size,
+            meta=np.array([], dtype=dtype),
+            dtype=dtype,
+        )
+    else:  # in-memory
+        return da.from_array(col.values, chunks=-1)
```
ilan-gold, line 1244:
```suggestion
    return da.from_array(col.values, chunks=-1)
```

ilan-gold: Further, why create a dask array on top of the values? You want the returned value to still be lazy here?  I think this should be some sort of flag then on `concat` - with backed datasets, we have basically no choice until something better is figured out but for this case, this is a very strong choice.

src/anndata/_core/xarray.py
```diff
@@ -52,13 +53,25 @@ def index(self) -> pd.Index:
...
     def index(self, val) -> None:
-        coord = get_index_dim(self)
-        self.coords[coord] = val
+        index_dim = self.index_dim
+        self.coords[index_dim] = (index_dim, val)
+        if isinstance(val, pd.Index) and val.name is not None and val.name != index_dim:
+            self.update(self.rename({self.index_dim: val.name}))
+            del self.coords[index_dim]
+        if "indexing_key" in self.attrs:
+            del self.attrs["indexing_key"]
```
ilan-gold, line 66: The codecov here (and elsewhere) should be addressed

src/anndata/compat/__init__.py
```diff
@@ -100,6 +100,32 @@ def __repr__():
             return "mock dask.array.core.Array"
 
 
+if find_spec("xarray") or TYPE_CHECKING:
+    import xarray
+    from xarray import DataArray as XArray
```
ilan-gold, line 105:
```suggestion
    from xarray import DataArray as XDataArray
```
for consistency (and elsewhere) - i like the prefix a lot, great call

ilan-gold (changes requested):
Nice! Some small things but overall looks good.  Moving `Dataset` to `core` is a good move, agreed, but exactly the sort of reason I want to move towards a `Protocol` instead on the next release.  I don't want the list of return types on `obs` to grow every time a new dataframe library comes out.  And it would be great to split off the `xarray` stuff into its own package IMO!

src/anndata/_core/merge.py
ilia-kats: I added a new argument to `concat` that controls this.

src/anndata/_core/xarray.py
ilia-kats: I've added a test module for the `Dataset2D` class. The other major piece of new code that is not covered by tests is merge.py:780-786, but that is impossible to test currently as that codepath (`axis==1`) would only be hit when concatenating along columns, which is never the case for obs/var/obsm/varm. I can remove that code and raise an exception there instead, if you think that would be better.

ci/scripts/min-deps.py
ilia-kats: Xarray 2024.10.0 depends on Pandas >= 2.1.0, but Anndata requires Pandas >= 2.0.0. I did not want to raise the Pandas version for Anndata, since Xarray is only an optional dependency), and instead the newer version is only required for the tests. This did not work with the previous version of the min-deps script: That would list pandas twice, with `pandas==2.0.0.*` and `pandas==2.1.0.*`, and pip would complain about unresolvable dependencies and refuse to do anything.

The new version of `min-deps.py` lists each dependency only once, and takes the highest listed version as the minimum version. However, `packaging.version.Version` cannot compare versions with wildcards in them, so instead of using `==1.0.*`, I am now using `~=1.0.0`, which has the same behavior during dependency resolution, but has a version that can be compared.

ilan-gold: But @ilia-kats why would `xarray` influence `min-deps`? It is not a test dependency

ilan-gold: Other than that, seems reasonable

ilan-gold (commented):
I will look into the `RangeIndex` business.  Looking good though.  Sorry for the delay, was on vacation for most of last week.

ci/scripts/min-deps.py
ilia-kats: xarray is a test dependency now. It was previously only a test dependency for `test-full`, which is `anndata[test,lazy]`. However, since this PR ist not really related to the lazy reading feature, I thought it makes sense to always test it. But I can remove the xarray dependency here. I'm not sure if that will do anything for the min-deps, though, since xarray is still a test dependency for `test-full`.

src/anndata/_core/merge.py
ilia-kats: Still doesn't work, I'm getting

```python
>       f" array types {[x.dtype for x in scalars_or_arrays]}"
    )
E   AttributeError: 'float' object has no attribute 'dtype'

../../envs/famo/lib/python3.11/site-packages/xarray/core/duck_array_ops.py:265: AttributeError
```

with xarray 2025.4.0

src/anndata/experimental/backed/_io.py
```diff
@@ -141,7 +141,10 @@ def callback(func: Read, /, elem_name: str, elem: StorageType, *, iospec: IOSpec
             }
             or "nullable" in iospec.encoding_type
         ):
-            if iospec.encoding_type == "dataframe" and elem_name in {"/obs", "/var"}:
+            if iospec.encoding_type == "dataframe" and (
+                elem_name[:4] in {"/obs", "/var"}
+                or elem_name[:8] in {"/raw/obs", "/raw/var"}
```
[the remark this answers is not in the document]
ilia-kats: This is supposed to cover `/obs`, `/obsm`, `/var`, and `/varm` without me having to type it all out.

src/anndata/_core/xarray.py
```diff
@@ -0,0 +1,145 @@
...
+            def __getitem__(self, idx):
+                coord = self._ds.index_dim
+                return self._ds.isel(**{coord: idx})
+
+        return IlocGetter(self)
+
+    def __getitem__(self, idx) -> Dataset2D:
+        ret = super().__getitem__(idx)
+        if len(idx) == 0 and not isinstance(idx, tuple):  # empty XDataset
+            ret.coords[self.index_dim] = self.xr_index
```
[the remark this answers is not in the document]
ilia-kats: Yes, we do:

```python
import pandas as pd
import xarray as xr

df = pd.DataFrame({"a": [1,2,3], "b": [4,5,6]})
xd = xr.Dataset.from_dataframe(test)
xd[[]]
```
results in
```
<xarray.Dataset> Size: 0B
Dimensions:  ()
Data variables:
    *empty*
```

ci/scripts/min-deps.py
ilan-gold: Right, so our CI only uses `test` for min-deps which is why I created `test-full`.  Maybe I should have made the old `test` into `test-min` and the current `test-full` into `test` to clarify better what would happen.  I think that would make more sense.  Then we could update the CI to use `test-min` for the min-deps job.  Does that make sense? That seems a bit clearer to me, although, to be clear, I think the update you made here makes sense.

ilan-gold: See https://github.com/ilia-kats/anndata/pull/1

src/anndata/experimental/backed/_io.py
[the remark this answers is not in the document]
ilan-gold: I see, great catch yes.  I'll add a test for this to my PR.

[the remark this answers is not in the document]
ilan-gold: PR open in your fork for a test!

src/anndata/_core/merge.py
ilan-gold: If this is still broken, can you post a repro or a test? For me things passed locally

ilia-kats: They pass with this workaround. Apply
```patch
diff --git a/src/anndata/_core/merge.py b/src/anndata/_core/merge.py
index 8a05a36f..855f29ce 100644
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -764,21 +764,9 @@ class Reindexer:
             fill_value = np.nan
         index_dim = el.index_dim
         if axis == 0:
-            # Dataset.reindex() can't handle ExtensionArrays
-            extension_arrays = {
-                col: arr for col, arr in el.items() if is_extension_array_dtype(arr)
-            }
-            el = el.drop_vars(extension_arrays.keys())
             el = el.reindex(
                 {index_dim: self.new_idx}, method=None, fill_value=fill_value
             )
-            for col, arr in extension_arrays.items():
-                el[col] = (
-                    index_dim,
-                    pd.Series(arr, index=self.old_idx).reindex(
-                        self.new_idx, fill_value=fill_value
-                    ),
-                )
             return el
         else:
             msg = "This should be unreachable, please open an issue."
```

and run `pytest test_concatenate.py::test_concatenate_fill_value`.

src/anndata/_core/merge.py
```diff
@@ -1254,45 +1351,57 @@ def concat_dataset2d_on_annot_axis(
...
+    else:
+        annotations = unify_dtypes(annotations)
+    for a in annotations:
+        old_key = a.index_dim
+        is_fake_index = old_key != a.true_index_dim
         # First create a dummy index
         a.coords[DS_CONCAT_DUMMY_INDEX_NAME] = (
             old_key,
-            pd.RangeIndex(a[a.attrs["indexing_key"]].shape[0]).astype("str"),
+            pd.RangeIndex(a.shape[0]),
```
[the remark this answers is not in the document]
ilan-gold: I see, we always end up concatenating these anyway using xarray which handles this for us.

src/anndata/_core/merge.py
ilia-kats: Yes, it seems to work with your fix.

ilan-gold:
@ilia-kats Is there anything else we are missing here? I will give this a final look tomorrow.  I'm hoping to bundle this, the upcoming zarr release as a minimum bound, and the upcoming xarray release as another minimum bound, into one new release candidate.

ilia-kats:
If your xarray PR is merged before the next release I would remove the workaround, and I would appreciate some advice on what to do about the docs (see [this](https://github.com/scverse/anndata/pull/1966#issuecomment-2801093818) comment). That's it, I think.

ilan-gold:
> I would appreciate some advice on what to do about the docs (see https://github.com/scverse/anndata/pull/1966#issuecomment-2801093818 comment)

Opened a PR on this

src/anndata/_core/merge.py
```diff
@@ -1099,7 +1185,11 @@ def _resolve_axis(
 def axis_indices(adata: AnnData, axis: Literal["obs", 0, "var", 1]) -> pd.Index:
     """Helper function to get adata.{dim}_names."""
     _, axis_name = _resolve_axis(axis)
-    return getattr(adata, f"{axis_name}_names")
+    attr = getattr(adata, axis_name)
+    if isinstance(attr, Dataset2D):
+        return attr.true_index
+    else:
+        return attr.index
```
ilan-gold, line 1192:
```suggestion
    attr = getattr(adata, axis_name)
    return attr.index
```
(or revert the change, I think), both in conjunction with removing:

```
concat_indices.name = DS_CONCAT_DUMMY_INDEX_NAME
concat_annot.index = concat_indices
```
since what this is trying to handle is when the "true indices" are not used i.e., dummy indices are present.  But we ensure that the "true indices" are used inside of `concat_dataset2d_on_annot_axis`.

src/anndata/_core/merge.py
```diff
@@ -1567,10 +1679,17 @@ def concat(  # noqa: PLR0912, PLR0913, PLR0915
...
+        )
         concat_indices.name = DS_CONCAT_DUMMY_INDEX_NAME
+        concat_annot.index = concat_indices
     if label is not None:
-        concat_annot[label] = label_col
+        concat_annot[label] = (
+            label_col
+            if not isinstance(concat_annot, Dataset2D)
+            else (DS_CONCAT_DUMMY_INDEX_NAME, label_col)
+        )
```
ilan-gold, line 1693: Can we move the previous implementation of `concat_annot[label] = label_col` into the `if are_any_annotations_dataframes` branch, and then move the xarray-specific logic into `concat_dataset2d_on_annot_axis`?

Separately, we should add a test for this, but if you want to leave that out of this PR, I can follow up anyway with something that tests both `label` and `keys` for xarray

ilan-gold (commented):
Two small things and then we should be good

src/anndata/_core/merge.py
ilan-gold: Could be wrong here, but tests passed locally for me and I think we explicitly test behavior around differing indexes

src/anndata/_core/merge.py
ilia-kats: I think that is already tested in `test_concatenate.py::test_batch_key`, unless you mean something else?

src/anndata/_core/merge.py
ilan-gold: You're right, again still in `tests/lazy` mode mentally, so I was using that as a quick check.

src/anndata/_core/merge.py
```diff
@@ -296,13 +308,40 @@ def try_unifying_dtype(
...
+                    )
+                    same_orders &= all(
+                        np.all(categories == dtype.categories)
+                        for dtype in dtypes
+                        if not pd.isnull(dtype) and len(dtype.categories) > 0
+                    )
+                    if same_orders:
+                        return next(iter(dtypes))
+                    else:
+                        return object
```
ilan-gold, line 344:
```suggestion
                    if same_orders:
                        return next(iter(dtypes))
                    return object
```

src/anndata/_core/xarray.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @index.setter
+    def index(self, val) -> None:
+        index_dim = self.index_dim
+        self.coords[index_dim] = (index_dim, val)
+        if isinstance(val, pd.Index) and val.name is not None and val.name != index_dim:
+            self.update(self.rename({self.index_dim: val.name}))
+            del self.coords[index_dim]
+        if "indexing_key" in self.attrs:
+            del self.attrs["indexing_key"]
```
ilan-gold, line 79: Why do we delete the `indexing_key` in the `if` branch here, `if "indexing_key" in self.attrs:`?  Maybe should comment, at first glance I don't see why `index` and `true_index` (for which `indexing_key` is the key, not necessarily a dimension/coordinate) are related.

ilan-gold (commented):
Just a small question, if you give a quick comment in the code on it (or here, and it's sufficiently obvious that I was just having a brain fart) I'll merge

src/anndata/_core/xarray.py
ilia-kats: I figured if someone is setting a new index on the object, the expectation is that this is the one definitive index without anything else. But I can remove that branch, I don't think that behavior is used anywhere at the moment.

src/anndata/_core/xarray.py
```diff
@@ -0,0 +1,145 @@
...
+    def index_dim(self) -> str:
+        if len(self.sizes) != 1:
+            msg = f"xarray Dataset should not have more than 1 dims, found {len(self.sizes)} {self.sizes}, {self}"
+            raise ValueError(msg)
+        return next(iter(self.coords.keys()))
+
+    @property
+    def true_index_dim(self) -> str:
+        index_dim = self.attrs.get("indexing_key", None)
+        return index_dim if index_dim is not None else self.index_dim
```
ilan-gold, line 45:
```suggestion
        return self.attrs.get("indexing_key", self.index_dim)
```

src/anndata/_core/xarray.py
ilan-gold: Ok, suggested a change above then, I see the logic a bit more now.
```suggestion
        index_dim = self.index_dim
        self.coords[index_dim] = (index_dim, val)
        if isinstance(val, pd.Index) and val.name is not None and val.name != index_dim:
            self.update(self.rename({self.index_dim: val.name}))
            del self.coords[index_dim]
        # without `indexing_key` explicitly set on `self.attrs`, `self.true_index_dim` will use the `self.index_dim`
        if "indexing_key" in self.attrs:
            del self.attrs["indexing_key"]
```

ilan-gold (approved):

docs/api.md
```diff
@@ -180,7 +180,7 @@ Types used by the former:
    experimental.StorageType
    experimental.backed._lazy_arrays.MaskedArray
    experimental.backed._lazy_arrays.CategoricalArray
-   experimental.backed._xarray.Dataset2D
+   _core.xarray.Dataset2D
```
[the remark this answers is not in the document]
flying-sheep: I already had this comment as a pending review but never submitted it :(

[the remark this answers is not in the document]
ilan-gold: Yes, apologies you are 100% right.

PR #1976, opened 2025-04-24
(fix): `zarr` bound

ilan-gold:
See https://github.com/zarr-developers/zarr-python/issues/2998

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
     # array-api-compat 1.5 has https://github.com/scverse/anndata/issues/1410
     "array_api_compat>1.4,!=1.5",
     "legacy-api-wrap",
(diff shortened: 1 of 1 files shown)


ilan-gold:
> should this be an upper bound?

Should it be exclusion? I thought upper bound would be safer for now

flying-sheep:
I personally prefer exclusion **when I know that the next release has the fix**.

When we don’t know when the fix will come, we need to do an upper bound like you did. But then it’s a temporary fix and we need to be on top of it and make it more precise ASAP.

ilan-gold:
I opened https://github.com/zarr-developers/zarr-python/pull/3015 but I don't know when it will be merged and Josh Moore said the next release might be very quick: https://github.com/zarr-developers/zarr-python/pull/3007#issuecomment-2823535240

PR #1984, opened 2025-05-13
Revert min-deps breakage and fix min deps

flying-sheep:


.github/workflows/test-cpu.yml
```diff
@@ -56,18 +56,18 @@ jobs:
...
       - name: Install dependencies
         if: matrix.dependencies-version == null
-        run: uv pip install --system --compile "anndata[dev,test-full] @ ." -c ci/constraints.txt
+        run: uv pip install --system --compile -r pyproject.toml --extra=dev --extra=test -e 'anndata @ .' -c ci/constraints.txt
 
       - name: Install minimum dependencies
         if: matrix.dependencies-version == 'minimum'
         run: |
           uv pip install --system --compile tomli packaging
-          deps=$(python3 ci/scripts/min-deps.py pyproject.toml --extra dev test-min)
```
flying-sheep, line 65: how in the hell was that working‽ there is no `--extra` option for `min-deps.py`, it’s called `--extras`!

flying-sheep: https://docs.python.org/3/library/argparse.html#allow-abbrev

I’m stunned that this is the default

ilan-gold (approved):

Issue #1697, opened 2024-09-26
support gene panel selection/adaptation


mojtababahrami:
### Please describe your wishes and possible alternatives to achieve the desired result.

I come across this issue many times when I have two Anndata objects A and B and I want to change B to have the same vars as A by keeping the shared genes and inflating the new genes that are in A but not in B with a default value (e.g. zeros). I searched the API but did not find a function to do this.
Now I do something like the following to achieve this:
```
adata = ad.concat([adata_train, adata_test], join='outer')
adata_test = adata[len(adata_train):]
adata_test = adata_test[:, adata_train.var.index]
assert (adata_train.var_names == adata_test.var_names).all()
```

It would be good to support it natively as I assume this is a very common use case (for example you train a model e.g. a simple PCA on A and you want to project/transform the B using that model). 

So I imagine something like this:
```
adata_test.adapt_vars(adata_train, fill_value=0)
```

ilan-gold:
Hmmm @mojtababahrami this is an interesting idea...I will need to think about it a bit.  My initial reaction is that should be fairly straightforward but there may be some rough edges.  Thanks for the issue!

PR #1986, opened 2025-05-14
Add adapt_vars_like to align .var between AnnData objects (issue #1697)

amalia-k510:
This adds a helper function `adapt_vars_like` that makes sure the `.var` (i.e., gene metadata) of a target AnnData matches that of a source object. It copies over .obs from the target and reindexes the data matrix .X so that any missing genes are filled with a default value (by default I set it to 0.0). Useful when working with datasets that have different gene sets and you need to bring them to a shared space. For example, before concatenation. Helps avoid issues when downstream functions expect identical .var across objects.


src/anndata/utils.py
```diff
@@ -450,3 +450,48 @@ def module_get_attr_redirect(
...
+def adapt_vars_like(
+    source: AnnData, target: AnnData, fill_value: float = 0.0
+) -> AnnData:
+    # source = AnnData object that defines the desired genes
+    # target = the data you want to reshape to match source
+    # fill_vlaue = what value to use for missing genes (default set to 0.0)
+    # returns a new AnnData object with the same genes as source
+    """
+    Make target have the same .var (genes) as source., missing genes are filled with fill_value.
+    """
```
ilan-gold, line 464: Try copying the format here of other docstrings and add this to the public API i.e., in `docs/api.md`.  The CI job (or locally you) can then check the doc rendering

src/anndata/utils.py
```diff
@@ -450,3 +450,48 @@ def module_get_attr_redirect(
...
+    if target.X is None:
+        msg = "target.X is None; cannot adapt vars without a data matrix."
+        raise ValueError(msg)
+    # this will define the gene list we want to match
+    new_var = source.var.copy()
+    # initializing a new dense np array of shape (number of target cells, number of genes in source)
+    # filled with fill_value
+    # this will become the new .X matrix.
+    # It makes sure all genes in source are represented, and placeholders are ready for copying shared ones
+    new_x = np.full((target.n_obs, new_var.shape[0]), fill_value, dtype=target.X.dtype)
```
ilan-gold, line 479: I think you can use the `Reindexer` class we have in `src/anndata/_core/merge.py` to handle the reindexing logic.  You'll just need to pass in the old/new indices

ilan-gold: This will allow us to handle different array types and dataframes.  Check out in that class how many different arrays there are, it's non-trivial definitely!

ilan-gold (commented):
Let's add tests as well!  Great start, on the right track no doubt, and the implementation itself looks good for what it is, but let's take it a step further!

Main comment (besides tests) would be to use the `ReIndexer` class to apply the logic to all parts of the `AnnData` object.  The function you end up writing will likely look somewhat similar to `anndata.concatenate` in that you'll need to apply `ReIndexer` to `var`, `X`, `layers`, `varm` and `varp` (the later three recursively to their respective sub-elements).

src/anndata/utils.py
amalia-k510: I updated the docstring format, but I’m still a bit unsure about how `api.md` is structured. I did add a section for the function, but I’d really appreciate it if you could take a quick look and let me know if anything needs adjusting.

src/anndata/utils.py
amalia-k510: I did end up switching, but maybe it is a bit too general now. Could you possibly review that as well?

tests/test_utils.py
```diff
@@ -55,3 +56,47 @@ def test_adata_unique_indices():
...
+    target = ad.AnnData(X=np.array([[7, 8]]), var=pd.DataFrame(index=["g3", "g4"]))
+    output = adapt_vars_like(source, target, fill_value=0.5)
+    np.testing.assert_array_equal(output.X, [[0.5, 0.5]])
+
+
+def test_adapt_vars_missing_genes():
+    source = ad.AnnData(X=np.ones((1, 3)), var=pd.DataFrame(index=["g1", "g2", "g3"]))
+    target = ad.AnnData(X=np.array([[1, 3]]), var=pd.DataFrame(index=["g1", "g3"]))
+    output = adapt_vars_like(source, target, fill_value=-1)
+    np.testing.assert_array_equal(output.X, [[1, -1, 3]])
```
ilan-gold, line 102: These two as well can probably be consolidated as well, maybe with the other two I pointed at, trying everything with both no fill value provided, and a fill value provided.

docs/api.md
```diff
@@ -118,6 +118,20 @@ Writing formats that cannot represent all aspects of {class}`AnnData` objects.
...
+Helper functions used internationally or for reshaping and aligng `AnnData` objects. Can be useful for cusotm workflows or edge cases.
+
+```{eval-rst}
+.. autosummary::
+	:toctree: generated/
+
+	utils.adapt_vars_like
+
+```
+
```
ilan-gold, line 134: Yup this looks great!!!

ilan-gold: The little `(utilities-api)=` at the top helps with intersphinx mapping i..e, accessing sections of docs from one project in another project's docs.  Thanks for adding it!

ilan-gold (commented):
Awesome stuff :) The test cases are super clear, thanks for them.

src/anndata/utils.py
```diff
@@ -450,3 +450,71 @@ def module_get_attr_redirect(
...
+        # computing the list of genes that are in source and target
+        shared = source.var_names.intersection(target.var_names)
+        # getting positions of the shared genes in source and target
+        source_idx = new_var.index.get_indexer(shared)
+        target_idx = target.var_names.get_indexer(shared)
+        # creating a new matrix of shape (number of cells, number of genes in source)
+        # filled with the fill_value
+        new_x = np.full((target.n_obs, new_var.shape[0]), fill_value)
+        # for the genes that are in both source and target, copy over the values
+        new_x[:, source_idx] = target.X[:, target_idx]
```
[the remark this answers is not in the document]
amalia-k510: I double-checked, and it does. I rewrote this function, so if you wouldn't mind checking it, I'd appreciate it!

src/anndata/utils.py
```diff
@@ -450,3 +451,82 @@ def module_get_attr_redirect(
...
+
+    # reindexing each layer matrix along the gene (column) axis so it matches new structure
+    new_layers = {
+        k: reindexer(v, axis=1, fill_value=fill_value) for k, v in target.layers.items()
+    }
+    # reindex varm which stores matrix-like annotation for each gene
+    # for each entry, reindex along the gene axis, cast it to a numpy array to make it uniform
+    # convert it to a plain python list to avoid type checker error
+    new_varm: dict[str, Sequence[Any]] = {
+        k: np.asarray(reindexer(v, axis=0, fill_value=fill_value)).tolist()
```
ilan-gold, line 509: There should be a better way to get around the type checker.  We definitely don't want to convert things like sparse matrices to list!

Check out our `typing.py` and `types.py` files for things that might help here (like `AxisStorable`)

src/anndata/utils.py
```diff
@@ -450,3 +451,82 @@ def module_get_attr_redirect(
...
+        for k, v in target.varm.items()
+    }
+    # creating new Anndata Object
+    # directly copying .obs without changes - we ar enot touching the cells here, just aligning features
+    new_adata = AnnData(
+        X=new_X,
+        obs=target.obs.copy(),
+        var=new_var,
+        varm=new_varm,
+        layers=new_layers,
```
ilan-gold, line 519: We'll need everything, including `obsp` and `varp` with `varp` also reindexed

ilan-gold (commented):
Check out the `gen_adata` function for a way to create "full" anndata objects that have all the different types and such to really thoroughly test that we are handling everything.

Otherwise, this is looking great.  Getting close to being done

src/anndata/utils.py
```diff
@@ -450,3 +451,82 @@ def module_get_attr_redirect(
...
+    new_var = source.var.copy()
+    # Initializing reindexer = help map the old gene indices to the new structure
+    reindexer = Reindexer(target.var.index, new_var.index)
+    # if target object actually has .X matrix (i.e. expression data), I reindex it to match the source
+
+    if target.X is not None:
+        new_X = reindexer(target.X, axis=1, fill_value=fill_value)
+    else:
+        # otherwise I just create a dummy matrix of the right shape filled with a constant value
+        new_X = np.full((target.n_obs, len(new_var)), fill_value)
```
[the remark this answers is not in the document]
amalia-k510: Yeah, `Reindexer(target.X)` won’t crash if `target.X` is `None`, it’ll just return `None`. But the point of the if/else is to make sure `new_X` is always a proper matrix with the right shape, filled with `fill_value`, so things downstream don’t break. If we drop the if, we’d end up passing None into places that probably assume a real array. That’s how I interpreted it at least but let me know if that assumption doesn’t hold.

src/anndata/utils.py
amalia-k510: Please let me know if the implementation I pushed aligns with the formatting conventions used in anndata.

src/anndata/utils.py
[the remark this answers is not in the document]
ilan-gold: Ok I think my point here was that if `target`has no `X` we should probably not create an empty one.  We can just set `X=None` when declaring the new `AnnData` instead of https://github.com/scverse/anndata/pull/1986/files#diff-22197e419767db6d7078531198e8c055d27d35510281a164e8343ec48fa9a938R523 setting it to this `np.full`

ilan-gold (commented):
Let's add some more tests like when `target.X is None` and "full" object tests from `gen_adata`, but other than that, this PR is looking pretty good

src/anndata/utils.py
ilan-gold: Yes!

src/anndata/utils.py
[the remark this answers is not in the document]
amalia-k510: Got it. I addressed that issue!

tests/test_utils.py
```diff
@@ -55,3 +56,127 @@ def test_adata_unique_indices():
...
+        pytest.param(
+            ad.AnnData(X=np.ones((1, 3)), var=pd.DataFrame(index=["g1", "g2", "g3"])),
+            ad.AnnData(X=np.array([[1, 3]]), var=pd.DataFrame(index=["g1", "g3"])),
+            -1,
+            np.array([[1, -1, 3]]),
+            id="missing_genes",
+        ),
+    ],
+)
+def test_adapt_vars_with_fill_value(source, target, fill_value, expected_X):
```
ilan-gold, line 109: Let's merge this test and `test_adapt_vars` and have a fill value of `None` for the `test_adapt_vars` ones in the param

ilan-gold:
Fixes #1697

ilan-gold (approved):
Let's add a raw test (code coverage is complaining) and merge those two tests, but other than that, this looks great.

PR #1988, opened 2025-05-15
Fix: Improve concat_on_disk varm handling + prototype for strict merge of .uns

amalia-k510:
This builds on @milos7250’s PR #1855 and fixes an issue in `concat_on_disk` where `.varm` was present in one input but missing in the other, which caused `test_concat_on_disk_varm_mapping` to fail. I also added a prototype for a stricter `.uns` merge option called `merge_same_strict`, which uses `merge_nested` and a `_raise_on_conflict` function that only accepts non-empty values that are identical across all datasets; otherwise, it throws a `ValueError`. The goal was to catch silent conflicts early, but I’m not sure if this behavior makes sense in practice. Open to feedback on whether it’s worth keeping as an optional merge strategy like `uns_merge="same_strict"` or dropping it entirely.

### DIFF
--- a/tests/test_concatenate_disk.py
+++ b/tests/test_concatenate_disk.py
@@ -292,37 +292,52 @@ def test_concat_on_disk_varm_and_uns_series(tmp_path):
     assert "meta" in adata.uns
 
 
-def test_varm_uns_missing_in_one_input(tmp_path):
-    a = AnnData(X=np.ones((2, 3)))
+def test_concat_on_disk_varm_and_uns_series(tmp_path):
+    a = AnnData(X=np.ones((3, 3)))
     b = AnnData(X=np.ones((2, 3)))
-    a.var_names = b.var_names = ["g1", "g2", "g3"]
 
-    a.varm["marker"] = pd.DataFrame([1, 2, 3], index=a.var_names, columns=["marker"])
-    # b.varm is empty
-    a.uns["meta"] = {"v": pd.DataFrame([1, 2], columns=["v"])}
-    # b.uns is empty
-    # constructing paths for writing the anndata files and thei final merged output
+    a.var_names = b.var_names = ["g1", "g2", "g3"]
+    a.obs_names = [f"a{i}" for i in range(3)]
+    b.obs_names = [f"b{i}" for i in range(2)]
+
+    marker_df = pd.DataFrame({"s": [1, 2, 3]}, index=["g1", "g2", "g3"])
+    a.varm["marker"] = marker_df
+    b.varm["marker"] = marker_df
+
+    a.uns["description"] = ["first", "second"]
+    a.uns["meta"] = {"version": [1, 2, 3]}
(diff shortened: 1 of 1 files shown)


src/anndata/experimental/merge.py
```diff
@@ -402,6 +417,24 @@ def _write_axis_annot(
     write_elem(output_group, axis_name, concat_annot)
 
 
+def _write_uns(groups, output_group, merge):
+    # function writes the .uns slot of merged AnnData objects to disk
+    # read and merge all .uns from input groups
```
ilan-gold, line 422: In favor of removing some of the comments although may are helpful

tests/test_concatenate_disk.py
```diff
@@ -252,3 +253,93 @@ def test_output_dir_exists(tmp_path):
...
+
+    concat_on_disk({"x": p1, "y": p2},
+                   out, axis=0,
+                   join="outer", merge="unique", 
+                   uns_merge="unique", index_unique="-")
+    result = ad.read_h5ad(out)
+    # checking with marker field is present after merge
+    assert "marker" in result.varm
+    # checking if meta field also survive even though not present in b
+    assert "meta" in result.uns
```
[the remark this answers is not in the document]
ilan-gold: See above comment about writing out `uns`

src/anndata/experimental/merge.py
```diff
@@ -402,6 +417,24 @@ def _write_axis_annot(
...
+    # making sure it is converted into a dataframe for consistency
+    if isinstance(uns, pd.Series):
+        uns = uns.to_frame()
+    # convert any nested series inside the merged dict to dataframes
+    # iterating over each element in dict and converting it to df
+    if isinstance(uns, dict):
+        for k, v in uns.items():
+            if isinstance(v, pd.Series):
+                uns[k] = v.to_frame()
+    write_elem(output_group, "uns", uns)
```
ilan-gold, line 434: I must be missing something here, but this would imply one could write out `uns` as a dataframe in some cases?

ilan-gold (commented):
Nice! Great tests!

src/anndata/experimental/merge.py
```diff
@@ -371,13 +371,28 @@ def _write_concat_sequence(
 
 
 def _write_alt_mapping(groups, output_group, alt_axis_name, alt_indices, merge):
-    alt_mapping = merge([read_as_backed(g[alt_axis_name]) for g in groups])
```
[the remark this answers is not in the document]
amalia-k510: Could you clarify what you meant by this comment or the idea behind it? I’ve tried modifying that line in a few different ways, but I keep running into the same issues regardless of what I change.

src/anndata/experimental/merge.py
amalia-k510: I modified this function, so let me know if the updated version is still confusing.

PR #1991, opened 2025-05-16
(fix): check xarray installation when reading in a dataframe

### LINKED ISSUE #1989: `ad.concat` is slow on lazy data on account of `tokenize`

### Report

Code:

```python
import anndata

paths = []

adatas = [anndata.experimental.read_lazy(path) for path in paths]

anndata.concat(adatas) # adatas are all lazily loaded
```

Looking at a profiler output, this takes about half a second per new extension array column to be dask-ified which is too long

### Versions

```python
numpy   2.1.3
dask    2025.1.0
----    ----
hatch-vcs       0.4.0
prompt_toolkit  3.0.50
PyYAML  6.0.2
pluggy  1.5.0
jaraco.functools        4.0.1
tblib   3.0.0
h5py    3.13.0
asttokens       3.0.0
msgpack 1.1.0
natsort 8.4.0
decorator       5.1.1
psutil  7.0.0
parso   0.8.4
hatchling       1.27.0
distributed     2025.1.0
zstandard       0.23.0
tornado 6.4.2
ipython 8.32.0
jaraco.collections      5.1.0
zarr    3.0.6
numcodecs       0.15.1
wcwidth 0.2.13
scanpy  1.11.0
more-itertools  10.3.0
python-dateutil 2.9.0.post0
traitlets       5.14.3
pandas  2.2.3
executing       2.2.0
rich    13.9.4
wrapt   1.17.2
awkward 2.7.4
Deprecated      1.2.18
session-info2   0.1.2
scipy   1.15.2
pytz    2025.1
locket  1.0.0
setuptools      75.8.0
jedi    0.19.2
cloudpickle     3.1.1
zict    3.0.0
legacy-api-wrap 1.4.1
stack-data      0.6.3
fsspec  2025.2.0
MarkupSafe      3.0.2
pathspec        0.12.1
donfig  0.8.1.post1
jaraco.context  5.3.0
pyarrow 19.0.1
awkward_cpp     44
crc32c  2.7.1
Jinja2  3.1.5
jaraco.text     3.12.1
click   8.1.8
pure_eval       0.2.3
toolz   1.0.0
setuptools-scm  8.1.0
packaging       24.2
charset-normalizer      3.4.1
xarray  2025.4.0
attrs   25.3.0
six     1.17.0
sortedcontainers        2.4.0
Pygments        2.19.1
----    ----
Updated 2025-05-15 14:12
```

ilan-gold:
- [x] Closes #1989
- [x] Tests added

### DIFF
--- a/src/anndata/experimental/backed/_io.py
+++ b/src/anndata/experimental/backed/_io.py
@@ -13,6 +13,7 @@ from anndata._types import AnnDataElem
 from testing.anndata._doctest import doctest_needs
 
 from ..._core.anndata import AnnData
+from ..._core.xarray import requires_xarray
 from ..._settings import settings
 from ...compat import ZarrGroup, is_zarr_v2
 from .. import read_dispatched
@@ -25,6 +26,7 @@ if TYPE_CHECKING:
 
 
 @doctest_needs("xarray")
+@requires_xarray
 def read_lazy(
     store: PathLike[str] | str | MutableMapping | ZarrGroup | h5py.Dataset,
     *,
@@ -81,13 +83,6 @@ def read_lazy(
     AnnData object with n_obs × n_vars = 490 × 33452
         obs: 'donor_id', 'self_reported_ethnicity_ontology_term_id', 'organism_ontology_term_id'...
     """
-    try:
-        import xarray  # noqa: F401
-    except ImportError as e:
-        msg = (
-            "xarray is required to use the `read_lazy` function. Please install xarray."
-        )
-        raise ImportError(msg) from e
     is_h5_store = isinstance(store, h5py.Dataset | h5py.File | h5py.Group)
     is_h5 = (
         isinstance(store, PathLike | str) and Path(store).suffix == ".h5ad"
--- a/src/anndata/_core/xarray.py
+++ b/src/anndata/_core/xarray.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from functools import wraps
 from typing import TYPE_CHECKING
 
 import pandas as pd
@@ -8,6 +9,19 @@ from ..compat import XDataset
 
 if TYPE_CHECKING:
     from ..compat import XDataArray
+
+
+def requires_xarray(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            import xarray  # noqa: F401
+        except ImportError as e:
+            msg = "xarray is required to read dataframes lazily. Please install xarray."
+            raise ImportError(msg) from e
+        return func(*args, **kwargs)
+
+    return wrapper
 
 
 class Dataset2D(XDataset):
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -12,7 +12,7 @@ from scipy import sparse
 
 import anndata as ad
 from anndata._core.file_backing import filename, get_elem_name
-from anndata._core.xarray import Dataset2D
+from anndata._core.xarray import Dataset2D, requires_xarray
 from anndata.abc import CSCDataset, CSRDataset
 from anndata.compat import DaskArray, H5Array, H5Group, XDataArray, ZarrArray, ZarrGroup
 
@@ -258,6 +258,7 @@ DUMMY_RANGE_INDEX_KEY = "_anndata_dummy_range_index"
 
 @_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("dataframe", "0.2.0"))
 @_LAZY_REGISTRY.register_read(H5Group, IOSpec("dataframe", "0.2.0"))
+@requires_xarray
 def read_dataframe(
     elem: H5Group | ZarrGroup,
     *,
@@ -297,6 +298,7 @@ def read_dataframe(
 
 @_LAZY_REGISTRY.register_read(ZarrGroup, IOSpec("categorical", "0.2.0"))
 @_LAZY_REGISTRY.register_read(H5Group, IOSpec("categorical", "0.2.0"))
+@requires_xarray
 def read_categorical(
     elem: H5Group | ZarrGroup,
     *,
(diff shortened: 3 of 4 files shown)


ilan-gold:
The codecov is complaining because we need a similar fix to https://github.com/scverse/scanpy/pull/3607#discussion_r2095543601 where we have "min-deps" i.e., no xarray dask etc.

flying-sheep:
This should be removed once our minimum supported pandas version no longer has that default (i.e. can no longer possibly throw this warning).

Is that the case?

hatch.toml
```diff
@@ -15,7 +15,7 @@ scripts.clean = "git restore --source=HEAD --staged --worktree -- docs/release-n
 
 [envs.hatch-test]
 default-args = [  ]
-features = [ "dev", "test-min" ]
+features = [ "test-min" ]
```
ilan-gold: `dev` was installing `test` which was installing `xarray`, so I removed it.  I couldn't see a reason to have it, but it's possible there is some sort of unknown use-case for it (to me)

ilan-gold: So this was also a pretty big whif on our test suite since we weren't testing without xarray

ilan-gold:
The awkward array docs are down (?) so we'll just rerun tomorrow

flying-sheep:
I guess they just started(?) exporting as `ak.Array`:

```nushell
❯ sphobjinv co json -u https://awkward-array.org/doc/stable/objects.inv - | get entries | where dispname =~ '[.]Array'
╭───┬────────────────────────────────────────┬────────┬──────┬──────────┬─────────────────────────────────────────────┬──────────────────────╮
│ # │                  name                  │ domain │ role │ priority │                     uri                     │       dispname       │
├───┼────────────────────────────────────────┼────────┼──────┼──────────┼─────────────────────────────────────────────┼──────────────────────┤
│ 0 │ reference/ak.builder.ArrayBuilder      │ std    │ doc  │ -1       │ reference/ak.builder.ArrayBuilder.html      │ ak._ext.ArrayBuilder │
│ 1 │ reference/generated/ak.Array           │ std    │ doc  │ -1       │ reference/generated/ak.Array.html           │ ak.Array             │
│ 2 │ reference/generated/ak.ArrayBuilder    │ std    │ doc  │ -1       │ reference/generated/ak.ArrayBuilder.html    │ ak.ArrayBuilder      │
│ 3 │ reference/generated/ak.types.ArrayType │ std    │ doc  │ -1       │ reference/generated/ak.types.ArrayType.html │ ak.types.ArrayType   │
╰───┴────────────────────────────────────────┴────────┴──────┴──────────┴─────────────────────────────────────────────┴──────────────────────╯
```

hatch.toml
flying-sheep: Good catch! But the actual issue is that `dev` depends on `test`, so please fix that instead and put `dev` back in here!

The purpose of `dev` is "additional stuff you want while developing", i.e. `hatch-vcs` for runtime version generation, maybe `irkernel` to be able to jump in with a temp notebook, this kind of things.

src/anndata/experimental/backed/_io.py
```diff
@@ -25,14 +26,15 @@
...
+@requires_xarray
 def read_lazy(
     store: PathLike[str] | str | MutableMapping | ZarrGroup | h5py.Dataset,
     *,
     load_annotation_index: bool = True,
 ) -> AnnData:
     """
     Lazily read in on-disk/in-cloud AnnData stores, including `obs` and `var`.
-    No array data should need to be read into memory with the exception of :class:`ak.Array`, scalars, and some older-encoding arrays.
+    No array data should need to be read into memory with the exception of :class:`awkward.Array`, scalars, and some older-encoding arrays.
```
flying-sheep: why did you do that? did they intermittently export this as `awkward.Array` instead of `ak.Array` as they did before and do now?

hatch.toml
ilan-gold: Great call!

tests/test_concatenate.py
```diff
@@ -217,7 +217,25 @@ def test_concat_interface_errors(obs_xdataset, var_xdataset):
 )
 @pytest.mark.parametrize(
     ("obs_xdataset", "var_xdataset", "force_lazy"),
```
flying-sheep: please make the purpose of these parameter sets clear by passing `id=` to all three parameter sets, e.g. what does lazy obs and var without `force_lazy` mean? `id="make_eager"`?

same in multiple cases below

ilan-gold:
@flying-sheep These tests are going to fail now because of an issue with xarray - see #2008 which is now based against this branch for test pass/failure

flying-sheep (changes requested):
See https://github.com/scverse/anndata/pull/1991/files#r2142319397

also why did you change away from `pytestmark = ...`? ideally we keep it that way, so pytest can collect all tests and then just skips the ones it can't run.

src/anndata/experimental/backed/_io.py
ilan-gold: Message on slack, it was when their docs were down.  I was just trying something out because the error was cryptic

tests/test_concatenate.py
ilan-gold: done!

ilan-gold:
@flying-sheep https://github.com/scverse/anndata/actions/runs/15609690266/job/43967402083?pr=1991 was the reason for `importorskip`.  I don't want params to be run.  Is there a way around this? `importorskip` was my solution

flying-sheep:
As said, it's not that important. You could just defer these parameters with `lambda: XDataset(...)` though, so it wouldn't be hard either.

ilan-gold:
> You could just defer these parameters with lambda: XDataset(...) though

I am not a huge fan of this paradigm so I will revert the change then.

flying-sheep (approved):
looks good, but why did you make everything into `importorskip` in d2be64ec26a944ffc3d58bd857cedc8a1d627b7b instead of just where necessary?

PR #1992, opened 2025-05-16
(fix): setting on `Dataset2D` "just works"

### LINKED ISSUE #1990: Fix setting on `Dataset2D` without specificed `coord`

### Report

Code:

```python
adata.obs["y"] = np.array(...)
```

should not add a coord when `.obs` is a `Dataset2D` (and the current behavior allows adding it as a coord if it is not the normal xarray tuple syntax).  When you do this at the moment, you run into our check against having multiple dimensions https://github.com/scverse/anndata/blob/ec650810368d45ceee7619eb38a59e8920869c60/src/anndata/_core/xarray.py#L38-L39 if you then try to use `adata.obs` subsequently.

Traceback:

```pytb
N/A
```

### Versions

```python
numpy   2.1.3
dask    2025.1.0
----    ----
hatch-vcs       0.4.0
prompt_toolkit  3.0.50
PyYAML  6.0.2
pluggy  1.5.0
jaraco.functools        4.0.1
tblib   3.0.0
h5py    3.13.0
asttokens       3.0.0
msgpack 1.1.0
natsort 8.4.0
decorator       5.1.1
psutil  7.0.0
parso   0.8.4
hatchling       1.27.0
distributed     2025.1.0
zstandard       0.23.0
tornado 6.4.2
ipython 8.32.0
jaraco.collections      5.1.0
zarr    3.0.6
numcodecs       0.15.1
wcwidth 0.2.13
scanpy  1.11.0
more-itertools  10.3.0
python-dateutil 2.9.0.post0
traitlets       5.14.3
pandas  2.2.3
executing       2.2.0
rich    13.9.4
wrapt   1.17.2
awkward 2.7.4
Deprecated      1.2.18
session-info2   0.1.2
scipy   1.15.2
pytz    2025.1
locket  1.0.0
setuptools      75.8.0
jedi    0.19.2
cloudpickle     3.1.1
zict    3.0.0
legacy-api-wrap 1.4.1
stack-data      0.6.3
fsspec  2025.2.0
MarkupSafe      3.0.2
pathspec        0.12.1
donfig  0.8.1.post1
jaraco.context  5.3.0
pyarrow 19.0.1
awkward_cpp     44
crc32c  2.7.1
Jinja2  3.1.5
jaraco.text     3.12.1
click   8.1.8
pure_eval       0.2.3
toolz   1.0.0
setuptools-scm  8.1.0
packaging       24.2
charset-normalizer      3.4.1
xarray  2025.4.0
attrs   25.3.0
six     1.17.0
sortedcontainers        2.4.0
Pygments        2.19.1
----    ----
Updated 2025-05-15 14:12
```

ilan-gold:
Since we're enforcing two dimensions on the object (in theory), we should probably be _super_ clear about that while also retaining dataframe-like operations i.e., when you try to "set a column" (in this case, xarray data variable that isn't a coord), it "just works" (sets the column and doesn't do anything else).  This behavior matches that of `pandas` in so far as having a named index doesn't mean it can be set/accessed via `df["my_named_index"]`

- [x] Closes #1990
- [x] Tests added

### DIFF
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -1338,6 +1338,7 @@ def concat_dataset2d_on_annot_axis(
     force_lazy: bool,
     label: str | None = None,
     label_col: pd.Categorical | None = None,
+    concat_indices: pd.Index | None = None,
 ) -> Dataset2D:
     """Create a concatenate dataset from a list of :class:`~anndata._core.xarray.Dataset2D` objects.
     The goal of this function is to mimic `pd.concat(..., ignore_index=True)` so has some complicated logic
@@ -1393,20 +1394,25 @@ def concat_dataset2d_on_annot_axis(
         xr.concat(annotations_re_indexed, join=join, dim=DS_CONCAT_DUMMY_INDEX_NAME),
     )
     ds.is_backed = have_backed
-    ds.coords[DS_CONCAT_DUMMY_INDEX_NAME] = pd.RangeIndex(
-        ds.coords[DS_CONCAT_DUMMY_INDEX_NAME].shape[0]
-    )
+    if concat_indices is not None:
+        concat_indices.name = DS_CONCAT_DUMMY_INDEX_NAME
+        ds.index = concat_indices
+    else:
+        ds.coords[DS_CONCAT_DUMMY_INDEX_NAME] = pd.RangeIndex(
+            ds.coords[DS_CONCAT_DUMMY_INDEX_NAME].shape[0]
+        )
     # Drop any lingering dimensions (swap doesn't delete)
     ds = ds.drop_dims(d for d in ds.dims if d != DS_CONCAT_DUMMY_INDEX_NAME)
     # Create a new true index and then delete the columns resulting from the concatenation for each index.
     # This includes the dummy column (which is neither a dimension nor a true indexing column)
-    index = xr.concat(
-        [a.true_xr_index for a in annotations_re_indexed],
-        dim=DS_CONCAT_DUMMY_INDEX_NAME,
-    )
-    # prevent duplicate values
-    index.coords[DS_CONCAT_DUMMY_INDEX_NAME] = ds.coords[DS_CONCAT_DUMMY_INDEX_NAME]
-    ds.coords[DS_CONCAT_DUMMY_INDEX_NAME] = index
+    if concat_indices is None:
+        index = xr.concat(
+            [a.true_xr_index for a in annotations_re_indexed],
+            dim=DS_CONCAT_DUMMY_INDEX_NAME,
+        )
+        # prevent duplicate values
+        index.coords[DS_CONCAT_DUMMY_INDEX_NAME] = ds.coords[DS_CONCAT_DUMMY_INDEX_NAME]
+        ds.coords[DS_CONCAT_DUMMY_INDEX_NAME] = index
     for key in {
         true_index
         for a in annotations_re_indexed
@@ -1691,14 +1697,17 @@ def concat(  # noqa: PLR0912, PLR0913, PLR0915
             ignore_index=True,
         )
         concat_annot.index = concat_indices
-        if label is not None:
-            concat_annot[label] = label_col
     else:
         concat_annot = concat_dataset2d_on_annot_axis(
-            annotations, join, force_lazy=force_lazy, label=label, label_col=label_col
+            annotations,
+            join,
+            force_lazy=force_lazy,
+            label=label,
+            label_col=label_col,
+            concat_indices=concat_indices,
         )
-        concat_indices.name = DS_CONCAT_DUMMY_INDEX_NAME
-        concat_annot.index = concat_indices
+    if label is not None:
+        concat_annot[label] = label_col
 
     # Annotation for other axis
     alt_annotations = [getattr(a, alt_axis_name) for a in adatas]
--- a/src/anndata/_core/xarray.py
+++ b/src/anndata/_core/xarray.py
@@ -140,14 +140,26 @@ class Dataset2D(XDataset):
         return pd.Index(columns)
 
     def __setitem__(self, key, value):
+        if key == self.index_dim:
+            msg = f"Cannot set {self.index_dim} as a variable. Use `index` instead."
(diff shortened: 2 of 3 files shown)


src/anndata/_core/merge.py
```diff
@@ -781,11 +781,8 @@ def _apply_to_dataset2d(self, el: Dataset2D, *, axis, fill_value=None):
                 {index_dim: self.new_idx}, method=None, fill_value=fill_value
             )
             for col, arr in extension_arrays.items():
-                el[col] = (
-                    index_dim,
-                    pd.Series(arr, index=self.old_idx).reindex(
-                        self.new_idx, fill_value=fill_value
-                    ),
+                el[col] = pd.Series(arr, index=self.old_idx).reindex(
+                    self.new_idx, fill_value=fill_value
```
flying-sheep, line 785: to be clear, this is just a cosmetic change, because `index_dim` is redundant?

src/anndata/_core/merge.py
```diff
@@ -1347,6 +1344,7 @@ def concat_dataset2d_on_annot_axis(
     force_lazy: bool,
     label: str | None = None,
     label_col: pd.Categorical | None = None,
+    concat_indices: pd.Index | None = None,
```
flying-sheep, line 1345: please describe this in the docstring.

src/anndata/_core/merge.py
ilan-gold: Yea, we could continue with the old behavior.

src/anndata/_core/xarray.py
```diff
@@ -143,3 +138,45 @@ def columns(self) -> pd.Index:
         if index_key is not None:
             columns.discard(index_key)
         return pd.Index(columns)
+
+    def __setitem__(self, key, value):
+        """
+        Setting can only be performed when the incoming value is \"standalone\" like :class:`nump.ndarray` to mimic pandas.
```
flying-sheep: What would be an example of a non-standalone value? Remote readonly or …?

```suggestion
        Setting can only be performed when the incoming value is “standalone” like :class:`nump.ndarray` to mimic pandas.
```

src/anndata/_core/xarray.py
```diff
@@ -143,3 +138,45 @@ def columns(self) -> pd.Index:
...
+            raise KeyError(msg)
+        if isinstance(value, tuple):
+            if isinstance(value[0], tuple):
+                if value[0][0] != self.index_dim:
+                    msg = f"Dimension tuple should have only {self.index_dim} as its dimension, found {value[0][0]}"
+                    raise ValueError(msg)
+                if len(value[0]) > 1:
+                    msg = "Dimension tuple is too long."
+                    raise ValueError(msg)
+            if not isinstance(value[0], tuple) and value[0] != self.index_dim:
```
flying-sheep: See https://github.com/scverse/anndata/pull/1992#discussion_r2120680878
```suggestion
            elif value[0] != self.index_dim:
```

src/anndata/_core/xarray.py
```diff
@@ -143,3 +138,45 @@ def columns(self) -> pd.Index:
...
+        if key == self.index_dim:
+            msg = f"Cannot set {self.index_dim} as a variable. Use `index` instead."
+            raise KeyError(msg)
+        if isinstance(value, tuple):
+            if isinstance(value[0], tuple):
+                if value[0][0] != self.index_dim:
+                    msg = f"Dimension tuple should have only {self.index_dim} as its dimension, found {value[0][0]}"
+                    raise ValueError(msg)
+                if len(value[0]) > 1:
+                    msg = "Dimension tuple is too long."
```
flying-sheep, line 161: What does that mean? The user is trying to set values in multiple dimensions at once? Should that be a `NotImplementedError` or does it make no sense at all?

src/anndata/_core/xarray.py
```diff
@@ -143,3 +138,45 @@ def columns(self) -> pd.Index:
...
+                    raise ValueError(msg)
+            if not isinstance(value[0], tuple) and value[0] != self.index_dim:
+                msg = f"Setting value tuple should have first entry {self.index_dim}, found {value[0]}"
+                raise ValueError(msg)
+
+        if not isinstance(value, tuple) and not isinstance(value, XDataArray):
+            # maintain setting behavior of a 2D dataframe i.e., one dim
+            value = (self.index_dim, value)
+        if isinstance(value, XDataArray):
```
flying-sheep: Either make the whole thing into a `match` block or:

```suggestion
        elif isinstance(value, XDataArray):
```

and then make this branch into an `else` branch:

```diff
-if not isinstance(value, tuple) and not isinstance(value, XDataArray):
+else:
    # maintain setting behavior of a 2D dataframe i.e., one dim
    value = (self.index_dim, value)
```

tests/test_xarray.py
```diff
@@ -89,3 +90,112 @@ def test_index(dataset2d):
...
+    ("da", "msg"),
+    [
+        pytest.param(
+            XDataArray(
+                np.arange(3),
+                coords={"foo": ("obs_names", np.arange(3))},
+                dims="obs_names",
+                name="bar",
+            ),
+            "DataArray should have coordinate obs_names",
```
flying-sheep: either:
- rename `msg` to `pattern` and add an `r` to each of the pattern string literals to hint that they’re regexes, or
- use `re.escape` in the test body to make sure the test uses the values as-is

src/anndata/_core/xarray.py
ilan-gold: "Standalone" was meant to distinguish from the tuple case in the next line

src/anndata/_core/xarray.py
ilan-gold: > The user is trying to set values in multiple dimensions at once?

Yup

> does it make no sense at all?

Also yup, the idea being a `Dataset2D` has one and only one dimension.

flying-sheep: maybe do sth. like this then:

```suggestion
                    msg = "Can only set values in 1 dimension, but dimension tuple is of length {len(value[0])}."
```

PR #1995, opened 2025-05-22
(feat): new zarr dtypes

ilan-gold:
- [x] Accounting for https://github.com/zarr-developers/zarr-python/pull/2874
- [x] Tests added


pyproject.toml
```diff
@@ -46,8 +46,7 @@ dependencies = [
     "packaging>=24.2",
     "array_api_compat>=1.7.1",
     "legacy-api-wrap",
-    # <3.1 on account of https://github.com/scverse/anndata/pull/1995
-    "zarr >=2.18.7, !=3.0.0, !=3.0.1, !=3.0.2, !=3.0.3, !=3.0.4, !=3.0.5, !=3.0.6, !=3.0.7, <3.1",
+    "zarr >=2.18.7, !=3.0.0, !=3.0.1, !=3.0.2, !=3.0.3, !=3.0.4, !=3.0.5, !=3.0.6, !=3.0.7, !=3.0.8, !=3.0.9, !=3.0.10",
```
flying-sheep: 3.0.0–3.0.7 have been yanked, so we don’t need to exclude them, they’re uninstallable, so this would be sufficient:

```suggestion
    "zarr >=2.18.7, !=3.0.8, !=3.0.9, !=3.0.10",
```

but why not just this?

```suggestion
    "zarr >=2.18.7, !=3.0.*",
```

ilan-gold: Ah much cleaner, yes!

ilan-gold:
> Seems like zarr.dtype.VariableLengthUTF8 wraps numpy’s new string dtype if available.

Hmm, I hadn't thought of it that way tbh, it was just "the new zarr API for dtypes" and this happens to be the only place we use it.  I will give more context next time, apologies!

flying-sheep:
Makes sense! I just wondered what that dtype is and found that it’s basically just our previous match statement, but done by zarr now.

ilan-gold:
More info here! https://zarr.readthedocs.io/en/stable/api/zarr/dtype/index.html This PR is basically what I held up the `anndata` release over - the new dtype stuff was big enough that I wanted a clean break.  But it is a very rich feature set! I haven't had time to dig into the new possibilities yet, but looks promising

Issue #1111, opened 2023-08-29
Idea: `__dataframe__` interchange protocol for anndata


ivirshup:
### Please describe your wishes and possible alternatives to achieve the desired result.

https://data-apis.org/dataframe-protocol/latest/index.html

It could be nice if AnnData supported the `__dataframe__` interchange protocol, especially when used by libraries which will use the `select_columns_by_name`, `get_column_by_name` interfaces.

## Use-case: plotting

The biggest use case is plotting. Both seaborn (https://github.com/mwaskom/seaborn/pull/3369) and altair (https://github.com/altair-viz/altair/pull/2888) now support inputs in the dataframe protocol.

In `scanpy` we typically use the `sc.get.obs_df` method to create a dataframe for plotting. A major painpoint for this in analysis code is that the user has to provide the keys they want to plot multiple times, once for creating the dataframe, and again to the plotting interface. Instead of having to do:

```python
sns.jointplot(
    data=sc.get.obs_df(adata, ["log1p_total_counts", "pct_counts_mito", "batch"]),
    x="log1p_total_counts",
    y="pct_counts_mito",
    hue="batch",
)
```

It could eventually be:

```python
sns.jointplot(
    data=adata,  # Likely something more like `DFInterface(adata, dim="obs", layer=...)` for now
    x="log1p_total_counts",
    y="pct_counts_mito",
    hue="batch",
)
```

This should also work for plots of gene expression values, especially if the underlying plotting library selects columns through the dataframe interface and the matrix was stored as CSC or dense.

This could even be a nice interface to on-disk data, especially when `X`/ `layers` is stored in `CSC`.

## Some more detail

* For dataframe interface for observations, available columns are a union of `.obs.columns`, `var_names`, keys like `obsm/pca/0`.
* We should be able to pick an alias for `var_names`
* We should be able to choose which `layer` is being accessed

## Implementation

I think it would make sense for this to start out as POC outside of the main implementation. It may require `pyarrow` as a dependency to work. In theory `pyarrow` be a dependency of `pandas` v3 early next year, so may not be an issue.

cc: @ilan-gold

ivirshup:
Very rough proof of concept:

<details>
<summary> </summary>

```python
import pandas as pd
from pandas.core.interchange.column import PandasColumn
from pandas.core.interchange.dataframe import PandasDataFrameXchg

import anndata as ad
import scanpy as sc

class ObsDF(pd.core.interchange.dataframe_protocol.DataFrame):
[...]

PR #1997, opened 2025-05-27
(feat): composable API for xarray dataset

ilan-gold:
Following up on a conversation with Luca, a composable API is like both more maintainable and more explainable.  It also represents a concrete step towards something like https://github.com/scverse/anndata/issues/1111.  This PR represents a first step as it forces us to more clearly enumerate the methods needed for internal compatibility that were hidden by subclassing from `Dataset` i.e., `copy` and `reindex`

- [ ] Closes #
- [x] Tests added


docs/conf.py
```diff
@@ -93,6 +93,14 @@
...
     ("py:class", "anndata.compat.ZappyArray"),
 ]
+nitpick_ignore_regex = [
+    # See https://github.com/python/cpython/blob/7775d93e2d142ab66c7d8fa7cff068b2522a37b0/Lib/_collections_abc.py#L792-L817
+    # for why these are necessary.
+    # The docstrings are just normal strings and thus misinterpreted.
+    # TODO: revert once this is resolved in python.
+    ("py:class", ".*d defaults to None.*"),
+    ("py:class", "a.*object providing.*a view on D.*"),
+]
```
flying-sheep: The fact that these strings end up in our docs at all is the problem, so instead of ignoring errors arising from that, please prevent the source of the issue

- either by not creating documentation for inherited methods (e.g. [like this](https://github.com/scverse/scanpy/blob/pa/settings-overhaul/docs/extensions/autosummary_skip_inherited.py)), and instead linking to `Mapping` and explaining that we inherit from it (e.g. setting [`"show-inheritance": True`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-autoclass-show-inheritance) in [`autodoc_default_options`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options))
- or by overriding them with our own docstrings

```suggestion
```

src/anndata/_core/xarray.py
```diff
@@ -1,86 +1,147 @@
...
 
     @property
     def true_xr_index(self) -> XDataArray:
-        return self[self.true_index_dim]
+        """The index {class}`~anndata.AnnData` is actually interested in e.g., cell names, for verification."""
+        return self.ds[self.true_index_dim]
 
     @property
     def true_index(self) -> pd.Index:
+        """{attr}`~anndata.experimental.backed.Dataset2D.true_xr_index` as a {class}`pandas.Index`"""
```
flying-sheep: MyST syntax isn’t supported in docstrings, please use rST syntax like you do e.g. in `index` above.

src/anndata/_core/xarray.py
```diff
@@ -1,86 +1,147 @@
...
 
 from ..compat import XDataArray, XDataset, XVariable
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
+    from typing import Any, Literal
 
-class Dataset2D(XDataset):
+
+class Dataset2D(Mapping[Hashable, "XDataArray | Dataset2D"]):
```
flying-sheep: Is there no key type in `xarray` that we can use here and in `__getitem__`?

src/anndata/_core/xarray.py
```diff
@@ -195,4 +269,79 @@ def __setitem__(self, key, value):
...
+        self,
+        index: pd.Index | None = None,
+        axis: Literal[0] = 0,
+        fill_value: Any | None = np.nan,
+    ) -> Dataset2D:
+        """Reindex the current object against a new index.
+
+        Parameters
+        ----------
+        index, optional
```
flying-sheep: no “optional” in these, please. always just the plain parameter name.

flying-sheep (changes requested):
There are a bunch of documentation issues, otherwise it looks fine!

please go through every single documented item and make sure the docs look right, there is broken syntax in some.

src/anndata/_core/merge.py
```diff
@@ -1377,33 +1344,41 @@ def concat_dataset2d_on_annot_axis(
...
+        dim=DS_CONCAT_DUMMY_INDEX_NAME,
     )
-    ds.is_backed = have_backed
+    ds_concat.attrs.pop("indexing_key", None)
+    ds_concat_2d = Dataset2D(ds_concat)
+    ds_concat_2d.is_backed = have_backed
     if concat_indices is not None:
         concat_indices.name = DS_CONCAT_DUMMY_INDEX_NAME
-        ds.index = concat_indices
+        ds_concat_2d.index = concat_indices
```
ilan-gold, line 1374: Add comment about why we need to re-wrap

src/anndata/_core/xarray.py
ilan-gold: Ah sorry, yea just get the two mixed up sometimes.

src/anndata/_core/xarray.py
ilan-gold: This was taken from them actually :/ I can add a comment, I had the same thought.  This is quite broad and not very helpful.

pyproject.toml
```diff
@@ -109,7 +109,7 @@ gpu = [ "cupy" ]
 cu12 = [ "cupy-cuda12x" ]
 cu11 = [ "cupy-cuda11x" ]
 # requests and aiohttp needed for zarr remote data
-lazy = [ "xarray>=2025.04.0", "aiohttp", "requests", "anndata[dask]" ]
+lazy = [ "xarray>=2025.04.0,<2025.06.0", "aiohttp", "requests", "anndata[dask]" ]
```
flying-sheep: Please always add a comment next to upper bounds that is sufficient to explain what needs to happen to remove the bound. An issue link is usually sufficient.

src/anndata/_core/xarray.py
```diff
@@ -1,86 +1,151 @@
...
+# See https://github.com/pydata/xarray/blob/main/xarray/core/dataset.py#L194 for typing
+class Dataset2D(Mapping[Hashable, "XDataArray | Dataset2D"]):
     """
-    A wrapper class meant to enable working with lazy dataframe data.
-    We do not guarantee the stability of this API beyond that guaranteed
-    by :class:`xarray.Dataset` and the `to_memory` function, a thin wrapper
-    around :meth:`xarray.Dataset.to_dataframe` to ensure roundtrip
-    compatibility here.
+
+    Bases :class:`~collections.abc.Mapping` [ :class:`~collections.abc.Hashable`, :class:`~xarray.DataArray` | :class:`~anndata.experimental.backed.Dataset2D` ]
```
flying-sheep: this should work to get rid of the spaces here

```suggestion
    r"""
    Bases :class:`~collections.abc.Mapping`\ [:class:`~collections.abc.Hashable`, :class:`~xarray.DataArray` | :class:`~anndata.experimental.backed.Dataset2D`\ ]
```

src/anndata/_core/xarray.py
```diff
@@ -195,4 +275,79 @@ def __setitem__(self, key, value):
...
+        return len(self.ds)
+
+    @property
+    def dtypes(self) -> pd.Series:
+        """
+        Return a Series with the dtypes of the variables in the Dataset2D.
+        """
+        return self.ds.dtypes
+
+    def equals(self, b: object) -> bool:
```
flying-sheep, line 353: This doesn't have a docstring, is this not supposed to be a public API? (ditto for `to_memory`)

src/anndata/_core/xarray.py
```diff
@@ -195,4 +275,79 @@ def __setitem__(self, key, value):
...
+            }
+            el = self.ds.drop_vars(extension_arrays.keys())
+            el = el.reindex({index_dim: index}, method=None, fill_value=fill_value)
+            for col, arr in extension_arrays.items():
+                el[col] = pd.Series(arr, index=self.index).reindex(
+                    index, fill_value=fill_value
+                )
+            return Dataset2D(el)
+        msg = "This should be unreachable, please open an issue."
+        raise Exception(msg)
```
flying-sheep: I guess if `Dataset2D` is a public API and this is a public API, this should be a `ValueError` and tell users to not do that instead.

Also plain `Exception` is iffy, as originally worded this would be an `AssertionError` I guess. Please start asking me if you're unsure which exception class is appropriate.

could you also make this an early exit and add `# pragma: no cover` instead of having the main function body inside of an if branch? My logic here is
- if your function covers multiple equally-importand cases, use `match` (unless the conditions are so complex that `match` is more verbose than an if-else tree)
- if your function has one or more parameter checks that can result in an error, handle these using early errors to leave the main function body unindented.

flying-sheep (changes requested):
this looks good, just some nitpicks!

Could you also check (manually is OK, no need for tests) if the Mapping mixin methods work as expected?

> `__contains__`, `keys`, `items`, `values`, `get`, `__eq__`, and `__ne__`

just curious if `self.ds` raises the appropriate exceptions to make these work.

pyproject.toml
ilan-gold: Ah, can be removed after the linked PR, things move quick: see #2008

src/anndata/_core/xarray.py
ilan-gold: > Also plain Exception is iffy, as originally worded this would be an AssertionError I guess. Please start asking me if you're unsure which exception class is appropriate.

This was just copied.  I should have caught this earlier when Ilia added it.

src/anndata/_core/xarray.py
ilan-gold: @flying-sheep why the `r"""` to start? Without it things seem to look ok for me locally:
Screenshot 2025-06-12 at 17 11 36

src/anndata/_core/xarray.py
flying-sheep: before I forget: `# pragma: no cover` or test coverage?

ilan-gold: thanks for the reminder!

src/anndata/_core/xarray.py
ilan-gold: Ah is it to escape the `\ `? Not sure I am familiar with the mechanism here

ilan-gold: Huh also the `r"""` was added in, I think by the formatter.  I don't remember doing it so still curious what is going on here, I thought that a raw string would not escape the `\` but must be misunderstanding

ilan-gold:
> Could you also check (manually is OK, no need for tests) if the Mapping mixin methods work as expected?

I decided to remove the inheritance from `Mapping` to lower the API surface area.  It turns out it wasn't really necessary and we now have something very close to the "anndata-dataframe" API which will make the 0.13 relatively straightforward

flying-sheep:
> I decided to remove the inheritance from Mapping to lower the API surface area.

Uh, but why? Inheriting from `Mapping` gave you that API surface for free. The fact that you had to make the code worse because you lost the `.items()` demonstrates that it's a net bonus.

And the upstream class does it too.

Having more classes that integrate with Python's type system when it makes sense instead of doing their own thing is good! E.g. I hate that `np.ndarray` can be 0d and therefore isn't a `Sequence`, makes everything more complicated.

src/anndata/_core/xarray.py
```diff
@@ -115,32 +180,61 @@ def iloc(self):
...
         """
+        coord = self.index_dim
 
         class IlocGetter:
             def __init__(self, ds):
                 self._ds = ds
 
             def __getitem__(self, idx):
-                coord = self._ds.index_dim
-                return self._ds.isel(**{coord: idx})
```
flying-sheep, line 125: why is `coord` now pulled from the function scope while `ds` is stored in `IlocGetter`'s `__dict__`? Shouldn't they both be stored in the same scope?

src/anndata/_core/xarray.py
```diff
@@ -217,4 +311,80 @@ def __setitem__(self, key, value):
         else:
             # maintain setting behavior of a 2D dataframe i.e., one dim
             value = (self.index_dim, value)
-        super().__setitem__(key, value)
+        self.ds.__setitem__(key, value)
+
+    def copy(self, *args, **kwargs):
```
flying-sheep: missing typing here and in other places.

ilan-gold:
> Uh, but why? Inheriting from Mapping gave you that API surface for free. The fact that it's a bonus is that you had to make the code worse because you lost the .items() demonstrates that it's a net bonus.

I think there are two reasons not to expose more API surface right now:

1. I intend to turn this set of methods into a runtime checkable protocol (or similar) once this is released.  So having `Mapping` there will create ambiguity as to what is necessary i.e.,  is `__eq__` really necessary or just part of the API?
2. My gut feeling is that people should less not more with the wrapper.  `xarray.Dataset` on its own is compatible with `AnnData` indpendent of the wrapper i.e., setting on `adata.obs` will work and be wrapper internally.  So I want to encourage the use of the dataset object on its own.

I understand the python type system argument, but we can always add `Mapping` in.  It will be harder to remove.

src/anndata/_core/xarray.py
ilan-gold: Can change, no reaosn

flying-sheep:
I feel like if something has `.__getitem__` and `.keys()` with `Mapping` semantics, it wants to be a `Mapping`, and I'm very surprised when it doesn't also have `.values()` and `.items()`. Inheriting from `Mapping` makes that all work as expected.

> is `__eq__` really necessary or just part of the API

That's a good point. Maybe just have it throw a `NotImplementedError`?

If you define it yourself, `Mapping` won't mix it in.

ilan-gold:
Here's an example of why I don't want to inherit from `Mapping`:

```python
import pandas as pd
df = pd.read_csv('https://raw.githubusercontent.com/mwaskom/seaborn-data/master/iris.csv')
ds = df.to_xarray()
ds.values() # what you'd expect, a bunch of `xarray.DataArrays` i.e., the columns themselves in dataframe parlance
df.values()
```
```pytb
Cell In[20], line 1
----> 1 df.values()

TypeError: 'numpy.ndarray' object is not callable
```

If people start to complain, I get it, but I'd rather address things as they arise than assume I know what we need.  With the current implementation, I know what we need to be `DataFrame` compatible, which is the purpose of the wrapper.

flying-sheep:
Yeah, pandas' API is very bad in parts. `pd.DataFrame.values` is at least deprecated in favor of `.to_numpy()` and `.array`.

PR #1998, opened 2025-05-27
(chore): derive CI matrix from hatch env

ilan-gold:
- [ ] Just like https://github.com/scverse/scanpy/pull/3607
- [ ] Tests added

### DIFF
--- a/.github/workflows/test-cpu.yml
+++ b/.github/workflows/test-cpu.yml
@@ -22,31 +22,48 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  pytest:
+  get-environments:
+    runs-on: ubuntu-latest
+    outputs:
+      envs: ${{ steps.get-envs.outputs.envs }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          filter: blob:none
+          fetch-depth: 0
+      - uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: false
+      - id: get-envs
+        run: |
+          ENVS_JSON=$(NO_COLOR=1 uvx hatch env show --json | jq -c 'to_entries
+            | map(
+              select(.key | startswith("hatch-test"))
+              | {
+                name: .key,
+                "test-type": (if (.key | test("pre|min")) then "coverage" else null end),
+                python: .value.python,
+              }
+            )')
+          echo "envs=${ENVS_JSON}" | tee $GITHUB_OUTPUT
+  test:
+    needs: get-environments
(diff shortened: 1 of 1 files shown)


.github/workflows/test-cpu.yml
```diff
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 env:
-  PYTEST_ADDOPTS: "-v --color=yes"
+  PYTEST_ADDOPTS: "-v --color=yes -n auto --internet-tests --junitxml=test-data/test-results.xml"
```
flying-sheep, line 11: `--internet-tests` doesn’t exist here, it’s a custom flag we define in scanpy’s test config

flying-sheep:
Ah, I just realized: Since we do

```bash
hatch run ${{ matrix.env.name }}:run
```

Instead of `hatch test`, hatch won’t do its CLI argument processing, so `--randomize` and `--parallel` don’t exist, and the normal defaults of the plugins apply.

E.g. when you don’t pass `--randomize`, hatch passes `-p no:randomly` to the CLI to disable the plugin. Which means that we’re already using the plugin in scanpy (and this PR).

ilan-gold:
> Doing hatch run hatch-test.*:run means that we’re already using the plugin in scanpy (and this PR).

Why did I have to add the `PYTEST_ADDOPTS` then?

ilan-gold:
Re: testing for coverage, I would be in favor of doing it for each type here.  `pre`, `min`, and `stable` could in theory have three different behaviors (unlikely) in which case we would want full coverage.

flying-sheep (approved):
looks good, but why are the tests so slow? (that’s not new, I just saw it now)

ilan-gold:
@flying-sheep just to be 100% clear, we also want to add in the hatch checks as well, no?

.github/workflows/test-cpu.yml
```diff
@@ -55,43 +67,20 @@ jobs:
...
 
-      - name: Run Pytest (coverage)
-        if: matrix.test-type == 'coverage'
-        run: coverage run -m pytest -n auto --cov --cov-report=xml
+      - name: Run tests
+        run: uvx hatch run ${{ matrix.env.name }}:run-cov -v --color=yes -n auto --cov --cov-report=xml --junitxml=test-data/test-results.xml
 
-      - name: Run Pytest (treat warnings as errors)
-        if: matrix.test-type == 'strict-warning'
-        run: pytest --strict-warnings -n auto
```
flying-sheep, line 85: oof, this should have stayed in.

PR #2005, opened 2025-06-04
(feat): turn off indptr caching for `hdf5` in dask

ilan-gold:
While profiling some code in dask, I noticed how much time was spent caching the indptr repeatedly for hdf5:

Screenshot 2025-06-04 at 17 15 59

This could explain some difference in performance between zarr and hdf5 in dask @Intron7 where I would guess (need to check) that the indptr is reused across chunks because the `BaseCompressedSparseDataset` seems to be shared across chunks from within a process.

An alternative to this idea (probably complementary since I think this feature is good-to-have regardless): we could somehow reuse `indptr` across the workers.  In practice, though, I don't know how this works.  I am not clear if the array would be copied once-per-worker (good), or once-per-chunk (bad).

To benchmark, checkout `git checkout ab46efc8` to get a commit where this feature is configurable and

```python

import dask.array as da
import zarr
import anndata as ad
import subprocess
import h5py
from fast_array_utils import stats
import dask.distributed as dd
from pathlib import Path

cluster = dd.LocalCluster(n_workers=8 )
client = dd.Client(cluster)
SPARSE_CHUNK_SIZE = 50_000

# 4 combinations here
user_zarr = False
should_cache_indptr = True

h5_path = Path("data/cell_atlas.h5ad")
zarr_path = Path("data/cell_atlas.zarr")

h5_path.parent.mkdir(exist_ok=True)
if not h5_path.exists():
    download(
        "https://datasets.cellxgene.cziscience.com/82eac9c1-485f-4e21-ab21-8510823d4f6e.h5ad",
        str(h5_path),
    )
if not zarr_path.exists():
    ad.read_h5ad(h5_path).write_zarr(zarr_path)
```

```python
%%time
subprocess.call(['sudo', 'sh', '-c', "sync; echo 3 > /proc/sys/vm/drop_caches"])
if use_zarr:
    f = zarr.open(zarr_path)
    adata = ad.AnnData(
        obs=ad.io.read_elem(f["obs"]), # only take the column necessary to reduce memory pressure
        var=ad.io.read_elem(f["var"]),
    )
    adata.X = ad.experimental.read_elem_lazy(
        f["X"], chunks=(SPARSE_CHUNK_SIZE, adata.shape[1]), load_indptr_in_parent=False
    )
else:
    with h5py.File(h5_path, "r") as f:
        adata = ad.AnnData(
            obs=ad.io.read_elem(f["obs"]), # only take the column necessary to reduce memory pressure
            var=ad.io.read_elem(f["var"]),
        )
        adata.X = ad.experimental.read_elem_lazy(
            f["X"], chunks=(SPARSE_CHUNK_SIZE, adata.shape[1]), should_cache_indptr=should_cache_indptr
        )


da.compute(stats.mean(adata.X))
```

For me on DENBI machine:

| indptr_cache    | format | time |
| -------- | ------- | ------- |
| True  | zarr    | 10.5
| True | h5ad     | 22.2
| False  | zarr    | 10.6
| False | h5ad     | 17.2

- [x] Tests added

### DIFF
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -395,10 +395,12 @@ def validate_indices(
 
 class BaseCompressedSparseDataset(abc._AbstractCSDataset, ABC):
     _group: GroupStorageType
+    _should_cache_indptr: bool
 
-    def __init__(self, group: GroupStorageType):
+    def __init__(self, group: GroupStorageType, *, should_cache_indptr: bool = True):
         type(self)._check_group_format(group)
         self._group = group
+        self._should_cache_indptr = should_cache_indptr
 
     @property
     def group(self) -> GroupStorageType:
@@ -616,8 +618,9 @@ class BaseCompressedSparseDataset(abc._AbstractCSDataset, ABC):
 
         It should therefore fit into memory, so we cache it for faster access.
         """
-        arr = self.group["indptr"][...]
-        return arr
+        if self._should_cache_indptr:
+            return self.group["indptr"][...]
+        return self.group["indptr"]
 
     @cached_property
     def _indices(self) -> H5Array | ZarrArray:
@@ -660,13 +663,19 @@ class _CSCDataset(BaseCompressedSparseDataset, abc.CSCDataset):
     """Internal concrete version of :class:`anndata.abc.CSRDataset`."""
 
 
-def sparse_dataset(group: GroupStorageType) -> abc.CSRDataset | abc.CSCDataset:
+def sparse_dataset(
+    group: GroupStorageType,
+    *,
+    should_cache_indptr: bool = True,
+) -> abc.CSRDataset | abc.CSCDataset:
     """Generates a backed mode-compatible sparse dataset class.
 
     Parameters
     ----------
     group
         The backing group store.
+    should_cache_indptr
+        Whether or not to cache the indptr for repeated reuse as a :class:`numpy.array`
 
     Returns
     -------
@@ -713,9 +722,9 @@ def sparse_dataset(group: GroupStorageType) -> abc.CSRDataset | abc.CSCDataset:
     """
     encoding_type = _get_group_format(group)
     if encoding_type == "csr":
-        return _CSRDataset(group)
+        return _CSRDataset(group, should_cache_indptr=should_cache_indptr)
     elif encoding_type == "csc":
-        return _CSCDataset(group)
+        return _CSCDataset(group, should_cache_indptr=should_cache_indptr)
     msg = f"Unknown encoding type {encoding_type}"
     raise ValueError(msg)
 
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -678,7 +678,7 @@ def sparse_dataset(
         Whether or not to cache the indptr for repeated reuse as a :class:`numpy.ndarray`.
         The default is `True` but one might set it to false if the dataset is repeatedly reopened
         using this command, and then only a subset is read in before closing again.
-        See https://github.com/scverse/anndata/blob/26d9f89a962724508f75b46a30569df225d0eb3b/src/anndata/_io/specs/lazy_methods.py#L85-L90
+        See https://github.com/scverse/anndata/blob/3c489b979086c39c59d3eb5dad90ebacce3b9a80/src/anndata/_io/specs/lazy_methods.py#L85-L95
         for the target use-case.
 
     Returns
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -675,7 +675,7 @@ def sparse_dataset(
     group
         The backing group store.
     should_cache_indptr
-        Whether or not to cache the indptr for repeated reuse as a :class:`numpy.array`
+        Whether or not to cache the indptr for repeated reuse as a :class:`numpy.ndarray`
 
     Returns
     -------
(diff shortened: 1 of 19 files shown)


ilan-gold:
Some eyeball confirmation this actually helps (beyond the fact that in testing on the cluster, data processing time went down by about a third) is given here.  Green is the io, yellow processing

With this PR:
Screenshot 2025-06-05 at 12 57 48
Previously:
image

src/anndata/_io/specs/registry.py
```diff
@@ -405,7 +405,9 @@ def read_elem(elem: StorageType) -> RWAble:
 
 
 def read_elem_lazy(
-    elem: StorageType, chunks: tuple[int, ...] | None = None, **kwargs
+    elem: StorageType,
+    chunks: tuple[int, ...] | None = None,
+    **kwargs,
```
ilan-gold, line 410: This is not an argument here because otherwise it would be passed down to all registered methods, which would fail because we explicitly check `if kwarg not in read_params`

ilan-gold:
Looking into this closer - the speedup is real without caching for hdf5 files, at least.  I need to check zarr.  I also need to double check the PCA step, which does not seem to benefit from the speedup (at least stand-alone).  So I think something inside one of these functions is what is causing this issue to surface.

```python
import anndata as ad
import dask.distributed as dd
import scanpy as sc
import h5py

cluster = dd.LocalCluster(n_workers=8)
client = dd.Client(cluster)
SPARSE_CHUNK_SIZE=50_000

adatas = []
for i in (range(4)):
    id = str(i + 1)
    PATH = f"/lustre/groups/ml01/workspace/goncalo.pinto/plate{id}_raw.h5ad"

    with h5py.File(PATH, "r") as f:
        adata = ad.AnnData(
            obs=ad.io.read_elem(f["obs"])[['pass_filter']], # only take the column necessary to reduce memory pressure
            var=ad.io.read_elem(f["var"]),
        )
        adata.X = ad.experimental.read_elem_lazy(
            f["X"], chunks=(SPARSE_CHUNK_SIZE, adata.shape[1]), should_cache_indptr=False
        )
        # 100m filtering
    pass_filter_mask = adata.obs["pass_filter"] == "full"
    adata = adata[pass_filter_mask, :].copy() # a copy is made anyway by normalize_total
    
    sc.pp.normalize_total(adata)
    sc.pp.log1p(adata)
    sc.pp.highly_variable_genes(adata, n_top_genes=2000)
```

I assume the big speedup comes from `highly_variable_genes` mean-var calls where they are computed jointly and therefore the chunks are only fetched once.

With caching (i.e., current behavior):

```
CPU times: user 3min 34s, sys: 54.6 s, total: 4min 29s
Wall time: 14min 11s
```

Without caching (this PR): 
```
CPU times: user 3min 21s, sys: 48.4 s, total: 4min 10s
Wall time: 10min 35s
```

ilan-gold:
Screenshot 2025-06-10 at 13 12 00

must have been an aberration with the pca - noticeably faster with even a simple `mean` operation

ilan-gold:
Confirmed with cache clearning on DENBI what I suspected: this only applicable for h5ad.  And hte difference is 20-30%

src/anndata/_core/sparse_dataset.py
```diff
@@ -660,13 +663,19 @@ class _CSCDataset(BaseCompressedSparseDataset, abc.CSCDataset):
     """Internal concrete version of :class:`anndata.abc.CSRDataset`."""
 
 
-def sparse_dataset(group: GroupStorageType) -> abc.CSRDataset | abc.CSCDataset:
+def sparse_dataset(
+    group: GroupStorageType,
+    *,
+    should_cache_indptr: bool = True,
```
ilan-gold, line 669: This is the only place this appears as a "public" api.  We could just use the private variable directly by assignment if this is too public

flying-sheep:
What do you mean with the quotes? Is this not actually really public? https://anndata.readthedocs.io/en/latest/generated/anndata.io.sparse_dataset.html

ilan-gold:
> What do you mean with the quotes? Is this not actually really public?

No it is public, but `sparse_dataset` is a less used API.  So we can go fully private if you want

flying-sheep:
I'm OK with `sparse_dataset` having this parameter, it changes behavior of the class this function exists for, so it's sufficiently far away from being an implementation detail here.

src/anndata/_core/sparse_dataset.py
```diff
@@ -660,13 +663,19 @@ class _CSCDataset(BaseCompressedSparseDataset, abc.CSCDataset):
...
+) -> abc.CSRDataset | abc.CSCDataset:
     """Generates a backed mode-compatible sparse dataset class.
 
     Parameters
     ----------
     group
         The backing group store.
+    should_cache_indptr
+        Whether or not to cache the indptr for repeated reuse as a :class:`numpy.ndarray`
```
flying-sheep: maybe explain when someone would choose to do that

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -83,7 +83,11 @@ def make_dask_chunk(
     # We need to open the file in each task since `dask` cannot share h5py objects when using `dask.distributed`
     # https://github.com/scverse/anndata/issues/1105
     with maybe_open_h5(path_or_sparse_dataset, elem_name) as f:
-        mtx = ad.io.sparse_dataset(f) if isinstance(f, H5Group) else f
+        mtx = (
+            ad.io.sparse_dataset(f, should_cache_indptr=False)
```
flying-sheep, line 92: please add a comment explaining why you set it to False here.

PR #2006, opened 2025-06-05
(feat): allow loading the indptr from the parent process

ilan-gold:
A complementary solution to #2005 

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_core/sparse_dataset.py
+++ b/src/anndata/_core/sparse_dataset.py
@@ -605,19 +605,25 @@ class BaseCompressedSparseDataset(abc._AbstractCSDataset, ABC):
         indices[orig_data_size:] = append_indices
 
         # Clear cached property
-        for attr in ["_indptr", "_indices", "_data"]:
+        for attr in ["_indices", "_data"]:
             if hasattr(self, attr):
                 delattr(self, attr)
+        if hasattr(self, "_indptr_cached"):
+            self._indptr_cached = None
 
-    @cached_property
+    @property
     def _indptr(self) -> np.ndarray:
         """\
         Other than `data` and `indices`, this is only as long as the major axis
 
         It should therefore fit into memory, so we cache it for faster access.
         """
-        arr = self.group["indptr"][...]
-        return arr
+        if getattr(self, "_indptr_cached", None) is None:
+            self._indptr_cached = self.group["indptr"][...]
+        return self._indptr_cached
+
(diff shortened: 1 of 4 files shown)


ilan-gold:
Looking into this closer - the speedup is real without caching for hdf5 files, at least.  I need to check zarr.  I also need to double check the PCA step, which does not seem to benefit from the speedup (at least stand-alone).  So I think something inside one of these functions is what is causing this issue to surface.

```python
import anndata as ad
import dask.distributed as dd
import scanpy as sc
import h5py

cluster = dd.LocalCluster(n_workers=8)
client = dd.Client(cluster)
SPARSE_CHUNK_SIZE=50_000

adatas = []
for i in (range(4)):
    id = str(i + 1)
    PATH = f"/lustre/groups/ml01/workspace/goncalo.pinto/plate{id}_raw.h5ad"

    with h5py.File(PATH, "r") as f:
        adata = ad.AnnData(
            obs=ad.io.read_elem(f["obs"])[['pass_filter']], # only take the column necessary to reduce memory pressure
            var=ad.io.read_elem(f["var"]),
        )
        adata.X = ad.experimental.read_elem_lazy(
            f["X"], chunks=(SPARSE_CHUNK_SIZE, adata.shape[1]), should_cache_indptr=False
        )
        # 100m filtering
    pass_filter_mask = adata.obs["pass_filter"] == "full"
    adata = adata[pass_filter_mask, :].copy() # a copy is made anyway by normalize_total
    
    sc.pp.normalize_total(adata)
    sc.pp.log1p(adata)
    sc.pp.highly_variable_genes(adata, n_top_genes=2000)
```

I assume the big speedup comes from `highly_variable_genes` mean-var calls where they are computed jointly and therefore the chunks are only fetched once.

With caching (i.e., current behavior):

```
CPU times: user 3min 34s, sys: 54.6 s, total: 4min 29s
Wall time: 14min 11s
```

Without caching (this PR): 
```
CPU times: user 3min 21s, sys: 48.4 s, total: 4min 10s
Wall time: 10min 35s
```

ilan-gold:
I could simply not figure out how to get this to work properly but even so I'm not sure we want to block on loading `inpdtr` in the parent process.  I asked here https://dask.discourse.group/t/per-worker-i-e-process-numpy-array/3967 but couldn't get it to work quickly and without warnings about the task graph.  I don't think this is worth pursuing futher.

PR #2019, opened 2025-06-30
Backend-Agnostic Refactor Using Array API Compatibility

amalia-k510:
This PR refactors the gen_adata function in benchmarks/utils.py to support backend-agnostic generation of AnnData objects, using the array-api-compat interface to infer and operate with the correct array namespace (e.g., NumPy or JAX). The goal is to allow tests and benchmarking utilities to seamlessly run across supported array backends without manual rewrites. The function now detects the namespace from a real array object instead of relying on a global fallback and uses backend-specific RNG to generate observation and variable metadata. I also updated the related test to parameterize over NumPy and JAX backends. However, I'm currently running into an issue where passing a scipy.sparse.csr_matrix into get_namespace() raises a TypeError, since sparse matrices are not supported by array-api-compat. As a workaround, I’m generating a dense array first to infer the namespace, and only falling back to sparse when explicitly requested in the attribute set. This works fine for NumPy but breaks for JAX, as sparse support isn’t available there. Still trying to figure out whether to raise, skip, or handle this more gracefully depending on the backend. Would appreciate any thoughts on best practices here.

### DIFF
--- a/src/anndata/_core/index_modified.py
+++ b/src/anndata/_core/index_modified.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 from collections.abc import Iterable, Sequence
 from functools import singledispatch
 from itertools import repeat
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import h5py
 import numpy as np
 import pandas as pd
+from array_api_compat import get_namespace as array_api_get_namespace
 from scipy.sparse import issparse
 
 from ..compat import AwkArray, CSArray, CSMatrix, DaskArray, XDataArray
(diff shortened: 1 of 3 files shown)


ilan-gold (commented):
3 things to simplify things:

1. Ignore the benchmarks for now, including the `gen_adata` there.  There is another `gen_adata` in https://github.com/scverse/anndata/blob/main/src/anndata/tests/helpers.py that you can add a jax type to much more easily (i.e., `X_type` or or `DEFAULT_KEY_TYPES`), and then you will get the entire test suit for free running against jax.  Specifically, I would add it as a type for `obsm`, `varm`, `X`, and `layers` while ignoring the otehrs.  It doesn't make sense to use it in `obs` or `var`, for example
2. No need to use the array_api where it's not needed (like for sparse as you point out).  I would remove it from all the other subsetting functions and only have it in the default `single_dispatch` case
3. The test file you have is great! I would start there if you feel overwhelmed by the full test suite

src/anndata/tests/helpers.py
```diff
@@ -103,6 +142,54 @@ def gen_vstr_recarray(m, n, dtype=None):
     )
 
 
+# def gen_vstr_recarray(
```
ilan-gold: No need to worry bout jax for this! I don't even think strings are in the array api: https://data-apis.org/array-api/latest/API_specification/data_types.html#data-types

src/anndata/_core/index_modified.py
```diff
@@ -0,0 +1,307 @@
...
+
+@singledispatch
+def _subset(a, subset_idx: Index):
+    # Select as combination of indexes, not coordinates
+    # Correcting for indexing behaviour of np.ndarray
+    xp = get_xp(a)
+    if all(
+        isinstance(x, Iterable) and not isinstance(x, (str, bytes)) for x in subset_idx
+    ):
+        subset_idx = xp.ix_(*subset_idx)
```
ilan-gold, line 216: Chekc type of `subset_idx` against what is required from https://data-apis.org/array-api/latest/API_specification/indexing.html#indexing

PR #2025, opened 2025-07-11
fix: make tests run shorter

### LINKED ISSUE #2023: Tests take far too long

### Report

We talked about it before, but tests should maybe take around 5 minutes or less, not 25!

I think it’s zarr stuff that takes that long, right? @ilan-gold could you describe why this is again here so we can work on a fix?

### Versions

```python
| Dependency        | Version |
| ----------------- | ------- |
| typing_extensions | 4.12.2  |
| wcwidth           | 0.2.13  |
| pure_eval         | 0.2.3   |
| prompt_toolkit    | 3.0.47  |
| traitlets         | 5.14.3  |
| jedi              | 0.19.1  |
| executing         | 2.0.1   |
| session-info2     | 0.1.2   |
| ipython           | 8.26.0  |
| asttokens         | 2.4.1   |
| Pygments          | 2.18.0  |
| parso             | 0.8.4   |
| decorator         | 5.1.1   |
| stack-data        | 0.6.3   |
| six               | 1.16.0  |

| Component | Info                                                 |
| --------- | ---------------------------------------------------- |
| Python    | 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 16.0.6 ] |
| OS        | macOS-15.5-arm64-arm-64bit                           |
| Updated   | 2025-07-10 17:55                                     |
```

ilan-gold:
Cutting down the size of generated anndata objects (that are then written to disk) cuts down test time by about 20% or so overall.

Separately, as can currently be seen, tests involving `zarr-python>=3` take 6-7X longer (see the timings for the split out tests).  There are some tests not marked `zarr_io` not marked (because the marking is either manual or by detection of the presence of a name) but hopefully most have been marked.

- [x] Closes #2023
- [x] Tests added

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,8 +37,7 @@ def backing_h5ad(tmp_path):
         ("h5ad", None),
         ("zarr", 2),
         pytest.param(
-            "zarr",
-            3,
+            ("zarr", 3),
             marks=pytest.mark.skipif(
                 is_zarr_v2(), reason="zarr v3 file format not supported with v2 package"
             ),


tests/test_concatenate_disk.py
```diff
@@ -52,7 +52,7 @@ def file_format(request) -> Literal["zarr", "h5ad"]:
 
 
 # 1000 is enough to guarantee that the feature is being used
-@pytest.fixture(params=[1_000, 100_000_000])
+@pytest.fixture(params=[50, 1_000_000])
```
flying-sheep: the comment mentions “1000”, but the new parameter set doesn’t contain that number.

ilan-gold: Ah true, I also cut down the size of the stores as well

tests/test_backed_sparse.py
```diff
@@ -40,12 +40,12 @@
 N = 50
 
 
-@pytest.fixture
+@pytest.fixture(params=[pytest.param(None, marks=pytest.mark.zarr_io)])
```
ilan-gold, line 43: This was a best attempt at marking certain parameters as `zarr_io` (separately from actually checking the `ids` as in the `_pytest.py` change).  Some tests that involve `zarr` do not have the name in their id (i.e., when they test only `zarr` and not both `zarr` and `h5ad`)

ilan-gold: The alternative would be marking each zarr-involved fixture as `zarr_io` but somehow that feels wasteful.

src/anndata/_settings.py
```diff
@@ -439,6 +439,9 @@ def validate_zarr_write_format(format: int):
     if format not in {2, 3}:
         msg = "non-v2 zarr on-disk format not supported"
         raise ValueError(msg)
+    if format == 3 and is_zarr_v2():
+        msg = "Cannot write v3 format against v2 package"
+        raise ValueError(msg)
```
ilan-gold, line 444: No idea what codecov is talking about here, this does get run locally

flying-sheep: the project coverage does seem to go down a bit, but all this might be an artifact of us having some minor coverage collection issues (when we switched to running the tests with hatch, the coverage changed a little in weird inconsistent ways … not for the better or worse, just weirdly different)

tests/conftest.py
```diff
@@ -31,7 +33,17 @@ def backing_h5ad(tmp_path):
 
 
 @pytest.fixture(
-    params=[("h5ad", None), ("zarr", 2), ("zarr", 3)], ids=["h5ad", "zarr2", "zarr3"]
+    params=[
+        ("h5ad", None),
+        ("zarr", 2),
+        pytest.param(
+            ("zarr", 3),
```
flying-sheep, line 40: does this work? I thought the syntax was `pytest.param("zarr", 3, marks=...)`

ilan-gold:
> seems like the zarr 2 job takes 1 minute as opposed to 8 or 9 for the zarr 3 job, but also only runs ~⅓ of the tests.

2-3X is better than 8-9X but still pretty bad!  And I don't think the number of tests is going to change anytime soon so this flood of slower tests is probably here to stay since we will likely eventually deprecate fully zarr v2 the package, but still need backwards compat testing for the format.

tests/conftest.py
ilan-gold: Evidently not when you're using it in `@pytest.fixture` as a param!

PR #2027, opened 2025-07-14
feat: add typing for the `settings` object

ilan-gold:
- [ ] Closes # I realized this is kind of a bad dev experience in vscode
- [x] Tests added


src/anndata/_settings.pyi
```diff
@@ -0,0 +1,49 @@
...
+    def __setattr__(self, option: str, val: object) -> None: ...
+    def __getattr__(self, option: str) -> object: ...
+    def __dir__(self) -> Iterable[str]: ...
+    def reset(self, option: Iterable[str] | str) -> None: ...
+    @contextmanager
+    def override(self, **overrides) -> Generator[None]: ...
+    @property
+    def __doc__(self): ...
+
+class AnnDataSettingsManager(SettingsManager):
```
flying-sheep: maybe prefix with `_` to make clearer that this class doesn’t actually exist / isn’t supposed to be used?

src/anndata/_settings.pyi
```diff
@@ -0,0 +1,49 @@
...
+from contextlib import contextmanager
+from dataclasses import dataclass
+from typing import Literal, TypeVar
+
+_T = TypeVar("_T")
+
+@dataclass
+class SettingsManager:
+    __doc_tmpl__: str = ...
+    def describe(
```
flying-sheep, line 12: hmm, too bad we have to repeat all that stuff here.

tests/test_settings.py
```diff
@@ -260,3 +264,18 @@ class TestEnum(Enum):
 )
 def test_describe(*, as_rst: bool, expected: str, settings: SettingsManager):
     assert settings.describe("test_var_3", as_rst=as_rst) == expected
+
+
+def test_hints():
+    settings = ad.settings
+    types_loader = importlib.machinery.SourceFileLoader(
+        "settings_types",
+        pathlib.Path(__file__).parent.parent.resolve() / "src/anndata/_settings.pyi",
```
flying-sheep: you only ever need to `.resolve()` if you want to pass the path to a command that might be executed with a different working directory.

Also, does this work?

```suggestion
        Path(ad.__file__).parent / "_settings.pyi",
```

src/anndata/_settings.pyi
ilan-gold: Yeah :/ But at least it's tested, which I was a bit worried about i.e., a new function or similar is addded

tests/test_settings.py
ilan-gold: Let's find out!

pyproject.toml
```diff
@@ -96,7 +96,8 @@ test-min = [
     "joblib",
     "boltons",
     "scanpy>=1.10",
-    "httpx",                # For data downloading
+    # TODO: Is 1.0dev1 a real pre-release? https://pypi.org/project/httpx/#history
```
flying-sheep, line 99: I’m pretty sure that as far as Python package installers are concerned, “pre-release” means “Is on PyPI and doesn’t have a stable version”, so yes.

Do you mean that “dev1” isn’t a version you’d make a release of? Well, they apparently do!

ilan-gold: I have concerns that it is even intentional.  I see nothing in their repo about it - can't find a branch, tag, or changelog notes.  Maybe I didn't look hard enough, but I wonder if it's even intentional

ilan-gold: I commented here: https://github.com/encode/httpx/issues/3602

PR #2032, opened 2025-07-17
feat: Usage of H5py Virtual Datasets for `concat_on_disk`

selmanozleyen:
When there is no need for reindexing and the backend is `hdf5` we can just use the [virtual datasets](https://docs.h5py.org/en/stable/vds.html) instead. This requires no in memory copies but instead it just links to the original file location instead. I was able to concat the tahoe datasets (314GB in total) in a few minutes and the result was a 12GB `.h5ad` file.

Other notes:
- The `indptr`'s were in total 780Mb for all the tahoe files so I just concatenate them in memory instead.
- Added TODO for being able to specify compression args since when I used the default approach the output file size grew up too much compared to the size of the inputs.

---
 TODOs:

- [x] Tests added (tests already cover this case)
- [x] Write about this feature in the docstrings


ilan-gold (commented):
1. I think this should be put behind an argument so it is opt-in
2. A single test should be enough to ensure results match
3. When the dataset is read back in, and a backing file has been deleted, does `h5` raise an error? Or will it error weirdly somehow within `anndata` once you try to access the data?

selmanozleyen:
> I think this should be put behind an argument so it is opt-in
> A single test should be enough to ensure results match

These are done 

> h5 raise an error?

According to the requirements documentation: https://support.hdfgroup.org/releases/hdf5/documentation/rfc/HDF5-VDS-requirements-use-cases-2014-12-10.pdf

> It will be the user’s responsibility to maintain the consistency of a VDS. If source files are
unavailable, the library will report an error or use fill values when doing I/O on the VDS
elements mapped to the elements in the missing source files.

From my experience it tries as hard to use a cache or something like that because when I delete original files it doesn't throw an error and the result file looks as if it didn't change. This behaviour isn't discussed much in https://docs.h5py.org

ilan-gold:
@selmanozleyen Do you want my review on this again or are the comments still somewhat unaddressed?

selmanozleyen:
> Do you want my review on this again or are the comments still somewhat unaddressed?

I found the behavior undefined/unpredictable when the source files are deleted. I couldn't find a way to overcome that because it's not stated clearly in the docs. Sometimes it errors sometimes it doesn't as in the CI tests.

If we are fine with just documenting this and merging I will remove the fail assertions then it will be read to merge

tests/test_concatenate_disk.py
```diff
@@ -156,7 +160,105 @@ def test_anndatas(
...
         file_format,
-        max_loaded_elems,
+        max_loaded_elems=max_loaded_elems,
+        virtual_concat=False,
+        axis=axis,
+        join=join_type,
+    )
+
+
+def test_anndatas_virtual_concat_missing_file(
```
ilan-gold: We shouldn't have multiple different functions that do almost identical things.  Please refactor

src/anndata/experimental/merge.py
```diff
@@ -215,41 +217,53 @@ def write_concat_sparse(  # noqa: PLR0917
...
 
 
 def _write_concat_mappings(  # noqa: PLR0913, PLR0917
     mappings,
     output_group: ZarrGroup | H5Group,
     keys,
     path,
+    *,
     max_loaded_elems,
+    virtual_concat,
```
ilan-gold: Typing! and `virtual_concat` is a boolean, so let's prefix this everywhere with a verb like `use` i.e., `use_virtual_concat`

src/anndata/experimental/merge.py
```diff
@@ -215,41 +217,53 @@ def write_concat_sparse(  # noqa: PLR0917
...
+    else:
+        init_elem = next(elems)
+        number_non_zero = sum(d.group["indices"].shape[0] for d in datasets)
+        indptr_dtype = "int64" if number_non_zero >= np.iinfo(np.int32).max else "int32"
+        write_elem(
+            output_group,
+            output_path,
+            init_elem,  # TODO: user should be able to specify dataset kwargs
+            dataset_kwargs=dict(indptr_dtype=indptr_dtype),
+        )
```
ilan-gold, line 326: I see this as the TODO but I'm not sure its relevance to 

```
Added TODO for being able to specify compression args since when I used the default approach the output file size grew up too much compared to the size of the inputs.
```

or in general, why the resultant dataset is 12GB.  Is this `obs` and `var` or?

src/anndata/experimental/merge.py
```diff
@@ -441,7 +468,11 @@ def concat_on_disk(  # noqa: PLR0912, PLR0913, PLR0915
     To adjust the maximum amount of data loaded in memory; for sparse
     arrays use the max_loaded_elems argument; for dense arrays
     see the Dask documentation, as the Dask concatenation function is used
-    to concatenate dense arrays in this function
+    to concatenate dense arrays in this function.
+
+    For sparse arrays, if the backend is hdf5 and there is no reindexing and
+    `virtual_concat` is True,
+    the virtual concatenation is used using the `h5py` virtual dataset support.
```
ilan-gold:
```suggestion
    For sparse arrays, if the backend is hdf5 and there is no reindexing and
    `virtual_concat` is True,
    virtual concatenation is used via docs.h5py.org/en/stable/vds.html.
```
Also maybe see if we can intersphinx this link? I suspect so but am not sure

ilan-gold (changes requested):
> I found the behavior undefined/unpredictable when the source files are deleted. I couldn't find a way to overcome that because it's not stated clearly in the docs.

Can we ask the developers of `h5py`?

> If we are fine with just documenting this and merging I will remove the fail assertions then it will be read to merge

I would like to at least open an issue with the `h5py` people and give it a day or two (I have some changes requested here anyway)

src/anndata/experimental/merge.py
selmanozleyen: I am not sure but should I do an analyis on this? In general passing datasets kwargs sounds like a good idea to me given that this function is already for advanced users and larger scales

src/anndata/experimental/merge.py
selmanozleyen: done

tests/test_concatenate_disk.py
selmanozleyen: done

PR #2033, opened 2025-07-17
fix: chunking for lazy reading

ilan-gold:
Two things after a report of a slowdown from Jan

1. Our default chunking for `hdf5` and `zarr` should be based on the on-disk chunks, just like `zarr` in `dask` by default - without on-disk chunking (as can happen for h5), we use our default striding.
2. We should allow for custom chunking for the `Dataset2D` object passed through from `read_elem_lazy` to its constituent dask arrays

Locally, for a compressed `h5` dataset, this yielded a ~40% speedup for a shape of `(118925, 53)`:

Before:
```python
In [5]: %timeit ds.to_memory()
621 ms ± 25.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
After:
```python
In [8]: %timeit ds.to_memory()
398 ms ± 5.79 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
And then setting `chunk=(-1,)` (i.e., one chunk per-column for the `Dataset2D` columns with dask arrays):
```python
In [4]: %timeit ds.to_memory()
261 ms ± 3.04 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

See #2036 for next steps beyond just ensuring we dont create small chunks unnecessarily

- [ ] Closes #
- [x] Tests added


src/anndata/_io/specs/lazy_methods.py
```diff
@@ -228,7 +237,6 @@ def read_h5_array(
 def read_zarr_array(
     elem: ZarrArray, *, _reader: LazyReader, chunks: tuple[int, ...] | None = None
 ) -> DaskArray:
-    chunks: tuple[int, ...] = chunks if chunks is not None else elem.chunks
```
ilan-gold, line 231: This is default behavior of dask anyway

tests/test_io_elementwise.py
```diff
@@ -650,3 +650,76 @@ def test_read_sparse_array(
...
+        ((1, 50), (1, 50)),
+        ((10, -1), (10, 50)),
+        ((10, None), (10, 50)),
+        (
+            None,
+            (
+                25,
+                25,
+            ),
+        ),
```
flying-sheep:
```suggestion
        (None, (25, 25)),
```

tests/lazy/test_read.py
```diff
@@ -182,3 +185,29 @@ def test_unconsolidated(tmp_path: Path, mtx_format):
...
+    path = tmp_path_factory.mktemp("foo.zarr")
+    g = zarr.open_group(path, mode="w", zarr_format=2)
+    write_elem(g, "foo", df, dataset_kwargs={"chunks": 25})
+    return zarr.open(path, mode="r")["foo"]
+
+
+@pytest.mark.parametrize(
+    ("chunks", "expected_chunks"),
+    [((1,), (1,)), ((-1,), (120,)), (None, (25,))],
+    ids=["small", "minus_one", "none"],
```
flying-sheep: I like the explicit IDs you came up with for the other test module!

```suggestion
    ids=["small", "minus_one_uses_full", "none_uses_ondisk_chunking"],
```

tests/test_io_elementwise.py
```diff
@@ -650,3 +650,76 @@ def test_read_sparse_array(
...
     mtx = ad.io.read_elem(f["mtx"])
     assert issubclass(type(mtx), CSArray)
+
+
+@pytest.mark.parametrize(
+    ("chunks", "expected_chunks"),
+    [((1,), (1,)), ((-1,), (120,)), (None, (25,))],
+    ids=["small", "minus_one_uses_full", "none_uses_ondisk_chunking"],
+)
+@pytest.mark.parametrize("arr", [np.arange(120), np.array(["a"] * 120)])
```
flying-sheep: maybe add IDs for the arrays too? `ints`, `strings` or so?

tests/test_io_elementwise.py
```diff
@@ -650,3 +650,76 @@ def test_read_sparse_array(
...
+            (
+                25,
+                25,
+            ),
+        ),
+    ],
+    ids=[
+        "small",
+        "minus_one_uses_full",
+        "none_on_axis_uses_full",
```
flying-sheep, line 685: why doesn’t it use ondisk chunking on that axis? is that just how zarr behaves and there’s no reason to change it?

flying-sheep (approved):
Awesome! I merely have nitpicks and one question.

tests/test_io_elementwise.py
ilan-gold: > why doesn’t it use ondisk chunking on that axis?

We coded this in, it probably should have been like you describe but could potentially change/remove for the next release.  See https://github.com/scverse/anndata/pull/2033/files#diff-f2c238a8a3a892e2c0b002229fc860e3bd0732cbb772ec69b7444a831d82862aL207-L209

flying-sheep: yeah, if -1 canonically means “full”, None should be “default” i.e. derive from on-disk, right?

IDK how breaking of a change that would be.

ilan-gold: I mean this is all `experimental` so technically we could break things, I suppose.  I would be fine with that actually, I like those semantics a lot.

ilan-gold: Ah ok, no, `None` actually does mean `full axis` in `dask`, so the current behavior is correct! Good to know I wasn't just randomly coding things.  I will add a comment!

flying-sheep: Hm, it is a bit confusing that one can pass
- `5` or `None` with `None` meaning “as on disk”, or
- `(…, 5)` or `(…, None)`, with `None` meaning “whole axis”

or did I get something wrong?

ilan-gold: If I understand you correctly, yes that is how things work.  But again, that is how `dask` works.  `None` in the `tuple` means "use specifically that full axis."  `None` as the argument itself means "dask default, which is on-disk chunking _along all axes_".

Issue #2021, opened 2025-07-03
Deadlocking with `zarr` (or more reproducibly, with `zarrs`), `AnnData` object, and `torch.utils.data.DataLoader`


ilan-gold:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of anndata.
- [x] (optional) I have confirmed this bug exists on the master branch of anndata.

### Report

I understand that this reproducer is somewhat convoluted, and I intend to simplify it once I get a chance, but wanted to document it.

The following code deadlocks, apparently:

Code:

```python
import anndata as ad
from torch.utils.data import DataLoader, IterableDataset
import zarr
import zarrs
zarr.config.set({"threading.max_workers":None, "codec_pipeline.path": "zarrs.ZarrsCodecPipeline"})

z = ad.AnnData(ad.io.sparse_dataset(zarr.open("tahoe_0.zarr", use_consolidated=False)["X"]))

class DummyDataset(IterableDataset):

    async def get(self):
        z_ = await (await zarr.AsyncGroup.open("tahoe_0.zarr", use_consolidated=False)).getitem("X")
        i = await z_.getitem("indptr")
        res = await i.getitem(Ellipsis)
        return res

    def __iter__(self):
        for i in range(32):
            yield zarr.core.sync.sync(self.get())

    def __len__(self):
        return 100

ds = DummyDataset()

dl = DataLoader(
    ds,
    batch_size=10,
    num_workers=16,
)
for idx, s in enumerate(dl):
    print(idx, s.shape)
```

I assume this has to do with some dangling reference from the `AnnData` object i.e., a rehash of #360 in a different form, but I am not sure.  By removing the `ad.AnnData` wrapper, the deadlock disappears.

Separately, using normal `zarr` instead of `zarrs` also makes the deadlock disappear, which makes this issue doubly wonderful.

I am running this in a jupyter notebook so it's possible the problem somehow lies there as well given that it has its own event loop.

Traceback:

```pytb
n/a
```

TODO: check if this happens in normal multiprocessing (i.e., is torch's shared memory business somehow at fault)

### Versions

```python
| Package | Version             |
| ------- | ------------------- |
| anndata | 0.12.0rc4           |
| torch   | 2.7.1 (2.7.1+cu126) |
| zarr    | 3.0.8               |
| zarrs   | 0.1.4               |

| Dependency             | Version     |
| ---------------------- | ----------- |
[...]

PR #2034, opened 2025-07-17
fix: no cache indptr `zarr` dask

ilan-gold:
Evidently the same problem that was present previously with only hdf5 is now present with zarr v3 as well.  Given that the performance is identical in zarr v2 with/without caching in dask, I think we should simply turn it off in general when using dask.  

I am wondering if this is related to #2021 or https://github.com/dask/dask/issues/11909.  I need to investigate more, but if zarr v3 is doing something unexpectedly with forking a process (which might force the `CSRDataset` class to bust its cache repeatedly, for example, in this case), that could account for these various issues.

But for now, I think this fix is fairly innocuous.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -132,7 +132,7 @@ def read_sparse_as_dask(
     path_or_sparse_dataset = (
         Path(filename(elem))
         if isinstance(elem, H5Group)
-        else ad.io.sparse_dataset(elem)
+        else ad.io.sparse_dataset(elem, should_cache_indptr=False)
     )
     elem_name = get_elem_name(elem)
     shape: tuple[int, int] = tuple(elem.attrs["shape"])


ilan-gold:
Here is the state of things as of this branch running:

```python
import anndata as ad
import dask.distributed as dd

cluster = dd.LocalCluster(n_workers=3)
client = dd.Client(cluster)

adata = ad.experimental.read_lazy("/Users/ilangold/sc_dataset.zarr")

%%timeit
adata.X[:adata.shape[0] // 2].compute() # only take second measurement so as not to include cold-cache
```

## This branch

All the green code is our code, and I can't even find a call to `indptr`

Screenshot 2025-07-23 at 11 31 15

and here are the timings:

```
2.62 s ± 210 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Main

The fetch for `indptr` clearly occupies a non-trivial amount of time

Screenshot 2025-07-23 at 11 34 53

and the timings reflect this:

```
2.25 s ± 71.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I also added a benchmark. Thanks for the push!

PR #2035, opened 2025-07-17
fix: integer indexing in `h5` is very slow for `ZarrOrHDF5Wrapper`

ilan-gold:
Hopefully resolves the issue Jan was seeing (which as it turns out, involved indexing with an integer array).  See the linked h5py issue for why indexing with an integer array is crushingly slow.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/benchmarks/benchmarks/dataset2d.py
+++ b/benchmarks/benchmarks/dataset2d.py
@@ -37,13 +37,13 @@ class Dataset2D:
         self.ds = ad.experimental.read_elem_lazy(store["obs"])
 
     def time_getitem_slice(self, *_):
-        self.ds.iloc[0 : (self.n_obs // 2)]
+        self.ds.iloc[0 : (self.n_obs // 2)].to_memory()
 
     def peakmem_getitem_slivce(self, *_):
-        self.ds.iloc[0 : (self.n_obs // 2)]
+        self.ds.iloc[0 : (self.n_obs // 2)].to_memory()
 
     def time_getitem_bool_mask(self, *_):
-        self.ds.iloc[np.random.randint(0, self.n_obs, self.n_obs // 2)]
+        self.ds.iloc[np.random.randint(0, self.n_obs, self.n_obs // 2)].to_memory()
 
     def peakmem_getitem_bool_mask(self, *_):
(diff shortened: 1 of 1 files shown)


ilan-gold:
> A local simple benchmark and an unit test to the error cases would be convincing enough for me

Our unit tests are already covering this change, hence why the coverage didn't go down (except for the errors, which can't happen given `xr.core.indexing.explicit_indexing_adapter`).

I can add a benchmark though, yeah!

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -46,12 +45,34 @@ def __init__(self, array: K):
...
+        res = xr.core.indexing.explicit_indexing_adapter(
             key,
             self.shape,
             xr.core.indexing.IndexingSupport.OUTER_1VECTOR,
-            lambda key: self._array[key],
+            self._getitem,
         )
+        return res
+
+    def _getitem(self, key: tuple):
```
flying-sheep: tuple of? return type?

flying-sheep (commented):
I took the liberty to add https://github.com/scverse/anndata/labels/benchmark so we can see the PR actually improving performance!

flying-sheep:
OK, weird. I’d expect that comment to only appear when the benchmark run is done, and if it says “Some benchmarks failed”, it should say which ones!

grafik

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: `tuple` was the best I could resolve from peeking at the internals of `xarray`.  I think there are other types like `slices` perhaps are possible.  This is how `xarray` types the object that produces this `tuple`: https://github.com/pydata/xarray/blob/fa03b5b4ae95a366f6de5b60f5cc4eb801cd51ec/xarray/core/indexing.py#L335-L337.  I could could add  https://github.com/pydata/xarray/blob/fa03b5b4ae95a366f6de5b60f5cc4eb801cd51ec/xarray/core/indexing.py#L423 to the `tuple` type but this is more of an educated guess than something that is actually typed by them properly

flying-sheep (approved):
awesome! also lol at the bug being from 2013!

PR #2048, opened 2025-07-23
(chore): Re-introduce strict warnings

flying-sheep:
No clue why the tests fail. Looks like an issue with some dependency in the min-vers job.

### DIFF
--- a/tests/test_structured_arrays.py
+++ b/tests/test_structured_arrays.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from itertools import combinations, product
 from typing import TYPE_CHECKING
 
@@ -7,6 +8,7 @@ import numpy as np
 
 import anndata as ad
 from anndata import AnnData
+from anndata.compat import is_zarr_v2
 from anndata.tests.helpers import gen_vstr_recarray
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


tests/test_io_elementwise.py
```diff
@@ -41,6 +41,16 @@
...
 
 
+pytestmark = [
+    pytest.mark.filterwarnings(
+        "default:.*Structured:zarr.core.dtype.common.UnstableSpecificationWarning"
+    ),
+    pytest.mark.filterwarnings(
+        "default:.*FixedLengthUTF32:zarr.core.dtype.common.UnstableSpecificationWarning"
+    ),
+]
```
ilan-gold: I would guess that this is somehow the source of the issue given the stack trace.  But why that is...

flying-sheep: ah thanks, seems like I missed it in the traceback!

flying-sheep:
@ilan-gold one failure left! Any idea where `NullTerminatedBytes` comes from and if that warning is expected or not?

> `zarr.core.dtype.common.UnstableSpecificationWarning`: The data type (`NullTerminatedBytes(length=10)`) does not have a Zarr V3 specification. That means that the representation of arrays saved with this data type may change without warning in a future version of Zarr Python. Arrays stored with this data type may be unreadable by other Zarr libraries. Use this data type at your own risk! Check https://github.com/zarr-developers/zarr-extensions/tree/main/data-types for the status of data type specifications for Zarr V3.

ilan-gold:
It means we're using https://github.com/zarr-developers/zarr-python/blob/a0c56fbbba12ca4402de3d83043e5f840406d670/src/zarr/core/dtype/npy/bytes.py#L166 but where exactly we are using it I don't know.  Maybe within a structured data type?

PR #2053, opened 2025-07-25
fix: unbound dask

ilan-gold:
~~Blocked by https://github.com/dask/dask/issues/12026~~

I think I need to be more careful with upper bounds...

All good now! The scipy PR I mention in the comment is important, but not something that is either core to our internal functionality or unworkable.  So I think this PR is good to go1

- [ ] Closes #
- [ ] Tests added


tests/test_dask_view_mem.py
```diff
@@ -70,7 +70,7 @@ def _alloc_cache():
 # the results seems more accurate with the experimental results
 # For example from dask.random we allocate 1mb
 @pytest.mark.usefixtures("_alloc_cache")
-@pytest.mark.limit_memory("1.5 MB")
+@pytest.mark.limit_memory("1.6 MB")
```
ilan-gold: I am not sure what happened here, but looking at the failure stack-trace, this isn't a hallucination i.e., the total bytes add up.  Why this changed, I don't know, but I don't know if it is worth jumping into this rabbit hole.

src/anndata/_core/merge.py
```diff
@@ -617,6 +617,8 @@ def _apply_to_dask_array(self, el: DaskArray, *, axis, fill_value=None):
         sub_el = _subset(el, make_slice(indexer, axis, len(shape)))
 
         if any(indexer == -1):
+            if isinstance(sub_el._meta, CSArray | CSMatrix) and np.isscalar(fill_value):
+                fill_value = np.array([[fill_value]])
```
flying-sheep, line 622: Does the TODO in the other file not apply here?

tests/test_concatenate.py
```diff
@@ -1811,7 +1814,7 @@ def test_concat_dask_sparse_matches_memory(join_type, merge_strategy):
     X = sparse.random(50, 20, density=0.5, format="csr")
     X_dask = da.from_array(X, chunks=(5, 20))
     var_names_1 = [f"gene_{i}" for i in range(20)]
-    var_names_2 = [f"gene_{i}{'_foo' if (i % 2) else ''}" for i in range(20, 40)]
+    var_names_2 = [f"gene_{i - 20}{'_foo' if (i % 2) else ''}" for i in range(20, 40)]
```
flying-sheep: So the intent of this test is to have partially matching `var_names`, but the test didn’t actually test that?

Also, why add and then subtract 20?

```suggestion
    var_names_2 = [f"gene_{i}{'_foo' if (i % 2) else ''}" for i in range(20)]
```

tests/test_views.py
```diff
@@ -170,6 +170,9 @@ def test_view_subset_shapes():
 
 
 def test_modify_view_component(matrix_type, mapping_name, request):
+    if "sparse_dask" in request.node.callspec.id:
+        msg = "sparse arrays in dask are generally expected to fail but in this case they do not"
+        pytest.xfail(msg)
```
flying-sheep: I’m very confused: `xfail` means:

> This is a feature that doesn’t work (yet or ever). We just ran test code, and **detected** that it failed, which is expected as the feature doesn’t work.

Where does this code run a test and detect that it fails? If you want to *mark* it as a test that will fail, use `pytest.mark.xfail`. If you want to xfail it in a branch in the test code, use `pytest.xfail` in a branch that checks some result value, not the test input.

flying-sheep (changes requested):
Some questions and an issue with the test. The code changes look good!

tests/test_concatenate.py
ilan-gold: Fair point! Yes that was the intent, but yours is cleaner

src/anndata/_core/merge.py
ilan-gold: Weirdly enough, this was actually giving a different error, although the reason is similar i.e., `scipy` expects very specific input.  In this case, the `fill_value` was being converted to a 1d array (?) and then there is a line within `dask` that does `scipy.csc_matrix(val)`, and then `scipy` was complaining it needs 2d input to initialize.  

I can submit a fix to `dask` for this, actually, now that I think of it.  But yes, different problem, similar weirdness.

tests/test_views.py
ilan-gold: So this one _was_ the same problem and should have been marked more clearly.  I copypasta-ed the message, but the `xfail` is right.  I'm changing it to a clear `raises` and have a `TODO` to remove it once the `scipy` PR is merged

src/anndata/_core/merge.py
ilan-gold: Fix submitted and comment added here

tests/test_views.py
```diff
@@ -189,13 +189,25 @@ def test_modify_view_component(matrix_type, mapping_name, request):
...
     assert not subset.is_view
-    assert getattr(subset, mapping_name)["m"][0, 0] == 100
+    # TODO: Remove `raises` after https://github.com/scipy/scipy/pull/23626.
+    import dask
 
-    assert init_hash == hash_func(adata)
+    is_dask_with_broken_view_setting = (
+        "sparse_dask" in request.node.callspec.id
+        and Version(dask.__version__) >= Version("2025.02.0")
+    )
```
flying-sheep, line 198: so to keep the `xfail` in there, you could do this, but the `raises` is also perfect!

```py
    # TODO: Remove `try`/`catch` after https://github.com/scipy/scipy/pull/23626.
    try:
        assert getattr(subset, mapping_name)["m"][0, 0] == 100
    except ValueError as e:
        if (is_dask_with_broken_view_setting or is_sparse_array_in_lower_dask_version) and "shape mismatch" in str(e):
            pytest.xfail("scipy bug")
        raise
```

ilan-gold: I think `raises` is cleaner since it lets the `assert init_hash == hash_func(adata)` evaluate, no?

flying-sheep: good point, that’s true!

PR #2054, opened 2025-07-28
(fix): skip X writing when X is None

### LINKED ISSUE #2052: Writing to H5AD has changed in v`0.12.1`
### Question

The behaviour when writing an `AnnData` with an empty `X` slot (possibly others?) seems to have changed in `0.12.1`. Previously, there was nothing written to the file:

```python
import anndata as ad
import h5py

adata = ad.AnnData(shape = [5, 10])
adata.write_h5ad("test.h5ad")

print(ad.__version__)
with h5py.File("test.h5ad", "r") as f:
    print(list(f.keys()))
    print(f["X"].attrs["encoding-type"])

# 0.11.4
# ['layers', 'obs', 'obsm', 'obsp', 'uns', 'var', 'varm', 'varp']
# KeyError: "Unable to synchronously open object (object 'X' doesn't exist)"
```

With `0.12.1` the group is included in the file but with a `"null"` encoding:

```python
# 0.12.1
# ['X', 'layers', 'obs', 'obsm', 'obsp', 'uns', 'var', 'varm', 'varp']
# null
```

The same is true if you have a `None` value in `uns`.

Trying to read a H5AD written with `0.12.1` with an earlier version gives this error:

```python
IORegistryError: No read method registered for IOSpec(encoding_type='null', encoding_version='0.1.0') from <class 'h5py._hl.dataset.Dataset'>. You may need to update your installation of anndata.
Error raised while reading key 'X' of <class 'h5py._hl.dataset.Dataset'> from /
```

Another change is that 1D arrays are now written with dimensions `[nx1]` instead of just `[n]`. This changes what you read from disk (I think this might depend on the version used for reading rather than writing).

Are these changes intended to continue going forward? If so, should the on disk spec (and versions) be updated to reflect this?

flying-sheep:
- [x] Closes #2052
- [x] Tests added

The removal of the `str` stuff is unrelated, I just didn’t want to step over it anymore without need (we don’t support the zarr version anymore that’s mentioned in the comment that addresses the `str` conversion)


src/anndata/_io/h5ad.py
```diff
@@ -115,6 +102,38 @@ def write_h5ad(
...
+    f: h5py.Group,
+    x: XDataType,
+    *,
+    as_dense: Container[str],
+    dataset_kwargs: Mapping[str, Any],
+) -> None:
+    if "X" in as_dense and isinstance(x, CSMatrix | BaseCompressedSparseDataset):
+        write_sparse_as_dense(f, "X", x, dataset_kwargs=dataset_kwargs)
+    elif x is None:
+        f.pop("X", None)
```
flying-sheep, line 121: adding this `pop` branch make Ruff complain about too high complexity, so I extracted `_write_x` and `_write_raw`.

The logic should otherwise be the same for raw, and almost the same for X:
1. check if X is to be written sparse-to-dense, if yes, do it, else
2. check if X is backed and the file path matches, if yes we’re done with X, else
3. (new) if X is None, remove `"X"` from the HDF5 group if it’s there, else
4. write X normally to the file

flying-sheep:
minimal AnnData files are byte-identical between 0.11.4 and this branch:

```console
$ uvx --with=anndata==0.11.4 python -c 'from anndata import AnnData; adata = AnnData(shape=(10, 20)); adata.write_h5ad("0.11.h5py")'
$ hatch -e hatch-test.stable run python -c 'from anndata import read_h5ad; read_h5ad("0.11.h5py").write_h5ad("0.12.h5py")'
$ md5sum 0.11.h5py 0.12.h5py 
a8cdc9539050b6523f529073636733cc  0.11.h5py
a8cdc9539050b6523f529073636733cc  0.12.h5py
```

maximalist anndata (with X, so regardless of this PR) isn’t:

```console
$ uvx --with=anndata[test]==0.11.4 --with=numba>=0.61 python -c 'from anndata.tests.helpers import gen_adata; adata = gen_adata((10, 20)); adata.write_h5ad("0.11.h5py")'
$ hatch -e hatch-test.stable run python -c 'from anndata import read_h5ad; read_h5ad("0.11.h5py").write_h5ad("0.12.h5py")'
$ md5sum 0.11.h5py 0.12.h5py 
74cbadb983b65dccd116c2342abdde30  0.11.h5py
a33f2f9244501426ca941a9b3299245e  0.12.h5py
```

flying-sheep:
The HDF5 diff test works locally, I’m requesting your review now, and if it fails on CI, I’ll of course install it somehow and I’ll make sure it runs before merging.

PR #2062, opened 2025-07-29
fix: allow categorical index in `read_lazy`

ilan-gold:
Fixed a bug reported by Jan.  We do have a warning about setting non-string indices, but it is not disallowed, in non-lazy `AnnData`.   So this fix matches the  behavior for `read_lazy`.
- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/docs/release-notes/2062.bugfix.md
+++ b/docs/release-notes/2062.bugfix.md
@@ -1,0 +1,1 @@
+Allow reading of non-string index in {func}`anndata.experimental.read_lazy` and {func}`anndata.experimental.read_elem_lazy` for {class}`anndata.experimental.backed.Dataset2D` {user}`ilan-gold`
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from contextlib import contextmanager
 from functools import partial, singledispatch
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


flying-sheep (commented):
>  We do have a warning about setting non-string indices, but it is not disallowed, in non-lazy AnnData

which isn’t a great thing. do we raise an error when writing that or …?

As said: `UUIDIndex` or (barring that) `AnonymousIndex` makes sense, `RangeIndex` or `Index(dtype=np.int*)` don’t.

ilan-gold:
> which isn’t a great thing. do we raise an error when writing that or …?

We have a warning about it:

https://github.com/scverse/anndata/blob/c72dfff5295e293a8d7c8e0b5d22de572fc72797/src/anndata/_core/anndata.py#L785-L800

I think this fix falls under the "people can in theory write it to disk, so we should read it" category.  If we want to start disallowing this behavior in some scope (writing? in-memory? both?), I think that's a separate conversation.

flying-sheep (changes requested):
I see, one more instance of “some old stuff where we weren’t strict enough”. This is some old stuff from #300.

I say we shouldn’t finish enshrining this as a feature. We say we don’t support numeric indices, so we shouldn’t betray that by allowing more and more support to creep in. 

Instead we should do something like converting to string on read (for old wrong-written indices) and write (to no longer write wrong indices) – with a warning each. Something that keeps things from breaking but doesn’t allow an unsupported index type to survive a full write-read cycle.

PR #2066, opened 2025-07-30
fix: fancy indexing fixes backed h5py error

### LINKED ISSUE #2064: duplicated indices when slicing dense backed view lead to .to_memory() TypeError

- [ ] (optional) I have confirmed this bug exists on the main branch of anndata.

### Report

Thanks for the great package!

I encountered something unexpected and I think it may be considered a bug.  If not a bug, then an "enhancement request"        :)

Here is what I observe in words:
I want to slice an anndata using indices that may not be in strictly increasing numerical order.  This seems to work in almost all cases, but I found a strange edge case where an error is thrown and I think it may be a bug.  To recreate the bug, you need to save an anndata with a dense X (not CSR).  Then you need to load the file in backed mode.  Then you need to slice it (in some order other than strictly numerically increasing) and call `.to_memory()`.

Here is an example that demonstrates the bug.

Code:

```python
import anndata
import numpy as np

ad = anndata.AnnData(X=np.ones((5, 10)))
ad.write('tmp.h5ad')
ad_backed = anndata.read_h5ad('tmp.h5ad', backed='r')
# ad_backed.X is <HDF5 dataset "X": shape (5, 10), type "<f8">
ad_backed[np.array([0, 2, 2])].to_memory()
```

Traceback:

```pytb
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/anndata/_core/anndata.py", line 1414, in to_memory
    attr = getattr(self, attr_name, None)
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/anndata/_core/anndata.py", line 556, in X
    X = _subset(X, (self._oidx, self._vidx))
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/anndata/_core/index.py", line 209, in _subset_dataset
    return d[tuple(ordered)][tuple(rev_order)]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/h5py/_hl/dataset.py", line 831, in __getitem__
    selection = sel.select(self.shape, args, dataset=self)
  File "/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/h5py/_hl/selections.py", line 82, in select
    return selector.make_selection(args)
  File "h5py/_selector.pyx", line 282, in h5py._selector.Selector.make_selection
  File "h5py/_selector.pyx", line 215, in h5py._selector.Selector.apply_args
TypeError: Indexing elements must be in increasing order
```

Lots of other combinations work as expected.  For example, if the X is a CSR, then this works fine.  It also all works in non-backed mode.  Example with a sparse X:

Code:

```python
import anndata
import numpy as np
import scipy.sparse as sp

ad = anndata.AnnData(X=sp.csr_matrix(np.ones((5, 10))))
ad.write('tmp2.h5ad')
ad_backed = anndata.read_h5ad('tmp2.h5ad', backed='r')
# ad_backed.X is CSRDataset: backend hdf5, shape (5, 10), data_dtype float64
ad_backed[np.array([0, 2, 2])].to_memory()
```

Output:

```pytb
/Users/sfleming/miniconda3/envs/cellarium/lib/python3.10/site-packages/anndata/_core/anndata.py:1756: UserWarning: Observation names are not unique. To make them unique, call `.obs_names_make_unique`.
  utils.warn_names_duplicates("obs")
AnnData object with n_obs × n_vars = 3 × 10
```

### Versions

```python
| Package | Version |
| ------- | ------- |
| anndata | 0.12.1  |

| Dependency                    | Version              |
| ----------------------------- | -------------------- |
| sphinxcontrib-devhelp         | 2.0.0                |
| h5py                          | 3.13.0               |
| ipython                       | 9.0.0                |
| prompt_toolkit                | 3.0.50               |
| stack-data                    | 0.6.3                |
| wrapt                         | 1.17.2               |
| hatchling                     | 1.27.0               |
| numcodecs                     | 0.15.1               |
| Deprecated                    | 1.2.18               |
| sphinxcontrib-htmlhelp        | 2.1.0                |
| msgpack                       | 1.1.0                |
| donfig                        | 0.8.1.post1          |
| six                           | 1.16.0               |
| PyYAML                        | 6.0.1                |
| sphinxcontrib-applehelp       | 2.0.0                |
| regex                         | 2023.12.25 (2.5.140) |
| setuptools                    | 75.8.0               |
| comm                          | 0.2.3                |
| awkward_cpp                   | 44                   |
| typing_extensions             | 4.12.2               |
| zipp                          | 3.17.0               |
| awkward                       | 2.7.4                |
| pure_eval                     | 0.2.3                |
| legacy-api-wrap               | 1.4.1                |
| python-dateutil               | 2.9.0.post0          |
| sphinxcontrib-qthelp          | 2.0.0                |
| MarkupSafe                    | 3.0.2                |
| pandas                        | 2.2.3                |
| charset-normalizer            | 3.4.1                |
| numpy                         | 2.2.3                |
| fsspec                        | 2025.2.0             |
| scipy                         | 1.15.2               |
| pytz                          | 2025.1               |
| jedi                          | 0.19.2               |
| traitlets                     | 5.14.3               |
| natsort                       | 8.4.0                |
| asttokens                     | 3.0.0                |
| executing                     | 2.2.0                |
| sphinxcontrib-serializinghtml | 2.0.0                |
| session-info2                 | 0.2                  |
| wcwidth                       | 0.2.13               |
| ipywidgets                    | 8.1.7                |
| crc32c                        | 2.7.1                |
| zarr                          | 3.1.1                |
| pluggy                        | 1.5.0                |
| sphinx-jinja2-compat          | 0.3.0                |
| Pygments                      | 2.19.1               |
| sphinxcontrib-jsmath          | 1.0.1                |
| parso                         | 0.8.4                |
| Jinja2                        | 3.1.5                |
| decorator                     | 5.2.1                |
| packaging                     | 24.2                 |
| importlib-metadata            | 7.0.1                |

| Component | Info                                                  |
| --------- | ----------------------------------------------------- |
| Python    | 3.11.11 (main, Dec 11 2024, 10:25:04) [Clang 14.0.6 ] |
| OS        | macOS-15.5-arm64-arm-64bit                            |
| Updated   | 2025-07-30 14:36                                      |
```

sjfleming:
- [x] Closes #2064 
- [x] Tests added

This is a first stab at fixing #2064 by adding `_safe_fancy_index_h5py` (and 3 related helper functions) to `anndata/_core/index.py`.  The function `_safe_fancy_index_h5py` only gets called in the case where there are repeated indices being requested (this is the only case that is currently causing a bug, so in all other cases, the existing code -- `d[tuple(ordered)][tuple(rev_order)]` -- is what runs).


src/anndata/_core/index.py
```diff
@@ -224,10 +224,91 @@ def _subset_dataset(d, subset_idx):
...
             rev_order[axis] = np.argsort(order)
+    # check for duplicates or multi-dimensional fancy indexing
+    has_duplicates = False
+    num_array_dims = 0
+    for axis_idx in ordered:
+        if isinstance(axis_idx, np.ndarray):
+            num_array_dims += 1
+            if len(np.unique(axis_idx)) != len(axis_idx):
+                has_duplicates = True
+                break
```
flying-sheep: I’d simplify:

```py
array_dims = [i for i in ordered if isinstance(i, np.ndarray)]
has_duplicates = any(len(np.unique(i)) != len(i) for i in array_dims)
```

and then later use `len(array_dims)`

src/anndata/_core/index.py
```diff
@@ -217,10 +217,91 @@ def _subset_dataset(d: h5py.Dataset, subset_idx: Index):
...
+        slices[axis] = idx
+        return result[tuple(slices)]
+
+
+def _safe_fancy_index_h5py(dataset, subset_idx):
+    # Handle multi-dimensional indexing while being memory-efficient
+    # This avoids h5py's limitation with multi-dimensional fancy indexing
+    # without loading the entire dataset into memory
+    if not isinstance(subset_idx, tuple):
+        subset_idx = (subset_idx,)
```
flying-sheep: could you please add types? I don‘t think it’s possible for `subset_idx` not to be a tuple at this point, so I think this `if not` block can be removed

src/anndata/_core/index.py
```diff
@@ -217,10 +217,91 @@ def _subset_dataset(d: h5py.Dataset, subset_idx: Index):
...
+
+    # Convert boolean arrays to integer arrays and handle sorting for h5py
+    processed_indices = []
+    reverse_indices = []
+
+    for idx in subset_idx:
+        processed_idx, reverse_idx = _process_index_for_h5py(idx)
+        processed_indices.append(processed_idx)
+        reverse_indices.append(reverse_idx)
```
flying-sheep: this can just be the following, I think:

```suggestion
    processed_indices, reverse_indices = zip(*map(_process_index_for_h5py, subset_idx))
```

src/anndata/_core/index.py
```diff
@@ -217,10 +217,91 @@ def _subset_dataset(d: h5py.Dataset, subset_idx: Index):
...
+    if len(np.unique(idx)) != len(idx):
+        # Has duplicates - use unique + inverse mapping approach
+        unique_idx, inverse_map = np.unique(idx, return_inverse=True)
+        return unique_idx, inverse_map
+    else:
+        # No duplicates - just sort and track reverse mapping
+        sort_order = np.argsort(idx)
+        sorted_idx = idx[sort_order]
+        reverse_order = np.argsort(sort_order)
+        return sorted_idx, reverse_order
```
flying-sheep: This block already exists in `_subset_dataset`. Can you deduplicate, i.e. reuse the previous results if it’s run on the same data, or introduce a helper function and call it in both spots if it isn’t?

tests/test_backed_hdf5.py
```diff
@@ -355,3 +346,145 @@ def test_backed_modification_sparse(adata, backing_h5ad, sparse_format):
...
+
+    # Load backed data
+    backed_adata = ad.read_h5ad(backed_pth, backed="r")
+
+    # Test simple multi-dimensional fancy indexing
+    obs_idx = np.array([0, 1, 2])  # no duplicates first
+    var_idx = np.array([1, 2])  # no duplicates first
+    mem_result_multi = mem_adata[obs_idx, var_idx]
+    backed_result_multi = backed_adata[obs_idx, var_idx]
+    assert_equal(mem_result_multi, backed_result_multi)
```
flying-sheep: each of these three blocks should be an entry in `pytest.mark.parametrize` instead of being copy&pasted:


```py
@pytest.mark.parametrize(
    ("obs_idx", "var_idx"),
    [
        pytest.param(np.array([0, 1, 2]), np.array([1, 2]), id="no_dupes"),
        pytest.param(np.array([0, 1, 0, 2]), slice(None), id="1d_dupes"),
        pytest.param(np.array([0, 1, 0, 2]), np.array([1, 2, 1]), id="2d_dupes"),
    ]
)
```

tests/test_backed_hdf5.py
```diff
@@ -355,3 +346,145 @@ def test_backed_modification_sparse(adata, backing_h5ad, sparse_format):
...
+        "single_dim_duplicates",
+        "multi_dim_no_duplicates",
+        "multi_dim_dup_first",
+        "multi_dim_dup_second",
+        "multi_dim_dup_both",
+        "boolean_arrays",
+        "mixed_slice_array",
+        "mixed_slice_array_list",
+        "unsorted_duplicates",
+    ],
```
flying-sheep: You don’t use `test_case_name` at all, and `description` is redundant. Please put the IDs inline instead, e.g.:

```py
pytest.param(
    np.array([0, 1, 0, 2]),
    id="single_dimension_with_duplicates",
),
```

and no `ids=` parameter here.

```suggestion
```

this makes everything shorter without losing a significant amount of information.

tests/test_backed_hdf5.py
```diff
@@ -355,3 +346,145 @@ def test_backed_modification_sparse(adata, backing_h5ad, sparse_format):
...
+            else:
+                expected = test_data[np.ix_(*indices)]
+        else:
+            # Single dimensional case
+            expected = test_data[indices]
+
+        # Assert arrays are equal
+        np.testing.assert_array_equal(
+            result, expected, err_msg=f"Failed for test case: {description}"
+        )
```
flying-sheep: `_safe_fancy_index_h5py` returns an `np.ndarray`, right? so you can probably dedent all of this as there are no more reads from the file after this?

src/anndata/_core/index.py
sjfleming: Thanks

src/anndata/_core/index.py
sjfleming: Yes

src/anndata/_core/index.py
sjfleming: Looks good

tests/test_backed_hdf5.py
sjfleming: Very clean, I like it!

tests/test_backed_hdf5.py
sjfleming: You are correct

src/anndata/_core/index.py
sjfleming: Used a helper function

sjfleming:
I took a shot at it.  Let me know what you think @flying-sheep !  Thanks

src/anndata/_core/index.py
```diff
@@ -180,48 +184,134 @@ def _subset_dask(a: DaskArray, subset_idx: Index):
...
+        *map(_process_index_for_h5py, subset_idx), strict=True
+    )
+
+    # Apply first dimension indexing
+    first_idx = processed_indices[0]
+    result = dataset[first_idx]
+
+    # Apply remaining dimensions
+    for dim_offset, idx in enumerate(processed_indices[1:], 1):
+        result = _apply_index_to_result(result, idx, dim_offset)
```
flying-sheep: why is the first one special and is directly applied rather than with `_apply_index_to_result`?

If there is no good reason, would this work? If not, please change the comment so it explains why it’s done this way.

```suggestion
    # Apply dimension indexing
    result = dataset
    for dim_offset, idx in enumerate(processed_indices):
        result = _apply_index_to_result(result, idx, dim_offset)
```

src/anndata/_core/index.py
```diff
@@ -180,48 +184,133 @@ def _subset_dask(a: DaskArray, subset_idx: Index):
...
+        else _index_order_and_inverse(idx)
+    )
+
+
+def _apply_index_to_result(
+    result: h5py.Dataset, idx: np.ndarray, axis: int
+) -> h5py.Dataset:
+    """Apply an index to a result array along a specific axis."""
+    if isinstance(idx, np.ndarray):
+        return result.take(idx, axis=axis)
```
flying-sheep: when I run this locally, I get

```pytb
AttributeError: 'Dataset' object has no attribute 'take'
```

In which version of h5py was this added/removed? Do we only support h5py versions where this exist?

src/anndata/_core/index.py
sjfleming: Due to my previously incorrect type-hint for `result`, this was confusing.  The reason is that `dataset` is an h5py dataset, while `result` is a numpy array, so that first dimension indexing allows us to pass a numpy array to the function.

src/anndata/_core/index.py
sjfleming: Sorry, looks like I type-hinted `result` (and the output) incorrectly.  They are both numpy arrays.  So `.take()` is from numpy.  When `_apply_index_to_result` gets called, `result` is a numpy array.

sjfleming:
Thanks @flying-sheep !  Yes I think your type changes make sense.  Please check and see whether my comments and changes address your concerns (which identified a problem with my previous type-hints for `_apply_index_to_result`)

flying-sheep:
I see, so actually it should just be 

```py
result = cast("np.ndarray", dataset[processed_indices[0]])
result = result[:, *processed_indices[1:]]
```

right?

And if the first index is `:`, this *will* load the entire dataset into memory.

I wonder what the best approach is then. Maybe finding out which 1D slice operation reduces the data the most, then apply that first?

sjfleming:
Hi @flying-sheep thanks for the simplification, much better.  I have again made it slightly more complicated by first slicing with the indexer that will make the dataset the smallest (as you suggested)... let me know what you think.

flying-sheep:
looks good to me! maybe @ilan-gold should have a look in case I missed something

src/anndata/_core/index.py
```diff
@@ -180,48 +187,153 @@ def _subset_dask(a: DaskArray, subset_idx: Index):
...
+def _index_order_and_inverse(
+    axis_idx: Index1DNorm,
+) -> tuple[Index1DNorm, NDArray[np.integer] | slice]:
+    """Order and get inverse index array."""
+    if not isinstance(axis_idx, np.ndarray):
+        return axis_idx, slice(None)
+    if axis_idx.dtype == bool:
+        axis_idx = np.flatnonzero(axis_idx)
+    order = np.argsort(axis_idx)
+    return axis_idx[order], np.argsort(order)
```
ilan-gold, line 256:
```suggestion
    return axis_idx[order], np.arange(len(order))
```
isn't `order` already sorted so `argsort` would just do an `arange`?

src/anndata/_core/index.py
```diff
@@ -180,48 +187,153 @@ def _subset_dask(a: DaskArray, subset_idx: Index):
...
+
+    # For h5py fancy indexing, we need sorted indices
+    # But we also need to track how to reverse the sorting
+    return (
+        # Has duplicates - use unique + inverse mapping approach
+        np.unique(idx, return_inverse=True)
+        if len(np.unique(idx)) != len(idx)
+        # No duplicates - just sort and track reverse mapping
+        else _index_order_and_inverse(idx)
+    )
```
ilan-gold, line 285:
```suggestion
    unique, inverse = np.unique(idx, return_inverse=True)
    return (
        # Has duplicates - use unique + inverse mapping approach
        unique, inverse
        if len(unique) != len(idx)
        # No duplicates - just sort and track reverse mapping
        else _index_order_and_inverse(idx)
    )
```

sjfleming: Yes, you are correct here that there is no reason to call `np.unique()` twice.  I implemented this suggestion.

src/anndata/_core/index.py
sjfleming: `order` is not already sorted here, it's just the index order that sorts `axis_idx`.  I did try the above suggested change, but the result was that tests would no longer pass.

sjfleming:
@ilan-gold I think I see what you mean about checking for duplicates twice.  I guess my thinking was that the `_safe_fancy_index_h5py` function should check for duplicates on its own in case it's ever called by another function for another purpose in the future.  And then I wanted to modify the existing functionality as little as possible, so I wanted `_subset_dataset` to only call any of the new code if it needed to... so I decided to check for duplicates in there as well.  It could be handled differently though.

ilan-gold:
@sjfleming Would you be up for adding a benchmark as well?  It's in our benchmarks folder; I'm just a little concerned about all the `np.unique` and `np.argsort` calls that are now run by default (I count 6).  I think at the moment, you could only add something for non-duplicated indexing but even there it would be helpful given the new default behavior.

Aside from that my only comment would be you could potentially add a flag to `_safe_fancy_index_h5py` to indicate that its results have already been checked.

sjfleming:
Hi @ilan-gold , I will look into making a benchmark.  You mean creating a new benchmark to cover these changes, right?  I didn't see an existing benchmark that looked like it would cover this use.

So I think that, the way this PR is written, if the requested indices have no repeats, then `_subset_dataset` performs (one or two) additional call(s) to `np.unique` here
https://github.com/sjfleming/anndata/blob/bdab9001bd096a63b4a6129e02b36af23d47046d/src/anndata/_core/index.py#L231

but if there are no repeats (which is probably almost always the case) the code skips the new `_safe_fancy_index_h5py` function altogether and just returns what it used to return:
https://github.com/sjfleming/anndata/blob/bdab9001bd096a63b4a6129e02b36af23d47046d/src/anndata/_core/index.py#L238

sjfleming:
Currently this seems like a major problem, so I'm trying to track down what's going on

```
| Change   | Before [c9c0be49] <main>   | After [2d5efbc3] <sf-backed-hdf5-fancy-indexing>   | Ratio   | Benchmark (Parameter)                                                                               |
|----------|----------------------------|----------------------------------------------------|---------|-----------------------------------------------------------------------------------------------------|
|          | 8.015625                   | 8.296875                                           | 1.04    | readwrite.H5ADWriteSuite.track_peakmem_write_compressed('pbmc3k')                                   |
| +        | 0.96875                    | 9.359375                                           | 9.66    | readwrite.H5ADWriteSuite.track_peakmem_write_full('pbmc3k')                                         |
```

sjfleming:
There seems to be a lot of run-to-run variability in these benchmarks (I assume just the ones related to peak memory).  Is this known?  (Is it just my laptop?)

```
(anndata) [sfleming@laptop:~/Documents/Github/anndata/benchmarks]$ asv run --python same --bench readwrite.H5ADWriteSuite.track_peakmem_write_full
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_Users_sfleming_miniconda3_envs_anndata_bin_python
[100.00%] ··· readwrite.H5ADWriteSuite.track_peakmem_write_full                                                                                                                                                   ok
[100.00%] ··· ============ ==========
               input_data            
              ------------ ----------
                 pbmc3k     6.296875 
              ============ ==========

(anndata) [sfleming@laptop:~/Documents/Github/anndata/benchmarks]$ asv run --python same --bench readwrite.H5ADWriteSuite.track_peakmem_write_full
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_Users_sfleming_miniconda3_envs_anndata_bin_python
[100.00%] ··· readwrite.H5ADWriteSuite.track_peakmem_write_full                                                                                                                                                   ok
[100.00%] ··· ============ ==========
               input_data            
              ------------ ----------
                 pbmc3k     15.15625 
              ============ ==========

(anndata) [sfleming@laptop:~/Documents/Github/anndata/benchmarks]$ asv run --python same --bench readwrite.H5ADWriteSuite.track_peakmem_write_full
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_Users_sfleming_miniconda3_envs_anndata_bin_python
[100.00%] ··· readwrite.H5ADWriteSuite.track_peakmem_write_full                                                                                                                                                   ok
[100.00%] ··· ============ ==========
               input_data            
              ------------ ----------
                 pbmc3k     4.234375 
              ============ ==========

(anndata) [sfleming@laptop:~/Documents/Github/anndata/benchmarks]$ asv run --python same --bench readwrite.H5ADWriteSuite.track_peakmem_write_full
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_Users_sfleming_miniconda3_envs_anndata_bin_python
[100.00%] ··· readwrite.H5ADWriteSuite.track_peakmem_write_full                                                                                                                                                   ok
[100.00%] ··· ============ ==========
               input_data            
              ------------ ----------
                 pbmc3k     2.921875 
              ============ ==========
```

sjfleming:
It seems possible that my first run just got very very unlucky, as unlikely as that sounds.

I may add some retry-averaging to the peak memory benchmarks if that is okay with you...

ilan-gold:
@sjfleming I'm not sure about the memory issue - we run these benchmarks semi-frequently and I haven't seen that much variability.  Thanks a million for these benchmarks by the way!

I would say the biggest concern with benchmarks (apologies for not clarifying this) was tanking performance for what was a previously valid indexing operation (sorted, no repeats).  As long as that is fine, or at least the penalty minimized, that would make me happy :)

ilan-gold:
@sjfleming So the benchmarks literally crashed here, not sure why!

Since the becnhmark took 174 minutes to run, could you either (a) enusre things run locally first but not in 174 minutes or (b) let me know that you want to use the CI for sanity checks and then ping me when you want the "real" benchmarks run on our machine? Then I'll remove the label and put it back when ready.  Thanks a bunch!

UPDATE: https://github.com/scverse/anndata/pull/2121/checks?check_run_id=53271517569 wow ok something is up with the machine, I think, different benchmarks are also all of a sudden taking forever

sjfleming:
@ilan-gold okay so the benchmarks I messed with in my attempt to "improve stability" [9734678](https://github.com/scverse/anndata/pull/2066/commits/973467864a06cb19c24220146bf5058e66d4332f) are now failing to return a value, so that was ill-conceived.  (It worked locally...)
Those are the ones with prefix `readwrite.....track_peakmem...` that now have all zeros.  I will revert this change.  It seems like I was misunderstanding how that memory calc worked: when I run on my laptop, I guess I'm seeing memory effects from a bunch of other processes.  When it runs on a dedicated machine in CI, it's probably fine.

There are new benchmarks I've written in `benchmarks/benchmarks/backed_hdf5.py` (`BackedHDF5.....timed_fancy_index...`) that will fail on `main` without this PR, because they cover new functionality.
Maybe it's a bad idea to include things like that?  Not sure what you think.

I can run the full benchmarking suite locally with `asv run --python same`. The benchmarks I've added take about 22 mins to run. The full run of all benchmarks (including those new ones) completes in 25 mins and everything looks okay.  Things also look good compared to `main` if I run on main and then do `asv compare`.

But it looks like I've added too much to the benchmarking.  I'm going to cut down on the number of new ones so as not to bog down the benchmarking suite.  I think I went way overboard.

sjfleming:
I've pared down the new benchmarks so that they run in 2 additional minutes.

sjfleming:
I've run benchmarking on my local machine using
```
asv run main --steps 1
asv run 655c3ba --steps 1
asv compare main 655c3ba
```
(`main` was commit 2fe7b396)
and the output looks alright to me (barring my local machine's issues measuring peak memory):
```
All benchmarks:

| Change   | Before [2fe7b396] <main>   | After [655c3ba9] <sf-backed-hdf5-fancy-indexing>   |   Ratio | Benchmark (Parameter)                                                                               |
|----------|----------------------------|----------------------------------------------------|---------|-----------------------------------------------------------------------------------------------------|
|          | 1872800.0                  | 1873429.0                                          |    1    | anndata.GarbargeCollectionSuite.track_peakmem_garbage_collection                                    |
|          | 703M                       | 706M                                               |    1.01 | backed_hdf5.BackedHDF5.peakmem_fancy_index_no_dupes                                                 |
| -        | 703M                       | 592M                                               |    0.84 | backed_hdf5.BackedHDF5.peakmem_index_with_dupes_obs                                                 |
|          | 705M                       | 669M                                               |    0.95 | backed_hdf5.BackedHDF5.peakmem_slice_obs                                                            |
|          | 679M                       | 661M                                               |    0.97 | backed_hdf5.BackedHDF5.peakmem_to_memory_subset                                                     |
|          | 273±2μs                    | 272±5μs                                            |    1    | backed_hdf5.BackedHDF5.time_fancy_index_no_dupes                                                    |
|          | 9.83±0.06ms                | 10.3±0.2ms                                         |    1.05 | backed_hdf5.BackedHDF5.time_fancy_index_no_dupes_to_memory                                          |
|          | 239±2μs                    | 244±7μs                                            |    1.02 | backed_hdf5.BackedHDF5.time_index_with_dupes_obs                                                    |
|          | 337±10μs                   | 327±8μs                                            |    0.97 | backed_hdf5.BackedHDF5.time_slice_obs                                                               |
|          | 4.72±0.07ms                | 4.90±0.07ms                                        |    1.04 | backed_hdf5.BackedHDF5.time_slice_obs_to_memory                                                     |
|          | 3.98±0.07ms                | 4.38±0.5ms                                         |    1.1  | backed_hdf5.BackedHDF5.time_to_memory_subset                                                        |
|          | 233M                       | 233M                                               |    1    | dataset2d.Dataset2D.peakmem_full_to_memory(<function Dataset2D.<lambda>> (0), (-1,))                |
|          | 233M                       | 228M                                               |    0.98 | dataset2d.Dataset2D.peakmem_full_to_memory(<function Dataset2D.<lambda>> (0), None)                 |
|          | 244M                       | 247M                                               |    1.01 | dataset2d.Dataset2D.peakmem_full_to_memory(<function Dataset2D.<lambda>> (1), (-1,))                |
|          | 246M                       | 246M                                               |    1    | dataset2d.Dataset2D.peakmem_full_to_memory(<function Dataset2D.<lambda>> (1), None)                 |
|          | 231M                       | 234M                                               |    1.01 | dataset2d.Dataset2D.peakmem_getitem_bool_mask(<function Dataset2D.<lambda>> (0), (-1,))             |
|          | 260M                       | 252M                                               |    0.97 | dataset2d.Dataset2D.peakmem_getitem_bool_mask(<function Dataset2D.<lambda>> (0), None)              |
|          | 248M                       | 243M                                               |    0.98 | dataset2d.Dataset2D.peakmem_getitem_bool_mask(<function Dataset2D.<lambda>> (1), (-1,))             |
|          | 243M                       | 243M                                               |    1    | dataset2d.Dataset2D.peakmem_getitem_bool_mask(<function Dataset2D.<lambda>> (1), None)              |
|          | 229M                       | 233M                                               |    1.02 | dataset2d.Dataset2D.peakmem_getitem_slice(<function Dataset2D.<lambda>> (0), (-1,))                 |
|          | 228M                       | 234M                                               |    1.03 | dataset2d.Dataset2D.peakmem_getitem_slice(<function Dataset2D.<lambda>> (0), None)                  |
|          | 227M                       | 248M                                               |    1.09 | dataset2d.Dataset2D.peakmem_getitem_slice(<function Dataset2D.<lambda>> (1), (-1,))                 |
|          | 231M                       | 229M                                               |    0.99 | dataset2d.Dataset2D.peakmem_getitem_slice(<function Dataset2D.<lambda>> (1), None)                  |
|          | 5.33±0.3ms                 | 5.36±0.08ms                                        |    1    | dataset2d.Dataset2D.time_full_to_memory(<function Dataset2D.<lambda>> (0), (-1,))                   |
|          | 28.5±0.3ms                 | 28.6±0.5ms                                         |    1    | dataset2d.Dataset2D.time_full_to_memory(<function Dataset2D.<lambda>> (0), None)                    |
|          | 8.51±0.2ms                 | 8.38±0.4ms                                         |    0.98 | dataset2d.Dataset2D.time_full_to_memory(<function Dataset2D.<lambda>> (1), (-1,))                   |
|          | 8.65±0.2ms                 | 8.79±0.06ms                                        |    1.02 | dataset2d.Dataset2D.time_full_to_memory(<function Dataset2D.<lambda>> (1), None)                    |
|          | 50.2±0.8ms                 | 49.6±0.2ms                                         |    0.99 | dataset2d.Dataset2D.time_getitem_bool_mask(<function Dataset2D.<lambda>> (0), (-1,))                |
|          | 542±20ms                   | 528±20ms                                           |    0.97 | dataset2d.Dataset2D.time_getitem_bool_mask(<function Dataset2D.<lambda>> (0), None)                 |
|          | 20.6±0.2ms                 | 22.3±2ms                                           |    1.08 | dataset2d.Dataset2D.time_getitem_bool_mask(<function Dataset2D.<lambda>> (1), (-1,))                |
|          | 19.3±0.3ms                 | 20.5±0.5ms                                         |    1.06 | dataset2d.Dataset2D.time_getitem_bool_mask(<function Dataset2D.<lambda>> (1), None)                 |
|          | 3.70±0.2ms                 | 3.81±0.1ms                                         |    1.03 | dataset2d.Dataset2D.time_getitem_slice(<function Dataset2D.<lambda>> (0), (-1,))                    |
|          | 20.4±0.7ms                 | 20.9±1ms                                           |    1.03 | dataset2d.Dataset2D.time_getitem_slice(<function Dataset2D.<lambda>> (0), None)                     |
|          | 6.23±0.2ms                 | 6.13±0.2ms                                         |    0.98 | dataset2d.Dataset2D.time_getitem_slice(<function Dataset2D.<lambda>> (1), (-1,))                    |
| +        | 6.33±0.09ms                | 7.12±0.3ms                                         |    1.12 | dataset2d.Dataset2D.time_getitem_slice(<function Dataset2D.<lambda>> (1), None)                     |
|          | 222M                       | 223M                                               |    1.01 | readwrite.H5ADBackedWriteSuite.peakmem_write_compressed('pbmc3k')                                   |
|          | 221M                       | 215M                                               |    0.97 | readwrite.H5ADBackedWriteSuite.peakmem_write_full('pbmc3k')                                         |
|          | 371±20ms                   | 344±2ms                                            |    0.93 | readwrite.H5ADBackedWriteSuite.time_write_compressed('pbmc3k')                                      |
|          | 95.3±2ms                   | 93.5±2ms                                           |    0.98 | readwrite.H5ADBackedWriteSuite.time_write_full('pbmc3k')                                            |
| -        | 25.328125                  | 19.40625                                           |    0.77 | readwrite.H5ADBackedWriteSuite.track_peakmem_write_compressed('pbmc3k')                             |
| -        | 26.046875                  | 15.359375                                          |    0.59 | readwrite.H5ADBackedWriteSuite.track_peakmem_write_full('pbmc3k')                                   |
|          | 113563819                  | 113570669                                          |    1    | readwrite.H5ADInMemorySizeSuite.track_actual_in_memory_size('pbmc3k')                               |
|          | 24109702                   | 24109702                                           |    1    | readwrite.H5ADInMemorySizeSuite.track_in_memory_size('pbmc3k')                                      |
|          | 24.1M                      | 24.1M                                              |    1    | readwrite.H5ADReadSuite.mem_readfull_object('pbmc3k')                                               |
|          | 196M                       | 193M                                               |    0.99 | readwrite.H5ADReadSuite.peakmem_read_backed('pbmc3k')                                               |
|          | 215M                       | 218M                                               |    1.01 | readwrite.H5ADReadSuite.peakmem_read_full('pbmc3k')                                                 |
|          | 69.4±2ms                   | 67.9±1ms                                           |    0.98 | readwrite.H5ADReadSuite.time_read_full('pbmc3k')                                                    |
| +        | 1.0                        | 1.6961942257217848                                 |    1.7  | readwrite.H5ADReadSuite.track_read_full_memratio('pbmc3k')                                          |
|          | 226M                       | 237M                                               |    1.05 | readwrite.H5ADWriteSuite.peakmem_write_compressed('pbmc3k')                                         |
|          | 231M                       | 229M                                               |    0.99 | readwrite.H5ADWriteSuite.peakmem_write_full('pbmc3k')                                               |
|          | 306±9ms                    | 294±3ms                                            |    0.96 | readwrite.H5ADWriteSuite.time_write_compressed('pbmc3k')                                            |
|          | 42.1±2ms                   | 40.1±0.9ms                                         |    0.95 | readwrite.H5ADWriteSuite.time_write_full('pbmc3k')                                                  |
| +        | 7.1875                     | 9.328125                                           |    1.3  | readwrite.H5ADWriteSuite.track_peakmem_write_compressed('pbmc3k')                                   |
| -        | 12.71875                   | 8.4375                                             |    0.66 | readwrite.H5ADWriteSuite.track_peakmem_write_full('pbmc3k')                                         |
|          | 243M                       | 250M                                               |    1.03 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '0:1000', False)            |
|          | 269M                       | 268M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '0:1000', True)             |
|          | 258M                       | 263M                                               |    1.02 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '0:9000', False)            |
|          | 287M                       | 285M                                               |    0.99 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '0:9000', True)             |
|          | 250M                       | 253M                                               |    1.01 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), ':9000:-1', False)          |
|          | 264M                       | 264M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), ':9000:-1', True)           |
|          | 277M                       | 284M                                               |    1.03 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '::-2', False)              |
|          | 281M                       | 272M                                               |    0.97 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), '::-2', True)               |
|          | 291M                       | 283M                                               |    0.97 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'alternating', False)       |
|          | 293M                       | 278M                                               |    0.95 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'alternating', True)        |
|          | 254M                       | 246M                                               |    0.97 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'arange', False)            |
|          | 250M                       | 259M                                               |    1.04 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'arange', True)             |
|          | 248M                       | 246M                                               |    0.99 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'array', False)             |
|          | 261M                       | 262M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'array', True)              |
|          | 249M                       | 249M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'first', False)             |
|          | 268M                       | 263M                                               |    0.98 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem((10000, 10000), 'first', True)              |
|          | 241M                       | 251M                                               |    1.04 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '0:1000', False)      |
|          | 250M                       | 244M                                               |    0.98 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '0:1000', True)       |
|          | 247M                       | 253M                                               |    1.02 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '0:9000', False)      |
|          | 251M                       | 246M                                               |    0.98 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '0:9000', True)       |
|          | 251M                       | 251M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), ':9000:-1', False)    |
|          | 251M                       | 250M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), ':9000:-1', True)     |
| +        | 230M                       | 256M                                               |    1.11 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '::-2', False)        |
|          | 249M                       | 243M                                               |    0.98 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), '::-2', True)         |
| +        | 212M                       | 249M                                               |    1.18 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'alternating', False) |
|          | 250M                       | 250M                                               |    1    | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'alternating', True)  |
|          | 248M                       | 243M                                               |    0.98 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'arange', False)      |
|          | 254M                       | 251M                                               |    0.99 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'arange', True)       |
|          | 252M                       | 232M                                               |    0.92 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'array', False)       |
|          | 253M                       | 244M                                               |    0.96 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'array', True)        |
|          | 250M                       | 251M                                               |    1.01 | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'first', False)       |
| -        | 248M                       | 223M                                               |    0.9  | sparse_dataset.SparseCSRContiguousSlice.peakmem_getitem_adata((10000, 10000), 'first', True)        |
|          | 2.23±0.2ms                 | 2.17±0.1ms                                         |    0.97 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '0:1000', False)               |
|          | 4.24±0.1ms                 | 4.34±0.08ms                                        |    1.02 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '0:1000', True)                |
|          | 5.93±0.07ms                | 6.08±0.06ms                                        |    1.03 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '0:9000', False)               |
|          | 16.4±0.6ms                 | 17.2±0.5ms                                         |    1.05 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '0:9000', True)                |
|          | 6.84±0.2ms                 | 7.16±0.2ms                                         |    1.05 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), ':9000:-1', False)             |
|          | 4.46±0.2ms                 | 4.37±0.05ms                                        |    0.98 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), ':9000:-1', True)              |
|          | 24.7±0.9ms                 | 25.0±1ms                                           |    1.01 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '::-2', False)                 |
|          | 17.2±0.6ms                 | 17.2±0.2ms                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), '::-2', True)                  |
|          | 30.5±1ms                   | 30.4±0.8ms                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'alternating', False)          |
|          | 22.2±0.6ms                 | 22.2±0.3ms                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'alternating', True)           |
|          | 6.16±0.07ms                | 6.22±0.2ms                                         |    1.01 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'arange', False)               |
|          | 4.61±0.2ms                 | 4.70±0.1ms                                         |    1.02 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'arange', True)                |
|          | 2.51±0.04ms                | 2.54±0.1ms                                         |    1.01 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'array', False)                |
|          | 6.86±0.08ms                | 7.03±0.2ms                                         |    1.03 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'array', True)                 |
|          | 2.00±0.03ms                | 1.99±0.04ms                                        |    0.99 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'first', False)                |
|          | 4.10±0.06ms                | 4.16±0.07ms                                        |    1.02 | sparse_dataset.SparseCSRContiguousSlice.time_getitem((10000, 10000), 'first', True)                 |
|          | 30.7±0.2μs                 | 30.6±0.4μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '0:1000', False)         |
|          | 31.1±0.6μs                 | 30.9±0.2μs                                         |    0.99 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '0:1000', True)          |
|          | 30.7±0.2μs                 | 30.8±0.2μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '0:9000', False)         |
|          | 30.8±0.2μs                 | 30.5±0.2μs                                         |    0.99 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '0:9000', True)          |
|          | 31.4±0.8μs                 | 30.8±0.2μs                                         |    0.98 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), ':9000:-1', False)       |
|          | 30.6±0.6μs                 | 31.0±0.5μs                                         |    1.01 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), ':9000:-1', True)        |
|          | 30.6±0.1μs                 | 30.6±0.3μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '::-2', False)           |
|          | 32.0±1μs                   | 30.8±0.3μs                                         |    0.96 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), '::-2', True)            |
|          | 87.0±1μs                   | 87.7±1μs                                           |    1.01 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'alternating', False)    |
|          | 88.5±3μs                   | 90.0±1μs                                           |    1.02 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'alternating', True)     |
|          | 45.3±0.5μs                 | 45.4±0.3μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'arange', False)         |
|          | 45.5±0.4μs                 | 46.4±1μs                                           |    1.02 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'arange', True)          |
|          | 41.3±0.6μs                 | 41.3±0.6μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'array', False)          |
|          | 41.1±0.2μs                 | 41.3±0.7μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'array', True)           |
|          | 31.6±0.3μs                 | 31.6±0.5μs                                         |    1    | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'first', False)          |
|          | 31.6±0.1μs                 | 31.2±0.3μs                                         |    0.99 | sparse_dataset.SparseCSRContiguousSlice.time_getitem_adata((10000, 10000), 'first', True)           |
```

@ilan-gold I think it's ready to be run on your CI.

ilan-gold:
> that will fail on main without this PR, because they cover new functionality. Maybe it's a bad idea to include things like that?

Right so my concern is really mostly what we have on main right now i.e., you pass in a string array, and now pay some performance penalty for, at least, finding out if it's sorted/unique.  And it looks like from what you have, to be reasonable.  The benchmark machine reutrned a job for me in 45 minutes yesterday as opposed to 3 hours, so hopefully whatever was happening is resolved.  I'll have yours running before the afternoon here - if I forget, feel free to ping.

ilan-gold:
Thanks for sticking with this @sjfleming !

ilan-gold:
@sjfleming I actually just noticed something: https://github.com/scverse/anndata/pull/2066/files#diff-4e5148021dbb054cace3f9749e52227b27fa61acf1137d979f90d148b82555c3R15-R21 only contains sparse benchmarking but this indexing is for dense, no?

If that is true, I would revert this PR (https://github.com/scverse/anndata/pull/2163) and then we would benchmark against main again with the fixed benchmarks.  But maybe I'm missing something.

Alternatively, we could do this manually locally, ~~but I'd kind of rather have the "clean" run on the benchmark machine.  Let me know what you think.~~ I think we need a separate PR with dense benchmarks.

ilan-gold:
I will handle the benchmarks.  I think I'm more irked about the time it takes (which is probably unrelated to this PR).  Thanks again, @sjfleming, sorry the panicked response.

sjfleming:
Awesome thanks @ilan-gold !  Yeah it looks like the benchmarks should probably be updated to include dense and sparse.

sjfleming:
Hi @ilan-gold , I was wondering if you know if this PR is already in a released version of `anndata`, and which version number that is?  If I was better at github I could probably figure this out, but I can't seem to

ilan-gold:
Yes, it was released.  I'm seeing there's something up with our release notes, bizarre!

ilan-gold:
(This PR had no release note which is also not great, but our fault, sorry)

PR #2067, opened 2025-07-31
Fix index types

flying-sheep:
Ilan made some good progress here, so this is intended to finish things up.

We have two “deal with indices” helpers in `anndata._core`: `.index._normalize_index` which accepts broad user-supplied types and unifies them into simple versions, and `.views._resolve_idxs` which simplifies multiple successive indexing operations (views of views).

Currently we have several “1D” (along the obs or var axis) index types which I’ll refer to by their name in the code:
1. `Index1D`: to capture user-supplied values like `list[str]`, `int`, `pd.Series[bool]` and so on.
2. `Index1DNorm` (new): stored in anndata views as `_oidx`/`_vidx`: Can be a `slice` or `np.ndarray` with dtype `np.bool_` or `np.integer`
3. what we pass to `_init_as_actual`: can be `Index1DNorm` or an `int`/`np.integer`. We should probably get rid of this and sliceify 0D indices in `_normalize_index`

TODO:
- [x] check if all advertised types work and clarify (e.g. which `pd.Index[???]` do we support?)
- [x] check if subsetting views works in all permutations (e.g. mask → slice or mask → ndarray)

### DIFF
--- a/src/anndata/compat/__init__.py
+++ b/src/anndata/compat/__init__.py
@@ -44,12 +44,15 @@ Index1D = (
     # normalized 1D idex
     | Index1DNorm
     # different containers for mask, obs/varnames, or numerical index
-    | pd.Series  # bool, int, str
-    | pd.Index
-    | NDArray[np.str_]
     | Sequence[int]
     | Sequence[str]
     | Sequence[bool]
+    | pd.Series  # bool, int, str
+    | pd.Index
+    | NDArray[np.str_]
+    | np.matrix  # bool
+    | CSMatrix  # bool
+    | CSArray  # bool
 )
 IndexRest = Index1D | EllipsisType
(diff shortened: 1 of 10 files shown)


src/anndata/compat/__init__.py
```diff
@@ -32,7 +34,23 @@ class Empty:
...
+Index1D = (
+    # 0D index
+    int
+    | str
+    | np.int64
+    # normalized 1D idex
+    | Index1DNorm
+    # different containers for mask, obs/varnames, or numerical index
+    | pd.Series  # bool, int, str
+    | pd.Index
```
flying-sheep: I don’t think we test these (at least not via `subset_func` and therefore with `test_views_of_views`)

Issue #2043, opened 2025-07-22
`DataFrame` API for `obs` and `var` keys


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

Our `pandas.DataFrame` is relatively small and is now clearly documented by the adapter `Dataset2D` class.  In theory anything that satisfies this `Protocol` should work as a `DataFrame`-like class in `anndata`:

https://github.com/scverse/anndata/blob/401e0d140b204bfdd1b46b45a3557ad3775f55d7/src/anndata/_core/xarray.py#L33-L51

To me the concrete steps would be

- [ ] Refactor the current `Dataset2D` class to inherit from a runtime-checkable `Protocol` and then replace all instances throughout the codebase of `pd.DataFrame` or `Dataset2D` with said `Protocol`.  This will likely entail adding a new method to the `Protocol` to handle `anndata.concat`.
- [ ] Remove the `xarray`/`Dataset` dependency from `anndata` into its own package
- [ ] Create new test cases that use both the `Protocol` and an actual `pandas.DataFrame` object to test the functionality in the absence of `Dataset2D`
- [ ] Create other readers/adatapters for other dataframe-like libraries (`dask.DataFrame`, `polar`, `cudf` etc.)

ilan-gold:
Note that this is different from https://github.com/scverse/anndata/issues/1111 which proposes making `anndata` interchangeable as a dataframe itself

srivarra:
@ilan-gold 

> Create other readers/adatpters for other dataframe-like libraries (dask.DataFrame, polars, cudf etc.) and their own read functions (likely other repos, or one big new repo)


I released a super early version of [`zarrwhals`](https://github.com/srivarra/zarrwhals) yesterday, which lets you take a Pandas / Dask / Polars DataFrames and write it out to Zarr, then read it back into those three formats.

ilan-gold:
This is awesome! It looks like the file format is quite close to ours in https://zarrwhals.readthedocs.io/en/latest/architecture/#group-attributes but I'm wondering why the `int64` specification - why choose to make that explicit instead of implicit on the zarr data type itself?

srivarra:
> but I'm wondering why the int64 specification - why choose to make that explicit instead of implicit on the zarr data type itself?


Are you asking why columns have the following metadata?

```json
{
    "encoding-type": "array",
    "narwhals_dtype": "Int64"
}

```

I did that for 2 reasons:

[...]

PR #2071, opened 2025-08-11
feat: support array-api

### LINKED ISSUE #697: Support JAX as a backend
I would support JAX as another backend.

Would probably be worth opening another issue for tracking JAX as a backend. For IO, I was thinking that could be done with #659

_Originally posted by @ivirshup in https://github.com/theislab/anndata/issues/695#issuecomment-1026039360_

### LINKED ISSUE #1195: General array-api support

### LINKED ISSUE #1731: make multidimensional sparse arrays work with the .varp and .obsp fields

`scipy.sparse` only supports 2D arrays. Whereas `varp` and `obsp` support N-D arrays.
Most of the time we would want them to be sparse.
But right now `anndata` prevent the `varp` and `obsp` fields to contain the pydata/sparse formats. 

Not all but many features are available on these matrix formats. Are there missing features that would be required as of now that would prevent this integration?

amalia-k510:
fixes #1731, fixes #1195, fixes #697

First step in getting anndata concat and test generation to work properly with JAX, (and Cubed potentially), without just converting everything into NumPy.

Random data creation and shape handling use xp.asarray so arrays stay in their original backend where possible. I also updated concat paths to actually check types before converting, added helpers for sparse detection and array API checks, and made sure backend arrays only get turned into NumPy when absolutely necessary. This fixes a bunch of concat-related test failures.

It’s still not perfect. Some pandas calls in concat still force conversion to NumPy, so the data gets copied instead of being used directly. Cubed support is only a placeholder right now. Type detection might still be a bit too broad, which can lead to extra conversions. Works for NumPy and JAX in tests, but I haven’t tried other backends.


src/anndata/_core/merge.py
```diff
@@ -658,7 +658,12 @@ def _apply_to_array(self, el, *, axis, fill_value=None):
 
         indexer = self.idx
 
-        # Indexes real fast, and does outer indexing
+        # Fallback to numpy: keep pandas
+        # Force to NumPy (materializes JAX/Cubed); fine for small tests,
+        # but may be slow or fail on large/lazy arrays
```
flying-sheep: This code doesn’t just run for tests though. Also are you sure that this is a good idea for arrays with pandas dtypes?

src/anndata/_core/merge.py
```diff
@@ -1399,6 +1404,37 @@ def concat_dataset2d_on_annot_axis(
     return ds_concat_2d
 
 
+def _is_sparse(x):
+    try:
+        return scipy.sparse.issparse(x)
+    except TypeError:
+        return False
```
flying-sheep: this is unused and I also don’t understand why this error would ever happen.

```suggestion
```

src/anndata/_core/merge.py
```diff
@@ -1399,6 +1404,37 @@ def concat_dataset2d_on_annot_axis(
...
+
+
+def _to_numpy_if_array_api(x):
```
flying-sheep: there should be no second copy of this that’s slightly different, only one!

src/anndata/_core/merge.py
```diff
@@ -1399,6 +1404,37 @@ def concat_dataset2d_on_annot_axis(
...
+        return np.asarray(x)
+    except TypeError:
+        # Not an array-API object (or lib not available) → return unchanged
+        return x
+
+
+def _normalize_nested(obj):
+    if isinstance(obj, dict):
+        return {k: _normalize_nested(v) for k, v in obj.items()}
+    if isinstance(obj, list | tuple):
```
flying-sheep: Usually you should never do `isinstance(obj, dict)` or `isinstance(obj, list)`, only `Mapping` and `Sequence` (or more precise types).

Also please explain what this is for, i.e. why you added it and call it?

src/anndata/tests/helpers.py
```diff
@@ -268,15 +280,60 @@ def maybe_add_sparse_array(
...
 
 
+# fixing the ndarray issue
+def _is_sparse_like(x):
+    try:
+        import scipy.sparse as sp
+
+        return sp.issparse(x)
+    except ImportError:
+        return False
```
flying-sheep: you should use `isinstance` with the `CS*` types in `anndata.compat` instead of this function

```suggestion
```

tests/test_helpers.py
```diff
@@ -28,6 +30,11 @@
 )
 from anndata.utils import axis_len
 
+jax.config.update(
+    "jax_enable_x64",
+    True,  # noqa: FBT003
+)  # Explicit conversion to JAX float64
```
flying-sheep: don’t call code on top-level.

instead, call it in fixture setups and undo what you did in fixture teardown:

```py
@pytest.fixture
def _give_me_a_name() -> None:
	# Explicit conversion to JAX float64
	jax.config.update("jax_enable_x64", True)  # noqa: FBT003
	yield
	# TODO: undo what you did above
```

flying-sheep (commented):
OK, I just went over general code style, nothing JAX-related

src/anndata/_core/merge.py
amalia-k510: Yeah, I was initially forcing everything to NumPy, but that’s no longer the case. I’ve updated it so the it should preserve arrays with pandas dtypes.

src/anndata/_core/merge.py
amalia-k510: I originally added it to catch JAX-related errors, but later fixed the underlying issue, so it’s no longer needed. I removed it.

tests/test_helpers.py
amalia-k510: Got it! I added it to the code and removed the top-level call.

src/anndata/_core/anndata.py
```diff
@@ -667,6 +667,12 @@ def X(self, value: XDataType | None):  # noqa: PLR0912
                     ImplicitModificationWarning,
                     stacklevel=2,
                 )
+                dest = self._adata_ref._X
+                # Handles read-only NumPy views from backend arrays like JAX by
+                # making a writable copy so in-place assignment on views can succeed.
+                if isinstance(dest, np.ndarray) and not dest.flags.writeable:
+                    dest = np.array(dest, copy=True)  # make a fresh, writable buffer
+                    self._adata_ref._X = dest
```
ilan-gold: I would actually just let the error be thrown in this case.  If something isn't writeable, I don't think that's our responsibility to handle

src/anndata/_core/merge.py
```diff
@@ -538,6 +539,15 @@ def resolve_merge_strategy(
     return strategy
 
 
+def safe_to_numpy(x):
+    """Convert to numpy array, handling JAX/Cupy arrays."""
+    if isinstance(x, pd.Series | pd.Index) or (
+        hasattr(x, "dtype") and is_extension_array_dtype(x.dtype)
+    ):
+        return x
+    return np.asarray(x)
```
ilan-gold: Ok nice this is the right direction no doubt! So what we want here probably is not to rely on `asarray` but `dlpack` to do the conversion. In short:

1. We should have a check in `_apply_to_array` to see if something is array-api compatible but not a numpy ndarray.
2. If this case is true, dlpack into numpy, recursively call `_apply_to_array`
3. Then use dlpack to take the output of the recursive call to the original type before we went to numpy.

Does that make sense?

ilan-gold: I think this is a nice paradigm to follow for situations where we have an existing numpy or cupy implementation and it isn't clear how to use the array-api to achieve our aims.  We should still try to use it as much as possible so that we can eventually remove numpy codepaths where possible, but this is a nice first step.

src/anndata/_core/merge.py
```diff
@@ -538,6 +544,48 @@ def resolve_merge_strategy(
...
+        return np.from_dlpack(x)
+    except TypeError as e:
+        msg = f"DLPack to NumPy failed: {e}"
+        raise TypeError(msg) from e
+
+
+def _dlpack_from_numpy(x_np, original_xp):
+    # cubed and other array later elif
+    if original_xp.__name__.startswith("jax"):
+        return jax.dlpack.from_dlpack(x_np)
```
ilan-gold: https://data-apis.org/array-api/latest/API_specification/generated/array_api.from_dlpack.html#array_api.from_dlpack

src/anndata/_core/merge.py
```diff
@@ -49,6 +52,9 @@
 
 T = TypeVar("T")
 
+with suppress(ImportError):
```
ilan-gold: See https://data-apis.org/array-api/latest/API_specification/generated/array_api.from_dlpack.html#array_api.from_dlpack

src/anndata/_core/merge.py
```diff
@@ -1018,16 +1083,22 @@ def concat_arrays(  # noqa: PLR0911, PLR0912
...
-        return np.concatenate(
+        # Use the backend of the first array as the reference
+        ref = arrays[0]
+        xp = get_namespace(ref)
+
+        # Convert all arrays to the same backend as `ref`
+        arrays = [ref] + [_same_backend(ref, x, copy=True)[1] for x in arrays[1:]]
+
+        # Concatenate with the backend’s API
+        value = xp.concatenate(
```
ilan-gold: This condition was previously hit by the fact that none of the above checks involving `any` were `True`.  Instead of changing this last default condition, I would create a new branch here specifically for the array-api, check that they all have the same backend, and the concatenate.  If they don't have the same backed, you just proceed to the `np` condition (which will fail presumably).  I wouldn't worry about mixing different backends, especially with the array-api for now.  If we use `cubed`, `dlpack` won't work there anyway

src/anndata/_core/merge.py
```diff
@@ -629,13 +627,22 @@ def _dlpack_to_numpy(x):
...
+    if hasattr(original_xp, "from_dlpack"):
+        try:
+            return original_xp.from_dlpack(x_np)
+        except Exception as e:
+            msg = f"Failed to call from_dlpack on backend {original_xp.__name__}: {e}"
+            raise TypeError(msg) from e
+
+    # fallback for known backends that put it elsewhere (JAX and later others)
     if original_xp.__name__.startswith("jax"):
+        import jax.dlpack
```
ilan-gold: `original_xp` should have a `from_dlpack` method! Does my comment here https://github.com/scverse/anndata/pull/2071/commits/383c4459c804df13b15fa14ba299a400e1e94b98#r2291442475 not apply?

tests/test_readwrite.py
```diff
@@ -708,10 +721,11 @@ def test_dataframe_reserved_columns(tmp_path, diskfmt, colname, attr):
         getattr(to_write, f"write_{diskfmt}")(adata_pth)
 
 
-def test_write_large_categorical(tmp_path, diskfmt):
+@pytest.mark.parametrize("xp", [np, jnp])  # xp = array namespace
```
ilan-gold: I would revert this - it's just for generating categories which gets pushed into pandas.  I don't think this triggers any internal array-api code

src/anndata/_core/index.py
```diff
@@ -107,9 +113,66 @@ def name_idx(i):
     elif isinstance(indexer, XDataArray):
         if isinstance(indexer.data, DaskArray):
             return indexer.data.compute()
-        return indexer.data
+        return indexer.dat
+
+    elif has_xp(indexer):
```
ilan-gold: This looks nearly identical to the numpy case.   In a case like this, I think you were/would be right to just merge the two if possible.  If it's not I would explain why.

src/anndata/_core/merge.py
```diff
@@ -647,21 +750,56 @@ def _apply_to_cupy_array(self, el, *, axis, fill_value=None):
...
-            shape[axis] = len(self.new_idx)
-            return np.broadcast_to(fill_value, tuple(shape))
 
         indexer = self.idx
 
-        # Indexes real fast, and does outer indexing
-        return pd.api.extensions.take(
-            el, indexer, axis=axis, allow_fill=True, fill_value=fill_value
-        )
+        if _is_pandas(el):
```
ilan-gold:
```suggestion
        if isinstance(el, np.ndarray):
```
I would have thought that `el` is a numpy array given that the old function name was `_apply_to_array`, no?

src/anndata/_core/merge.py
```diff
@@ -647,21 +750,56 @@ def _apply_to_cupy_array(self, el, *, axis, fill_value=None):
...
+            # Convert to NumPy via DLPack
+            el_np = _dlpack_to_numpy(el)
+
+            # Recursively call this same function
+            out_np = self._apply_to_array_api(el_np, axis=axis, fill_value=fill_value)
+
+            # reverting back to numpy as it is hard to reindex on JAX and others
+            return _dlpack_from_numpy(out_np, xp)
+
+        # numpy case
```
ilan-gold: I think the logic here is a little confused.  This function used to be for `numpy`, but you added the `_is_pandas` check above, which I don't think applies here.  But the logic you've written "# numpy case" and down works great for non-numpy array-api compatible arrays as well!

So I would leave the numpy case as before (i.e., remove `_is_pandas` and check `isinstance(el, np.ndarray)`), and then in the case it is not a numpy array, use this logic under "# numpy case"! You can then get rid of the `if not isinstance(el, np.ndarray) and _is_array_api_compatible(el):` branch.  One reason I cautioned against falling back to `numpy` behavior is that some things like `jax` arrays that are API compatible might be on the GPU! You can't transfer a JAX array on the GPU to numpy :/

src/anndata/_core/views.py
```diff
@@ -474,6 +582,20 @@ def _resolve_idx_slice(
         return np.arange(*old.indices(l))[new]
 
 
+@_resolve_idx.register(jnp.ndarray)
```
ilan-gold: `_resolve_idx` has no default case.  I would move 

1. change `_resolve_idx_jnp` to `_resolve_idx_array_api` and make it generic over the array api.  If this function is tested right now via `jax` it should still be testable in and work when you use the array-api to do these operations instead of `jnp`
2. Remove `@_resolve_idx.register(jnp.ndarray)` and then immediately
3. Move the body of this function into the default implementation i.e., `_resolve_idx` with the `@singledispatch` above it.  This will allow you to _by default_ check if something is array-api compatible and then use your implementation. 
4. Otherwise just raise the `NotImplementedError` that is already there

src/anndata/_io/specs/registry.py
```diff
@@ -34,6 +40,54 @@
...
 
 
+def is_sparse_like(x):
+    try:
+        return sp.issparse(x)
+    except AttributeError:
+        return False
+
+
+def to_numpy_if_array_api(x):
```
ilan-gold: You probably want to check `device` first to weed out gpu arrays https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.device.html and/or bring them back to the CPU

src/anndata/_core/views.py
```diff
@@ -556,7 +554,42 @@ def _resolve_idxs(
...
+    import array_api_compat as aac
+
+    from ..compat import has_xp
+
+    # handling array-API–compatible arrays (e.g. JAX, etc)
+    if has_xp(old):
+        xp = aac.array_namespace(old)
+        # skip early if numpy
+        if xp.__name__.startswith("numpy"):
+            return old[new]
```
ilan-gold: Why special case this? If `isinstance(old, numpy.ndarray)` then `singledispatch` handles this for us.  If something is resolved to use `numpy` by `array_api_compat` (I don't see why this would ever be the case here, and I don't see anything in the docs about falling back to `numpy`)

src/anndata/_core/views.py
```diff
@@ -556,7 +554,42 @@ def _resolve_idxs(
...
+        # handle boolean mask; i.e. checking whether old is a boolean array
+        if hasattr(old, "dtype") and str(old.dtype) in ("bool", "bool_", "boolean"):
+            # retrieving the where function (like np.where or jnp.where)
+            where_fn = getattr(xp, "where", None)
+            # if where exists, use it to convert the boolean mask to integer indices
+            if where_fn is not None:
+                old = where_fn(old)[0]
+            else:
+                # if no where function is found, fallback to NumPy
+                old = np.where(np.asarray(old))[0]
```
ilan-gold: How could `where_fn` ever be `None`? https://data-apis.org/array-api/latest/API_specification/generated/array_api.where.html#where is part of the array-api

src/anndata/_core/views.py
```diff
@@ -556,7 +554,42 @@ def _resolve_idxs(
...
+
+    # handling array-API–compatible arrays (e.g. JAX, etc)
+    if has_xp(old):
+        xp = aac.array_namespace(old)
+        # skip early if numpy
+        if xp.__name__.startswith("numpy"):
+            return old[new]
+
+        # handle boolean mask; i.e. checking whether old is a boolean array
+        if hasattr(old, "dtype") and str(old.dtype) in ("bool", "bool_", "boolean"):
```
ilan-gold: If `old` is not array-api compatible, shouldn't we error out? `old.dtype` would exist and https://data-apis.org/array-api/latest/API_specification/generated/array_api.isdtype.html#isdtype could handle checking for `bool`, no need for strings or anything

src/anndata/_core/views.py
```diff
@@ -556,7 +554,42 @@ def _resolve_idxs(
...
+            if where_fn is not None:
+                old = where_fn(old)[0]
+            else:
+                # if no where function is found, fallback to NumPy
+                old = np.where(np.asarray(old))[0]
+
+        # if new is a slice object, converting it into a range of indices using arange
+        if isinstance(new, slice):
+            # trying to get arange from the backend
+            arange_fn = getattr(xp, "arange", None)
```
ilan-gold: Why `getattr` (same with `where`)? Shouldn't `xp.arange` exist: https://data-apis.org/array-api/latest/API_specification/generated/array_api.arange.html#arange?

src/anndata/_io/specs/registry.py
```diff
@@ -64,21 +64,38 @@ def to_numpy_if_array_api(x):
...
     try:
+        # importing the array API compatibility layer
         import array_api_compat as aac
 
+        # getting the array namespace
         xp = aac.array_namespace(x)
-        # Already a NumPy array
+        # Skip if it's already NumPy
         if xp.__name__.startswith("numpy"):
             return x
```
ilan-gold: How could this condition be reached if we have the above `isinstance(x, np.ndarray | ...)` check?

src/anndata/_io/specs/registry.py
```diff
@@ -34,6 +40,71 @@
...
+        | np.generic
+        | pd.DataFrame
+        | pd.Series
+        | pd.Index
+        | ExtensionArray
+        | DaskArray
+        | sp.spmatrix
+        | AnnData,
+    ):
+        return x
```
ilan-gold: Given this check, I would think this function should be called `to_writeable`, no?

src/anndata/_io/specs/registry.py
```diff
@@ -386,6 +457,10 @@ def write_elem(
         elif k in store:
             del store[k]
 
+        # Normalize array-API (e.g., JAX/CuPy) payloads buried in mappings/lists
+        if not isinstance(elem, AnnData):
+            elem = normalize_nested(elem)
```
ilan-gold: Wouldn't we want to also normalize `AnnData` objects so they're sub elements are also corrected?

src/anndata/_core/merge.py
```diff
@@ -905,6 +1038,24 @@ def concat_arrays(  # noqa: PLR0911, PLR0912
...
+
+    elif all(hasattr(a, "__array_namespace__") for a in arrays):
+        # All arrays are array-api compatible
+
+        # use first as a reference to check if all of the arrays are the same type
+        xp = get_namespace(arrays[0])
+
+        if not all(get_namespace(a) is xp or a.shape == 0 for a in arrays):
+            msg = "Cannot concatenate array-api arrays from different backends."
+            raise ValueError(msg)
```
ilan-gold, line 909: I think the check around making sure they have the same namespace is encapsulated in https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace since it can take in an array of array objects - you could use that instead of just using the first one

src/anndata/_core/merge.py
```diff
@@ -584,6 +687,8 @@ def apply(self, el, *, axis, fill_value=None):  # noqa: PLR0911
         """
         if self.no_change and (axis_len(el, axis) == len(self.old_idx)):
             return el
+        if _is_pandas(el) or isinstance(el, pd.DataFrame):
+            return self._apply_to_df_like(el, axis=axis, fill_value=fill_value)
```
ilan-gold: Why did you add this?

src/anndata/_core/merge.py
```diff
@@ -647,21 +752,44 @@ def _apply_to_cupy_array(self, el, *, axis, fill_value=None):
...
+            return xp.broadcast_to(fv, shape)
+
+        # Check: is array-api compatible, but not NumPy
+        if not isinstance(el, np.ndarray) and _is_array_api_compatible(el):
+            # Convert to NumPy via DLPack
+            el_np = _dlpack_to_numpy(el)
+            # Recursively call this same function
+            out_np = self._apply_to_array_api(el_np, axis=axis, fill_value=fill_value)
+            # reverting back to numpy as it is hard to reindex on JAX and others
+            return _dlpack_from_numpy(out_np, xp)
```
ilan-gold: Why convert to numpy here? Just let the array pass through, that's the whole point of using the array api :) If a jax array is on the GPU you're going to bring it to the CPU here, but why?

ilan-gold (changes requested):

src/anndata/_core/merge.py
ilan-gold: Apologies if it wasn't clear in my [previous comment](https://github.com/scverse/anndata/pull/2071#discussion_r2285643685) - we should only go to numpy if the existing array is on the CPU _and_ we couldn't come up with a generic way of doing this via the array-api.  But you made a way of doing this via the array-api which is great, so no need to convert to numpy!

ilan-gold (commented):
Apologies if this is going in circles, but I'm kind of reviewing locally to changes, and losing the big picture sometimes!

src/anndata/_core/merge.py
amalia-k510: I think I added this clause to explicitly test for the pandas dataframe when I was testing and ended up redundant. I'll remove the _is_pandas() check to avoid unnecessary duplication and keep the logic clearer.

src/anndata/_core/views.py
```diff
@@ -558,38 +558,24 @@ def _resolve_idx(old: Index1DNorm, new: Index1DNorm, l: Literal[0, 1]) -> Index1
...
+    # error out to make sure that object is array-API compatible, will xFAIL it later
+    if not has_xp(old):
+        msg = f"Expected array-API–compatible array, got {type(old)}"
+        raise TypeError(msg)
+
+    xp = aac.array_namespace(old)
+
+    # handle boolean mask; check dtype string representation for bool-like types
+    # not sure what is causing it but i get the error stating that array_api_compat does not have isdtype (can't seem to resolve the issue)
+    # I have 2023 version of array_api_compat installed
```
ilan-gold: Yea, this isn't a matter of installation - each library has its own version.  You can request a specific version: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__array_namespace__.html

src/anndata/_io/specs/registry.py
```diff
@@ -47,7 +47,7 @@ def is_sparse_like(x):
         return False
 
 
-def to_numpy_if_array_api(x):
+def to_writeable(x):
```
ilan-gold, line 39: For this function, we have GPU CI so probably makes sense to make jax-specific tests that check for ability to write

ilan-gold (commented):
Nice, let's get the tests passing and merge fixed now (I can't even see about the tests because main and this branch are quite far off)!

tests/test_dask.py
```diff
@@ -310,6 +322,28 @@ def test_dask_to_memory_unbacked(array_func, mem_type):
...
+
+    random_state = np.random.default_rng()
+    arr = random_state.binomial(100, 0.005, (20, 15)).astype("float32")
+
+    # TODO: need to change type for cupy
+    orig = ad.AnnData(to_dask(arr))
+    orig = orig[orig.shape[0] // 2]
+    path = tmp_path / f"test.{diskfmt}"
+    if isinstance(orig.X, jax.Array):
+        pytest.xfail("JAX arrays aren’t supported for h5ad or zarr writes yet")
```
ilan-gold: Should probably not even get to this point, I don't think dask officially support jax in any way.  What's up here?

src/anndata/_io/specs/registry.py
```diff
@@ -35,6 +41,68 @@
...
+                    x = xp.to_device(
+                        x, "cpu"
+                    )  # not sure about this, would we want to move the entire array to CPU?
+            # otherwise, if the device is not a string, check if it's not "cpu"
+            elif str(device) != "cpu":
+                x = xp.to_device(x, "cpu")
+
+        # Convert to NumPy using DLPack (safe now)
+        if hasattr(x, "__to_dlpack__"):
+            return np.from_dlpack(x)
```
ilan-gold: See https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.array.__dlpack__.html

tests/test_awkward.py
```diff
@@ -130,6 +130,30 @@ def test_copy(key):
         getattr(adata, key)["awk"]["d"]
 
 
+# @pytest.mark.parametrize("key", ["obsm", "varm"])
```
ilan-gold: Remove commented out stuff

hatch.toml
```diff
@@ -42,8 +42,9 @@ overrides.matrix.deps.python = [
 ]
 overrides.matrix.deps.features = [
     { if = [ "stable", "pre" ], value = "test" },
+    { if = [ "array-api" ], features = [ "test-min", "jax" ] },
```
ilan-gold:
```suggestion
    { if = [ "array-api" ], value = [ "test-min", "jax" ] },
```

pyproject.toml
```diff
@@ -97,19 +97,28 @@ test-min = [
...
     "dask[array]>=2023.10.1,!=2024.8.*,!=2024.9.*,!=2025.2.*,!=2025.3.*,!=2025.4.*,!=2025.5.*,!=2025.6.*,!=2025.7.*,!=2025.8.*",
 ]
+jax = [
+    "jax>=0.6.0",
+    "jaxlib>=0.6.0",
+]
 
 [tool.hatch.version]
+# https://github.com/dask/dask/issues/11290
+# https://github.com/dask/dask/issues/11752
```
ilan-gold: Please move these comments back

pyproject.toml
```diff
@@ -97,19 +97,28 @@ test-min = [
     "pyarrow",
     "anndata[dask]",
 ]
-test = [ "anndata[test-min,lazy]", "fast-array-utils>=1.2.3" ]
+test = [
+    "anndata[test-min,lazy]",
+    "anndata[jax]",
```
ilan-gold: I woulkd remove this from `test` and then make a separate `array-api-test` that has `jax` in it directly i.e., 

```
test-array-api = [
    "anndata[test-min]",
    "jax>=0.6.0",
    "jaxlib>=0.6.0"
]
```
we don't need `lazy` in there for now IMO

pyproject.toml
```diff
@@ -108,6 +118,10 @@ lazy = [ "xarray>=2025.06.1", "aiohttp", "requests", "anndata[dask]" ]
 dask = [
     "dask[array]>=2023.10.1,!=2024.8.*,!=2024.9.*,!=2025.2.*,!=2025.3.*,!=2025.4.*,!=2025.5.*,!=2025.6.*,!=2025.7.*,!=2025.8.*",
 ]
+jax = [
+    "jax>=0.6.0",
+    "jaxlib>=0.6.0",
+]
```
ilan-gold: Please remove

tests/test_x.py
```diff
@@ -13,12 +13,17 @@
 from anndata.tests.helpers import GEN_ADATA_NO_XARRAY_ARGS, assert_equal, gen_adata
 from anndata.utils import asarray
 
+# jax option (more later)
+jax = pytest.importorskip("jax")
```
ilan-gold: Do you need this?

tests/test_x.py
```diff
@@ -192,3 +201,12 @@ def test_fail_on_non_csr_csc_matrix():
...
         ad.AnnData(X=X)
+
+
+def test_create_anndata_with_jax():
+    from contextlib import suppress
+
+    with suppress(ImportError):
+        import jax.numpy as jnp
+
+    ad.AnnData(X=jnp.ones((10, 10)))
```
ilan-gold: Mark this as `skipif` if `jax` is not installed, don't suppress the `ImportError`

tests/test_x.py
```diff
@@ -57,6 +63,9 @@ def test_setter_view(orig_array_type, new_array_type):
     if isinstance(orig_X, np.ndarray) and sparse.issparse(to_assign):
         # https://github.com/scverse/anndata/issues/500
         pytest.xfail("Cannot set a dense array with a sparse array")
+
+    if isinstance(orig_X, jnp.ndarray) or isinstance(to_assign, jnp.ndarray):
+        pytest.xfail("JAX arrays are immutable and do not support in-place assignment")
```
ilan-gold: Instead of xfailing here, use https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions to assert the exact error, putting the line that fails in the context manager and then `return`

tests/test_views.py
```diff
@@ -379,6 +399,9 @@ def test_not_set_subset_X_dask(matrix_type_no_gpu, subset_func):
 
 @IGNORE_SPARSE_EFFICIENCY_WARNING
 def test_set_scalar_subset_X(matrix_type, subset_func):
+    if "jax" in matrix_type.__module__:
+        msg = "JAX arrays don't support item assignment via boolean indexing (e.g. adata.X[bool_idx] = val)."
+        pytest.xfail(msg)
```
ilan-gold: This isn't exactly the test you want - I would just `xfail` on an instance check of `adata.X`.  Also same deal, I would have a `raises` and `return`

tests/test_views.py
```diff
@@ -354,6 +371,9 @@ def test_not_set_subset_X(matrix_type_base, subset_func):
 @IGNORE_SPARSE_EFFICIENCY_WARNING
 def test_not_set_subset_X_dask(matrix_type_no_gpu, subset_func):
     adata = ad.AnnData(matrix_type_no_gpu(asarray(sparse.random(20, 20))))
+    if jnp is not None and isinstance(adata.X, jnp.ndarray):
+        msg = "JAX does not support in-place mutation, so this test is irrelevant."
+        pytest.skip(msg)
```
ilan-gold: Again might be worth a context with `pytest.raises` and a return.

pyproject.toml
```diff
@@ -177,6 +191,7 @@ markers = [
     "gpu: mark test to run on GPU",
     "zarr_io: mark tests that involve zarr io",
     "dask_distributed: tests that need a distributed client with multiple processes",
+    "array_api: mark tests that test Array API compatibility (JAX, Cubed)",
```
ilan-gold: I think this needs an implementation / usage

ilan-gold (commented):
So close - some comments in specific places but can be applied broadly

src/anndata/_core/merge.py
```diff
@@ -105,6 +118,39 @@ def not_missing(v) -> bool:
...
+        # This tells us whether we are dealing with NumPy, JAX, Cubed, etc.
+        xp_x = get_namespace(x)
+        xp_y = get_namespace(y)
+    except AttributeError as err:
+        # if either x or y does not support __array_namespace__, we cannot do backend detection
+        msg = "Both arrays must support __array_namespace__ for backend detection."
+        raise TypeError(msg) from err
+
+    # Special-case: Dask cannot be converted via DLPack
+    if is_dask_collection(x) or is_dask_collection(y):
```
ilan-gold: Just use `isinstance(x, DaskArray)` (we provide `DaskArray` in `compat.py`)

src/anndata/_core/merge.py
```diff
@@ -526,17 +581,65 @@ def merge_only(ds: Collection[Mapping]) -> Mapping:
...
+def _dlpack_to_numpy(x):
+    try:
+        return np.from_dlpack(x)
+    except TypeError:
+        return np.asarray(x)
+
+
+def _dlpack_from_numpy(x_np, original_xp):
+    # TODO: cubed and other array later elif
+    if hasattr(original_xp, "from_dlpack"):
```
ilan-gold: My `cmd + F` of these two functions doesn't return anything - are they used? What about `_same_backend`?

src/anndata/_core/merge.py
```diff
@@ -901,6 +1029,24 @@ def concat_arrays(  # noqa: PLR0911, PLR0912
...
+            msg = "Cannot concatenate array-api arrays from different backends."
+            raise ValueError(msg) from err
+
+        return xp.concatenate(
+            [
+                f(x, fill_value=fill_value, axis=1 - axis)
+                for f, x in zip(reindexers, arrays, strict=True)
+            ],
+            axis=axis,
+        )
```
ilan-gold: This is where I would think `_same_backend` would be used?

src/anndata/_core/merge.py
```diff
@@ -1706,6 +1854,17 @@ def concat(  # noqa: PLR0912, PLR0913, PLR0915
...
+    # Had to introduce this hacky check to convert initially placed -1 to np.nan
+
+    # Convert -1 to np.nan if .X is integer type and contains -1
+    if (
+        isinstance(X, np.ndarray)
+        and np.issubdtype(X.dtype, np.integer)
+        and (X == -1).any()
+    ):
+        X = X.astype(float)
+        X[X == -1] = np.nan
```
ilan-gold: Interesting, converting to `float` silently feels wrong.  I think using `-1` is intentional.  What's the issue with it?

ilan-gold (commented):
Few small comments. Maybe it makes sense to sit together and comb through outstanding issues since they are small and scattered at this point? Feel free to reply on slack!

Very very nice and exciting!

src/anndata/_core/merge.py
amalia-k510: These helper functions (`_dlpack_to_numpy`, `_dlpack_from_numpy`, and `_same_backend`) were part of an earlier attempt to manually convert between backends during merging. However, after focusing on getting JAX support stable, they ended up no longer being used.

Right now they are unused and the tests are temporarily xfailed. I am not fully certain whether the final design should still include explicit DLPack conversion helpers in case if we have two arrays that are of different format.

ilan-gold:
@flying-sheep An overview of how Amalia/I structured this PR:

## Indexes

You incoming index could come from a previously-unsupported-but-array-api-compatible library.  Nice! We support that:

1. In `__getitem__`, the incoming array-api indexer is sanitized just as numpy ones were before
2. The indexer is checked for array-api compatibility within `_init_as_view` afterwards and then wrapped in an `IndexManager` which will manage devices, caching copies of hte indexer per-device when accessed for that device. Ideally this prevents us from making `IndexManager` public since you never end up declaring an `AnnData` object with it
3. A `numpy` version of the indexer (using using `dlpack` inside `IndexManager`) is generated and used for the `pandas.DataFrame` (which could be relaxed once we have dataframes on different devices: https://github.com/scverse/anndata/issues/2043)
4. The `IndexManager` is stored in `._{v,o}idx`

Then when the view is materialized (i.e., `adata[subset].obsm["foo"]`) our `_subset` method does the following:

1. If we know ahead of time (true for most existing classes) that the incoming indexer should be `numpy`-ified in order to subset an element, a decorator handles converting `IndexManager` to `numpy.ndarray`
2. For an array-api element, the indexer needed to subset the element on its device for its library is materialized using `dlpack` from the given library (0-copy if the indexer already exists on the device and cached otherwise).  `dlpack` provides a 0-copy way of going between different array-api libraries within a given device so if this element is from a different library but exists on the same device as an already subsetted element, the `IndexManager` can use the cached version!

## Array elements

As alluded before, `_susbet` is implemented for the array-api.  We have a custom `np.ix_` function written for the array api (really just a vendored version) and otherwise it is quite simple.

For writing, there is a custom function that brings everything to numpy (via dlpack) before writing to disk which for now makes sense but in the future we could relax this if we ever do GPU-direct writing.


## Types

I basically had to set up two sets of types because we now have internal functions that handle `IndexManager` as well as external types that are almost identical except for `IndexManager`.  Not sure there's a better way to handle this...open to ideas!

## Tests

We tried to add explicit tests especially around edge-cases.  Also, most tests have been updated to be parametrized by a `jax.Array` if they were already parametrized on the array type where it makes sense to do so. To this end there is now an `array_api` `pytest.mark` that we can select for.  This is done for a separate suite of tests (so as not to pollute dependencies with `jax`) as well as for the GPU tests (so we test `jax` on GPU as well).

Hope this helps!

pyproject.toml
```diff
@@ -5,7 +5,7 @@ requires = [ "hatchling", "hatch-vcs" ]
 [project]
 name = "anndata"
 description = "Annotated data."
-requires-python = ">=3.12"
+requires-python = ">=3.12,<=3.14.0"
```
flying-sheep: please revert.
1. uv ignores this anyways, so it’s non-functional
2. if it worked, this would be a regression since we already support it
3. we decided not to do upper bounds

pyproject.toml
```diff
@@ -103,7 +103,14 @@ test-min = [
...
 ]
-
+test-array-api = [
+    { include-group = "test-min" },
+    { include-group = "test-jax" },
+]
+test-jax = [
+    "jax>=0.6.0",
+    "jaxlib>=0.6.0",
+]
```
flying-sheep, line 112: I guess the reason why you made our test matrix more complicated is that Jax has some weird dependencies or so? Could you please shine a light on that?

tests/test_x.py
```diff
@@ -10,16 +10,30 @@
...
+
+
 UNLABELLED_ARRAY_TYPES = [
     pytest.param(sparse.csr_matrix, id="csr"),
     pytest.param(sparse.csc_matrix, id="csc"),
     pytest.param(sparse.csr_array, id="csr_array"),
     pytest.param(sparse.csc_array, id="csc_array"),
     pytest.param(asarray, id="ndarray"),
-]
+] + (
```
flying-sheep: please never use + for lists, always unpacking.

tests/test_views.py
flying-sheep: This is some nasty long test! Please add subtests to stuff like this when you’re making it even more complex.

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+        ),
+    ],
+    ids=[
+        "integer-array",
+        "slice",
+        "ellipsis",
+        "two-axis-integer-arrays",
+        "two-axis-boolean-arrays",
+        "mixed-array-type",
+    ],
```
flying-sheep: `ids=[…]` is for when both params and `ids` fit on one line together (or at least almost fit) like when giving ids to `[True, False]`.

When they don’t fit on one line together, use `pytest.param(..., id=...)` instead.

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+)
+@pytest.mark.array_api
+def test_index_into_jax(index):
+    adata = ad.AnnData(X=np.ones((10, 10)))
+    adata_as_jax = ad.AnnData(X=jnp.ones((10, 10)))
+    assert_equal(adata[index], adata_as_jax[index])
```
flying-sheep: Maybe use random values instead? I feel like it can easily happen that broken indexing code manages to make shape match but not order.

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+
+
+@pytest.mark.array_api
+def test_normalize_index_jax_boolean():
+    index = pd.Index([f"cell_{i:02d}" for i in range(10)])
+    mask = jnp.array([True, False, True, False, True, False, True, False, True, False])
```
flying-sheep:
```suggestion
    mask = jnp.array([True, False] * 5)
```

src/anndata/tests/helpers.py
```diff
@@ -64,6 +66,23 @@
         *(pd.UInt8Dtype, pd.UInt16Dtype, pd.UInt32Dtype, pd.UInt64Dtype),
     )
 
+
+def get_jnp_or_none() -> None | ModuleType:
```
flying-sheep: Please get rid of this function, and do the `jax.config.update` in `testing.anndata._pytest`, which you already modify in this PR. Then:

1. either get rid of the following constant and just import it in the individual test functions instead, and for the `isinstance(thing, jnp.ndarray)`, you can add our classic `if TYPE_CHECKING or find_spec(...)` import.
2. or do this: https://play.ty.dev/8c635b98-93c4-418c-812f-74374d87de90

src/anndata/compat/__init__.py
```diff
@@ -476,3 +563,7 @@ def _safe_transpose(x):
         return _transpose_by_block(x)
     else:
         return x.T
+
+
+def has_xp(x):
```
flying-sheep: Please unify approach with `SupportsArrayApi`.

If both “has array namespace” and “supports all of array api that we use” need to be different checks, make sure either

1. to add `array-api-typing` to the dev dependencies, then make both into functions that return `TypeGuard[xpt.SomeProtocol]`, e.g. this would return `TypeGuard[xpt.HasArrayNamespace]`
2. or you make both into runtime checkable protocols and use isinstance (which might mean less code reuse, but unlocks using them in singledispatch’s .register functions)

src/anndata/compat/__init__.py
```diff
@@ -70,6 +146,17 @@ class Empty(Enum):
...
+_Index = (
+    _IndexRest
+    | tuple[_Index1D, _IndexRest]
+    | tuple[_IndexRest, _Index1D]
+    | tuple[_Index1D, _Index1D, EllipsisType]
+    | tuple[EllipsisType, _Index1D, _Index1D]
+    | tuple[_Index1D, EllipsisType, _Index1D]
+    | CSMatrix
+    | CSArray
+)
```
flying-sheep: You can fix the duplication using generics:

```py
_M = TypeVar("_M", IndexManager, Never, default=Never)

Index1DNorm: TypeAlias = slice | ... | _M
Index1D: TypeAlias = ... | IndexNorm[_M] | ... | _M
```

Users will only see `Index`, wich is `Index[Never]` due to the default. `Never` just collapses into nothing in a union.

We can internally use `Index1D[IndexManager]`.

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+    assert out.shape == (10,)
+    assert out.dtype == jnp.bool_
+
+
+@pytest.mark.array_api
+def test_normalize_index_jax_float_valid():
+    index = pd.Index([f"cell_{i:02d}" for i in range(10)])
+    idx = jnp.array([0, 2, 4])
+    out = _normalize_index(idx, index)
+    assert (out == jnp.array([0, 2, 4])).all()
```
flying-sheep: maybe this for better errors (here and below)?

```suggestion
    assert out.tolist() == [0, 2, 4]
```

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+
+
+@pytest.mark.array_api
+def test_normalize_index_jax_flatten_2d():
+    index = pd.Index([f"cell_{i}" for i in range(5)])
+
+    # column vector (5,1)
```
flying-sheep: subtests

tests/test_views.py
```diff
@@ -892,22 +931,119 @@ def test_index_float_sequence_raises_error(index):
...
+    # row vector (1,5)
+    idx_row = jnp.array([[0, 1, 2, 3, 4]])
+    out_row = _normalize_index(idx_row, index)
+    assert out_row.shape == (5,)
+    assert isinstance(out_row, jnp.ndarray)
+    assert (out_row == jnp.array([0, 1, 2, 3, 4])).all()
+
+
+@pytest.mark.array_api
+def test_index_jax_both_axes():
```
flying-sheep: This should probably be a parametrization of `tests/test_base.py::test_slicing`, but that test isn’t parametrized yet.

Mind fixing that?

tests/test_concatenate.py
```diff
@@ -1509,8 +1522,23 @@ def expected_shape(
...
     "shape", [pytest.param((8, 0), id="no_var"), pytest.param((0, 10), id="no_obs")]
 )
+@pytest.mark.parametrize(
+    "array_type",
+    [pytest.param(np.array, id="np")]
+    if jnp is None
+    else [
+        pytest.param(np.array, id="np"),
+        pytest.param(jnp.array, id="jax", marks=pytest.mark.array_api),
+    ],
```
flying-sheep: no need for the `... if jnp is None else` part, the `pytest.mark.array_api` param is skipped without jax like the others.

src/anndata/tests/helpers.py
```diff
@@ -584,16 +605,26 @@ def func_wrapper(*args, _elem_name: str | None = None, **kwargs):
 
 
 @report_name
-def _assert_equal(a, b):
+def _assert_equal(a, b, exact):
```
flying-sheep: types

tests/test_io_utils.py
```diff
@@ -110,3 +113,52 @@ class Foo:
 
     with pytest.raises(IORegistryError, match=pattern):
         ad.io.read_elem(group)
+
+
+def test_to_writeable_numpy_passthrough():
```
flying-sheep: all tests you added in this file should be a single parametrized test instead

tests/test_helpers.py
```diff
@@ -323,3 +303,36 @@ def test_as_cupy_dask(request: pytest.FixtureRequest, dask_matrix_type) -> None:
...
+    adata.layers["counts"] = jnp.asarray(adata.X, dtype=jnp.float32)
+
+    assert isinstance(adata.obsm["pca"], jnp.ndarray)
+    assert adata.obsm["pca"].shape == (5, 2)
+
+    assert isinstance(adata.varm["gene_scores"], jnp.ndarray)
+    assert adata.varm["gene_scores"].shape == (5, 3)
+
+    assert isinstance(adata.layers["counts"], jnp.ndarray)
+    assert adata.layers["counts"].shape == adata.X.shape
```
flying-sheep: subtests

src/anndata/compat/__init__.py
```diff
@@ -24,12 +24,82 @@
...
+        """Return numpy array for compatibility with numpy/pandas/sparse operations."""
+        if (1, 0) not in self._manager:
+            # (1, 0) is the default python interpreter key from __dlpack_device__ i.e., where numpy arrays live.
+            # As of 2023 dlpack, it must be possible for a library to export to this, see: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#array_api.array.__dlpack__
+            # However, https://github.com/numpy/numpy/issues/20742 means we can't roundtrip jax arrays using dlpack so better to just let numpy do its thing in asarray.
+            self.add_array(np.asarray(self.get_default()))
+        res = np.from_dlpack(self._manager[(1, 0)])
+        return res.copy() if copy else res
+
+    def keys(self):
```
flying-sheep: type all methods and functions you add, including None return types.

also are all methods of this class actually used?

src/anndata/_core/views.py
```diff
@@ -446,8 +454,37 @@ def _resolve_idxs(
...
 @singledispatch
-def _resolve_idx(old: Index1DNorm, new: Index1DNorm, l: Literal[0, 1]) -> Index1DNorm:
-    raise NotImplementedError
+def _resolve_idx(
+    old: Index1DNorm | IndexManager, new: Index1DNorm, l: Literal[0, 1]
+) -> Index1DNorm | IndexManager:
+
+    from ..compat import has_xp
+
+    if not has_xp(old):  # pragma: no cover
```
flying-sheep: if you use runtime checkable protocols instead of typeguard functions (see other comments), you can .register them instead of having to add code in the fallback branch here.

src/anndata/_core/index.py
```diff
@@ -61,46 +77,75 @@ def name_idx(i):
...
+        ):
             if indexer.shape != index.shape:
                 msg = (
                     f"Boolean index does not match AnnData’s shape along this "
                     f"dimension. Boolean index has shape {indexer.shape} while "
                     f"AnnData index has shape {index.shape}."
                 )
                 raise IndexError(msg)
-            return indexer
-        else:  # indexer should be string array
```
flying-sheep, line 101: why remove that comment? is it no longer true? if so what is `indexer` here now?

src/anndata/_core/index.py
```diff
@@ -61,46 +77,75 @@ def name_idx(i):
...
+        ) or (
+            not is_pandas
+            and not is_numpy_string
+            and xp.isdtype(
+                indexer.dtype, ("signed integer", "unsigned integer", "real floating")
+            )
+        ):
+            return (
+                np.asarray(indexer) if is_pandas else indexer
+            )  # Might not work for range indexes
```
flying-sheep: please fix formatting issues like this by pulling the comment up into the line before the return

src/anndata/_core/index.py
```diff
@@ -61,46 +77,75 @@ def name_idx(i):
         return indexer
     elif isinstance(indexer, str):
         return index.get_loc(indexer)  # int
-    elif isinstance(
+    elif (use_xp := has_xp(indexer)) or isinstance(
```
flying-sheep: This else branch was already pretty spaghetti, now I don’t know what’s going on anymore. Please at least pull out into one function, but if you can also just refactor it into one big match statement, that would be even cleaner. If not, it might be cleaner to break it up into multiple functions that allow to follow the high-level structure of what’s happening instead of just seeing one huge lock of if statements.

Basically, please remove `# noqa: PLR0911, PLR0912` in a way that makes everything clearer.

flying-sheep (changes requested):
I see that this was an enormous piece of work, great job!

My main concern is about making `_normalize_index` even more complex. We need to refactor this one to have any chance at maintaining this. Please do that, I’ll delay reviewing it until then.

I checked out all the rest, looks great apart from my usual nitpicks.

Please remember to just habitually type every single parameter and return type, we should really not increase our technical debt there.

ilan-gold:
@flying-sheep Will make a new PR into this one for `_normalize_index` soon, have something locally that is much clearer, hopefully

pyproject.toml
ilan-gold: Yes they have both `numpy` and `scipy` from what I can tell, so we'd be relying on them not outpacing us in their upgrades: https://github.com/jax-ml/jax/blob/a5854e44cb01c1b6aa24d43ab997e0ed77d8dab6/jaxlib/setup.py#L63-L67

tests/test_concatenate.py
ilan-gold: So if you don't check this, the `jnp.array` call actually occurs and then it tries looking for `array` on `None`.  There might be cleaner ways of doing this but this is what I landed on in `xarray` as well (I think based on what they were doing) when creating params from maybe-installed packages directly

ilan-gold: I guess the other option would be to wrap all of these in a `jnp_array` function that hides `jnp.array` from the `param` i.e., `pytest.param(jnp_array, id="jax", marks=pytest.mark.array_api)`

src/anndata/tests/helpers.py
ilan-gold: > either get rid of the following constant and just import it in the individual test functions instead

The big thing here is param creation so `jnp.xyz` needs to make `xyz` available when `pytest.param` is run or it needs to be checked.  And we don't want to create "empty" params.  I think I should begin by wrapping up `jnp.array` calls in a function and relying on the skip behavior and then go from there

.github/workflows/test-gpu.yml
```diff
@@ -73,14 +73,14 @@ jobs:
...
-          uv pip install -e ".[cu12]" --group=test -c ci/constraints.txt
+          uv pip install -e ".[cu12]" --group=test --group=test-jax -c ci/constraints.txt
 
       - name: Env list
         run: uv pip list
 
       - name: Run tests
         run: |
-          uv run coverage run -m pytest -m gpu -n auto
+          uv run coverage run -m pytest -m "array_api or gpu"
```
flying-sheep: sory if this was answered elsewhere, but why is this not parallel anymore?

src/anndata/_core/anndata.py
```diff
@@ -616,14 +627,20 @@ def X(self, value: _XDataType | None):  # noqa: PLR0912
         # If indices are both arrays, we need to modify them
         # so we don’t set values like coordinates
         # This can occur if there are successive views
+        # Handle IndexManager by extracting array_api compat if possible
+        # Otherwise fall back to numpy
+        oidx, vidx = (
+            idx.get_for_array(self._adata_ref._X)
+            if (is_mgr := isinstance(idx, IndexManager)) and has_xp(self._adata_ref._X)
+            else (np.array(idx) if is_mgr else idx)
```
flying-sheep: Isn’t this clearer?

```suggestion
            (
                idx.get_for_array(self._adata_ref._X)
                if has_xp(self._adata_ref._X)
                else np.array(idx)
            )
            if isinstance(idx, IndexManager) 
            else idx
```

src/anndata/_core/index.py
```diff
@@ -169,11 +271,112 @@ def unpack_index(index: Index) -> tuple[Index1D, Index1D]:
...
+        return func(
+            a,
+            _index_manager_to_numpy_idx_in_tuple(subset_idx),
+        )
+
+    return _ensure
+
+
+def array_api_ix(*args: SupportsArrayApi) -> tuple[SupportsArrayApi, ...]:
+    """Vendored version of numpy's _ix for the array-api"""
```
flying-sheep: You mean `ix_`?


```suggestion
    """Construct an open mesh from multiple sequences.
    
    Vendored version of `numpy.ix_` for the array-api.
    
    For each sequence `args[i]`, it returns an array with `.ndim == len(args)`,
    `.size == .shape[i] == len(args[i])` (i.e. `shape[…] == 1` for each other dimension)
    """
```

src/anndata/_io/specs/registry.py
```diff
@@ -379,6 +396,9 @@ def write_elem(
         elif k in store:
             del store[k]
 
+        # Normalize array-API (e.g., JAX/CuPy) even if not AnnData
+        elem = normalize_nested(elem)
```
flying-sheep, line 400: So we move everything to CPU before writing. Seems reasonable. We can still add some performance optimization here later if someone implements some amazing IO stuff for their array type.

src/anndata/compat/__init__.py
```diff
@@ -17,19 +17,85 @@
...
+        dtype: np.dtype | None = None,
+        copy: bool | None = None,  # noqa: FBT001
+    ) -> np.ndarray:
+        """Return numpy array for compatibility with numpy/pandas/sparse operations."""
+        if (1, 0) not in self._manager:
+            # (1, 0) is the default python interpreter key from __dlpack_device__ i.e., where numpy arrays live.
+            # As of 2023 dlpack, it must be possible for a library to export to this, see: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#array_api.array.__dlpack__
+            # However, https://github.com/numpy/numpy/issues/20742 means we can't roundtrip jax arrays using dlpack so better to just let numpy do its thing in asarray.
+            self.add_array(np.asarray(self.get_default()))
+        res = np.from_dlpack(self._manager[(1, 0)])
```
flying-sheep, line 63: I don’t fully get this: is `np.asarray(self.get_default()).__dlpack_device__() == (1, 0)` always true, i.e. will the `add_array` make the `res = …` line succeed?

(also `self._manager[(1, 0)]` is spelled `self._manager[1, 0]`!)

flying-sheep (approved):
looks great, I just have some clarifying questions!

src/anndata/compat/__init__.py
ilan-gold: > np.asarray(self.get_default()).__dlpack_device__() == (1, 0)  always true, i.e. will the add_array make the res = … line succeed?

Yup, numpy doesn't support other devices AFAIK

ilan-gold: > (also self._manager[(1, 0)] is spelled self._manager[1, 0]!)

I feel like I'm missing something, I don't see `self._manager[1, 0]` in the codebase.  What do you mean?

flying-sheep: just a minor style thing: you wouldn’t do `some_array[(0, :)]`, so why do it for `_manager`?

.github/workflows/test-gpu.yml
```diff
@@ -53,8 +53,9 @@ jobs:
         run: nvidia-smi
 
       - name: Install yq # https://cirun.slack.com/archives/C09SNDRB3A8/p1766512487317849?thread_ts=1766512112.938459&cid=C09SNDRB3A8
+        # https://github.com/mikefarah/yq/issues/2592 forces version pin
         run: |
-          sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
+          sudo wget -qO /usr/local/bin/yq http://github.com/mikefarah/yq/releases/download/v4.52.2/yq_linux_amd64
```
flying-sheep: hmm, this doesn’t seem to work. maybe caused by dependency update? (also always use https)

```suggestion
          sudo wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/v4.52.2/yq_linux_amd64
```

src/anndata/compat/__init__.py
flying-sheep, line 63:
```suggestion
        res = np.from_dlpack(self._manager[1, 0])
```

ilan-gold: Because `_manager` is a dict mapping keys of dlpack device (which are tuples) to arrays

ilan-gold (approved):

PR #2087, opened 2025-08-21
feat: Support sparrays in dask

### LINKED ISSUE #2085: Support scipy sparse arrays in dask

Maybe after #2084, because it needs some fiddling with dask versions: it went into 2025.3.0, but for some undocumented(!) reason we upper bound dask.

See also #1088

flying-sheep:
- [x] Closes #2085
- [x] Tests added


flying-sheep:
@ilan-gold I changed the helper so it doesn‘t just convert things into `csr_matrix`.

Does it make sense that Dask arrays can contain any of `cs{rc}_{array,matrix}` (and therefore the way I adapted the failing test `test_dask_to_memory_unbacked`)

src/anndata/tests/helpers.py
```diff
@@ -878,29 +888,31 @@ def _(a):
...
-def _(a):
-    return a.map_blocks(sparse.csr_matrix)
+def _(a: DaskArray) -> DaskArray:
+    return a.map_blocks(_as_sparse_for_dask)
+
+
+def _as_sparse_for_dask(a: NDArray | CSArray | CSMatrix) -> CSArray | CSMatrix:
+    """Convert into a a sparse container that dask supports (or complain)."""
+    if isinstance(a, CSArray) and not DASK_CAN_SPARRAY:  # convert sparray to spmatrix
+        msg = "Dask <2025.3 without fast-array-utils doesn’t support sparse arrays"
```
ilan-gold, line 909: Should this message mention something like `convert to spmatrix`? Or did you want to do that here (given the comment)?

src/anndata/tests/helpers.py
```diff
@@ -878,29 +888,31 @@ def _(a):
...
+        raise TypeError(msg)
+    if isinstance(a, CSArray | CSMatrix):
+        return a
+    return (sparse.csr_array if DASK_CAN_SPARRAY else sparse.csr_matrix)(a)
```
ilan-gold: I don't want to touch this default case, for now, or at least we should make sure it is not being hit.  If there are situations in which it is being hit, I would think we should add tests for `csr_array` as well as `csr_matrix` (since we are aware that they can have non-overlapping issues)

src/anndata/tests/helpers.py
flying-sheep: this function leaves sparrays as sparrays, so this checks that we never run it with a nonsensical argument.

src/anndata/tests/helpers.py
flying-sheep: OK, now I split it up into two `partial`s instead.

src/anndata/tests/helpers.py
```diff
@@ -878,29 +882,37 @@ def _(a):
...
+        msg = "Dask <2025.3 without fast-array-utils doesn’t support sparse arrays"
+        raise TypeError(msg)
+    return typ(a)
 
-@as_sparse_dask_array.register(DaskArray)
-def _(a):
-    return a.map_blocks(sparse.csr_matrix)
+
+as_sparse_dask_array = partial(_as_sparse_dask, typ=sparse.csr_array)
+as_sparse_dask_matrix = partial(_as_sparse_dask, typ=sparse.csr_matrix)
```
ilan-gold, line 915: This is the reverse of what we had before, no? (I'm open to changing the names, but again, I'd rather not have the test behavior changed).

The other option would be to have `as_sparse_dask_matrix` alongside `as_sparse_dask_array` wherever the old `as_sparse_dask_array` appears

flying-sheep: > The other option […]

that’s exactly what I’m doing!

ilan-gold: Awesome just wanted to be sure :)

ilan-gold (approved):

PR #2088, opened 2025-08-21
ci: Check PRs for semantic labels

flying-sheep:
… and adapts towncrier fragments to match the labels

Also:
- cleans up CI (remove useless options, update actions)
- adds dependabot for automatic GH actions updates
- uploads test results to codecov

Does not:
- check if PR title matches fragment (TODO? how to deal with breaking vs “!”)?


pyproject.toml
```diff
@@ -250,10 +250,13 @@ single_file = false
...
+# Valid fragments should be a subset of conventional commit types (except for `breaking`):
+# https://github.com/commitizen/conventional-commit-types/blob/master/index.json
+# style, refactor, test, build, ci: should not go into changelog
+fragment.feat.name = "Features"
+fragment.fix.name = "Bug fixes"
+fragment.docs.name = "Documentation"
+fragment.perf.name = "Performance"
+fragment.chore.name = "Miscellaneous changes"
+fragment.revert.name = "Revert"
+fragment.breaking.name = "Breaking changes"   # add `!` to commit type (e.g. “feature!:”)
```
flying-sheep, line 262: please double-check if I got them all right. I didn’t add a test, but these don’t change so it should be fine.

PR #2093, opened 2025-08-25
Add missing keys methods

lazappi:
Add `layers_keys()`, `obsp_keys()` and `varp_keys()` as they were missing for these slots and it's more convenient than `list(adata.layers.keys())`

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -704,7 +704,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):  # noqa: PLW1641
 
     Return layers’ names::
 
-        adata.layers.keys()
+        adata.layers_keys()
     """
 
     @property
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -140,7 +140,7 @@ def test_assert_equal():
     #     exact=False,
     # )
     adata2 = adata.copy()
-    to_modify = next(iter(adata2.layers.keys()))
+    to_modify = next(iter(adata2.layers_keys()))
     del adata2.layers[to_modify]
     with pytest.raises(AssertionError) as missing_layer_error:
         assert_equal(adata, adata2)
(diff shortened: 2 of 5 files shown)


flying-sheep:
I completely forgot that these APIs exist.

I think the only reason `obs_keys` makes the slightest bit of sense is that `obs.keys()` is an `pandas.Index` which doesn’t subclass `Sequence`.

The other ones make no sense at all, `KeysView` is a much better return type than `list` (as it’s an ordered immutable set, which therefore supports fast `__contains__` and other set operations)

I think we should deprecate these APIs, what do you think @ilan-gold?

ilan-gold:
I will rarely say "no" to deprecating APIs :) I think people can manage with `obs.keys()`

lazappi:
For context, this came up because the accessors that return lists get converted automatically in R while `KeysView` needs to be converted manually. This is a niche case though which shouldn't change your decision. In **{anndataR}** we have all the `*_keys()` methods I added here but we could also remove them if they are going to be deprecated in Python.

flying-sheep:
> get converted automatically in R

You mean by `reticulate`? It doesn’t even convert sets or bytes, right? How do `reticulate` users deal with cases like that? I think there should be a reticulate utility that converts
- Python mappings to R named lists (error if the keys aren’t strings)
- other Python iterables to lists

I also think there’s nothing wrong with anndataR having some convenience methods. We won’t add anything that conflicts with these names and does something completely different, right?

lazappi:
> You mean by `reticulate`? It doesn’t even convert sets or bytes, right? How do `reticulate` users deal with cases like that? I think there should be a reticulate utility that converts
> 
> * Python mappings to R named lists (error if the keys aren’t strings)
> * other Python iterables to lists

Yeah. Normal lists and dicts get converted but not other types. It's easy enough to add a conversion method if we need to to though.

> I also think there’s nothing wrong with anndataR having some convenience methods. We won’t add anything that conflicts with these names and does something completely different, right?

Makes sense. We're trying to keep things as close as possible but there are some little things where it's worth having some minor changes.

Feel free to close this if you are going to deprecate the current methods. I think it makes sense to either have them for all slots or no slots though.

flying-sheep:
> I think it makes sense to either have them for all slots or no slots though.

for sure!

PR #2095, opened 2025-08-27
fix: most immutable setting situtations

ilan-gold:
This is a PR to help push things along a bit here.  There are two commits here:

1. Bring back `backed` mode in its former state
2. Update the logic in initialization so that (a) `Raw` is initialized with an `AnnData` object that has a valid `layers` (hence the change of where the `Raw` declaration happens) and ~~(b) better handle the interaction between `layers` and `X`~~
3. EDIT: various small sharp edges

~~There are still quite a few errors but at least they fall into only a few categories I can address a bit more clearly now, I think.~~ Most of the remaining issues have to do with setting on a view or similar now, which is a bit fraught because setting on `X` is special-cased

One of the issues with the current logic is that `layers` now may or may not contain `None` but in `backed` mode, it never does even if `X` is not `None`, which is a problem for testing but I think also a bit of a rough edge.  I'm going to play around with holding a `SparseDataset` class in memory at `None` and seeing how that goes.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -590,7 +590,11 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):  # noqa: PLW1641
                 X[oidx, vidx] = value
             else:
                 self._set_backed("X", value)
-        self.layers[None] = value
+        # TODO: should we add support for `layers[None] = None`?
+        if value is not None:
+            self.layers[None] = value
+        else:
+            self.layers.pop(None)
 
     @X.deleter
     def X(self) -> None:


src/anndata/_core/anndata.py
```diff
@@ -448,19 +451,24 @@ def _init_as_actual(  # noqa: PLR0912, PLR0913, PLR0915
...
-            self._raw = Raw(self, raw.X, raw.var, raw.varm)
 
         # clean up old formats
         self._clean_up_old_format(uns)
 
         # layers
+        if X is not None:
+            if layers is None or not bool(layers):
+                layers = {None: X}
+            elif bool(layers):
```
ilan-gold: I think this can be `else`?

ilan-gold:
>  I'm going to play around with holding a SparseDataset class in memory at None and seeing how that goes.

So the issue here is apparently doing anything after the object is a view and the file is closed.  You can't just reopen it so something like 

```python
backed_v = backed_adata[obs_idx, var_idx]
backed_v.to_memory() # closes file handle
backed_v.write_h5ad(final_pth)
```

doesn't work because `backed_v` has a closed file handle and thus can no longer read data.  We could change the behavior of `to_memory` to not close the file, but that would be a breaking change, and I'm not confident enough to say it is even a good idea.

src/anndata/_core/anndata.py
flying-sheep: yeah, if `None` and `False` mean the same, just don’t distinguish between them, they’re both falsy. Please read the last sentence of the first paragraph here: https://docs.python.org/3/reference/expressions.html#boolean-operations

Your logic here is:

```py
if layers:
    layers[None] = X
else:
    layers = {None: X}
```

is mutating the `layers` parameter OK?

src/anndata/_core/anndata.py
```diff
@@ -537,15 +544,60 @@ def shape(self) -> tuple[int, int]:
...
+        else:
+            oidx, vidx = self._oidx, self._vidx
+        if (
+            np.isscalar(value)
+            or (hasattr(value, "shape") and (self.shape == value.shape))
+            or (self.n_vars == 1 and self.n_obs == len(value))
+            or (self.n_obs == 1 and self.n_vars == len(value))
+        ):
+            if self.is_view:
+                X = self.file["X"]
```
flying-sheep, line 587: this is called even when `not self.isbacked` …

src/anndata/_core/anndata.py
ilan-gold: Hmm good point, wouldn't want to touch that.  Let me try to work around that

src/anndata/_core/anndata.py
ilan-gold: Yup, looking now

src/anndata/tests/helpers.py
```diff
@@ -848,9 +848,17 @@ def fmt_name(x):
         "varp",
         "raw",
     ]:
+        a_elem, b_elem = getattr(a, attr), getattr(b, attr)
+        # TODO: This is helpful in backed mode where `X is not None` but `None not in layers`.
+        # Does this filter make sense in general? Is there a case where we explicitly want to check `None in layers`?
```
flying-sheep, line 853: Yeah, backed mode complicates things.

My plan for the PR was that these two invariants should hold:
- `adata.X is adata.layers.get(None)`
- `(adata.X is None) is (None not in adata.layers)`

I get why you’re going the route of making `X`’s getter/setter more complex instead of `layers`, but I think my initial idea is more obvious from an API standpoint, don’t you think?

ilan-gold: Totally agree, and I absolutely would go that way, but to do that we would need to deprecate parts of our current `X` functionality.  Things like
```python
import anndata as ad, numpy as np
adata = ad.AnnData(np.random.rand(3, 10), layers = { "a_layer":np.random.rand(3, 10) })
# updates adata.X
adata[:2].X = 1
# does not update adata.layers["a_layer"]
adata[:2].layers["a_layer"] = np.ones((2, 10))
```
actually do two different things so if were to route the former through the later, we would be dropping functionality that the former offers.  That's the same reason `adata.X is adata.layers.get(None)` doesn't hold; if it were to hold, we would be dropping functionality from what I can tell (in this case, `backed` mode more or less, where we never hold an array-like reference to the object, instead reopening the file on access).  Like I mentioned [here](https://github.com/scverse/anndata/pull/2095#issuecomment-3228407172), I would love to be able to remedy this, but we would be breaking even more currently standardized behavior then.

Maybe I'm missing what you're saying, but from my perspective, that was most of the idea with this PR - to bring back this functionality (largely).

ilan-gold: In summation

1. backed mode currently has various edge cases, like never holding anything other than a reference to a file handle for `X`
2. Setting on `adata.X` has fundamentally different semantics than other elems

If we broke both of these, then I could see your invariants holding / complexity being removed.

flying-sheep: yeah, that’s what I meant with 

> I get why you’re going the route of making `X`’s getter/setter more complex instead of `layers`

→ in order to make it all work transparently, we’d have to modify the `Layers*` classes more heavily to make all of this work the same.

but anyway, your approach here is “in backed mode, `.layers[None]` behaves as if `.X` was `None`”?

tests/test_views.py
```diff
@@ -496,8 +496,15 @@ def test_view_delattr(attr, subset_func):
...
-    assert_equal(getattr(subset, attr), getattr(empty, attr))
+    # Should now have same value as default, except for `layers`, which still has the `None` key for `subset`
+    if attr == "layers":
+        assert_equal(
+            {k: v for k, v in getattr(subset, attr).items() if k is not None},
+            getattr(empty, attr),
+        )
+    else:
+        assert_equal(getattr(subset, attr), getattr(empty, attr))
+
```
flying-sheep, line 507: I wonder if we should make `Layers` with only `None` inside falsey.

That way `del ad.layers; assert not layers` would work.

PR #2097, opened 2025-08-29
fix: reallow h5 slash writing

### LINKED ISSUE #2039: Request: Migration support for `/` character restriction in keys
## Description
I understand the need to restrict `/` characters in variable names to fix the storage issues described in #1834. However, this change will affect existing datasets and workflows. Could we explore adding migration support to help users transition smoothly?

In my use case of flow cytometry data with markers like `LIVE/DEAD`, `CD14/19`, they appear very frequently as naming conventions (e.g., https://readfcs.lamin.ai/quickstart) and maybe for other complex biological data types as well. I need to now manually fix the `/` in all datasets in both `.var.columns` and `.uns`. This will continue to impact future public datasets.

## Problems with current approach
- **No deprecation warning**: Users get no advance notice their code will break
- **No migration utility**: No tool provided to help users update existing datasets  
- **No backwards compatibility**: Existing `.h5ad` files with `/` in keys become unusable without migration
- **Silent breaking change**: Users only discover the problem when their production code fails

## Thoughts
I really appreciate the work being done to fix the storage issues! But I think it'd be best to have a transition phase with a deprecation warning. 

In addition, would any of the following be feasible?

1. **Add a backwards compatibility flag**
2. **Automatic character escaping** - Handle the escaping automatically at the storage layer so users don't need to change their code
3. **Add a deprecation warning and provide a utility function** for existing datasets

This would help users like me transition our existing datasets without having to manually hunt down and rename all the problematic variable names.

Thanks for considering this!

ilan-gold:
- [x] Closes #2039
- [x] Tests added

### DIFF
--- a/tests/test_readwrite.py
+++ b/tests/test_readwrite.py
@@ -967,13 +967,22 @@ def test_h5py_attr_limit(tmp_path):
 @pytest.mark.parametrize(
     "elem_key", ["obs", "var", "obsm", "varm", "layers", "obsp", "varp", "uns"]
 )
-def test_forward_slash_key(elem_key, tmp_path):
+@pytest.mark.parametrize("store_type", ["zarr", "h5ad"])
+def test_forward_slash_key(
+    elem_key: Literal["obs", "var", "obsm", "varm", "layers", "obsp", "varp", "uns"],
+    tmp_path: Path,
+    store_type: Literal["zarr", "h5ad"],
+):
     a = ad.AnnData(np.ones((10, 10)))
     getattr(a, elem_key)["bad/key"] = np.ones(
         (10,) if elem_key in ["obs", "var"] else (10, 10)
     )
-    with pytest.raises(ValueError, match="Forward slashes"):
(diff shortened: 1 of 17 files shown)


tests/test_readwrite.py
```diff
@@ -967,13 +967,34 @@ def test_h5py_attr_limit(tmp_path):
...
 )
-def test_forward_slash_key(elem_key, tmp_path):
+@pytest.mark.parametrize(
+    ("store_type", "disallow_forward_slash_in_h5ad"),
+    [
+        pytest.param("zarr", False, id="zarr"),
+        pytest.param("h5ad", True, id="h5ad-no-slash-allowed"),
+        pytest.param("h5ad", False, id="h5ad-slash-allowed"),
+    ],
+)
```
flying-sheep: I’d just make this 2 params covering all 4 cases.

flying-sheep (commented):
looks great! don’t forget to add the issue tracking the removal to the next version’s milestone.

flying-sheep (approved):

PR #2115, opened 2025-09-15
Fix coverage

flying-sheep:
One more thing was necessary to have this work like `hatch test -pc`.


flying-sheep:
just requesting your review so you can see. it works!

PR #2121, opened 2025-09-16
perf: use `name` in `map_blocks` to bypass tokenization

### LINKED ISSUE #1989: `ad.concat` is slow on lazy data on account of `tokenize`

ilan-gold:
- [x] Closes #1989 
- [x] Tests added (benchmark)

### DIFF
--- a/benchmarks/benchmarks/sparse_dataset.py
+++ b/benchmarks/benchmarks/sparse_dataset.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import tempfile
 from types import MappingProxyType
 
 import numpy as np
@@ -81,7 +80,9 @@ class SparseCSRContiguousSlice:
 
 
 class SparseCSRDask:
-    def setup(self):
+    filepath = "data.zarr"
+
+    def setup_cache(self):
         X = sparse.random(
             10_000,
             10_000,
@@ -89,9 +90,11 @@ class SparseCSRDask:
             format="csr",
             random_state=np.random.default_rng(42),
         )
-        self.tmpdir = tempfile.TemporaryDirectory("tmp.zarr")
-        self.group = zarr.group(self.tmpdir.name)
-        write_elem(self.group, "X", X)
+        g = zarr.group(self.filepath)
+        write_elem(g, "X", X)
+
+    def setup(self):
+        self.group = zarr.group(self.filepath)
         self.adata = AnnData(X=read_elem_lazy(self.group["X"]))
 
     def time_concat(self):
@@ -105,6 +108,3 @@ class SparseCSRDask:
 
     def peakmem_read(self):
         AnnData(X=read_elem_lazy(self.group["X"]))
-
-    def teardown(self):
-        self.tmpdir.cleanup()


src/anndata/_io/specs/registry.py
```diff
@@ -459,11 +459,11 @@ def read_elem_lazy(
 
     >>> g = zarr.open(zarr_path)
     >>> adata.X = ad.experimental.read_elem_lazy(g["X"])
-    >>> adata.X
-    dask.array<make_dask_chunk, shape=(2700, 32738), dtype=float32, chunksize=(1000, 32738), chunktype=scipy.csr_matrix>
+    >>> adata.X.chunksize  # default chunk settings
+    (1000, 32738)
```
flying-sheep: why this change? isn’t it good to see all the info in the docs?

ilan-gold: Yes but the `make_dask_chunk` is now some uuid name so I figured it better to check something rather than nothing.  I am going to alter this anyway, because using `name` with the sparse-dask stuff doesn't seem to have an impact on performance.

flying-sheep (commented):
Can you explain the idea behind the `uuid`? Why is this the right thing to do for this use case and not other uses of `map_blocks`? Is there anything in the dask docs that recommends this pattern for certain use cases?

ilan-gold:
> Can you explain the idea behind the uuid? 

I just want to be 100% sure the name I create is unique.  [From the docs](https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html):

```
The key name to use for the output array. Note that this fully specifies the output key name, and must be unique. If not provided, will be determined by a hash of the arguments.
```

> Why is this the right thing to do for this use case

I want to avoid the tokenization of the input function, which appears to be expensive in certain cases.  I had originally added `name` everywhere we do `map_blocks` but this was overkill it seems and does not seem to have affected performance.

flying-sheep (approved):
I see! Weird that adding it everywhere didn’t affect performance if adding it here does. Maybe we can form an intuition on where it’s expensive so we don’t have to guess/measure as much?

ilan-gold:
Requested review based on the last non-merge commit @flying-sheep because I added an optimization for something I noticed in #2156 , although I don't know why it fails for Selman (it worked for me but was slow, hence the additional commit after your approval)

benchmarks/benchmarks/sparse_dataset.py
```diff
@@ -77,3 +77,34 @@ def peakmem_getitem_adata(self, *_):
         res = self.adata[self.index]
         if isinstance(res, DaskArray):
             res.compute()
+
+
+class SparseCSRDask:
```
ilan-gold, line 82: I considered use `name` for the sparse block mapping but it had no appreciable effect: https://github.com/scverse/anndata/runs/51718340360 is the result and https://github.com/scverse/anndata/blob/59041d4c25f7443e8999c15695213d066053a713/src/anndata/_io/specs/lazy_methods.py#L172-L178 shows that the commit on which the benchmark was run contained the `name` parameter. I leave the benchmark in anyway

ilan-gold:
Ok @flying-sheep I know you reviewed, but the PR is grossly simplified now and hopefully the becnhmark results make some sense i.e., the big change is `concat`

Issue #1854, opened 2025-02-12
concat_on_disk fails to write alternative axis mapping and uns


milos7250:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of anndata.
- [ ] (optional) I have confirmed this bug exists on the master branch of anndata.

### Report

When concatenating any anndata files that contain a mapping for alternative axis (e.g. concatenating along `obs` and having `varm`), the command fails. Moreover, the `uns_merge` argument is completely ignored. 

Code:

```python
import anndata as ad

ad.experimental.concat_on_disk(
    in_files={"1": "sample1.h5ad", "2": "sample2.h5ad"},
    out_file="output.h5ad",
    max_loaded_elems=int(1e9),
    axis=0,
    join="outer",
    merge="unique",
    uns_merge="unique",
    index_unique="-",
)
```

Traceback:

```pytb
Traceback (most recent call last):
  File "/mnt/shared/scratch/mmicik/rna-scripts/merge_adatas.py", line 68, in <module>
    ad.experimental.concat_on_disk(
  File "/mnt/apps/users/mmicik/conda/envs/rna-python/lib/python3.12/site-packages/anndata/experimental/merge.py", line 630, in concat_on_disk
    _write_alt_mapping(groups, output_group, alt_axis_name, alt_index, merge)
  File "/mnt/apps/users/mmicik/conda/envs/rna-python/lib/python3.12/site-packages/anndata/experimental/merge.py", line 381, in _write_alt_mapping
    write_elem(output_group, alt_axis_name, alt_mapping)
  File "/mnt/apps/users/mmicik/conda/envs/rna-python/lib/python3.12/site-packages/anndata/_io/specs/registry.py", line 487, in write_elem
    Writer(_REGISTRY).write_elem(store, k, elem, dataset_kwargs=dataset_kwargs)
  File "/mnt/apps/users/mmicik/conda/envs/rna-python/lib/python3.12/site-packages/anndata/_io/utils.py", line 249, in func_wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/apps/users/mmicik/conda/envs/rna-python/lib/python3.12/site-packages/anndata/_io/specs/registry.py", line 354, in write_elem
    return write_func(store, k, elem, dataset_kwargs=dataset_kwargs)
[...]

PR #2122, opened 2025-09-16
fix: merge strategies for `concat_on_disk`

### LINKED ISSUE #2110: Concatenating on disk doesn't respect the `merge` argument.

### Report

# Description
I have a set of `zarr` AnnData with two separate variables in the `var` object: `symbol` and `ensembl_id`.

If I load all of these objects and merge them via `ad.concat`, we can merge the `var` columns via `merge="same"` or `merge="first"`, etc.

However, if we use the `ad.experimental.concat_on_disk`, the `merge` argument is ignored, and no parameter allows us to merge the values.

# Code

Does not work:
```python
ad.experimental.concat_on_disk(
        in_files=adata_files, 
        out_file=output_file,
        join="outer",
        merge="same"  # or first, unique, etc. doesn't work
    )
```

Does work:
```
adata_objs = [
    ad.read_zarr(f) for f in adata_files
]
adata = ad.concat(adata_objs, join="outer", merge="same")
```

### Versions

```python
# the above gave me an issue related to a missing NVIDIA file? 
# Attaching the following instead:
>>> uv pip show anndata
Name: anndata
Version: 0.13.0.dev26+gc5184ae20
Location: /host_root/usr/repos/P06086_ad_neuroinflammation/.venv/lib/python3.12/site-packages
Requires: array-api-compat, h5py, legacy-api-wrap, natsort, numpy, packaging, pandas, scipy, zarr
Required-by: cellxgene-census, mudata, muon, pyclustree, rapids-singlecell, scanpy, scarches, schpl, scib, scib-metrics, scvi-tools, somacore, tiledb-cloud, tiledbsoma
```

ilan-gold:
I'm putting this in the soonest upcoming milestone because

a) I think it's a bug that we didn't have any sort of warning or anything about this and
b) It's in `experimental` and this isn't even an API change

~~As for the nature of the changes, I still need to look into why `read_elem_lazy` produces different results in tokenization for dask dense than `read_elem` on the off-axis.  Otherwise, this could be fully lazy~~ Bug resolved

- [x] Closes #2110 and part of #1854
- [x] Tests added


src/anndata/experimental/merge.py
```diff
@@ -376,17 +376,27 @@ def _write_concat_sequence(  # noqa: PLR0913, PLR0917
         raise NotImplementedError(msg)
 
 
-def _write_alt_mapping(groups, output_group, alt_axis_name, alt_indices, merge):
-    alt_mapping = merge([read_as_backed(g[alt_axis_name]) for g in groups])
-    # If its empty, we need to write an empty dataframe with the correct index
-    if not alt_mapping:
-        alt_df = pd.DataFrame(index=alt_indices)
-        write_elem(output_group, alt_axis_name, alt_df)
```
ilan-gold, line 384: So you might wonder why completely ignoring an argument and also writing to the wrong key for a given function wouldn't produce an error previously.  Crazily enough it was that

1. `alt_mapping` was always `{}` here because `merge` is always `None` in the tests
2. We then write a dataframe (because `not {}` is `True`) to the wrong key `alt_axis_name` instead of `f"{alt_axis_name}m"` but its index is correct
3. We never actually checked the off-axis dataframe in the tests because `merge` was `None` on the groundtruth as well for `anndata.concat` so all of the other columns were dropped in the groundtruth concatenation

src/anndata/_core/merge.py
```diff
@@ -143,11 +143,12 @@ def equal_dask_array(a, b) -> bool:
...
-    if isinstance(a._meta, CSMatrix | CSArray):
-        # TODO: Maybe also do this in the other case?
-        return da.map_blocks(equal, a, b, drop_axis=(0, 1)).all()
-    else:
-        return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all()
+    if a.chunksize == b.chunksize:
+        if isinstance(a._meta, CupySparseMatrix | CSMatrix | CSArray):
+            # TODO: Maybe also do this in the other case?
+            return da.map_blocks(equal, a, b, drop_axis=(0, 1)).all()
+        return da.equal(a, b, where=~(da.isnan(a) + da.isnan(b))).all().compute()
```
ilan-gold, line 150: Genuinely don't know how this was ever working with `da.isnan(a) == da.isnan(b)` would basically always return all True, so the `where=~()` would always be False, which meant nothing was checked for equality.  Or am I missing something?

So I changed to `da.isnan(a) + ds.isnan(b)` so that we check where either or b are not nan.  But maybe this check is wrong?

src/anndata/_core/merge.py
```diff
@@ -143,11 +143,12 @@ def equal_dask_array(a, b) -> bool:
...
-    if isinstance(a._meta, CSMatrix | CSArray):
-        # TODO: Maybe also do this in the other case?
-        return da.map_blocks(equal, a, b, drop_axis=(0, 1)).all()
-    else:
-        return da.equal(a, b, where=~(da.isnan(a) == da.isnan(b))).all()
+    if a.chunksize == b.chunksize:
+        if isinstance(a._meta, CupySparseMatrix | CSMatrix | CSArray):
+            # TODO: Maybe also do this in the other case?
+            return da.map_blocks(equal, a, b, drop_axis=(0, 1)).all()
+        return da.equal(a, b, where=~(da.isnan(a) * da.isnan(b))).all().compute()
```
ilan-gold: Genuinely don't know how this was ever working with da.isnan(a) == da.isnan(b) which would basically always return all True, so the where=~() would always be False, which meant nothing was checked for equality. Or am I missing something?

So I changed to da.isnan(a) * ds.isnan(b) so that we check for equality where either a or b are not nan. But maybe this check is wrong?

flying-sheep: Yeah, I also think the operation is supposed to mean “check if everything is equal, treating `nan` as equal. So I think you’re right, we should

> check for equality where either a or b are not nan

Except that I’d phrase that as `a XOR b` to be precise.

I’m only confused why you’re using number arithmetic for bools. `*` should be `&`, right?

```suggestion
        return da.equal(a, b, where=~(da.isnan(a) & da.isnan(b))).all().compute()
```

So 

```pycon
>>> # before (last element wrong)
>>> cmp = lambda a, b: np.equal(a, b, where=~(np.isnan(a) == np.isnan(b)))
>>> cmp([0, np.nan, 0, 0], [0, np.nan, np.nan, 1])
array([ True,  True, False,  True])
>>> # after (all correct)
>>> cmp = lambda a, b: np.equal(a, b, where=~(np.isnan(a) & np.isnan(b)))
>>> cmp([0, np.nan, 0, 0], [0, np.nan, np.nan, 1])
array([ True,  True, False, False])
```

src/anndata/experimental/merge.py
```diff
@@ -408,6 +424,30 @@ def _write_axis_annot(  # noqa: PLR0917
     write_elem(output_group, axis_name, concat_annot)
 
 
+def _write_uns(
+    groups: Collection[H5Group, ZarrGroup], output_group: ZarrGroup | H5Group, merge
+):
+    uns = merge([read_elem(g["uns"]) for g in groups])
+    write_elem(output_group, "uns", uns)
```
flying-sheep: this doesn’t seem to be used, maybe you forgot to add it (and remove the NotImplementedError)

src/anndata/experimental/merge.py
```diff
@@ -408,6 +424,30 @@ def _write_axis_annot(  # noqa: PLR0917
...
+
+
+def _write_alt_pairwise(
```
flying-sheep, line 427: how does it work? create disconnected graph components (i.e. a [block diagonal matrix](https://en.wikipedia.org/wiki/Block_matrix#Block_diagonal_matrices))?

<details>
<summary>unnecesary reasoning why it’s false by default, pls ignore</summary>


I think we intentionally didn’t merge `{obs,var}p` because merging usually invalidates them. e.g. if you have these `distances`:

```py
dists_a = [0, 1, 1,  # c1.1
           1, 0, 1,  # c1.2
           1, 1, 0]  # c1.3
dists_b = [0, 4, 4,  # c2.1
           4, 0, 4,  # c2.2
           4, 4, 0]  # c2.3
```

they’d be merged like this:

```py
[0, 1, 1, 0, 0, 0,  # c1.1
 1, 0, 1, 0, 0, 0,  # c1.2
 1, 1, 0, 0, 0, 0,  # c1.3
 0, 0, 0, 0, 4, 4,  # c1.1
 0, 0, 0, 4, 0, 4,  # c1.2
 0, 0, 0, 4, 4, 0]  # c1.3
```

i.e. it would contain two completely disconnected graph components.

So I think people should have to opt into merging `{obs,var}p`

</details>

src/anndata/experimental/merge.py
```diff
@@ -490,7 +530,8 @@ def concat_on_disk(  # noqa: PLR0912, PLR0913, PLR0915
         DataFrames are padded with missing values.
     pairwise
         Whether pairwise elements along the concatenated dimension should be included.
-        This is False by default, since the resulting arrays are often not meaningful.
+        This is False by default, since the resulting arrays are often not meaningful, and is ignored when True.
+        If you are interested in this feature, please open an issue.
```
flying-sheep: I’m confused: isn’t the behavior of this PR right now

- raises NotImplementedError when set to `True`
- `{obs,var}p` *are* merged when it’s set to false?

doesn’t seem to make sense to me!


Shouldn’t this PR instead
- document this as working
- remove the NotImplementedError
- wrap the call to `_write_alt_pairwise` in `if pairwise`?

src/anndata/_core/merge.py
```diff
@@ -143,11 +143,12 @@ def equal_dask_array(a, b) -> bool:
...
+    return equal(a.compute(), b.compute())
```
flying-sheep, line 155: maybe this should raise a warning if it’s hit, seems like unintentional behavior

src/anndata/_core/merge.py
ilan-gold: No reason for `*` just was thinking of array arithmetic but tests pass with `&` and it's clearer

src/anndata/experimental/merge.py
ilan-gold: Meant to delete this, I looked over this PR and deleted some other things related to `uns` but must have missed this.

src/anndata/experimental/merge.py
ilan-gold: The reason I added this was simply that it's default behavior in `anndata.concat` for `merge` and wanted to bring us to feature parity with that.  I didn't think about it deeply tbh

src/anndata/experimental/merge.py
ilan-gold: No, I wanted to match the behavior of `anndata.concat` here so this PR raises `NotImplementedError` only for the  pairwise mappings along the concatenation axis, but just like `anndata.concat` automatically merges the off-axis.

src/anndata/_core/merge.py
ilan-gold: Yeah, I guess I should have been more careful here.  Previously this simply failed because the chunks weren't aligned.  Maybe we should error here? This equality check is new behavior though, so maybe maintaining the error (but making it more informative) is the way to go.

flying-sheep: sounds good!

src/anndata/experimental/merge.py
flying-sheep: Ah I see `_write_alt_pairwise` is about the off-axis. Makes sense!

But the rest is still valid: https://github.com/scverse/anndata/blob/1d4cb8751da11df1f086038f0e9e42d5808e6f0d/src/anndata/experimental/merge.py#L580-L582

… means that “[pairwise] is ignored when True.” is a lie. It’s not ignored, it throws an error.

src/anndata/_core/merge.py
ilan-gold: Looking into this a bit more, the issue here is that if we don't add this full equality check, it's possible that (reindexed) arrays can have different chunk sizes leading to potential ambiguity - even if they do have different chunkings, the underlying data might be equal.  So if you have `same` strategy, you want to retain those arrays but if we only check for chunk alignment, this would never be detected.

I feel ok adding this "new" behavior of full equality check because previously there was an uninformative error yielded from dask.  So I think we should go with a warning, and this way we can be sure the outputs match `anndata.concat`

src/anndata/experimental/merge.py
ilan-gold: I've updated the message now to hopefully be clearer

flying-sheep (approved):

PR #2133, opened 2025-09-25
feat: Support pandas 3.0 upcoming changes

### LINKED ISSUE #2221: Pandas 3.0 compatibility
Pandas will default to `pd.options.future.infer_string=True` in the 3.0: https://pandas.pydata.org/docs/whatsnew/v2.3.0.html

This will change a lot of `DataFrame` initialization code so `string` columns/indices are created instead of `object` ones. This is a good thing!

We already have code to write these columns, depending on `ad.settings.allow_write_nullable_strings`, which is `False` by default.

# Goals
- keep a way to write the old format
- sensible migration to allowing to write nullable string columns by default in 0.14

# Plan

1. add `allow_write_nullable_strings=None` option that makes it default to `pd.options.future.infer_string`
2. change `allow_write_nullable_strings=False` to mean “try writing them as non-nullable and fail only if we can’t”
3. in anndata 0.13’s release note, we add a fat warning that this setting will flip in 0.14
4. in anndata 0.14 we flip `allow_write_nullable_strings`’ default to True

PS: Originally, we planned that we only change `allow_write_nullable_strings=False` to mean “try writing them as non-nullable”, but I’m not sure now why we should wait. Why not allow this today? There’s a small performance hit when checking `string` columns, but that would have been an exception before and now it just starts working.

# further considerations

## `mode.copy_on_write = True`

Weirdly, this doesn’t cause any test failures! According to the docs it implies `mode.chained_assignment = "raise"`, but the warning we suppressed in #545 doesn’t get elevated to an exception. And tests like `test_base.py::test_slicing_*` still pass. Maybe everything is fine?  

We could also do `pd.options.mode.copy_on_write = "warn"` as the docs say …

## `"str"` vs `"string"`

Pandas
- originally added the `"string"` dtype (`pd.StringDtype(na_value=pd.NA)`), and then
- added the `"str"` dtype in 2.3, which will become the default in 3.0 (`pd.StringDtype(na_value=pd.nan)`)

We’re using `dtype="string"` (i.e. `pd.StringDtype(na_value=pd.NA)`) in some locations, and it’s easy to dynamically use the current default (`"O"` or `pd.StringDtype(na_value=pd.nan)`) by specifying `dtype="str"`.

They both compare `== "string"`, which is nice, but have slightly different [missing value semantics](https://pandas.pydata.org/pdeps/0014-string-dtype.html#missing-value-semantics).

flying-sheep:
- [x] Fixes #2221
- [x] Tests added
- [x] Changelog entry added

TODO:
- [ ] save `storage` in data as well?


src/anndata/_core/aligned_df.py
```diff
@@ -59,7 +59,7 @@ def mk_index(l: int) -> pd.Index:
         df = pd.DataFrame(
             anno,
             index=None if length is None else mk_index(length),
-            columns=None if anno else [],
+            columns=None if anno else pd.array([], dtype=str),
```
flying-sheep: The `infer_string` pandas option makes it so `dtype='str'`/`dtype=str` means `'string'` when it’s true and pandas’s version ≥2.3, and `object` otherwise.

I specify this in multiple places to guarantee that anndata’s behavior is dependent on this option in more places than it was before.

src/anndata/_core/file_backing.py
```diff
@@ -174,7 +175,11 @@ def _(x: AwkArray, *, copy: bool = False):
 
 @to_memory.register(Dataset2D)
 def _(x: Dataset2D, *, copy: bool = False):
-    return x.to_memory(copy=copy)
+    with warnings.catch_warnings():
+        warnings.filterwarnings(  # TODO: somehow only do this in tests but still localized
+            "default", r"Dtype inference.*pandas.*deprecated", FutureWarning
+        )
+        return x.to_memory(copy=copy)
```
flying-sheep: This is an xarray bug I assume. I changed this so our tests don’t break from xarray’s code encountering this, while still running into it if our code triggers it.

No idea how to do this better, as
- this will override if a user silences this warning.
- ignoring it here means that user’s won’t see this (valid) xarray-triggered warning.

src/anndata/_core/merge.py
```diff
@@ -953,8 +953,13 @@ def gen_inner_reindexers(els, new_index, axis: Literal[0, 1] = 0) -> list[Reinde
...
             raise NotImplementedError(msg)
         common_keys = intersect_keys(el.fields for el in els)
+        # TODO: replace dtype=object once this is fixed: https://github.com/scikit-hep/awkward/issues/3730
         reindexers = [
-            Reindexer(pd.Index(el.fields), pd.Index(list(common_keys))) for el in els
+            Reindexer(
+                pd.Index(el.fields, dtype=object),
+                pd.Index(list(common_keys), dtype=object),
+            )
+            for el in els
```
flying-sheep, line 966: see comment

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -301,7 +306,12 @@ def read_dataframe(
     if not use_range_index:
         dim_name = elem.attrs["_index"]
         # no sense in reading this in multiple times since xarray requires an in-memory index
-        index = elem_dict[dim_name].compute()
+        if isinstance(elem_dict[dim_name], DaskArray):
+            index = elem_dict[dim_name].compute()
+        elif isinstance(elem_dict[dim_name], MaskedArray):
+            index = elem_dict[dim_name][BasicIndexer((slice(None),))]
```
flying-sheep, line 312: @ilan-gold you said you didn’t really think this makes sense, but there was some confusion in the conversation so maybe this is correct after all.

src/anndata/_io/specs/methods.py
```diff
@@ -1135,41 +1134,52 @@ def read_partial_categorical(elem, *, items=None, indices=(slice(None),)):
...
             "and not supported in anndata < 0.11, still use by many people. "
             "Opt-in to writing these arrays by toggling the setting to True."
         )
         raise RuntimeError(msg)
     g = f.require_group(k)
+    if isinstance(v, pd.arrays.StringArray | pd.arrays.ArrowStringArray):
+        if v.dtype.na_value is pd.NA:
+            g.attrs["na_value"] = "pandas.NA"
+        elif np.isnan(v.dtype.na_value):
+            g.attrs["na_value"] = "numpy.nan"
```
flying-sheep: as said above, maybe we should store the “storage” attribute of the dtype, but that seems more like a runtime characteristic, and `pandas` should pick the most appropriate one by itself, I think.

src/anndata/_io/specs/methods.py
```diff
@@ -1211,12 +1209,29 @@ def _string_array(
...
+    elem: GroupStorageType, *, _reader: Reader
+) -> pd.api.extensions.ExtensionArray:
+    values = _reader.read_elem(elem["values"])
+    mask = _reader.read_elem(elem["mask"])
+    if PANDAS_SUPPORTS_NA_VALUE:
+        dtype = pd.StringDtype(
+            na_value=pd.NA
+            if _read_attr(elem.attrs, "na_value", default="pandas.NA") == "pandas.NA"
+            else np.nan
+        )
```
flying-sheep: so far, pandas only supported `pd.NA`, that’s why a missing `na_value` attr defaults to this.

src/anndata/tests/helpers.py
```diff
@@ -331,8 +331,8 @@ def gen_adata(  # noqa: PLR0913
         random_state = np.random.default_rng()
 
     M, N = shape
-    obs_names = pd.Index(f"cell{i}" for i in range(shape[0]))
-    var_names = pd.Index(f"gene{i}" for i in range(shape[1]))
+    obs_names = pd.Index([f"cell{i}" for i in range(shape[0])], dtype=str)
+    var_names = pd.Index([f"gene{i}" for i in range(shape[1])], dtype=str)
```
flying-sheep: this is why I add `lambda:` in a lot of places: `gen_adata` needs to be called when the `infer_strings` option is as we want it.

src/testing/anndata/_pytest.py
```diff
@@ -11,17 +11,22 @@
...
+from packaging.version import Version
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterable, Sequence
     from pathlib import Path
 
     from ._doctest import WarningFilter
 
+# Use a marker present in the environment so VS Code’s tests behave identical
+IS_PRE = Version(version("zarr")).is_prerelease
```
flying-sheep, line 29: Yeah, I had no better idea. I want this option tied to the `pre` env, without relying on anything that can’t be inferred when running the tests through VS Code (i.e. no env vars, no CLI options)

tests/test_io_backwards_compat.py
```diff
@@ -49,13 +49,20 @@ def test_backwards_compat_files(archive_dir: Path) -> None:
     assert_equal(from_h5ad, from_zarr, exact=True)
 
 
-def test_no_diff(tmp_path: Path, archive_dir: Path) -> None:
+def test_no_diff(
+    request: pytest.FixtureRequest, tmp_path: Path, archive_dir: Path
+) -> None:
+    if pd.options.future.infer_string:
+        reason = "even old files get read as `string` dtype if this is active"
+        request.applymarker(pytest.mark.xfail(reason=reason))
```
flying-sheep: This is the biggest issue of this PR right now:

AFAIK there is no non-nullable string dtype for pandas, therefore we can’t
1. read non-nullable string data into it, and therefore we also can’t
2. round-trip without changes (breaking the `diff`

(we just treated `object` as “non-nullable string dtype” when in reality it can of course contain any Python object at runtime. We should just not have these when pandas 3 hits)

flying-sheep (commented):
@ilan-gold treat these comments as narrative explanation of what this PR does before you review it!

docs/release-notes/2133.breaking.md
```diff
@@ -0,0 +1,16 @@
+..
+    this isn’t actually a breaking change, but it needs to be far up in the changelog.
+    maybe we should move the “attention” directive up when compiling release notes?
+
+Switch {attr}`anndata.settings.allow_write_nullable_strings` to new default `None`,
+which means the writing of pandas string arrays will be controlled by `pandas.options.future.infer_string`
+(see the {doc}`pandas:user_guide/migration-3-strings`).
+Setting `allow_write_nullable_strings=False` explicitly now means to try writing the array in the non-nullable string array format,
+which was the only supported format up to anndata 3.11.
```
ilan-gold: 1. I think it's ok to call this a breaking change even if it is not but as you wish
2. Why isn't this (and other stuff) on one line when part of the same sentence? More for curiosity

src/anndata/_core/file_backing.py
ilan-gold: Can you provide some more context here? Bug report?

src/anndata/_io/specs/lazy_methods.py
ilan-gold: Now this does make sense.

But this raises a bigger question - how do we handle `NA` in our indexes? This seems like something we'll have to confront and definitely motivates UUIDs.  Maybe we should reject writing indexes with NAs?

src/anndata/_io/specs/methods.py
```diff
@@ -1211,12 +1230,29 @@ def _string_array(
...
+def _read_nullable_string(
+    elem: GroupStorageType, *, _reader: Reader
+) -> pd.api.extensions.ExtensionArray:
+    values = _reader.read_elem(elem["values"])
+    mask = _reader.read_elem(elem["mask"])
+    if PANDAS_SUPPORTS_NA_VALUE:
+        dtype = pd.StringDtype(
+            na_value=pd.NA
+            if _read_attr(elem.attrs, "na_value", default="pandas.NA") == "pandas.NA"
+            else np.nan
```
ilan-gold: So what would the practical implication of choosing the "wrong one?" That we don't round trip it in-memory?  In this case, I would think that roundtripping on-disk is what matters (are other languages going to know what to do with this?).

src/anndata/tests/helpers.py
ilan-gold: This is a great thing to fix anyway.  This isn't the only case we'll want `gen_adata` run under various conditions

ilan-gold (commented):
Nice, very clean.  Would like to think through the fact that we write indexes to disk as nullables.  Is this possible now?

docs/release-notes/2133.breaking.md
flying-sheep: I want
1. the lines to be short (to be within line length limit)
2. no exact length-based breaks (so small changes don’t force us to either make things ugly by having some different-length lines or reformat whole blocks)

So I just do line breaks at commas or colons if possible, which leads to small diffs for small changes while lines stay short.

IDK if that’s the best way!

src/anndata/_core/file_backing.py
flying-sheep: I figured out how to fix it instead! I’m setting the `_index` column’s dtype manually before converting it to an index.

src/anndata/_io/specs/lazy_methods.py
flying-sheep: >  Maybe we should reject writing indexes with NAs?

which would mean the runtime cost of checking them, but that’s not a big deal: we already have a setting to turn runtime checking of indices off.

I think it could be an OK idea to use anonymous observations (all-NA indices). What do you think?

src/anndata/_io/specs/methods.py
flying-sheep: See https://pandas.pydata.org/pdeps/0014-string-dtype.html#missing-value-semantics

Which makes me think:

Pandas
- originally added the `"string"` dtype (`pd.StringDtype(na_value=pd.NA)`), and then
- added the `"str"` dtype (`pd.StringDtype(na_value=pd.nan)`) in 2.3, which will become the default in 3.0

We’re using `dtype="string"` (i.e. `pd.StringDtype(na_value=pd.NA)`) in some locations, and this PR so far dynamically uses the current default (`"O"` or `pd.StringDtype(na_value=pd.nan)`) by specifying `dtype="str"`.

They both compare `== "string"`, which is nice, but as you pointed out, they have differences.

So what should we do? Make sure we use the `np.nan` variant (`"str"`) everywhere in the future? That would mean changing away from `dtype="string"` where we use it to a helper function that returns `pd.StringDtype(na_value=np.nan)` on pandas 2.3+ and otherwise `"string"`/`pd.StringDtype(na_value=pd.NA)`

ilan-gold: For i/o, I think it shouldn't matter though, no? It's a runtime characteristic as you point out.

So it's more what we promise to the user.  Once the user has the object, they could change things without breaking functionality (because we always check `== "string"` in theory).  Then when we write, `pd.NA` or `NaN` shouldn't matter (it's not like we're calling `str.len()` or something to my knowledge where that might matter).

I don't think we need to be completely consistent across the code base (i.e., helpers), but I think our `read_xxx` functions should always return the same thing, and in that case, we wouldn't need to store the NA type.  

Maybe I'm missing something; could you come up with a concrete failure case where storing this semantic _is_ important for some concrete behavior?

flying-sheep:
```py
>>> import pandas as pd
>>> assert pandas.__version__  # must be ≥2.3
2.3.1
>>> pd.options.future.infer_string = True
>>> pd.array(["a", None], dtype="str") == "a"
array([ True, False])
>>> pd.array(["a", None], dtype="string") == "a"
<BooleanArray>
[True, <NA>]
Length: 2, dtype: boolean
```

flying-sheep:
We gotta be quick! https://github.com/pandas-dev/pandas/releases/tag/v3.0.0rc0

src/anndata/_io/specs/methods.py
ilan-gold: Right, but isn't this runtime behavior? Can we just document that this is going to change with the upgrade to pandas 3?

I was asking more a way that anndata would fail.  If someone wrote with `str` and then read with `string`, as long as we're clear on what we return, I think we should be ok, no?

src/anndata/_io/specs/lazy_methods.py
ilan-gold: > I think it could be an OK idea to use anonymous observations (all-NA indices). What do you think?

What does this mean? I haven't heard the term "anonymous observations" before?

flying-sheep: It’s kinda silly that `ad.concat([AnnData(shape=(2, 5)), AnnData(shape=(3, 5))])` raises a warning about duplicate observations.

Clearly the user doesn’t care about observations being identifiable by `obs_names`.

Ideally we’d use UUIDs to give them completely random identities, but opting into anonymous identities by setting `obs_names` to all NaN would also work.

src/anndata/_io/specs/methods.py
flying-sheep: Yeah, I think so!

src/anndata/_io/specs/lazy_methods.py
ilan-gold: > Ideally we’d use UUIDs to give them completely random identities, but opting into anonymous identities by setting obs_names to all NaN would also work.

And then we'd write what? I actually thinking we should write `None` but that would be a spec change i.e., writing these anonymous indices. This problem of different indices types also exists right now but also is potentially solved?  We state in the spec:

```
The group MUST contain an array for the index
```

But it's not clear if this is our concept of array or zarr/hdf5.  So writing nullable strings is actually valid maybe? I guess it's up to interpretation.

Judging by things like

```
The group MAY contain an entry X, which MUST be either a dense or sparse array
```

I would assume "array" is meant in the `anndata` sense i.e., things like categorical arrays are actually valid on disk indices.  I know we have this warning/transformation for setting a new `obs`/`var dataframe i.e., `msg = "Transforming to str index."` in `_gen_dataframe`, but I actually now am not sure why other types of indices modulo their serializability are disallowed (with the exception of integers, of course).

flying-sheep: we say things like

> each column of a DataFrame is saved as a separate array

so “array” seems to mean “our serialization of an array”, not necessarily a zarr/hdf5 array 

----

> And then we'd write what?

our string array serialization

… which actually isn’t mentioned in the spec 😢, only “Nullable integers and booleans”

we need to fix that too, but this PR is balooning already, so I wouldn’t want to do that in here.

ilan-gold: Agreed, but glad we talked through it, I put some ideas down in #2241

docs/release-notes/2133.breaking.md
ilan-gold:
```suggestion
which was the only supported format up to anndata 0.11.
```
or?

src/anndata/_io/specs/methods.py
```diff
@@ -1185,18 +1216,6 @@ def _read_nullable(
...
-def _string_array(
-    values: np.ndarray, mask: np.ndarray
-) -> pd.api.extensions.ExtensionArray:
-    """Construct a string array from values and mask."""
-    arr = pd.array(
-        values.astype(NULLABLE_NUMPY_STRING_TYPE),
-        dtype=pd.StringDtype(),
-    )
-    arr[mask] = pd.NA
-    return arr
```
ilan-gold, line 1197: Why don't we either preserve this behavior or make a new default? I think always returning the same `type` is fine even if it differs from the runtime interpretation of `NA`

ilan-gold (commented):
I see this is `0.13` - do we need to rush that release out to get ahead of this? We have no upper bound on `pandas` right now.   The `pandas.options.future.infer_string` seems relevant for `pandas<3` - do we need to break this PR up?

src/anndata/_io/specs/methods.py
flying-sheep: `Version(version("numpy")) < Version("2")` is always False, so there’s nothing to be preserved, right?

/edit: unless we backport to 0.12 of course

flying-sheep:
> do we need to rush that release out to get ahead of this?

I think that would be smart, yes

> The `pandas.options.future.infer_string` seems relevant for pandas<3 - do we need to break this PR up?

I doubt that a lot of people are using it, I assume it’s mostly used by library authors like us.

Also no idea how or why we’d want to break up this PR? This PR changes anndata’s behavior for when `infer_string=True`, which is automatically the case in pandas 3 and opt-in in pandas 2.3.

ilan-gold:
> Also no idea how or why we’d want to break up this PR? This PR changes anndata’s behavior for when infer_string=True, which is automatically the case in pandas 3 and opt-in in pandas 2.3.

So perhaps "break up" was the wrong word.  I could see a separate PR for the next `0.12` release doing:

1. Make lines like https://github.com/scverse/anndata/pull/2133/files#diff-be76039ad63e3b28abf862eadc28ec96a9b8333fd2e5c66c3b88e30f95b1f67eR62 dependent on the value and/or presence of `infer_string` i.e., leave the old behavior in place if the value is not there because the version of `pandas` is too low or it is `False`, and if `infer_string` is `True`, use `str` type
2. For writing, leave the old behavior in place independent of `infer_string` i.e., rely on `allow_write_nullable_strings` behavior
3. Upper bound pandas by `3`

I think this would make the "nasty" default behavior of having unwriteable anndata objects (because `infer_string=True` and `allow_write_nullable_strings=False`) opt-in, which means you can also opt-in to not having that by setting `allow_write_nullable_strings=True` in addition to `infer_string=True`.

Does this make sense? Am I missing something?

Then for 0.13 we can add in  `allow_write_nullable_strings=None`.

flying-sheep:
Line changes like the one you linked to only change behavior with `infer_string=True`. `dtype="str"` creates an `object` dtype as we’ve always done with `infer_string=False`.

----

I think the “only” thing that would have to change to backport this into 0.12 is numpy 1 compatibility. Pandas 2.3 supports numpy 1 still. And as you said, semantic changes to `allow_write_nullable_strings`

src/anndata/_io/specs/methods.py
flying-sheep: OK, let’s finish this conversation in the main thread.

flying-sheep:
OK, summary of what we said so far:

1. let’s finish talking about “should we store NA vs. NaN semantics, and if so already in 0.12.x or only in 0.13”
2. we merge this
3. I manually backport everything but the changes to `allow_write_nullable_strings`, which means people will have to set both `infer_string=True` and `allow_write_nullable_strings=True` if they want the new behavior. (also of course if we decide not to store NA/NaN semantics in 0.12, I’ll also not backport that part)

flying-sheep:
OK, I do think we should store the NA-semantics so people can round-trip them. I think we don’t need to sneak the `na_value` writing into 0.12, since in 0.12 things will still be very opt-in, and the differences are subtle enough that people who use this should be OK with small changes like that.

I think they should be defined so cross-language implementers have freedom to ignore them:

> `na_value` (optional): `"NA"` | `"NaN"` (more valid options might get defined in the future)
>
> Defines semantics for missing values if available in the target data model. `"NA"` means that a comparison between a missing value and a defined value produces a missing value (e.g. `"x" == NA ➡️ NA`), whereas `"NaN"` means that a comparison between a missing value and defined value produces a binary result (e.g. `"x" == NaN ➡️ false`).
> 
> ## Reading
> Implementers SHOULD return arrays that behave according to the specified semantics if the target data model allows for it. A missing `na_value` attribute SHOULD default to `"NA"`. Trying to read an unknown `na_value` SHOULD NOT produce an error. Implementers may freely decide which semantics to fall back to when reading an unknown `na_value` (anndata.
>
> ## Writing
> Implementers SHOULD store one of the defined options for `na_value` if the source data model corresponds to one of them. If the source data model instead behaves very differently from all currently defined ones, implementers SHOULD store no `na_value` attribute, but should please reach out to get a new value defined here.

src/anndata/compat/__init__.py
```diff
@@ -221,6 +221,14 @@ def __repr__():
     PANDAS_STRING_ARRAY_TYPES += [ArrowStringArrayNumpySemantics]
 
 
+def pandas_str_dtype() -> pd.StringDtype | np.dtypes.ObjectDType:
```
ilan-gold: Can we cache the return of this so it's only done once?

src/anndata/_core/aligned_df.py
```diff
@@ -91,9 +91,9 @@ def _gen_dataframe_df(
     if not is_string_dtype(anno.index[~anno.index.isna()]):
         msg = "Transforming to str index."
         warn(msg, ImplicitModificationWarning)
-        anno.index = anno.index.astype("str")
+        anno.index = anno.index.astype(pandas_str_dtype())
```
ilan-gold: Presumably this line is problematic for the failing tests because (I think) they all involve converting a range index to an `object` type

ilan-gold (approved):
Aside from the `index` issue, I think this works.

src/anndata/_core/aligned_df.py
ilan-gold: I have been thinking about this a bit, and I'm actually not sure what we should do here.  I would say maybe we should go with `NA` semantics as the default because of its status as our default elsewhere in the codebase, but indexes are special.  I could see `NaN` semantics being more useful because maybe you want actual booleans from `NA` comparison? This is just a vibe, I definitely would go with `NA` if I had to choose something with no further thought

src/anndata/compat/__init__.py
flying-sheep: yeah, kinda, as it’s dependent on `pd.options.future.infer_string`

src/anndata/_core/aligned_df.py
flying-sheep: maybe we need to create a more involved function that converts smartly:

- `float` becomes `StringDtype(na_value=np.nan)` (or on pandas 2.2 `object` but preserving `np.nan`s)
- `pd.IntegerArray` / `pd.BooleanArray` becomes `StringDtype(na_value=pd.NA)` always
- `object` … idk how to handle the fact that it might have `np.nan`s, `pd.NA`s or either. Maybe just choose `"string"`/`StringDtype(na_value=np.NA)` for simplicity.

ilan-gold (approved):

katosh:
Hi! I've been following this PR while waiting for it to land before my own PR on the HTML representation. Out of curiosity, I did some experimenting and wanted to share an observation in case it's helpful (apologies if I'm misunderstanding something!):

It seems like NA values in indices might be similar to how anndata already handles non-unique indices (that anndata allows with warnings). Pandas always maintains internal integer positions via `.iloc`, so NA values don't fundamentally break indexing:

```python
import pandas as pd

idx = pd.Index(['a', 'b', pd.NA, 'c', pd.NA], dtype='string')
idx.is_unique  # False - pandas treats NAs as duplicates
idx[~idx.isna()].has_duplicates  # False - actual uniqueness

df = pd.DataFrame({'val': [1,2,3,4,5]}, index=idx)
df.iloc[2]  # Still works fine
```

I also noticed the CI error message (`TypeError: expected unicode string, found 1`) mentioned "found 1" which made me wonder if it was about integer indices (RangeIndex) rather than nullable strings?

```python
df = pd.DataFrame({'x': ['a', 'b', 'c']})
df.index._values  # array([0, 1, 2], dtype=int64) integers, not strings
```

But I might be missing context here.

Anyway, thanks for working on this! Looking forward to pandas 3.0 compatibility!

PR #2134, opened 2025-09-27
fix: write iospec directly with sparse dataset

ilan-gold:
- [x] See added `test_write_dispatched_csr_dataset` for what would have failed - there was no io-spec in the `CS{R,C}Dataset` writer, and was instead appended at the end.  I am not sure why this was done but it prevents creating dispatched write functions properly on sparse datasets unless you check the type of the `elem` @ivirshup any history to this? It's been in since #554
- [x] Tests added


flying-sheep (approved):
I can’t think of any reason why it would be like that 🤷

PR #2143, opened 2025-10-02
chore: upgrade to spec0

ilan-gold:
See https://scientific-python.org/specs/spec-0000/ for more info.  The concerning part is 
image

which means we can't have both ~1.13~ 1.12 scipy and 2.0 numpy simultaneously ~1.13~ 1.12 scipy doesn't support 2.0 numpy.

Few options:

1. Manually "edit" min job to either downgrade numpy or upgrade scipy if we change the pyproject.toml to reflect spec0 properly
2. Add more min jobs that have different dependency chains (i.e., do both of the above separately)
3. Leave as-is with the non spec0 deps in the pyproject.toml

- [ ] Closes #
- [ ] Tests added


flying-sheep:
I guess we should probably switch all our `TypeAlias`es (implicit and explicit) to `type` syntax?

ilan-gold:
> I guess we should probably switch all our TypeAliases (implicit and explicit) to type syntax?

There's a wonderful python rabbit hole about https://github.com/scverse/anndata/pull/2143/files#diff-a9ba7a4b902a1d3c9af7d69f66e65700aa74d225800741030956adb3d0b12167R35-R36 if you want to jump into it, but aside from that I think we can.  `get_args` + `type` != `get_args` + `TypeAlias`

flying-sheep:
I went for the `typing` vs `types` distinction for that reason:

- runtime-checked type unions and aliases should be in `.types` or `.compat` (`CSRBase = sps.csr_matrix | sps.csr_array` `isinstance(m, CSRBase)`)
- types for annotation (and maybe internal use with `get_args(MyType.__value__)`) should be in `.typing`

flying-sheep:
ah, this was the issue: dda018f739558b6d664257211a02e7faf08d5bb0

we have multiple registries, this can’t be a class variable.

flying-sheep (approved):

flying-sheep:
~~regarding numpy 2 vs scipy 1.13, it’s a weird situation.~~

~~we can’t drop numpy 1 jobs (and upgrade to numpy-2-only APIs) while supporting scipy 1.13, so effectively we’d test with numpy 1 until then. I guess SPEC 0 support therefore means “if you depend on scipy, you gotta support numpy 1 until scipy 1.13 support is dropped”.~~

~~I guess I have to re-introduce numpy 1 support in fast-array-utils since it supports scipy.~~

pyproject.toml
```diff
@@ -29,18 +29,17 @@ classifiers = [
...
-    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Topic :: Scientific/Engineering :: Bio-Informatics",
     "Topic :: Scientific/Engineering :: Visualization",
 ]
 dependencies = [
-    "pandas >=2.1.0, !=2.1.2",
+    "pandas >=2.2.2",
     "numpy>=1.26",
```
flying-sheep: wait a second.

> 1.13 scipy doesn't support 2.0 numpy.

That’s not true: https://github.com/scipy/scipy/blob/v1.13.1/pyproject.toml#L38

```suggestion
    "numpy>=2",
```

flying-sheep (changes requested):

flying-sheep (approved):

PR #2158, opened 2025-10-20
fix: remove `read_dataset` usage in `read_lazy{_elem}` stack

ilan-gold:
- [x] https://github.com/scverse/anndata/pull/2121 pushed me to remove these calls, and in theory, things will be faster pending benchmark machine, but also, it is just better to use `read_elem` for stability reasons
- [x] Tests added

### DIFF
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -309,7 +309,6 @@ def read_dataframe(
     else:
         dim_name = DUMMY_RANGE_INDEX_KEY
         index = pd.RangeIndex(len(elem_dict[elem.attrs["_index"]])).astype("str")
-
     elem_xarray_dict = dict(
         _gen_xarray_dict_iterator_from_elems(elem_dict, dim_name, index)
     )


src/anndata/_io/specs/lazy_methods.py
```diff
@@ -303,11 +304,12 @@ def read_dataframe(
     # which is used below as well.
     if not use_range_index:
         dim_name = elem.attrs["_index"]
-        # no sense in reading this in multiple times
+        # no sense in reading this in multiple times since xarray requires an in-memory index
         index = elem_dict[dim_name].compute()
     else:
         dim_name = DUMMY_RANGE_INDEX_KEY
         index = pd.RangeIndex(len(elem_dict[elem.attrs["_index"]])).astype("str")
+
```
ilan-gold:
```suggestion
```

benchmarks/benchmarks/dataset2d.py
```diff
@@ -14,36 +14,58 @@
...
         )
-        self.ds = ad.experimental.read_elem_lazy(store["obs"], chunks=chunks)
+        self.ds = ad.experimental.read_elem_lazy(self.store["df"], chunks=chunks)
         self.n_obs = self.ds.shape[0]
 
+    def time_read_lazy_default(self, *_):
+        ad.experimental.read_elem_lazy(self.store["df"])
+
+    def peakmem_read_lazy_default(self, *_):
+        ad.experimental.read_elem_lazy(self.store["df"])
```
flying-sheep, line 67: What are these for? They shouldn’t do much since they’re lazy right?

tests/test_dask_view_mem.py
```diff
@@ -72,7 +72,7 @@ def _alloc_cache():
 # As of 2025.09.* dask, this needs a bit more than the previous 1.5mb.
 # TODO: Why?
 @pytest.mark.usefixtures("_alloc_cache")
-@pytest.mark.limit_memory("1.7 MB")
+@pytest.mark.limit_memory("2.2 MB")
```
flying-sheep, line 75: where does the increased memory come in here? does it scale with data size or is it more flat than that?

ilan-gold: This test has always been flaky, not sure why today is the day things happened but from https://github.com/scverse/anndata/actions/runs/18714329715/job/53369994329 we have 
```
- 1.0MiB allocated here:
        blockdims_from_blockshape:/home/runner/.local/share/hatch/env/virtual/anndata/88DSKUwu/hatch-test.min/lib/python3.11/site-packages/dask/array/core.py:1270
...
- 1.0MiB allocated here:
        _apply_random:/home/runner/.local/share/hatch/env/virtual/anndata/88DSKUwu/hatch-test.min/lib/python3.11/site-packages/dask/array/random.py:821
```

benchmarks/benchmarks/dataset2d.py
ilan-gold: I don't want things to get slower for opening one of these objects, and was hoping they would get faster by removing `read_dataset`

ilan-gold:
> but also, it is just better to use read_elem for stability reasons

It was not just about performance.  My original hope was that we could use a dask array for the index, but as the comment states, it's crazy slow.  So this PR is more around not using `read_dataset` since it lives in the `read_elem` stack anyway.  I think it's just better to use `read_elem` as it's a public API.

flying-sheep:
I agree, that seems good! Still

> the benchmarks don’t seem to reflect it.

is there any reason why this slowness can’t be seen here? https://github.com/scverse/anndata/pull/2158#issuecomment-3425844429

0.88 is the greatest speedup for h5ad, which isn’t an improvement from crazy slow to normal/fast in my book. That’s why I’m asking!

ilan-gold:
> is there any reason why this slowness can’t be seen here? https://github.com/scverse/anndata/pull/2158#issuecomment-3425844429

What I meant when I said

> My original hope was that we could use a dask array for the index, but as the comment states, it's crazy slow.

was that I tried using dask for the string reading, but it was slow, so this PR _only_ swaps out `read_dataset` for `read_elem`

> 0.88 is the greatest speedup for h5ad, which isn’t an improvement from crazy slow to normal/fast in my book. That’s why I’m asking!

Right this PR only refactors to remove `read_dataset` as a call.  It's not a performance issue but an API issue at this point.  It would be like calling `dask.array.from_zarr` directly when we have this machinery backed into `read_elem` already and should use it

flying-sheep (approved):
Aah got it!

PR #2159, opened 2025-10-20
perf: add ability to write downcasted indices

### LINKED ISSUE #2153: Downcast `indices` for CSR matrices if possible on-disk

Long standing issue @felix0097 raised a while ago, but we should begin a spec "change" process to begin to allow writing out `indices` of a CSR matrix (or CSC, although less valuable there) whose max value is checked + dtype automatically set optimally.  Nowhere are dtypes specified in https://anndata.readthedocs.io/en/latest/fileformat-prose.html#sparse-array-specification-v0-1-0 so while this isn't a breaking change, it could potentially complicate things downstream.

In other words, the values in `indices` of a CSR matrix are regularly less than `max(uint16)` (because we often dont have more than 30000 or so genes) but are often written as `{u}int32/64` so allowing users to write data optimized for this fact  without breaking downstream pipelines is in our interest. The process for this would be

1. Add a setting to allow this behavior via `anndata.settings.downcast_indices_in_sparse = True` or similar - I would guess the behavior would be "take the max of the incoming `indices` optionally and then write out as the minimum needed dtype"
2. Write tests within `anndata` that ensure reading this data back in doesn't break cupy/scipy sparse
3. Release with this setting as `False`
4. Potentially in the future set to `True`

The downside/complication has been traditionally scipy sparse handling of differing `indptr` + `indices` dtypes but I think that is a manageable problem if we limit ourselves to just io here, and I imagine the performance would be better, even if the data has to be re-upcast into int32 for this compatibility issue, given the lessened io. 

cc @lazappi @ivirshup @keller-mark, happy to produce some dummy data for this for y'all to test.  Let me know if you are aware of any downsides here or if you have commments!

ilan-gold:
TODO:

- [ ] gather feedback in #2153
- [x] ~~tests for + run on GPU CI~~ We don't support io directly into GPU anyway!

Checks
- [x] Closes #2153
- [x] Tests added


tests/test_io_elementwise.py
```diff
@@ -424,6 +424,113 @@ def test_write_indptr_dtype_override(store, sparse_format):
...
+    format: Literal["csr", "csc"],
+):
+    minor_axis_index = np.array([num_minor_axis - 1])
+    major_axis_index = np.array([10])
+    row_cols = (
+        (minor_axis_index, major_axis_index)
+        if format == "csc"
+        else (major_axis_index, minor_axis_index)
+    )
+    shape = [num_minor_axis, 20] if format == "csc" else [20, num_minor_axis]
```
flying-sheep:
```suggestion
    shape = (num_minor_axis, 20) if format == "csc" else (20, num_minor_axis)
```

flying-sheep (approved):
looks great!

When reading this in, will the dtype be preserved? And if so, how do certain APIs deal with tiny int dtypes? Like anndata’s concatenation and scanpy’s algorithms?

ilan-gold:
> When reading this in, will the dtype be preserved? And if so, how do certain APIs deal with tiny int dtypes? Like anndata’s concatenation and scanpy’s algorithms?

`scipy` will convert for us to match `indptr` from what I can tell from the issues and the behavior of the tests passing (i.e., the data is read back in and matches the input).  So this isn't really a scanpy problem.  In the future when supporting different sparse matrices, (like finch tensor), I think we will hopefully be able to preserve types.  I don't see anything in https://graphblas.org/binsparse-specification/ that would indicate that the data types have to match

flying-sheep (approved):

PR #2167, opened 2025-10-23
perf: auto sharding for zarr v3

### LINKED ISSUE #2148: Use `auto` sharding from `zarr`

https://zarr.readthedocs.io/en/stable/api/zarr/index.html#zarr.create_array has the `shards` kwargs whose type (should be linked) actually apparently includes the `auto` option.  This is far simpler that using `write_dispatched`. We should

1. Add a setting to https://anndata.readthedocs.io/en/stable/generated/anndata.settings.html for `auto` sharding
2. Decide how this should interact with https://anndata.readthedocs.io/en/stable/generated/anndata.settings.html#anndata.settings.zarr_write_format since you actually can write zarr v2 sharded data (I think error out is the clear choice given possible read compatibility issues in other languages, to say the least)
3. Update the zarr-v3 guide accordingly https://anndata.readthedocs.io/en/stable/tutorials/zarr-v3.html

ilan-gold:
cc @jan-engelmann this is about as good as I think we will get on auto-solving the "many small files problem" without fully answering the open question of how to do this best.  Note the `auto` feature is still experimental in `zarr` itself, hence my point about it being an "open question".

- [x] Closes #2148
- [x] Tests added

### DIFF
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -518,6 +518,7 @@ def write_basic_dask_dask_dense(
     is_h5 = isinstance(f, H5Group)
     if not is_h5:
         dataset_kwargs = zarr_v3_compressor_compat(dataset_kwargs)
+        dataset_kwargs = zarr_v3_sharding(dataset_kwargs)
     if is_zarr_v2() or is_h5:
         g = f.require_dataset(k, shape=elem.shape, dtype=elem.dtype, **dataset_kwargs)
     else:
--- a/tests/test_dask.py
+++ b/tests/test_dask.py
@@ -117,7 +117,7 @@ def test_dask_write(adata, tmp_path, diskfmt):
 @pytest.mark.dask_distributed
 @pytest.mark.parametrize(
     "auto_shard_zarr_v3",
-    [True, False],
+    [pytest.param(True, id="shard"), pytest.param(False, id="no-shard")],
 )
 def test_dask_distributed_write(
     adata: AnnData,


jan-engelmann:
yes this would be nice. will try it out when I get a chance

src/anndata/_settings.py
```diff
@@ -444,6 +444,13 @@ def validate_zarr_write_format(format: int):
         raise ValueError(msg)
 
 
+def validate_zarr_sharding(auto_shard: bool):  # noqa: FBT001
+    validate_bool(auto_shard)
+    if auto_shard and is_zarr_v2():
```
flying-sheep: what happens when setting this to `True` first and `zarr_write_format` to 2 after?

ilan-gold:
> Why do a setting at all? is there any potential that this breaks people’s code? Or is this just about it being experimental?

I am proceeding extremely carefully about on-disk stuff after both the forward-slash and `None` bugs.  I think all of the `anndata` readers in my head should support sharding (that support zarr).  I also don't want to spam people with zarr warnings, although I imagine we could filter them.

flying-sheep:
Maybe at least mention in the settings doc that this setting will probably be turned on by default once zarr no longer warns here.

src/anndata/_io/specs/methods.py
```diff
@@ -694,6 +703,9 @@ def write_recarray_zarr(
     else:
         dataset_kwargs = dataset_kwargs.copy()
         dataset_kwargs = zarr_v3_compressor_compat(dataset_kwargs)
+        # https://github.com/zarr-developers/zarr-python/issues/3546
```
flying-sheep, line 696: what does “broken” mean? crash or data corruption?

src/anndata/_settings.py
```diff
@@ -134,9 +135,13 @@ def check_and_get_int(option, default_value):
...
-    _registered_options: dict[str, RegisteredOption] = field(default_factory=dict)
-    _deprecated_options: dict[str, DeprecatedOption] = field(default_factory=dict)
-    _config: dict[str, object] = field(default_factory=dict)
+    _registered_options: OrderedDict[str, RegisteredOption] = field(
+        default_factory=OrderedDict
+    )
+    _deprecated_options: OrderedDict[str, DeprecatedOption] = field(
+        default_factory=OrderedDict
+    )
+    _config: OrderedDict[str, object] = field(default_factory=OrderedDict)
```
flying-sheep: why? Python dicts are guaranteed to maintain insertion order unless things get deleted. Do we delete things?

src/anndata/_settings.py
```diff
@@ -455,8 +477,8 @@ def validate_zarr_write_format(format: int):
 )
 
 
-def validate_sparse_settings(val: Any) -> None:
-    validate_bool(val)
+def validate_sparse_settings(val: Any, remove_unused_categories: bool) -> None:  # noqa: FBT001
+    validate_bool(val, remove_unused_categories)
```
flying-sheep: I’m confused, isn’t the second argument a SettingsManager?

src/anndata/_io/specs/methods.py
ilan-gold: updated title :)

ilan-gold: (crash)

PR #2172, opened 2025-10-24
fix: remove global lock on zarr dense stores from dask

### LINKED ISSUE #2166: Incorrect values saved to zarr with dask distributed

### Report

Code:

```python
import anndata
import dask.array as da
import numpy as np
import pandas as pd
from distributed import Client

if __name__ == "__main__":
    with Client():
        X = da.arange(12).reshape(3, 4)
        X = X.rechunk((2, 2))  # works if this line is removed
        data1 = anndata.AnnData(X=X,
                                obs=pd.DataFrame(index=['1', '2', '3'],
                                                 data=dict(a=[1, 2, 3, ])),
                                var=pd.DataFrame(index=['1', '2', '3', '4']))
        data1 = data1[[0, 2]]
        data1.write_zarr("test.zarr")
        data2 = anndata.read_zarr("test.zarr")
        np.testing.assert_array_equal(data1.X, data2.X)
```

### Versions

```python
| Package | Version                  |
| ------- | ------------------------ |
| anndata | 0.1.0.dev1582+g390ad2944 |

| Dependency        | Version     |
| ----------------- | ----------- |
| natsort           | 8.4.0       |
| traitlets         | 5.14.3      |
| pandas            | 2.3.3       |
| numcodecs         | 0.16.3      |
| six               | 1.17.0      |
| PyYAML            | 6.0.3       |
| setuptools        | 80.9.0      |
| decorator         | 5.2.1       |
| Pygments          | 2.19.2      |
| typing_extensions | 4.15.0      |
| executing         | 2.2.1       |
| jedi              | 0.19.2      |
| parso             | 0.8.5       |
| zarr              | 3.1.3       |
| session-info2     | 0.2.3       |
| python-dateutil   | 2.9.0.post0 |
| donfig            | 0.8.1.post1 |
| h5py              | 3.14.0      |
| legacy-api-wrap   | 1.4.1       |
| ipython           | 9.6.0       |
| crc32c            | 2.7.1       |
| pytz              | 2025.2      |
| numpy             | 2.3.3       |
| prompt_toolkit    | 3.0.52      |
| comm              | 0.2.3       |
| pure_eval         | 0.2.3       |
| ipywidgets        | 8.1.7       |
| scipy             | 1.16.2      |
| stack-data        | 0.6.3       |
| packaging         | 25.0        |
| asttokens         | 3.0.0       |
| wcwidth           | 0.2.14      |

| Component | Info                                                                              |
| --------- | --------------------------------------------------------------------------------- |
| Python    | 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:38:53) [Clang 18.1.8 ] |
| OS        | macOS-15.7.1-arm64-arm-64bit                                                      |
| Updated   | 2025-10-23 12:52
```

ilan-gold:
~~I tried to do some forensic work to establish what this was for originally, but couldn't quite figure it out.  #1079 doesn't contain enough info to reproduce the issue without the lock, and all tests pass with this removed as per this PR.  I remembered https://scverse.zulipchat.com/#narrow/channel/393966-scanpy-anndata-dev/topic/meeting.202023-07-25/near/378887685  but said at the time I thought it was for hdf5 (which made sense, and everyon seemed to agree on that) so not sure how this translated into zarr.~~

The lock appears to have been necessary for v2 zarr python, the package, but causes issues with v3 the package.  So I have isolated the behavior, and opened an issue.  But I also think getting rid of an unexplainable lock is probably a good idea!

Here's a quick v2 example showing why the lock is necessary - it fails without the lock with a similar error to that linked in #1079:

<details>

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr<3",
#   "dask[array]==2023.5.1",
#   "distributed",
#   "numpy==1.26"
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues
from __future__ import annotations

import dask.array as da
import numpy as np
import zarr
from distributed import Client

if __name__ == "__main__":
    with Client():
        X = da.random.random((3, 4), chunks=(2, 2))
        data1 = X[np.array([0, 2]), :]
        g = zarr.open("test.zarr", mode="w")
        arr = g.create_dataset("my_arr", shape=data1.shape, dtype=data1.dtype)
        da.store(data1, arr)
        data2 = da.from_zarr(arr)
        np.testing.assert_array_equal(data1, data2)
```
</details>

- [x] Closes #2166
- [x] Tests added: ~~See https://github.com/scverse/anndata/issues/2166#issuecomment-3443506595 for why I'm hesitant to add a test.  If we do, it would have to be without the `local_cluster_addr` which seems bad given our previous experience.  If I had something to check from the old PR, I would be able to point at that but can't now.~~ Added a new CI job and handle the case when run in `xdist`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -118,11 +118,14 @@ def local_cluster_addr(
     # Adapted from https://pytest-xdist.readthedocs.io/en/latest/how-to.html#making-session-scoped-fixtures-execute-only-once
     import dask.distributed as dd
 
-    def make_cluster() -> dd.LocalCluster:
-        return dd.LocalCluster(n_workers=1, threads_per_worker=1)
+    def make_cluster(worker_id: str) -> dd.LocalCluster:
+        # If we're not using multiple pytest-xdist workers, let the cluster have multiple workers.
+        return dd.LocalCluster(
+            n_workers=1 if worker_id != "master" else 2, threads_per_worker=1
+        )
 
     if worker_id == "master":
-        with make_cluster() as cluster:
+        with make_cluster(worker_id) as cluster:
             yield cluster.scheduler_address
             return
 
@@ -138,7 +141,7 @@ def local_cluster_addr(
         yield address
         return
 
-    with make_cluster() as cluster:
+    with make_cluster(worker_id) as cluster:
         fn.write_text(cluster.scheduler_address)
         lock.release()
         yield cluster.scheduler_address
--- a/tests/lazy/test_concat.py
+++ b/tests/lazy/test_concat.py
@@ -220,11 +220,11 @@ def test_concat_to_memory_var(
 @pytest.mark.xdist_group("dask")
 @pytest.mark.dask_distributed
 def test_concat_data_with_cluster_to_memory(
-    adata_remote: AnnData, join: Join_T, local_cluster_addr: str, worker_id: str
+    adata_remote: AnnData, join: Join_T, local_cluster_addr: str
 ) -> None:
     import dask.distributed as dd
 
-    with dd.Client(local_cluster_addr if worker_id != "master" else None):
+    with dd.Client(local_cluster_addr):
         ad.concat([adata_remote, adata_remote], join=join).to_memory()
 
 
--- a/tests/test_dask.py
+++ b/tests/test_dask.py
@@ -121,7 +121,6 @@ def test_dask_distributed_write(
     tmp_path: Path,
     diskfmt: Literal["h5ad", "zarr"],
     local_cluster_addr: str,
-    worker_id: str,
 ) -> None:
     import dask.array as da
     import dask.distributed as dd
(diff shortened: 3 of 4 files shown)


.github/workflows/test-cpu.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         env: ${{ fromJSON(needs.get-environments.outputs.envs) }}
-        io_mark: ["zarr_io", "not zarr_io"]
+        io_mark: ["zarr_io", "not zarr_io", "dask_distributed"]
```
flying-sheep: what’s the motivation for splitting the tests between dask distributed and the rest? (we split the zarr_io off because of performance reasons)

ilan-gold: I didn't explicitly try this but we have had issues running distributed clusters in tests with `xdist` before so I wanted to split it off and have it run in CI.  I will remove `-n auto` for the `dask_distributed` tests.

> If the idea is to test with different schedulers, this wouldn’t do it, right?

https://github.com/scverse/anndata/pull/2172/files#diff-8f4b462bf701012d1ca6188eb9ea3bd4e09f04c9266b69c2d1947ebe18a04ef6R227 This should handle that though.

flying-sheep: I see, now it makes sense! Please add a comment explaining the 2 reasons why the 3-way split of the tests exists!

ilan-gold: Done, thanks for the push, the test also now runs without `-n auto`, so glad you caught that

Issue #2173, opened 2025-10-25
Add type-annotation, enable mypy to check it


szabgab:
### Please describe your wishes and possible alternatives to achieve the desired result.

As far as I can tell some of the code already has type annotation, but `mypy` is not being used neither in the pre-commit hook nor in the CI to verify it.

IMHO it could be a nice improvement to the long-term quality of the project if type-annotation was more wide-spread in the code and if it was verified by running mypy in the CI.

Running mypy on the current code-base fails with various errors. Staring with reports about duplicate module named  "anndata", but if those are fixed then many other errors show up. So the strategy to introduce it could be:

* Create a configuration file that will exclude certain files and errors allowing mypy to pass.
* Add mypy to the pre-commit hook.
* Setup CI to run CI.

This will already ensure that no new issues will be added to the code-base.

Then one-by-one enable certain errors and add type annotation to the code while making sure mypy passes.
If necessary making changes to the code as well while maintaining passing tests and mypy still passing.

Related: #397,  #1859

I'd be happy to send a couple of PRs to get this rolling.

ilan-gold:
Hi @szabgab,

I wouldn't turn away a PR that begins to introduce it incrementally.

That being said, I would let @flying-sheep weigh in with the final say since he has used it more recently than I and knows more about the ecosystem (I don't use type checking much outside what vscode offers by default).

> Staring with reports about duplicate module named "anndata"

What is this about?  I'm just curious

szabgab:
As far as I can tell the duplicate "anndata" modules are `src/anndata/` and `src/testing/anndata/` so either of these would overcome that:

```
mypy --exclude src/anndata/__init__.py src/
mypy --exclude src/testing/anndata/__init__.py src/
```

flying-sheep:
I think it would make sense. It was a bad idea for fast-array-utils, since it has so many overloads that mypy is really bad at unifying. But I think AnnData is more straightforward. If I were you, I’d wait until after #2143 before tackling this!

adamgayoso:
Have you all considered making anndata a py typed library? I think this would help with the scverse ecosystem. 

One recent example is the introduction of `obs -> DataFrame | Dataset2D`. There are many instances across scanpy/anndata of doing something like `adata.obs.{pandas_only_method}`, which would have raised errors with the recent type change.

flying-sheep:
[...]

PR #2174, opened 2025-10-25
Add type-checking

szabgab:
* Add .mypy.ini
* Add mypy to pre-commit
* Add mypy to GitHub Actioons

- [X] Starts working on #2173
- [X] Tests added

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -21,7 +21,7 @@ jobs:
           python-version: ${{ env.max_python_version }}
       - name: Install mypy
         run: |
-          pip install mypy
+          pip install 'mypy @ https://github.com/python/mypy.git'
       - name: Run mypy
         run: |
           mypy --show-traceback src
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -21,7 +21,7 @@ jobs:
           python-version: ${{ env.max_python_version }}
       - name: Install mypy
         run: |
-          pip install 'mypy @ https://github.com/python/mypy.git'
+          pip install 'mypy @ git+https://github.com/python/mypy.git'
       - name: Run mypy
         run: |
           mypy --show-traceback src
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -7,17 +7,18 @@ on:
 jobs:
   mypy:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        python: ["3.12"]
 
     steps:
       - name: Checkout
         uses: actions/checkout@v5
+      - name: Extract max Python version from classifiers
+        run: |
+          classifiers=$(yq .project.classifiers pyproject.toml -oy  | grep --only-matching --perl-regexp '(?<=Python :: )(\d\.\d+)')
+          max_version=$(echo "$classifiers" | sort -V | tail -1)
+          echo "max_python_version=$max_version" >> $GITHUB_ENV
       - uses: actions/setup-python@v6
         with:
-          python-version: ${{ matrix.python }}
+          python-version: ${{ env.max_python_version }}
       - name: Install mypy
         run: |
           pip install mypy
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -23,4 +23,4 @@ jobs:
           pip install mypy
       - name: Run mypy
         run: |
-          mypy src
+          mypy --show-traceback src
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,4 +1,4 @@
-name: Run mypy
+name: mypy
 
 on:
   push:
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,0 +1,26 @@
+name: Run mypy
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  mypy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["3.12"]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ matrix.python }}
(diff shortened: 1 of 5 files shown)


.github/workflows/mypy.yml
```diff
@@ -0,0 +1,26 @@
+name: Run mypy
```
ilan-gold: I think we have a pre-commit bot that already runs on PRs, so in theory should be running the added mypy to the `.pre-commit-config.yaml`

ilan-gold: There it is! https://results.pre-commit.ci/run/github/100038377/1761382588.jTYvjMUUTWKRTaM3owMDkw

szabgab: You are right, Now I can see it is triggered in `.github/workflows/check-pr.yml`.
What is unclear to me why is it failing on the pre-commit.ci server. It seems that it does not take in account the `.mypy.ini` that excludes the duplicate files.

szabgab: If I run `pre-commit run -a`  locally, I get the same error as we can see on pre-commit.ci

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

src/anndata/_settings.pyi: error: Duplicate module named "anndata._settings" (also at "src/anndata/_settings.py")
Found 1 error in 1 file (errors prevented further checking)
```

szabgab: I've commented out mypy in the pre-commit hook to make ti pass on the pre-commit.ci server.
We still have it on GitHub Actions.
Either someone will figure out what's the problem on the pre-commit.ci server or this can be revisited later.

Zethson:
I don't understand why we would add a GHA workflow instead of just adding mypy to the pre-commit config. The latter should suffice.

I also suggest that we put the mypy config into the pyproject.toml file.

Thanks @szabgab for getting this started!

szabgab:
I've moved the mypy configuration to `pyproject.toml ` and rebased the PR.

I still could not figure out why does mypy fail on the pre-commit.ci while passes on GHA. That's the reason I kept GHA.

flying-sheep:
> I still could not figure out why does mypy fail on the pre-commit.ci while passes on GHA. That's the reason I kept GHA.

Because pre-commit creates isolated environments. You’d need to duplicate all dependencies in `.pre-commit-config.yaml` to make it work, like here:

https://github.com/scverse/fast-array-utils/blob/6af5d9a339a78604ff62230e6462f3a7a94fec3f/.pre-commit-config.yaml#L24-L41

szabgab:
It seems that rebasing to `main`  broke `mypy`. Specifically this commit 470db9cc09bf5861234dbf0d9cb575cd3b68c12d
The class `name[]` syntax seems something that mypy does not accept.

flying-sheep:
Starting with version 1.12, mypy supports this syntax: https://mypy.readthedocs.io/en/latest/changelog.html#support-python-3-12-syntax-for-generics-pep-695

I see an internal error, so it’s for sure a mypy bug though!

flying-sheep:
Seems like it’s either fixed on master or the “49 files” mypy checks are `docs` and stuff. No way it just succeeds.

szabgab:
So maybe you could go ahead and merge this now.
We might later move the mypy run to the pre-commit hook, though it seems it is also disabled on `fast-array-utils`

.pre-commit-config.yaml
```diff
@@ -39,3 +40,25 @@ repos:
       - id: codespell
         additional_dependencies:
           - tomli
+# See https://github.com/scverse/anndata/pull/2174
```
Zethson, line 43: We can also exclude them like this:

https://github.com/laminlabs/lamindb/blob/main/.pre-commit-config.yaml#L71

.github/workflows/mypy.yml
```diff
@@ -0,0 +1,27 @@
+name: mypy
```
Zethson, line 1: I still think that this whole workflow shouldn't exist and we can make it work with pre-commit. See below for the hammer - just exclude the unimportant conftest files.

.github/workflows/mypy.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          classifiers=$(yq .project.classifiers pyproject.toml -oy  | grep --only-matching --perl-regexp '(?<=Python :: )(\d\.\d+)')
+          max_version=$(echo "$classifiers" | sort -V | tail -1)
+          echo "max_python_version=$max_version" >> $GITHUB_ENV
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.max_python_version }}
+      - name: Install mypy
+        run: |
+          pip install 'mypy @ git+https://github.com/python/mypy.git'
```
Zethson, line 24: Even if we were to use this workflow, we shouldn't install from Github. We can just use a release.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,6 @@
 ci:
   autoupdate_commit_msg: "ci: pre-commit autoupdate"
+  skip: [mypy]  # too big
```
Zethson, line 3:
```suggestion
```

Zethson (changes requested):

.github/workflows/mypy.yml
ilan-gold: Re: what Lukas is saying in the review, it should probably be the reverse. We should figure out why things aren't working on the `pre-commit` ci instead of trying to use github actions.

.pre-commit-config.yaml
flying-sheep: @Zethson did you test that it works? `pre-commit.ci` often cries when the mypy environment contains a halfway useful subset of a package’s dependencies, so I basically have to add this `skip` to almost everything.

Zethson: No I did not but I have never had to use it anywhere. But you're probably right then

Zethson:
@szabgab are you still interested in finishing this PR? I think the general idea has value.

Zethson:
We've recently added a config to the template https://github.com/scverse/cookiecutter-scverse/pull/530 which should be used here. @szabgab are you still up for finish this PR? Otherwise, I'll close this PR

PR #2175, opened 2025-10-26
chore: Setter does not return values

szabgab:
This is a setter function that does not return any value. The type annotation should say so.

Caught by mypy.

- [X] Related to #2173
- [X] Tests added - irrelevant


flying-sheep:
thanks!

PR #2176, opened 2025-10-26
chore: Add explicit return

szabgab:
Instead of relying the implicit return of None it is better to explicitly deal with the else case.

Caught by mypy

- [X] Related to #2173
- [X] Tests added - irrelevant


flying-sheep:
Thanks! I’ll clean up the whole function instead, it’s too nested!

See https://github.com/scverse/anndata/pull/2228

PR #2183, opened 2025-10-28
fix: use `threads` scheduler in `da.store`

ilan-gold:
I discovered this while looking into something unrelated, but am glad.  It seems reasonable and also explainable.

- [x] A different fix for #2166
- [x] Tests added


flying-sheep (approved):
I guess there’s no way in which user code could break when not using the distributed scheduler?

ilan-gold:
> I guess there’s no way in which user code could break when not using the distributed scheduler?

So went my thinking.  This feels safer.

PR #2186, opened 2025-10-30
feat: create parent dirs when writing into h5ad

marcfranquesa:
Solves the following error raised when writing an h5ad file if a parent directory does not exist:

```
FileNotFoundError: [Errno 2] Unable to synchronously create file (unable to open file: name = 'dir1/dir2/adata.h5ad', errno = 2, error message = 'No such file or directory', flags = 13, o_flags = 242)
```

It is specifically raised by the following line https://github.com/scverse/anndata/blob/920f8a91d63213264e24d73cbb621e2414351377/src/anndata/_io/h5ad.py#L79

I am not sure if this is the general desired behavior, if not, I would happily modify the implementation (e.g. adding a new function argument to `write_h5ad`).

I believe it makes more sense to create the directories by default. Let me know otherwise. I will happily add a test if this behavior is desired.

### DIFF
--- a/src/anndata/_io/h5ad.py
+++ b/src/anndata/_io/h5ad.py
@@ -72,7 +72,7 @@ def write_h5ad(
             adata.strings_to_categoricals(adata.raw.var)
     dataset_kwargs = {**dataset_kwargs, **kwargs}
     filepath = Path(filepath)
-    filepath.mkdir(parents=True, exist_ok=True)
+    filepath.parent.mkdir(parents=True, exist_ok=True)
     mode = "a" if adata.isbacked else "w"
     if adata.isbacked:  # close so that we can reopen below
         adata.file.close()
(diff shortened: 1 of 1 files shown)


ilan-gold:
Hmm @marcfranquesa I generally hesitate to introduce more "magic" file behavior.  What happens if the writing of the `h5ad` file fails? Then you have a magically created directory.  I wonder if this somehow could break something else unexpectedly. I don't see any reason `mdkir -p` would not be threadsafe, but that was just a first glance.  So that isn't an issue.

> I believe it makes more sense to create the directories by default.

What's the use-case here? Just convenience?

marcfranquesa:
Hi Ilan, thank you for the reply and explanation.

> What's the use-case here? Just convenience?

Yes.

The PR can be closed.

ilan-gold:
@flying-sheep any thoughts?

PR #2190, opened 2025-10-31
ci: fix metadata, test Python 3.14

flying-sheep:
Makes the following metadata changes:

- Isaac isn’t an active maintainer
- Ilan is a (very) significant contributor
- We support Python 3.14

and some changes to the test deps


flying-sheep:
We’re not bumping a dep, this is the `test-min` extra.

But sure, we can keep testing with an older `awkward` version. The two working sets  are `awkward<2.8.6`+`pyarrow<21` and `awkward>=2.8.6`+`pyarrow`.

So maybe we should instead try to only constrain to `pyarrow<21` in the `deps=mi job? That would allow people to install incompatible versions of awkward and pyarrow if they really insist, but it’s not our job to protect them from that.

PR #2201, opened 2025-11-04
chore: backport dask helpers to 0.12

ilan-gold:
- [x] See https://github.com/scverse/scanpy/pull/3870
- [ ] Tests added
- [x] Release note added (or unnecessary): Not necessary

### DIFF
--- a/docs/release-notes/2204.fix.md
+++ b/docs/release-notes/2204.fix.md
@@ -1,0 +1,1 @@
+Attach {class}`h5py.File` object to {class}`~anndata.AnnData` at the `file` attribute returned from  {func}`~anndata.experimental.read_lazy` {user}`ilan-gold`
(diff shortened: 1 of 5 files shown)


tests/test_dask.py
flying-sheep, line 1: shouldn’t this be switched to `as_sparse_dask_matrix` instead of removing it?

ilan-gold: Sorry maybe I'm missing something - this comment is on the file.  What is the "it" being removed?

flying-sheep: all changes in that file are about removing `as_sparse_dask_array`.

ilan-gold: got it yes!

tests/test_dask.py
```diff
@@ -326,7 +326,7 @@ def test_dask_to_memory_unbacked(array_func, mem_type):
     "array_func",
     [
         pytest.param(as_dense_dask_array, id="dense_dask_array"),
-        pytest.param(as_sparse_dask_array, id="sparse_dask_array"),
+        pytest.param(as_sparse_dask_matrix, id="sparse_dask_array"),
```
flying-sheep:
```suggestion
        pytest.param(as_sparse_dask_matrix, id="sparse_dask_matrix"),
```

PR #2202, opened 2025-11-04
feat: switch to `warnings.deprecated` and `skip_file_prefixes`

flying-sheep:
- [ ] Closes #
- [x] Tests added

Sadly we can’t wrap `warnings.deprecated`, as type checkers need the original one. So we can’t modify its signature. We could modify its behavior more than just switching out the default `category` if we want.

TODO:

- [x] make sure `AnnData.concatenate` is still visible in the docs, and `obs_keys` and so on are still skipped.


ilan-gold (approved):

flying-sheep:
@ilan-gold I was a bit overeager to request your review, there were a few changes needed: https://github.com/scverse/anndata/compare/82c9dd567a770ff85908e2eddba5b57cc79d4d4c..pa/warnings

ilan-gold (approved):
Seems like the changes are reasonable given the problem.  I would say to add a comment but if we're going to fix the problem inside scanpy, then no worries.

flying-sheep:
why am I trying to backport this, it’s a feature that’s not supported on Python 3.11

PR #2213, opened 2025-11-11
ci: adapt tests to scanpy 1.12 changes

flying-sheep:
Moves the ancient `10x_pbmc68k_reduced.h5ad` from before https://github.com/scverse/scanpy/pull/3883 to our test data

Fixes the currently broken tests

TODO:
- [x] check which exact version the file was written in. If that version isn’t in `tests/data/archives`, move it there.


flying-sheep:
kinda crazy that we kept that old anndata file around for so long. We could probably remove that ancient code around now if we had moved things sooner

PR #2218, opened 2025-11-11
fix: no dangling file handles in `concat_on_disk`

### LINKED ISSUE #2198: `concat_on_disk` leaves open file handles when it fails

For example, it failed when I had Awkward Arrays in my `.obsm`, since that's not supported yet. However, when I tried to rewrite one of the files I was trying to concatenate in the same Jupyter Notebook, I found that I could not because the file handle was still open
```bash
OSError: Unable to synchronously create file (unable to truncate a file which is already open)
```

Would it be possible to add some cleanup code to close file handles when `concat_on_disk` errors out?

ilan-gold:
- [x] Closes #2198 
- [x] Tests added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,9 +42,10 @@ dependencies = [
     "h5py>=3.11",
     "natsort",
     "packaging>=24.2",
-    "array_api_compat>=1.7.1",
+    "array-api-compat>=1.7.1",
     "legacy-api-wrap",
     "zarr >=2.18.7, !=3.0.*",
+    "typing-extensions; python_version<'3.13'",
 ]
 dynamic = [ "version" ]
 
@@ -77,7 +78,7 @@ doc = [
 ]
 dev-doc = [ "towncrier>=24.8.0" ] # release notes tool
 test-min = [
-    "pytest>=8.2,<8.3.4",
+    "pytest>=8.2",
     "pytest-cov",           # only for VS Code
     "pytest-randomly",
     "pytest-memray",
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -509,9 +509,6 @@ def merge_only(ds: Collection[Mapping]) -> Mapping:
 # Interface
 ###################
 
-# Leaving out for now, it's ugly in the rendered docs and would be adding a dependency.
-# from typing_extensions import Literal
-# UNS_STRATEGIES_TYPE = Literal[None, "same", "unique", "first", "only"]
 MERGE_STRATEGIES = {
     None: lambda x: {},
     "same": merge_same,


src/anndata/experimental/merge.py
```diff
@@ -616,101 +639,105 @@ def concat_on_disk(  # noqa: PLR0912, PLR0913, PLR0915
...
-    # Combining indexes
-    concat_indices = pd.concat(
-        [pd.Series(_df_index(g[axis_name])) for g in groups], ignore_index=True
-    )
-    if index_unique is not None:
-        concat_indices = concat_indices.str.cat(
-            label_col.map(str, na_action="ignore"), sep=index_unique
+    with (
+        as_group(out_file, mode="w") as output_group,
+        ListContextManager([as_group(f, mode="r") for f in in_files]) as groups,
```
flying-sheep: Just use [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) here instead of defining your own class, it’s only very slightly more verbose.

Also this whole function is enormous. Is it possible to reduce the block here to basically the following to avoid all that indentation and diff, or would that create too many parameters?

```py
def concat_on_disk(...) -> ...:
    ...
    with ExitStack() as stack, as_group(out_file, mode="w") as output_group:
        groups = [stack.enter_context(as_group(f, mode="r")) for f in in_files]
        return _concat_on_disk_inner(...)


def _concat_on_disk_inner(...) -> ...:
    ...
```

tests/test_concatenate_disk.py
```diff
@@ -278,10 +284,59 @@ def test_output_dir_exists(tmp_path):
...
+    in_pth = tmp_path / f"in.{file_format}"
+    in_pth2 = tmp_path / f"in2.{file_format}"
+    out_pth = tmp_path / f"out.{file_format}"
+
+    adata = AnnData(X=np.ones((2, 1)))
+    getattr(adata, f"write_{file_format}")(in_pth)
+    getattr(adata, f"write_{file_format}")(in_pth2)
+
+    # Intentionally write an unsupported array type, which could leave dangling file handles:
+    # https://github.com/scverse/anndata/issues/2198
```
flying-sheep: this comment is wrongly copy-pasted, right?

In the other test it makes sense.

tests/test_concatenate_disk.py
```diff
@@ -278,10 +284,59 @@ def test_output_dir_exists(tmp_path):
...
+    try:
+        concat_on_disk([in_pth, in_pth2], out_pth)
+    except NotImplementedError:
+        for path in [in_pth, in_pth2, out_pth]:
+            # should not error out because there are no file handles open
+            f = h5py.File(path, mode="w")
+            f.close()
+
+
+def test_write_using_paths(tmp_path, file_format):
```
flying-sheep: What’s the purpose of this test? Which function’s path support is being tested here?

tests/test_dask.py
```diff
@@ -134,35 +134,34 @@ def test_dask_distributed_write(
...
+        assert_equal(curr.varm["a"], orig.varm["a"])
+        assert_equal(curr.obsm["a"], orig.obsm["a"])
+        assert_equal(curr.X, orig.X)
+
+        assert isinstance(curr.X, np.ndarray)
+        assert isinstance(curr.obsm["a"], np.ndarray)
+        assert isinstance(curr.varm["a"], np.ndarray)
+        assert isinstance(orig.X, DaskArray)
+        assert isinstance(orig.obsm["a"], DaskArray)
+        assert isinstance(orig.varm["a"], DaskArray)
```
flying-sheep: does this all need the group open?

```suggestion
    with pytest.raises(AssertionError):
        assert_equal(curr.obsm["a"], curr.obsm["b"])

    assert_equal(curr.varm["a"], orig.varm["a"])
    assert_equal(curr.obsm["a"], orig.obsm["a"])
    assert_equal(curr.X, orig.X)

    assert isinstance(curr.X, np.ndarray)
    assert isinstance(curr.obsm["a"], np.ndarray)
    assert isinstance(curr.varm["a"], np.ndarray)
    assert isinstance(orig.X, DaskArray)
    assert isinstance(orig.obsm["a"], DaskArray)
    assert isinstance(orig.varm["a"], DaskArray)
```

src/anndata/experimental/merge.py
ilan-gold: Wow all my googling (which is what occupied a large swath of the time on this PR) and I couldn't not find that, I was quite surprised something didn't exist.

I can refactor yes, I agree about the block indentation.

tests/test_concatenate_disk.py
ilan-gold: Misnamed, it should be `using_groups`

src/anndata/experimental/merge.py
flying-sheep: I usually just go through the handful of stuff in the relevant library:
- context manager stuff: `contextlib`
- iteration stuff: `itertools`
- functional programming: `functools`

ilan-gold: I'll have to stop relying on the PageRank / AI overlords at google so much...definitely the way to go

flying-sheep (approved):

PR #2220, opened 2025-11-12
fix: column subsetting on `Dataset2D` without index loaded

ilan-gold:
- [x] Closes issue raised in https://github.com/biocontext-ai/anndata-mcp/issues/4#issuecomment-3521738900
- [x] Tests added

### DIFF
--- a/tests/test_xarray.py
+++ b/tests/test_xarray.py
@@ -69,6 +69,7 @@ def test_true_index_dim_column_subset(dataset2d, df):
     df_expected = dataset2d[cols].to_memory()
     # account for the fact that we manually set `true_index_dim`
     df.index = df[col]
+    df.index.name = None
     pd.testing.assert_frame_equal(df_expected, df[cols])
 
 


tests/test_xarray.py
```diff
@@ -57,7 +59,20 @@ def test_backed_property(dataset2d):
...
+
+    # Ensure we can actually select columns properly that are not the index column
+    cols = [next(col_iter), next(col_iter)]
+    df_expected = dataset2d[cols].to_memory()
+    # account for the fact that we manually set `true_index_dim`
+    df_expected.index = df.index
+    assert np.all(df_expected == df[cols])
+
+
+def test_index_dim(dataset2d, df):
```
flying-sheep:
```suggestion
def test_index_dim(dataset2d):
```

tests/test_xarray.py
```diff
@@ -57,7 +59,20 @@ def test_backed_property(dataset2d):
...
+    col_iter = iter(dataset2d.keys())
+    col = next(col_iter)
+    dataset2d.true_index_dim = col
+
+    # Ensure we can actually select columns properly that are not the index column
+    cols = [next(col_iter), next(col_iter)]
+    df_expected = dataset2d[cols].to_memory()
+    # account for the fact that we manually set `true_index_dim`
+    df_expected.index = df.index
+    assert np.all(df_expected == df[cols])
```
flying-sheep: Something like this?

```suggestion
    pd.testing.assert_frame_equal(df_expected, df[cols])
```

flying-sheep (commented):
Looks good!

Will that result in `dataset2d[[a, b]]` having 3 columns sometimes or is the `true_index_dim` stored in a way that accounts for that?

ilan-gold:
> Will that result in dataset2d[[a, b]] having 3 columns sometimes or is the true_index_dim stored in a way that accounts for that?

Right, so in the case where the "true index" (like cell labels) lies in the "columns" i.e., it is not in `xarray.Dataset.coords` so `Dataset2D.index_dim != Dataset2D.true_index_dim`, then that selection will now always have `[a, b, true_index_dim]` as the columns plus the index, but `to_memory` will materialize only `[a, b]` with `true_index_dim` as the index on the resultant `Dataframe`.  I updated the test to make that clearer: https://github.com/scverse/anndata/pull/2220/commits/4a935012b33a7b7eb79e96172858cd2bddbf9716. Thanks for the push!

PR #2223, opened 2025-11-14
docs: Add `.cruft.json`, remove news section, update key contributors

Zethson:
- Adds `.cruft.json` to allow us to enable template sync
- Removes unused prettier config files
- Removes the not really used/updated news section
- Adds Ilan to the key contributors & remove links to people that didn't work anyways

This could be multiple PRs. Let me know if you want them split up but I wanted to be pragmatic.
I've also only added Phil in the authors field as I didn't want to add all of them. It should be fine because this is often manually fixed anyways.


.cruft.json
```diff
@@ -0,0 +1,43 @@
...
+            "project_description": "Annotated data.",
+            "author_full_name": "Philipp Angerer",
+            "author_email": "philipp.angerer@helmholtz-munich.de",
+            "github_user": "scverse",
+            "github_repo": "anndata",
+            "license": "BSD 3-Clause License",
+            "ide_integration": true,
+            "_copy_without_render": [
+                ".github/workflows/build.yaml",
+                ".github/workflows/test.yaml",
```
ilan-gold, line 18: Wouldn't we want to change the name of our files to get updates from these? I think our tests now are called `test-cpu.yml`. As far as I understand it, this section is not used after the initial directory creation.  Still my question stands.

ilan-gold (commented):
I'm kind of in interested in going "all-in" on the cookie cutter or not doing it at all but don't feel strongly.  I'd be interested in hearing why not. (So my concern is not the size of this PR)

.cruft.json
Zethson: copy without render means that cookiecutter won't do cookiecutter variable interpolation for these files.

You'll get updates for these files in any case. What will happen is that it'll try to add these files in addition to your existing ones. You can sort it out manually then or rename your existing one to match the template names.

Zethson:
> I'm kind of in interested in going "all-in" on the cookie cutter or not doing it at all but don't feel strongly. I'd be interested in hearing why not. 

I don't have the time to recreate the project with the template in mind or to do manual diffs to match both. I suggest that you'll ensure that you adhere to the template with the next 0.7.0 sync. The PR diff will help.

ilan-gold:
> I don't have the time to recreate the project with the template in mind or to do manual diffs to match both. I suggest that you'll ensure that you adhere to the template with the next 0.7.0 sync. The PR diff will help.

Seems reasonable.

PR #2226, opened 2025-11-17
perf: don't use `DataArray`

ilan-gold:
Declaring `DataArray` with `coords` caused xarray to check the `coords` one-by-one to ensure they match.  But we don't need that and the performance hit is big if the `coords` (i.e., `index` in `pandas`-speak) are very big.

I added the benchmark with all the columns then to highlight this fact.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -1181,6 +1181,7 @@ def make_dask_col_from_extension_dtype(
     A :class:`dask.Array`: representation of the column.
     """
     import dask.array as da
+
     from anndata._io.specs.lazy_methods import (
         compute_chunk_layout_for_axis_size,
         get_chunksize,


benchmarks/benchmarks/dataset2d.py
```diff
@@ -18,11 +18,11 @@ class Dataset2D:
     params = (
         ("zarr", "h5ad"),
         ((-1,), None),
-        ("cat", "numeric", "string-array", "nullable-string-array"),
+        ("cat", "numeric", "string-array", "nullable-string-array", "all"),
     )
 
     def setup_cache(self):
-        n_obs = 10000
+        n_obs = 100000
```
flying-sheep:
```suggestion
        n_obs = 100_000
```

benchmarks/benchmarks/dataset2d.py
```diff
@@ -45,6 +45,15 @@ def setup_cache(self):
                     ad.io.write_elem(store, "df", df)
                 if writing_string_array_on_disk:
                     assert store["df"]["a"].attrs["encoding-type"] == "string-array"
+        for store in [
+            h5py.File("data_all.h5ad", mode="w"),
+            zarr.open("data_all.zarr", mode="w", zarr_version=2),
+        ]:
+            df = pd.DataFrame(array_types, index=[f"cell{i}" for i in range(n_obs)])
+            # write a string array
+            df["string-array"] = df["string-array"].to_numpy()
```
flying-sheep: I don’t understand what this is for. Is that basically something like `.asdtype(object)`? And why?

benchmarks/benchmarks/dataset2d.py
flying-sheep:
```suggestion
        array_type: Literal["cat", "numeric", "string-array", "nullable-string-array", "all"],
```

benchmarks/benchmarks/dataset2d.py
ilan-gold: In the wake of the pandas 3 stuff, that would be a safer bet.  I will update here, but the goal was to trigger https://github.com/scverse/anndata/blob/71966500949adcac4e49d2233f06e9f11f438e19/src/anndata/_io/specs/methods.py#L557-L559

flying-sheep (approved):

PR #2228, opened 2025-11-18
chore: simplify unify_dtypes

flying-sheep:
Supersedes #2176

### DIFF
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -280,7 +280,7 @@ def unify_dtypes(
     return dfs
 
 
-def try_unifying_dtype(  # noqa: PLR0911
+def try_unifying_dtype(
     dtypes: Sequence[np.dtype | ExtensionDtype], *, has_missing: bool
 ) -> ExtensionDtype | type[object] | None:
     """Determine unified dtype if possible.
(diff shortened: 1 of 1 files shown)


src/anndata/_core/merge.py
```diff
@@ -242,106 +242,92 @@ def as_cp_sparse(x) -> CupySparseMatrix:
...
+        )
+        if len(categories_intersection) < len(categories):
+            return object
+
+        if all(
+            dtype.ordered and np.all(categories == dtype.categories)
+            for dtype in dtypes
+            if len(dtype.categories) > 0
+        ):
+            return next(iter(dtypes))
```
ilan-gold: What if we just move this up? Is there a case where this condition could be met but `len(categories_intersection) < len(categories)` that we actually need to account for?

flying-sheep: you know more about this than me! https://github.com/pydata/xarray/issues/10247

I just refactored this without changing the semantics

ilan-gold: I do think so.  This checks for all dtypes having same categories and being ordered.  If that is true, then the dtypes are the same and we return any one of them.  Otherwise `object`.  I was more asking because I saw it as a simplification.

src/anndata/_core/merge.py
```diff
@@ -242,106 +242,92 @@ def as_cp_sparse(x) -> CupySparseMatrix:
...
+        return object
```
ilan-gold:
```suggestion
        if all(
            dtype.ordered and np.all(categories == dtype.categories)
            for dtype in dtypes
            if len(dtype.categories) > 0
        ):
            return next(iter(dtypes))
        return object
```

ilan-gold (approved):

PR #2250, opened 2025-12-12
fix: backport pandas `infer_string` fixes

### LINKED ISSUE #2248: Backport Pandas 3.0 stuff

See #2221 to 0.12 without changing any default behavior - see corresponding PR for discussion about what should be backported vs. not.

ilan-gold:
Changes:

1. If `infer_string` is true, conversion of non-standard indexes to string proceed as in #2133
2. No io changes i.e., roundtripping is not guaranteed
3. No change to semantics of `allow_write_nullable_strings` - this ensures everything is opt-in with `infer_string`
4. All read-time defaults are the same (so we use `NULLABLE_NUMPY_STRING_TYPE` when applicable + default to reading as `StringDtype` with `pd.NA`)

- [x] Closes #2248
- [x] Tests added

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,8 +122,8 @@ packages = [ "src/anndata", "src/testing" ]
 data_file = "test-data/raw-coverage"
 source_pkgs = [ "anndata" ]
 omit = [ "src/anndata/_version.py", "**/test_*.py" ]
-concurrency = [ "multiprocessing" ]
-parallel = true
+patch = [ "subprocess" ]
+
 [tool.coverage.xml]
 output = "test-data/coverage.xml"
 [tool.coverage.paths]


src/anndata/_io/specs/methods.py
```diff
@@ -1216,12 +1206,22 @@ def _string_array(
...
+@_REGISTRY.register_read(ZarrGroup, IOSpec("nullable-string-array", "0.1.0"))
+def _read_nullable_string(
+    elem: GroupStorageType, *, _reader: Reader
+) -> pd.api.extensions.ExtensionArray:
+    values = _reader.read_elem(elem["values"])
+    mask = _reader.read_elem(elem["mask"])
+    dtype = pd.StringDtype()
+
+    arr = pd.array(
+        values.astype(NULLABLE_NUMPY_STRING_TYPE),
```
flying-sheep, line 1220: I think this might be causing the current test failures.

I didn’t really build the PR with numpy 1 in mind, so backporting it might be slightly involved due to that additional degree of freedom.

flying-sheep (approved):
looks good! would be nice if you documented in the PR text what you left out/changed from #2133

IIRC something like
- Keep `NULLABLE_NUMPY_STRING_TYPE` for numpy 1 compatibility
- Don’t change semantics of `allow_write_nullable_strings`
- …?

src/anndata/_io/specs/methods.py
ilan-gold: This doesn't have an impact on the PR - the failures were just related to my turning on `infer_string` for pre-release and forgetting everythign stops roundtripping at that point - I added an explicit test for `infer_string` now

Issue #2252, opened 2025-12-16
Keep providing a non-nullable string type?


flying-sheep:
With pandas 3, string arrays will be inferred to become one of a dedicated pair of types (`pd.arrays.{Arrow,}StringArray`) and no longer behave like numpy `object` arrays, specifically they now have clear missing value / “nullable” behavior, see [PDEP 14](https://pandas.pydata.org/pdeps/0014-string-dtype.html).

We added experimental support for writing `{Arrow,}StringArray` as “nullable string arrays” behind a flag in the past, and with anndata 0.13, this will become active by default.

However, this means that reading a “non-nullable” array from disk means it becomes a “nullable” one in memory, and subsequent writes will write them as nullable.

This raises two questions:
- is there a in-memory overhead?
- can we avoid the on-disk overhead?

  we could either try a different serialization (oof, late!) or have a way for users to indicate (maybe using some string dtype variant)

Since the answer to the first question has no options, and is important for the second, I tried to find it below:

# Results, TL;DR

no in-memory overhead, if we can trust the self-reported memory usage

`NumpyExtensionArray` with `np.dtypes.StringDType()` is small and non-nullable. We should test that it gets written using the old non-nullable string array serialization, and maybe advise users to convert string columns to it if that boolean mask on disk is too big. Opinions?

# Results, Details

I tried to figure out how much memory the different options use (see below), but didn’t get a convincing result, I think since [the memory just doesn’t get freed immediately](https://bloomberg.github.io/memray/memory.html#memory-is-not-freed-immediately)

`pd.arrays.NumpyExtensionArray` is just a slim wrapper around a numpy array and doesn’t take noticably more memory. The same currently applies to `pd.arrays.StringArray` (i.e. `pd.StringDtype(storage="python")`).

So here are the self-reported stats (using pandas’ memory_usage which also tracks object overhead):

| Array class | dtype | #bytes |
|--------|--------|--------|
| `NumpyExtensionArray` | `"U10"` | 40 MB |
| `NumpyExtensionArray` | `object` | 59 MB |
| `NumpyExtensionArray` | `np.dtypes.StringDType(na_object=None)` | 16 MB |
| `NumpyExtensionArray` | `np.dtypes.StringDType()` | 16 MB |
| `StringArray` | `pd.StringDtype(storage="python")` | 59 MB  |
| `ArrowStringArray` | `pd.StringDtype(storage="pyarrow")` | 18 MB |

[...]

PR #2254, opened 2025-12-16
fix: disallow nullable on-disk categoricals for strings

ilan-gold:
- [x] See https://github.com/scverse/anndata/issues/2252#issuecomment-3660515124 
- [ ] Tests added

### DIFF
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -1107,9 +1107,7 @@ def write_categorical(
     _writer.write_elem(
         g,
         "categories",
-        v.categories._values
-        if not pd.api.types.is_string_dtype(v.categories)
-        else v.categories.to_numpy(),
+        v.categories.to_numpy(),
         dataset_kwargs=dataset_kwargs,
     )
 


src/anndata/_io/specs/methods.py
```diff
@@ -1093,7 +1093,12 @@ def write_categorical(
 
     _writer.write_elem(g, "codes", v.codes, dataset_kwargs=dataset_kwargs)
     _writer.write_elem(
-        g, "categories", v.categories._values, dataset_kwargs=dataset_kwargs
+        g,
+        "categories",
+        v.categories._values
+        if not pd.api.types.is_string_dtype(v.categories)
+        else np.array(v.categories),
+        dataset_kwargs=dataset_kwargs,
```
flying-sheep: I wonder if we should call into `write_vlen_string_array{,_zarr}` instead of converting this multiple times. `_values` is also iffy …

Also I noticed that we write `T` arrays as non-nullable even though we should check if they’re nullable and decide based on that (but that’s maybe out-of-scope for this PR)

https://github.com/scverse/anndata/blob/46a8c6065c6b9f834382c83e299e0e29a2e1a606/src/anndata/_io/specs/methods.py#L581

ilan-gold: > _values is also iffy …

Oh it's crazy that we use this, no doubt.

> I wonder if we should call into write_vlen_string_array{,_zarr} instead of converting this multiple times. _values is also iffy …

What would this solve? I think `v.categories` is a pandas object and when we do `np.array` (I guess I assumed this), it should go to object dtype in which case `write_vlen_string_array` is dispatched.  So you're saying we should do `np.array(v.categories, dtype=object)` then?

flying-sheep: Your current code (both if branches) are just `v.categories.to_numpy()`, right? I don’t understand why that ternary exists at all.

ilan-gold: I can't say I know exactly what `v.categories._values` is for or expected return type, just that it works at the moment.  So I didn't want to touch it yet (hence draft).  But yes, `to_numpy` on the other end for strings is better.

I want to minimize numpy conversion because if arrow types land in zarr, we should preserve the type of the categories (they could be not string) in which case we would use `.array` and not `._values`

ilan-gold: Using `to_numpy` now on the string side :)

ilan-gold: (happy to do both sides if you feel strongly, just trying to minimze behavior changes)

flying-sheep: `._values` is very well documented: https://github.com/pandas-dev/pandas/blob/9c40b37d7102a2a21ec45e31a4a21dc12756d59b/pandas/core/series.py#L794

It’s “`.to_numpy()` if it’s a NumpyExtensionArray, else `.array`”.

So I’m saying let’s get rid of private APIs, no matter how well-documented, and use public ones instead.

ilan-gold: Alright, sounds good!

flying-sheep: Behaviorally, I’d say we should use what causes the fewest repetitions of “go through the whole array and convert every single element”.

As pointed out in #2272, hdf5 understands numpy string dtype arrays, but I think pandas doesn’t do them yet.

PR #2270, opened 2025-12-30
fix: `compressor=None` kwarg handling with zarr v3

ilan-gold:
- [x] Fixes a small bug I noticed in #1927 
- [x] Tests added


src/anndata/_io/specs/methods.py
```diff
@@ -94,9 +94,21 @@
 #     return False
 
 
-def zarr_v3_compressor_compat(dataset_kwargs) -> dict:
-    if not is_zarr_v2() and (compressor := dataset_kwargs.pop("compressor", None)):
```
ilan-gold, line 98: Previously, the `not is_zarr_v2()` would evaluate to `True` but `compressor` being `None` would be `False`y so the `pop` happens but the assignment to `compressors` doesn't.  Oops!

flying-sheep: Welp, always a gamble if `None` is a valid option or just means “default”.

tests/test_readwrite.py
```diff
@@ -377,37 +377,50 @@ def check_compressed(key, value):
...
 
-        # Don't use Blosc since it's defaults can change:
-        # https://github.com/zarr-developers/zarr-python/pull/3545
-        compressor = ZstdCodec(level=3, checksum=True)
-    not_compressed = []
+            # Don't use Blosc since it's defaults can change:
+            # https://github.com/zarr-developers/zarr-python/pull/3545
+            compressor = ZstdCodec(level=3, checksum=True)
+    else:
+        compressor = None
```
flying-sheep: I’d do

```py
if not use_compression:
    compressor = None
elif zarr_write_format == 2 or is_zarr_v2():
    ...
else:
    ...
```

tests/test_readwrite.py
```diff
@@ -377,37 +377,50 @@ def check_compressed(key, value):
...
     ad.io.write_zarr(pth, adata, compressor=compressor)
 
     def check_compressed(value, key):
         if not isinstance(value, ZarrArray) or value.shape == ():
             return None
-        (read_compressor,) = value.compressors
+        if len(value.compressors) > 0:
+            (read_compressor,) = value.compressors
+        else:
+            read_compressor = None
```
flying-sheep: How about this? Or can `value.compressors` also be `None` and that means something else?

```python
(read_compressor,) = value.compressors or [None]
```

tests/test_readwrite.py
```diff
@@ -417,11 +430,11 @@ def check_compressed(value, key):
...
-    if not_compressed:
+    if wrongly_compressed:
         sep = "\n\t"
         msg = (
-            f"These elements were not compressed correctly:{sep}"
-            f"{sep.join(not_compressed)}"
+            f"These elements were not (un)compressed correctly:{sep}"
+            f"{sep.join(wrongly_compressed)}"
         )
         raise AssertionError(msg)
```
flying-sheep: This could simply be

```py
assert not wrongly_compressed, "Some elements were not (un)compressed correctly"
```

pytest rewrites assertions well.

PR #2272, opened 2026-01-04
fix: decode bytes in nullable-string-array index for read_lazy

### LINKED ISSUE #2271: experimental.read_lazy: obs_names/var_names contain "b'...'" instead of decoded strings

### Report

When using `anndata.experimental.read_lazy()`, the `obs_names` and `var_names` indices contain string representations of bytes (e.g., `"b'cell_A'"`) instead of proper decoded strings (e.g., `"cell_A"`).

This differs from `read_h5ad()` which correctly returns decoded strings.

**Code:**

```python
import tempfile
import numpy as np
import h5py
from anndata import AnnData
import anndata as ad
from anndata.experimental import read_lazy

# Create test file
with tempfile.NamedTemporaryFile(suffix='.h5ad', delete=False) as f:
    path = f.name

adata = AnnData(np.zeros((5, 3)))
adata.obs.index = ['cell_A', 'cell_B', 'cell_C', 'cell_D', 'cell_E']
adata.var.index = ['gene_X', 'gene_Y', 'gene_Z']
adata.write_h5ad(path)

# Read with read_lazy
with h5py.File(path, 'r') as f:
    lazy = read_lazy(f)
    print('read_lazy obs_names:', list(lazy.obs_names))

# Compare with regular read
regular = ad.read_h5ad(path)
print('read_h5ad obs_names:', list(regular.obs_names))
```

**Output:**

```
read_lazy obs_names: ["b'cell_A'", "b'cell_B'", "b'cell_C'", "b'cell_D'", "b'cell_E'"]
read_h5ad obs_names: ['cell_A', 'cell_B', 'cell_C', 'cell_D', 'cell_E']
```

**Expected:** `read_lazy` should return `['cell_A', 'cell_B', ...]` like `read_h5ad` does.

**Impact:**
- Index lookups fail (e.g., `adata["cell_A", :]` won't match `"b'cell_A'"`)
- Display of lazy AnnData shows incorrect index values
- Any code comparing index values will behave unexpectedly

**Possible root cause:** The xarray/Dataset2D integration may be converting bytes from HDF5 using `str(bytes_value)` instead of `bytes_value.decode('utf-8')`.

### Versions

Validated on commit `4a62c2e3128b950e7d699506b91d8e572da20a96` (main branch, 2025-01-04):

| Package | Version                            |
| ------- | ---------------------------------- |
| anndata | 0.13.0.dev181+g22818754d.d20251227 |
| h5py    | 3.13.0                             |
| xarray  | 2025.11.0                          |
| numpy   | 2.1.0                              |
| pandas  | 3.0.0rc0                           |
| scipy   | 1.15.2                             |

| Component | Info                                                                             |
| --------- | -------------------------------------------------------------------------------- |
| Python    | 3.12.9 \| packaged by conda-forge \| (main, Mar  4 2025, 22:44:42) [Clang 18.1.8] |
| OS        | macOS-15.6.1-arm64-arm-64bit                                                     |

katosh:
# Fix `read_lazy` nullable-string-array index decoding

- [x] Closes #2271
- [x] Tests added
- [x] Release note unnecessary (bug not yet released)

## Summary

Fixes `read_lazy()` returning `obs_names`/`var_names` as byte-string representations (`"b'cell_A'"`) instead of properly decoded strings (`"cell_A"`).

## Fix

HDF5 stores strings as bytes. When `MaskedArray.__getitem__` calls `values.astype(StringDType)` on these bytes, it uses `str()` which produces `"b'...'"` instead of decoding.

The fix uses h5py's `.astype("T")` in `read_nullable` to decode strings at the source:

```python
# lazy_methods.py
values = elem["values"]
if encoding_type == "nullable-string-array" and isinstance(elem, H5Group):
    values = values.astype("T")
```

This follows h5py's recommendation to use `.astype("T")` over `.asstr()` when using numpy ≥2.

## Compatibility

The `H5AstypeView` type (returned by `.astype()`) is imported with a fallback chain for h5py version compatibility:
- `AsTypeView` (newer h5py)
- `AstypeWrapper` (h5py 3.11)
- Dynamic fallback using `in_memory()` (future h5py versions)

## Test

Added `test_nullable_string_index_decoding` as a regression test that verifies index values are properly decoded strings after `read_lazy()`.

## Edit History

**Jan 4:** Initial fix decoded bytes in `MaskedArray.__getitem__` using `np.char.decode()`.

**Jan 5:** Refactored to use h5py's `.asstr()` at the source in `read_nullable` instead of decoding in `MaskedArray.__getitem__`. This is consistent with how regular string arrays are handled and avoids per-access decode overhead. Also updated `ZarrOrHDF5Wrapper` to handle `AsStrView` which lacks the `chunks` attribute.

**Jan 5:** Switched from `.asstr()` to `.astype("T")` per h5py recommendation for numpy ≥2. Updated typing to use `H5AstypeView` with nested try-except fallback chain for h5py version compatibility. The `in_memory()` fallback is only reached in future h5py versions where both `AsTypeView` and `AstypeWrapper` are removed.

### DIFF
--- a/src/anndata/_io/specs/lazy_methods.py
+++ b/src/anndata/_io/specs/lazy_methods.py
@@ -368,7 +368,7 @@ def read_nullable(
         values = values.astype("T")
     return MaskedArray(
         values=values,
-        mask=elem.get("mask", None),
+        mask=elem["mask"],
         dtype_str=encoding_type,
         base_path_or_zarr_group=base_path_or_zarr_group,
         elem_name=elem_name,
--- a/src/anndata/compat/__init__.py
+++ b/src/anndata/compat/__init__.py
@@ -76,17 +76,17 @@ H5File = h5py.File
 
 # h5py recommends using .astype("T") over .asstr() when using numpy ≥2
 if TYPE_CHECKING:
-    from h5py._hl.dataset import AsTypeView as H5AstypeView
+    from h5py._hl.dataset import AsTypeView as H5AsTypeView
 else:
     try:
         try:
-            from h5py._hl.dataset import AsTypeView as H5AstypeView
+            from h5py._hl.dataset import AsTypeView as H5AsTypeView
         except ImportError:
             # h5py 3.11 uses AstypeWrapper (lowercase 't')
-            from h5py._hl.dataset import AstypeWrapper as H5AstypeView
+            from h5py._hl.dataset import AstypeWrapper as H5AsTypeView
     except ImportError:  # pragma: no cover
         warn("AsTypeView changed import location", DeprecationWarning)
-        H5AstypeView = type(
+        H5AsTypeView = type(
             h5py.File.in_memory().create_dataset("x", shape=(), dtype="S1").astype("U1")
         )
 
--- a/src/anndata/experimental/backed/_lazy_arrays.py
+++ b/src/anndata/experimental/backed/_lazy_arrays.py
@@ -13,7 +13,7 @@ from anndata._io.specs.lazy_methods import get_chunksize
 from ..._settings import settings
 from ...compat import (
     H5Array,
-    H5AstypeView,
+    H5AsTypeView,
     XBackendArray,
     XDataArray,
     XZarrArrayWrapper,
@@ -36,7 +36,7 @@ if TYPE_CHECKING:
         from xarray.core.indexing import ExplicitIndexer
 
 
-class ZarrOrHDF5Wrapper[K: (H5Array, H5AstypeView, ZarrArray)](XZarrArrayWrapper):
+class ZarrOrHDF5Wrapper[K: (H5Array | H5AsTypeView, ZarrArray)](XZarrArrayWrapper):
     def __init__(self, array: K) -> None:
         # AstypeView from h5py .astype() lacks chunks attribute
         self.chunks = getattr(array, "chunks", None)
@@ -71,7 +71,7 @@ class ZarrOrHDF5Wrapper[K: (H5Array, H5AstypeView, ZarrArray)](XZarrArrayWrapper
         if (
             isinstance(key, np.ndarray)
             and np.issubdtype(key.dtype, np.integer)
-            and isinstance(self._array, H5Array | H5AstypeView)
+            and isinstance(self._array, H5Array | H5AsTypeView)
         ):
             key_mask = np.zeros(self._array.shape).astype("bool")
             key_mask[key] = True
@@ -87,7 +87,7 @@ class CategoricalArray[K: (H5Array, ZarrArray)](XBackendArray):
     """
 
     _codes: ZarrOrHDF5Wrapper[K]
-    _categories: ZarrArray | H5Array
+    _categories: K
     shape: tuple[int, ...]
     base_path_or_zarr_group: Path | ZarrGroup
     elem_name: str
@@ -95,7 +95,7 @@ class CategoricalArray[K: (H5Array, ZarrArray)](XBackendArray):
     def __init__(
         self,
         codes: K,
-        categories: ZarrArray | H5Array,
+        categories: K,
         base_path_or_zarr_group: Path | ZarrGroup,
         elem_name: str,
         *args,
@@ -135,10 +135,10 @@ class CategoricalArray[K: (H5Array, ZarrArray)](XBackendArray):
 
 
 # circumvent https://github.com/tox-dev/sphinx-autodoc-typehints/issues/580
-type K = H5Array | H5AstypeView | ZarrArray
+type K = H5Array | H5AsTypeView | ZarrArray
 
 
-class MaskedArray[K: (H5Array, H5AstypeView, ZarrArray)](XBackendArray):
+class MaskedArray[K: (H5Array | H5AsTypeView, ZarrArray)](XBackendArray):
     """
     A wrapper class meant to enable working with lazy masked data.
     We do not guarantee the stability of this API beyond that guaranteed
@@ -154,11 +154,11 @@ class MaskedArray[K: (H5Array, H5AstypeView, ZarrArray)](XBackendArray):
 
     def __init__(
         self,
-        values: ZarrArray | H5Array | H5AstypeView,
+        values: K,
         dtype_str: Literal[
             "nullable-integer", "nullable-boolean", "nullable-string-array"
         ],
-        mask: ZarrArray | H5Array,
+        mask: K,
         base_path_or_zarr_group: Path | ZarrGroup,
         elem_name: str,
     ):


src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -37,7 +37,8 @@
 
 class ZarrOrHDF5Wrapper[K: (H5Array, ZarrArray)](XZarrArrayWrapper):
     def __init__(self, array: K) -> None:
-        self.chunks = array.chunks
+        # AsStrView from h5py .asstr() lacks chunks attribute
```
flying-sheep: So you need to both update the type parameter and any isinstance(self._array, H5Array) check.

Please make sure you add a `pytest.param` to `test_safe_fancy_index_h5py_function` to make sure your change works with that.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -361,8 +361,12 @@ def read_nullable(
         Path(filename(elem)) if isinstance(elem, H5Group) else elem
     )
     elem_name = get_elem_name(elem)
+    values = elem["values"]
+    # HDF5 stores strings as bytes; use .asstr() to decode on access
+    if encoding_type == "nullable-string-array" and isinstance(elem, H5Group):
+        values = values.asstr()
     return MaskedArray(
```
flying-sheep, line 369: Same here, you need to update `MaskedArray`’s type parameter and `__init__` arg type

tests/lazy/test_read.py
```diff
@@ -234,3 +234,44 @@ def test_chunks_df(
...
+    """
+    import h5py
+
+    expected_obs = ["cell_A", "cell_B", "cell_C", "cell_D", "cell_E"]
+    expected_var = ["gene_X", "gene_Y", "gene_Z"]
+
+    # Create AnnData with custom index names
+    adata = AnnData(np.zeros((len(expected_obs), len(expected_var))))
+    adata.obs.index = pd.Index(expected_obs)
+    adata.var.index = pd.Index(expected_var)
```
flying-sheep: Just do adata = AnnData(obs=dict(obs_names=[…]), var=dict(var_names=[...])) instead of this whole block.

Or, to use string columns instead of indices, just name the columns anything but `obs_names` and `var_names`.

tests/lazy/test_read.py
```diff
@@ -234,3 +234,44 @@ def test_chunks_df(
...
+        lazy = read_lazy(f)
+        obs_names = list(lazy.obs_names)
+        var_names = list(lazy.var_names)
+
+    # Check for byte-string representation pattern first (more informative error)
+    for name in obs_names + var_names:
+        assert not name.startswith("b'"), (
+            f"Index value {name!r} contains byte-string representation. "
+            "Bytes should be decoded to strings, not converted via str()."
+        )
```
flying-sheep: I don’t think this is necessary, pytests’ assertion rewriting should produce nice enough errors just given the code below.

tests/lazy/test_read.py
```diff
@@ -234,3 +234,44 @@ def test_chunks_df(
     for k in ds:
         if isinstance(arr := ds[k].data, DaskArray):
             assert arr.chunksize == expected_chunks
+
+
+def test_nullable_string_index_decoding(tmp_path: Path):
```
flying-sheep, line 239: This test currently passes on main, probably because we write indices as non-nullable if possible. Please change it so it actually tests your PR’s changes (e.g. maybe it works using string columns, not indices)

src/anndata/experimental/backed/_lazy_arrays.py
katosh: Thanks for the thorough review! I wanted to clarify regarding the suggestion to add a `pytest.param` to `test_safe_fancy_index_h5py_function`.

## Understanding the Code Paths

I think there might be some confusion about which code paths are involved, so let me lay this out:

### 1. `test_safe_fancy_index_h5py_function` (in `tests/test_backed_hdf5.py`)

This test exercises `_safe_fancy_index_h5py()` from `src/anndata/_core/index.py`:

```python
def test_safe_fancy_index_h5py_function(h5py_test_data, indices, description):
    with h5py.File(h5_path, "r") as f:
        dataset = f["test"]  # Regular h5py.Dataset
        result = _safe_fancy_index_h5py(dataset, indices)
```

This function is designed for **regular backed AnnData** and expects an `h5py.Dataset`. It handles h5py's limitations with multi-dimensional fancy indexing.

### 2. Our changes (in `src/anndata/experimental/backed/_lazy_arrays.py`)

Our fix is in `ZarrOrHDF5Wrapper._getitem()`, which handles indexing for **lazy/xarray-backed AnnData**:

```python
def _getitem(self, key):
    if (
        isinstance(key, np.ndarray)
        and np.issubdtype(key.dtype, np.integer)
        and isinstance(self._array, H5Array | H5AsStrView)  # Our change
    ):
        # Boolean mask workaround
```

These are **separate code paths** that don't interact directly.

## Considerations

| Aspect | `_safe_fancy_index_h5py` | `ZarrOrHDF5Wrapper._getitem` |
|--------|--------------------------|------------------------------|
| Location | `_core/index.py` | `experimental/backed/_lazy_arrays.py` |
| Used by | Regular backed AnnData | Lazy/xarray AnnData (`read_lazy`) |
| Input type | `h5py.Dataset` | `H5Array`, `H5AsStrView`, `ZarrArray` |
| Our changes | None | Added `H5AsStrView` support |

## Options Going Forward

1. **Current state (recommended)**: Our `test_nullable_string_index_decoding` test covers the `AsStrView` code path end-to-end through `read_lazy()`. This exercises `ZarrOrHDF5Wrapper._getitem` with the `H5AsStrView` type.

2. **Add explicit unit test**: I could add a dedicated test for `ZarrOrHDF5Wrapper` with `AsStrView` and integer array indexing, if you'd prefer more granular coverage.

3. **Extend `_safe_fancy_index_h5py`**: If you'd like `_safe_fancy_index_h5py` to also support `AsStrView`, that would be a separate change beyond the scope of this bug fix.

## Additional Note

I verified that `AsStrView` actually supports integer array indexing directly (unlike regular `h5py.Dataset` in some cases):

```python
with h5py.File(path, 'r') as f:
    asstr = f['test'].asstr()
    idx = np.array([0, 2, 4])
    print(asstr[idx])  # Works: ['a' 'c' 'e']
```

So the existing boolean mask workaround in `ZarrOrHDF5Wrapper._getitem` is defensive but not strictly required for `AsStrView`.

Please let me know which direction you'd prefer, or if I've misunderstood something!

flying-sheep: Yeah, I meant testing that the `isinstance` in `ZarrOrHDF5Wrapper._getitem` is covered. But I guess since it’s about speed and not correctness, we can’t test.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -361,8 +361,12 @@ def read_nullable(
         Path(filename(elem)) if isinstance(elem, H5Group) else elem
     )
     elem_name = get_elem_name(elem)
+    values = elem["values"]
+    # HDF5 stores strings as bytes; use .asstr() to decode on access
+    if encoding_type == "nullable-string-array" and isinstance(elem, H5Group):
+        values = values.asstr()
```
flying-sheep: H5Py recommends using [`.astype("T")`](https://docs.h5py.org/en/latest/high/dataset.html#h5py.Dataset.astype) over `.asstr()` when using numpy ≥2, which we do.

src/anndata/compat/__init__.py
```diff
@@ -73,6 +73,10 @@ class Empty(Enum):
 H5Group = h5py.Group
 H5Array = h5py.Dataset
 H5File = h5py.File
+# AsStrView was renamed from AsStrWrapper in h5py 3.12
+_H5AsStrView = getattr(h5py._hl.dataset, "AsStrView", None)
+_H5AsStrWrapper = getattr(h5py._hl.dataset, "AsStrWrapper", None)
+H5AsStrView: type = _H5AsStrView or _H5AsStrWrapper
```
flying-sheep: With `astype("T")` instead, we’d have a `AsTypeView`! And the way you did this isn’t very typing friendly, please do:


```suggestion
if TYPE_CHECKING:
    from h5py._hl.dataset import AsTypeView as H5AsTypeView
else:
    try:
        from h5py._hl.dataset import AsTypeView as H5AsTypeView
    except ImportError:
        warn("AsTypeView changed import location", DeprecationWarning)
        H5AsTypeView = type(
            h5py.File.in_memory().create_dataset("x", shape=(), dtype="u8").astype("i8")
        )
```

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -152,11 +154,11 @@ class MaskedArray[K: (H5Array, ZarrArray)](XBackendArray):
 
     def __init__(
         self,
-        values: ZarrArray | H5Array,
+        values: ZarrArray | H5Array | H5AsStrView,
         dtype_str: Literal[
             "nullable-integer", "nullable-boolean", "nullable-string-array"
         ],
-        mask: ZarrArray | H5Array,
+        mask: ZarrArray | H5Array | None,
```
flying-sheep: That makes no sense, `ZarrOrHDF5Wrapper` doesn’t accept `None`, but is unconditionally created with `mask`.

If `mask=None` is ever passed here, we need to change more than this type.

katosh: Ok, I see. I will revert this for now. I only added it because of https://github.com/scverse/anndata/blob/af09beb495052348a4fc1ffcb40ecaf8ef552a07/src/anndata/_io/specs/lazy_methods.py#L366

flying-sheep: Thanks for pointing it out, must be wrong then!

katosh:
Thanks for the thorough review, @flying-sheep!

I've updated the implementation to use `.astype("T")` instead of `.asstr()` as recommended. Also renamed `H5AsStrView` to `H5AstypeView` throughout to reflect this change.

Regarding the `mask: ... | None` type annotation: I reverted that since the `elem.get("mask", None)` pattern was already in the codebase and not introduced by this PR. Happy to address that separately if needed.

src/anndata/compat/__init__.py
```diff
@@ -74,6 +74,18 @@ class Empty(Enum):
...
+if TYPE_CHECKING:
+    from h5py._hl.dataset import AsTypeView as H5AstypeView
+else:
+    try:
+        from h5py._hl.dataset import AsTypeView as H5AstypeView
+    except ImportError:
+        warn("AsTypeView changed import location", DeprecationWarning)
+        # Create in-memory file to get the type (in_memory() not available in older h5py)
+        with h5py.File("_", mode="w", driver="core", backing_store=False) as _f:
+            H5AstypeView = type(_f.create_dataset("x", shape=(), dtype="S1").astype("U1"))
```
flying-sheep: The idea was that the warning and the `in_memory()` are only called in *future* h5py versions, not currently tested ones. Something like this should work!

```suggestion
    try:
        try:
            from h5py._hl.dataset import AsTypeView as H5AstypeView
        except ImportError:
            from h5py._hl.dataset import AsTypeWrapper as H5AstypeView
    except ImportError:
        warn("AsTypeView changed import location", DeprecationWarning)
        H5AstypeView = type(
            h5py.File.in_memory().create_dataset("x", shape=(), dtype="S1").astype("U1")
        )
```

katosh: You'r right. This is typing friendly and covers past and *future* h5py versions where both `AsTypeView` and `AsTypeWrapper` might be removed.

I updated this in 2129825ed5b2525d1491ea2fad49a3c96dcd771b.

katosh:
The `hatch-test.stable` failures are due to a flaky doctest in `concat_on_disk` an HDF5 truncated file error unrelated to this PR:

```
OSError: Unable to synchronously open file (truncated file: eof = 475136, sblock->base_addr = 0, stored_eof = 22528845)
```

All other tests pass, including `hatch-test.min` which validates the h5py compatibility fix. Should we just re-run of the stable tests?

flying-sheep:
@katosh please allow us to commit to your fork, I want to be able to hit that “merge upsteam changes” button so I can see if the codecov changes are a fluke or real somehow.

katosh:
Of course! I have you and @ilan-gold write access to the repo, and I see you already merged it 👍

src/anndata/compat/__init__.py
```diff
@@ -74,6 +74,22 @@ class Empty(Enum):
...
+if TYPE_CHECKING:
+    from h5py._hl.dataset import AsTypeView as H5AstypeView
+else:
+    try:
+        try:
+            from h5py._hl.dataset import AsTypeView as H5AstypeView
+        except ImportError:
+            # h5py 3.11 uses AstypeWrapper (lowercase 't')
+            from h5py._hl.dataset import AstypeWrapper as H5AstypeView
+    except ImportError:
```
flying-sheep:
```suggestion
    except ImportError:  # pragma: no cover
```

flying-sheep:
Thanks @katosh!

PR #2274, opened 2026-01-05
fix: don’t double-open h5ad files in backed mode

### LINKED ISSUE #719: Unclosed HDF5 file handle
The HDF5 file object is not explicitly closed here. In some cases, it causes the file to be locked even after `adata.file.close()`. It can be fixed by using the context manager.

https://github.com/theislab/anndata/blob/283b0c13b235b41a9c8fd0cb3c4712698858aa33/anndata/_io/h5ad.py#L134

flying-sheep:
- [x] Closes #719
- [x] Tests added

I’m sneakily expanding the `filename` argument to also take a `h5py.File` object. There are probably less hacky ways to fix this, but this is pretty minimalist.

### DIFF
--- a/docs/release-notes/2274.fix.md
+++ b/docs/release-notes/2274.fix.md
@@ -1,0 +1,1 @@
+Only open HDF5 file once in backed mode {user}`flying-sheep`
--- a/src/anndata/_io/h5ad.py
+++ b/src/anndata/_io/h5ad.py
@@ -188,6 +188,7 @@ def read_h5ad_backed(
     d["raw"] = _read_raw(f, attrs={"var", "varm"})
 
     adata = AnnData(**d)
+    assert adata.file._file is f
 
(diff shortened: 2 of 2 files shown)


flying-sheep:
> Shouldn't the types of read_h5ad and read_h5ad_backed change?

No, why? The only type changes I made is that `AnnData.__init__(filetype=...)` now accepts `h5py.File`, but as said, this is more of a hack than a feature I want to advertise.

If you have a better idea how to pass in the file object without making AnnData’s signature more complicated for a feature on its way out, I’m happy to listen!

ilan-gold (approved):
Hmm sorry you're right.  The first step in backed mode is still "open the file" and this just ensures that the `__init__` method doesn't reopen it.  I thought it enabled passing in your own file handle for some reason to `read_xxx`.

Issue #1699, opened 2024-09-26
Add official `testing.anndata` API


flying-sheep:
### Please describe your wishes and possible alternatives to achieve the desired result.

AnnData.tests.helpers is an undocumented mess

We should clearly distinguish

- [ ] between non-pytest helpers (`testing.anndata`) and a pytest plugin (`testing.anndata.pytest`). Only the second would import `pytest` and e.g. define the GPU marker (see https://github.com/scverse/rapids_singlecell/pull/267)
- [ ] between public and private helpers

And we should

- [ ] document how exactly things like `assert_equal` work and what they support. (As well as that they’re singledispatch and therefore extensible)

ilan-gold:
I've started looking into this and thought of a few more considerations:

1. Our `gen_adata` function uses types to filer keys in `varm`, `layers` etc. I'm not sure this is an ideal API but I am also fine with it.  The issue is what those default keys are - we currently have two optional dependencies as defaults (`awkward` and `xarray`).  I think both are valid (especially if https://github.com/scverse/anndata/issues/2043 lands, we'll need a nice test-case for it).  So, do we make a new extra for this functionality? Turn them off if they aren't installed and make it clear in the docs they are defaults otherwise? Remove them?
2. https://github.com/scverse/anndata/pull/2071 contains a nice simplification I think, so probably best to wait for it to land before getting carried away with this.
3. `assert_equal` is in relatively good condition, I think and its guarantee to me seems to be clear - we aren't checking for types but rather numerical equality so don't expect that.  We also probably don't want to make performance guarantees here since we denseify fairly aggressively.

`assert_equal` and `gen_adata` is probably all I would export - we cover data generation and assertion, you cover the test

PR #2276, opened 2026-01-05
ci: switch to dependency groups

flying-sheep:
- [ ] Closes #
- [x] Tests added

Considerations: scanpy and/or integration testing and/or other packages might try installing `anndata[test]` since they depend on anndata’s unofficial test utils.

Which isn’t a good idea, that extra is for `anndata`’s own test (and therefore being migrated to a dependency group)

So *if* the anndata test utils need anything the dependent packages’s tests don’t already depend on anyway, we have two routes:
1. merge this, fix other packages’ CI by adding the necessary deps for the unofficial test utils themselves, then implement #1699, and finally switch the other packages to the official test utils
2. Implement #1699, switch dependent packages over to the official test utils, merge this.

the second one might me one step less, but I don’t think it’s going to be any easier or harder.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -32,7 +32,7 @@ jobs:
       ASV_DIR: "./benchmarks"
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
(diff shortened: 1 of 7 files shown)


flying-sheep:
shall I set the milestone to 0.12.8 and backport this?

ilan-gold (approved):

PR #2280, opened 2026-01-05
fix: use urlretrieve for reliable download in concat_on_disk doctest

### LINKED ISSUE #2282: Flaky doctest failures due to HDF5 file corruption during cellxgene downloads

## Report

The `read_lazy` and `concat_on_disk` doctests have intermittently failed with HDF5 corruption errors when downloading example files from cellxgene. Multiple failures have been observed since October 2025.

### Observed Failures

**1. `concat_on_disk` doctest** ([run 18879594928](https://github.com/scverse/anndata/actions/runs/18879594928), Oct 28, 2025)

```
FAILED src/anndata/experimental/merge.py::anndata.experimental.merge.concat_on_disk
UNEXPECTED EXCEPTION: KeyError('Unable to synchronously open object (free block size is zero?)')
```

**2. `concat_on_disk` doctest** ([run 19856037175](https://github.com/scverse/anndata/actions/runs/19856037175), Dec 2, 2025)

```
FAILED src/anndata/experimental/merge.py::anndata.experimental.merge.concat_on_disk
UNEXPECTED EXCEPTION: OSError('Unable to synchronously open file (truncated file: eof = 10682368, sblock->base_addr = 0, stored_eof = 22528845)')
```

**3. `concat_on_disk` doctest on main** ([run 20139516066](https://github.com/scverse/anndata/actions/runs/20139516066), Dec 11, 2025)

```
FAILED src/anndata/experimental/merge.py::anndata.experimental.merge.concat_on_disk
UNEXPECTED EXCEPTION: KeyError('Unable to synchronously open object (bad symbol table node signature)')
```

**4. `concat_on_disk` doctest** ([run 20722954348](https://github.com/scverse/anndata/actions/runs/20722954348), Jan 5, 2026)

```
FAILED src/anndata/experimental/merge.py::anndata.experimental.merge.concat_on_disk
UNEXPECTED EXCEPTION: OSError('Unable to synchronously open file (truncated file: eof = 475136, sblock->base_addr = 0, stored_eof = 22528845)')
```

**5. `read_lazy` doctest** ([run 20722393723](https://github.com/scverse/anndata/actions/runs/20722393723), Jan 5, 2026)

```
FAILED src/anndata/experimental/backed/_io.py::anndata.experimental.backed._io.read_lazy
UNEXPECTED EXCEPTION: KeyError('Unable to synchronously open object (bad symbol table node signature)')
Error raised while reading key 'obs' of <class 'h5py._hl.group.Group'> from /
```

### Affected doctests

Both doctests download ~22MB H5AD files from cellxgene to `sc.settings.datasetdir`:
- `a93eab58-3d82-4b61-8a2f-d7666dcdb7c4.h5ad`
- `d170ff04-6da0-4156-a719-f8e1bbefbf53.h5ad`

The download code in the doctests:
```python
>>> def get_cellxgene_data(id_: str):
...     out_path = sc.settings.datasetdir / f"{id_}.h5ad"
...     if out_path.exists():
...         return out_path
...     file_url = f"{base_url}/{id_}.h5ad"
...     sc.settings.datasetdir.mkdir(parents=True, exist_ok=True)
...     # concat_on_disk uses: out_path.write_bytes(httpx.get(file_url).content)
...     # read_lazy uses: urlretrieve(file_url, out_path)
...     return out_path
```

### Potential contributing factors

1. **Download method**: `concat_on_disk` uses `httpx.get().content` which buffers in memory before writing
2. **Parallel test execution**: pytest-xdist runs tests in parallel (`gw0`, `gw1`, etc.) on the same machine
3. **Shared cache**: Both doctests download to the same `sc.settings.datasetdir` path with the same filenames
4. **Non-atomic writes**: Both `httpx` and `urlretrieve` write directly to the target path

### Versions

```python
CI environment (from failed runs):
- Python 3.13
- pytest-xdist 3.8.0
- hatch-test.stable environment
- Ubuntu 24.04
```

katosh:
- [x] Closes #2282
- [ ] Tests added
- [x] Release note unnecessary (doctest fix only)

## Summary

Improves robustness of `concat_on_disk` and `read_lazy` doctests after observing intermittent HDF5 corruption errors during cellxgene file downloads. See #2282 for details.

### Changes

1. **Replace `httpx.get().content` with `urlretrieve`** in `concat_on_disk`:
   - Downloads directly to file (no memory buffering)
   - Stdlib only (no extra dependency)
   - Consistent with `read_lazy` doctest

2. **Use distinct cache file prefixes**:
   - `read_lazy`: `lazy_{id_}.h5ad`
   - `concat_on_disk`: `concat_{id_}.h5ad`
   - Prevents potential conflicts if doctests run in parallel via pytest-xdist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -173,6 +173,7 @@ qualname_overrides = {
     "numpy.dtypes.StringDType": ("py:attr", "numpy.dtypes.StringDType"),
     "pandas.DataFrame.iloc": ("py:attr", "pandas.DataFrame.iloc"),
     "pandas.DataFrame.loc": ("py:attr", "pandas.DataFrame.loc"),
+    "pandas.core.dtypes.dtypes.BaseMaskedDtype": "pandas.api.extensions.ExtensionDtype",
     # should be fixed soon: https://github.com/tox-dev/sphinx-autodoc-typehints/pull/516
     "types.EllipsisType": ("py:data", "types.EllipsisType"),
     "pathlib._local.Path": "pathlib.Path",
--- a/docs/release-notes/2287.fix.md
+++ b/docs/release-notes/2287.fix.md
@@ -1,0 +1,1 @@
+Fix {obj}`numpy.uint` support in {func}`anndata.experimental.read_lazy` and {func}`anndata.experimental.read_elem_lazy` {user}`flying-sheep`
--- a/src/anndata/_io/utils.py
+++ b/src/anndata/_io/utils.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from functools import WRAPPER_ASSIGNMENTS, wraps
+from functools import WRAPPER_ASSIGNMENTS, cache, wraps
 from itertools import pairwise
 from typing import TYPE_CHECKING, Literal, cast
+
+import numpy as np
+import pandas as pd
 
 from .._core.sparse_dataset import BaseCompressedSparseDataset
 from ..utils import warn
@@ -11,6 +14,8 @@ from ..utils import warn
 if TYPE_CHECKING:
     from collections.abc import Callable, Mapping
     from typing import Any, Literal
+
+    from pandas.core.dtypes.dtypes import BaseMaskedDtype
 
     from .._types import StorageType, _WriteInternal
     from ..compat import H5Group, ZarrGroup
@@ -117,6 +122,30 @@ def check_key(key):
     else:
         msg = f"{key} of type {typ} is an invalid key. Should be str."
         raise TypeError(msg)
+
+
+@cache
+def pandas_nullable_dtype(dtype: np.dtype) -> BaseMaskedDtype:
+    """Infer nullable dtype from numpy dtype.
+
+    There is no public pandas API for this, so this is the cleanest way.
+    See <https://github.com/pandas-dev/pandas/issues/63608>
+    """
+    try:
+        from pandas.core.dtypes.dtypes import BaseMaskedDtype
+    except ImportError:
+        pass
+    else:
+        return BaseMaskedDtype.from_numpy_dtype(dtype)
+
+    match dtype.kind:
+        case "b":
+            array_type = pd.arrays.BooleanArray
+        case "i" | "u":
+            array_type = pd.arrays.IntegerArray
+        case _:
+            raise NotImplementedError
+    return array_type(np.ones(1, dtype), np.ones(1, bool)).dtype
 
 
 # -------------------------------------------------------------------------------
--- a/src/anndata/experimental/backed/_lazy_arrays.py
+++ b/src/anndata/experimental/backed/_lazy_arrays.py
@@ -10,6 +10,7 @@ from anndata._core.index import _subset
 from anndata._core.views import as_view
 from anndata._io.specs.lazy_methods import get_chunksize
 
+from ..._io.utils import pandas_nullable_dtype
 from ..._settings import settings
 from ...compat import (
     H5Array,
@@ -24,8 +25,9 @@ if TYPE_CHECKING:
     from pathlib import Path
     from typing import Literal
 
+    from numpy.typing import NDArray
     from pandas._libs.missing import NAType
-    from pandas.core.dtypes.base import ExtensionDtype
+    from pandas.core.dtypes.dtypes import BaseMaskedDtype
 
     from anndata.compat import ZarrGroup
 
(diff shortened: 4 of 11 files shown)


katosh:
### Note on download robustness

This fix addresses the immediate issue (parallel test conflicts via distinct prefixes, and avoiding the memory buffer from `httpx.get().content`), but `urlretrieve` still isn't fully robust. Interrupted downloads leave corrupt files in the cache, and there's no integrity check.

### Existing download utilities

I looked for existing functionality that could be reused:

| Option | Status |
|--------|--------|
| **anndata** | No download utilities |
| **scanpy `_download`** | Private function; cleans up on exception but not atomic (writes directly to target path) |
| **pooch `retrieve`** | Requires `known_hash` upfront; not an anndata dependency |

Since the doctest already imports scanpy, `scanpy.readwrite._download` could potentially be used, but it's private API and still not atomic.

### Fully robust alternative

For reference, a more robust implementation would look like:

```python
>>> import httpx
>>> def get_cellxgene_data(id_: str):
...     out_path = sc.settings.datasetdir / f"concat_{id_}.h5ad"
...     if out_path.exists():
...         return out_path
...     file_url = f"{base_url}/{id_}.h5ad"
...     tmp_path = out_path.with_suffix(".h5ad.part")
...     sc.settings.datasetdir.mkdir(parents=True, exist_ok=True)
...     with httpx.stream("GET", file_url, timeout=30, follow_redirects=True) as r:
...         r.raise_for_status()
...         expected = int(r.headers.get("Content-Length", 0))
...         written = 0
...         with tmp_path.open("wb") as f:
...             for chunk in r.iter_bytes():
...                 f.write(chunk)
...                 written += len(chunk)
...     if expected and written != expected:
...         tmp_path.unlink(missing_ok=True)
...         raise IOError(f"Incomplete download: {written}/{expected} bytes")
...     tmp_path.rename(out_path)
...     return out_path
```

This would add: atomic write (`.part` → rename), Content-Length check, proper error raising, timeout, cleanup on failure.

That said, this might be overkill for a doctest example. The current fix should resolve the parallel test conflict, and if flaky failures persist, `# doctest: +SKIP` might be the simplest path forward.

ilan-gold (commented):
Seems reasonable, thanks! Do we need the `httpx` dependency anymore in `test-min`?

katosh:
Awesome! You are right, there seem to be no other uses of `httpx`. Thank you for reviewing!

flying-sheep:
We’re kinda recommending pooch everywhere, so maybe adding it to the test deps isn’t a bad idea.

@ilan-gold what do you think?

ilan-gold:
That seems like a great idea!

katosh:
Thanks for the suggestion! Updated to use `pooch.retrieve()` instead of `urlretrieve`.

Benefits: atomic downloads (temp file + rename), built-in caching, and better error handling.

Also simplified the filenames to just `{id_}.h5ad` since pooch handles concurrent access more safely. Both doctests now share the same cache.

src/anndata/experimental/merge.py
```diff
@@ -554,19 +554,18 @@ def concat_on_disk(  # noqa: PLR0913
     See :func:`anndata.concat` for the semantics.
     The following examples highlight the differences this function has.
 
-    First, let’s get some “big” datasets with a compatible ``var`` axis:
+    First, let's get some "big" datasets with a compatible ``var`` axis:
```
flying-sheep:
```suggestion
    First, let’s get some “big” datasets with a compatible ``var`` axis:
```

src/anndata/experimental/backed/_io.py
```diff
@@ -62,17 +62,16 @@ def read_lazy(
...
-    ...     out_path = sc.settings.datasetdir / f"{id_}.h5ad"
-    ...     if out_path.exists():
-    ...         return out_path
-    ...     file_url = f"{base_url}/{id_}.h5ad"
-    ...     sc.settings.datasetdir.mkdir(parents=True, exist_ok=True)
-    ...     urlretrieve(file_url, out_path)
-    ...     return out_path
+    ...     return pooch.retrieve(
+    ...         f"{base_url}/{id_}.h5ad",
+    ...         known_hash=None,
```
flying-sheep: As you wrote:

> `pooch` requires `known_hash` upfront

so please actually use it here and in the other example.

katosh: Thanks for the suggestion! I looked into adding hash verification, but there's a fundamental issue:

**The function accepts arbitrary dataset IDs:**

```python
def get_cellxgene_data(id_: str):
    ...
    pooch.retrieve(url, known_hash=???)  # What hash do we use?
```

To use `known_hash`, we'd need either:
1. **A lookup table** mapping every cellxgene dataset ID → hash (impractical, there are thousands of datasets)
2. **Fetch hash from cellxgene API** (the API doesn't provide checksums)

Even if we hardcoded hashes for just the two IDs used in the doctests, we'd be creating a confusing API where the function only works for those specific IDs, defeating the purpose of parameterizing by `id_`.

**Additional concerns:**

- These are living datasets that cellxgene can update anytime, invalidating any hardcoded hashes
- The existing benchmark code (`benchmarks/benchmarks/readwrite.py`) uses `known_hash=None` for the same reasons

Happy to reconsider if there's a way to fetch hashes dynamically from cellxgene!

flying-sheep: We’re not creating a reusable API, as we define these functions ad-hoc in the doctests. So just add a second parameter `hash: str` and pass in the hash when you call the function!

katosh: Done!

PR #2287, opened 2026-01-08
fix: allow reading nullable uint arrays in `read_{,elem_}lazy`

flying-sheep:
- [ ] Closes #
- [x] Tests added

Use `pandas.core.dtypes.dtypes.BaseMaskedDtype.from_numpy_dtype` (with fallback to array constructors) instead of string manipulation to get `MaskedArray`’s dtype.

Extracted and improved from #2279

### DIFF
--- a/tests/lazy/test_concat.py
+++ b/tests/lazy/test_concat.py
@@ -166,17 +166,21 @@ def test_concat_to_memory_obs(
 
 
 def test_concat_to_memory_obs_dtypes(
+    subtests: pytest.Subtests,
     lazy_adatas_for_concat: list[AnnData],
     join: Join_T,
-):
+) -> None:
     concated_remote = ad.concat(lazy_adatas_for_concat, join=join)
     # check preservation of non-categorical dtypes on the concat axis
-    assert concated_remote.obs["int64"].dtype == "int64"
-    assert concated_remote.obs["uint8"].dtype == "uint8"
-    assert concated_remote.obs["nullable-int"].dtype == "int32"
-    assert concated_remote.obs["float64"].dtype == "float64"
-    assert concated_remote.obs["bool"].dtype == "bool"
-    assert concated_remote.obs["nullable-bool"].dtype == "bool"
+    for name in concated_remote.obs.columns:
+        dtype = name.removeprefix("nullable-")
+        with subtests.test(col=name):
+            try:
+                assert concated_remote.obs[name].dtype == dtype
+            except AssertionError:
+                if "cat" in name:
+                    pytest.xfail("categorical dtypes are not preserved")
+                raise
 
 
 def test_concat_to_memory_var(


ilan-gold:
Is this a public API? https://pandas.pydata.org/docs/search.html?q=BaseMaskedDtype

flying-sheep:
Hm, no: https://pandas.pydata.org/docs/reference/index.html

> The `pandas.core`, `pandas.compat`, and `pandas.util` top-level modules are PRIVATE. Stable functionality in such modules is not guaranteed.

I’ll fall back to the previous approach then.

src/anndata/_io/utils.py
```diff
@@ -119,6 +124,32 @@ def check_key(key):
...
+    else:
+        return BaseMaskedDtype.from_numpy_dtype(dtype)
+
+    match dtype.kind:
+        case "b":
+            array_type = pd.arrays.BooleanArray
+        case "i" | "u":
+            array_type = pd.arrays.IntegerArray
+        case "f":
+            array_type = pd.arrays.FloatingArray
```
ilan-gold: I think this should error out.  These arrays are tied to the strings like `nullable-integer` via our format spec: https://anndata.readthedocs.io/en/latest/fileformat-prose.html#nullable-integers-booleans-and-strings and there is no `nullable-floating`

flying-sheep:
> what we had wasn't random "string manipulation" but the actual on-disk file format names

*That* code is not what I referred to! I’m talking about this fragile construct:

https://github.com/scverse/anndata/blob/088fa16d9f402369224709f6238f67e12eed7941/src/anndata/experimental/backed/_lazy_arrays.py#L194-L197

What you mean is that the previous code dispatched on `_dtype_string` whereas mine dispatches on `values.dtype.kind`.

I believe that what I do is perfectly safe and robust since the anndata version with this code only supports `nullable-integer` 0.1.0, `nullable-boolean` 0.1.0, and `nullable-string-array` 0.1.0. A hypothetical future anndata version that supports constructing a masked array from integers or bools in a different way will gain support for the new protocols together with new code dispatching on that.

My changes just make sure that we can confidently say we support `UInt64` without looking into our tests (will `str.capitalize` create that? Probably not? maybe?)

flying-sheep:
Nope, it doesn’t:

```python
import pandas as pd, numpy as np
nullable_dtype = lambda dtype: pd.array([], dtype=str(pd.api.types.pandas_dtype(dtype)).capitalize()).dtype
nullable_dtype(np.dtype(np.uint64))
```

```pytb
Cell In[3], line 1
----> 1 nullable_dtype(np.dtype(np.uint64))

Cell In[2], line 1, in <lambda>(dtype)
----> 1 nullable_dtype = lambda dtype: pd.array([], dtype=str(pd.api.types.pandas_dtype(dtype)).capitalize()).dtype 

File /usr/lib/python3.13/site-packages/pandas/core/construction.py:311, in array(data, dtype, copy)
    309 # this returns None for not-found dtypes.
    310 if dtype is not None:
--> 311     dtype = pandas_dtype(dtype)
    313 if isinstance(data, ExtensionArray) and (dtype is None or data.dtype == dtype):
    314     # e.g. TimedeltaArray[s], avoid casting to NumpyExtensionArray
    315     if copy:

File /usr/lib/python3.13/site-packages/pandas/core/dtypes/common.py:1663, in pandas_dtype(dtype)
   1656     with warnings.catch_warnings():
   1657         # TODO: warnings.catch_warnings can be removed when numpy>2.3.0
   1658         # is the minimum version
   1659         # GH#51523 - Series.astype(np.integer) doesn't show
   1660         # numpy deprecation warning of np.integer
   1661         # Hence enabling DeprecationWarning
   1662         warnings.simplefilter("always", DeprecationWarning)
-> 1663         npdtype = np.dtype(dtype)
   1664 except SyntaxError as err:
   1665     # np.dtype uses `eval` which can raise SyntaxError
   1666     raise TypeError(f"data type '{dtype}' not understood") from err

TypeError: data type 'Uint64' not understood
```

flying-sheep:
OK, I’ll change this into a `fix` because it’s an actual bug:

```py
ad.AnnData(np.zeros((4, 6)), dict(nint=pd.array([1, 0, None, 1], dtype=pd.UInt32Dtype()))).write_zarr("/tmp/test.ad.zarr")
ad.experimental.read_lazy("/tmp/test.ad.zarr")
```

```pytb
TypeError: data type 'Uint32' not understood
Error raised while reading key 'obs' of <class 'zarr.core.group.Group'> from /
```

ilan-gold:
We should add a test. Ideally it wouldn't even be a new test but some setting or something.  I'm looking into how this ever even slipped through, because in theory it should be tested.

ilan-gold:
Thanks!

flying-sheep:
probably slipped through because the nullable-integer branch in our test utils was a bit inflexible, but I fixed that.

PR #2288, opened 2026-01-08
feat: add `LazyCategoricalDtype` for lazy categorical columns

### LINKED ISSUE #2283: feat: Efficient category count and partial loading for lazy AnnData
## Summary

Request an API to efficiently get category count (without loading values) and partial category loading for lazy AnnData columns.

## Motivation

For lazy AnnData (from `read_lazy()`), accessing category information via `col.dtype.categories` works but always loads ALL category values. For datasets with many categories (e.g., cell IDs, barcodes), this is inefficient when you only need the count or a preview.

**Current behavior:**
```python
from anndata.experimental import read_lazy

lazy_adata = read_lazy("dataset.zarr")  # 1M cells, 100k categories
col = lazy_adata.obs["cell_id"]

categories = col.dtype.categories  # ~25ms - loads all 100k values
n_categories = len(col.dtype.categories)  # Must load all just to count
```

**Desired behavior:**
```python
# Efficient count-only (no data loading):
n_categories = get_category_count(col)  # ~0.1µs, metadata only

# Partial loading for previews:
first_100 = get_categories(col, n=100)  # Load only first 100
```

## Benchmark Results

Tested with **1M cells and 100k categories** (median of 3 runs):

| Method | Zarr | H5AD | Data loaded |
|--------|------|------|-------------|
| `col.values` (full column) | 80 ms | 87 ms | All codes (1M) + categories |
| `col.dtype.categories` | 25 ms | 34 ms | Category group (values + mask) |
| `read_elem(values)` | 9 ms | 16 ms | Category values only |
| `read_elem_partial(n=100)` | 9.1 ms | 0.1 ms | First 100 categories |
| `values.shape[0]` (metadata) | 0.12 µs | 0.08 µs | Count only (no loading) |

**Key findings:**
- **Count from metadata is ~200,000x faster** than loading values
- `col.dtype.categories` reads the whole category group (values + mask), ~2.5x slower than reading values directly
- Partial loading via `read_elem_partial` is very fast for H5AD but shows no benefit for Zarr (possibly not optimized yet?)

<details>
<summary>Benchmark code</summary>

```python
import anndata as ad
from anndata.experimental import read_lazy
from anndata._io.specs.registry import read_elem, read_elem_partial
import numpy as np
import pandas as pd
import tempfile
import time
import os

# Create AnnData with 1M cells and 100k categories
n_obs = 1_000_000
n_cats = 100_000
adata = ad.AnnData(
    X=np.random.rand(n_obs, 10).astype(np.float32),
    obs=pd.DataFrame({
        'cell_id': pd.Categorical([f'Cell_{i % n_cats}' for i in range(n_obs)])
    })
)

with tempfile.TemporaryDirectory() as tmpdir:
    path = os.path.join(tmpdir, 'test.zarr')
    adata.write_zarr(path)
    lazy_adata = read_lazy(path)
    col = lazy_adata.obs['cell_id']

    # Navigate to storage
    cat_arr = col.variable._data.array
    cats_storage = cat_arr._categories
    values = cats_storage["values"] if hasattr(cats_storage, "keys") else cats_storage

    # Benchmark each method
    start = time.perf_counter()
    _ = col.dtype.categories
    print(f"col.dtype.categories: {(time.perf_counter() - start)*1000:.1f}ms")

    start = time.perf_counter()
    _ = read_elem(values)
    print(f"read_elem(values): {(time.perf_counter() - start)*1000:.1f}ms")

    start = time.perf_counter()
    _ = read_elem_partial(values, indices=slice(0, 100))
    print(f"read_elem_partial(n=100): {(time.perf_counter() - start)*1000:.1f}ms")

    start = time.perf_counter()
    _ = values.shape[0]
    print(f"values.shape[0]: {(time.perf_counter() - start)*1e6:.2f}µs")
```

</details>

<details>
<summary>Test environment</summary>

- **Hardware:** Apple Silicon (arm64)
- **Python:** 3.12.9
- **anndata:** 0.13.0.dev
- **zarr:** 3.1.5
- **numpy:** 2.4.0
- **pandas:** 3.0.0rc1

</details>

## Current Workaround (in PR #2236)

The HTML representation module implements a workaround for efficient category display by navigating to the underlying storage:

```python
from anndata.experimental.backed._lazy_arrays import CategoricalArray
from anndata._io.specs.registry import read_elem_partial

# Navigate xarray internals to find CategoricalArray
cat_arr = col.variable._data.array  # Fragile internal path

# Access storage directly (bypasses cached_property)
cats = cat_arr._categories  # zarr.Group or h5py.Group
values = cats["values"] if hasattr(cats, "keys") else cats

# Get count from metadata (no loading)
n_categories = values.shape[0]

# Partial loading
first_n = read_elem_partial(values, indices=slice(0, n))
```

**Why this workaround is fragile:**
- Navigates xarray internals (`col.variable._data.array`) which may change
- Relies on private `_categories` attribute
- Assumes storage structure (`cats["values"]` pattern)

### Option A: Methods on CategoricalArray

```python
class CategoricalArray:
    @property
    def n_categories(self) -> int:
        """Number of categories (metadata only, no loading)."""
        values = self._categories["values"] if hasattr(self._categories, "keys") else self._categories
        return values.shape[0]

    def get_categories(self, n: int | None = None) -> np.ndarray:
        """Get category values, optionally limited to first N."""
        values = self._categories["values"] if hasattr(self._categories, "keys") else self._categories
        if n is not None:
            return read_elem_partial(values, indices=slice(0, n))
        return read_elem(values)
```

### Option B: Utility function

```python
def get_lazy_column_categories(
    col: DataArray,
    n: int | None = None,
) -> tuple[np.ndarray, int]:
    """Efficiently get categories from a lazy column."""
```

### Option C: Integration with Dataset2D

```python
class Dataset2D:
    def get_column_categories(self, column: str, n: int | None = None) -> np.ndarray | None:
        """Get categories efficiently, with optional partial loading."""

    def get_column_category_count(self, column: str) -> int | None:
        """Get category count from metadata (no data loading)."""
```

## Use Cases

1. **HTML representation** - Display category count ("100,000 categories") and preview actual values (first N) without loading all (PR #2236, [visual demo](https://htmlpreview.github.io/?https://gist.githubusercontent.com/katosh/4a2399d1472c733b041ef8dfd5b489b9/raw/repr_html_visual_test.html#9b-lazy-anndata-experimental))
2. **Data inspection** - Quick count and preview of categories in large datasets
3. **Validation** - Check category count matches color arrays in `uns`

## References

- [PR #2236](https://github.com/scverse/anndata/pull/2236): HTML representation with lazy AnnData support
- [`CategoricalArray`](https://github.com/scverse/anndata/blob/a892252b/src/anndata/experimental/backed/_lazy_arrays.py#L80-L132)
- [Current workaround `_get_lazy_categories`](https://github.com/scverse/anndata/blob/aef9083f/src/anndata/_repr/utils.py#L261-L323)

katosh:
# feat: add `LazyCategoricalDtype` for lazy categorical columns

- [x] Closes #2283
- [x] Tests added
- [x] Release note added

## Summary

Add `LazyCategoricalDtype` extending `pd.CategoricalDtype` with lazy loading support for categorical columns in lazy AnnData. This enables efficient access to categorical metadata without loading all categories into memory.

```python
lazy_adata = ad.experimental.read_lazy("large_dataset.h5ad")
dtype = lazy_adata.obs["cell_type"].dtype  # LazyCategoricalDtype

# Cheap metadata access (no I/O)
dtype.n_categories     # 100000
dtype.ordered          # False

# Partial reads (efficient)
dtype.head_categories()     # first 5 categories
dtype.head_categories(10)   # first 10 categories
dtype.tail_categories()     # last 5 categories
dtype.tail_categories(10)   # last 10 categories

# Full load (cached after first access)
dtype.categories       # pd.Index with all categories
```

## Motivation

When working with lazy AnnData objects containing many categories (e.g., 100k+ cell IDs as categories), loading all categories just to display a preview or check metadata is inefficient. This is particularly important for:

1. **repr/HTML display** - showing category info without triggering full loads
2. **Data exploration** - quickly inspecting category names
3. **Memory efficiency** - avoiding unnecessary allocations

## API Design

Following [Ilan's suggestion](https://github.com/scverse/anndata/issues/2283#issuecomment-2608398505), the API uses familiar pandas naming conventions:

| Property/Method | Returns | Behavior |
|-----------------|---------|----------|
| `.categories` | `pd.Index` | Full load, cached (standard pandas) |
| `.ordered` | `bool` | Standard pandas |
| `.n_categories` | `int` | Cheap metadata access |
| `.head_categories(n=5)` | `np.ndarray` | First n categories (partial read) |
| `.tail_categories(n=5)` | `np.ndarray` | Last n categories (partial read) |

The `head`/`tail` naming follows pandas `DataFrame.head()`/`DataFrame.tail()` conventions.

## Implementation Details

- `LazyCategoricalDtype` extends `pd.CategoricalDtype` to maintain compatibility
- Categories are loaded lazily on first `.categories` access and cached
- `head_categories`/`tail_categories` use `read_elem_partial` for efficient partial reads
- Works with both zarr and h5ad backends

## Benchmark Results

Tested with **100k categories** (median of 5 runs):

| Method | H5AD | Zarr |
|--------|------|------|
| `n_categories` | 0.05 ms | 0.11 ms |
| `head_categories(10)` | 0.19 ms | 8.82 ms |
| `categories` (full) | 30.32 ms | 19.19 ms |

**Speedups vs full load:**

| Method | H5AD | Zarr |
|--------|------|------|
| `n_categories` | **621x** | **168x** |
| `head_categories(10)` | **160x** | 2.2x |

Note: zarr speedup for partial reads is limited because categories are currently written without explicit chunking.


src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+        *,
+        ordered: bool = False,
+    ):
+        # Create instance without calling parent __init__ with categories
+        instance = object.__new__(cls)
+        return instance
+
+    def __init__(
+        self,
+        categories_array: ZarrArray | H5Array | None = None,
```
ilan-gold: Why can this be `None`? The only thing I can think of that would cause this would be `CategoricalDtype()` which we never do to my knowledge (or something similar).  I think removing this would make things a lot simpler.

ilan-gold: Also isn't `{Zarr,H5}Group` valid?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+        *,
+        ordered: bool = False,
+    ):
+        self._categories_array = categories_array
```
ilan-gold: I would call this `_categories_elem` to make it clear it could be a group or array

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+    def n_categories(self) -> int:
+        """Number of categories (cheap, metadata only)."""
+        if self._cached_n_categories is not None:
+            return self._cached_n_categories
+        if self.__categories is not None:
+            return len(self.__categories)
+        if self._categories_array is not None:
+            n = self._get_categories_array().shape[0]
+            self._cached_n_categories = n
+            return n
```
ilan-gold: Why not just always do these three lines i.e., use `shape[0]`? Then there is no ned for caching

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+            return np.asarray(self.__categories[:n])
+
+        if self._categories_array is None:
+            return np.array([])
+
+        from anndata._io.specs.registry import read_elem_partial
+
+        arr = self._get_categories_array()
+        total = self.n_categories
+        return read_elem_partial(arr, indices=slice(0, min(n, total)))
```
ilan-gold: If `arr` is just a `{H5,Zarr}Array`, just use their raw slicing methods
```suggestion
        return arr[0:min(n, total))]
```

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+
+        if self._categories_array is None:
+            return np.array([])
+
+        from anndata._io.specs.registry import read_elem_partial
+
+        arr = self._get_categories_array()
+        total = self.n_categories
+        start = max(total - n, 0)
+        return read_elem_partial(arr, indices=slice(start, total))
```
ilan-gold: Duplicated with `head_categories` - please deduplicate

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+    def __repr__(self) -> str:
+        if self.__categories is not None:
+            # Fully loaded - use standard repr
+            return f"CategoricalDtype(categories={self.__categories!r}, ordered={self.ordered})"
+        return f"LazyCategoricalDtype(n_categories={self.n_categories}, ordered={self.ordered})"
+
+    @property
+    def name(self) -> str:
+        """String identifier for this dtype."""
+        return "category"
```
ilan-gold: Why do we need this? 
```suggestion
```

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+
+    def __hash__(self) -> int:
+        # Need to be hashable for pandas internals
+        return hash((id(self._categories_array), self._ordered_flag))
```
ilan-gold: What exactly fails without this?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+        -------
+        np.ndarray or ExtensionArray
+            The first n categories.
+
+        Examples
+        --------
+        .. code-block:: python
+
+            dtype.head_categories()  # first 5
+            dtype.head_categories(10)  # first 10
```
ilan-gold: We don't need this for methods, that's what the docstring explanation is for :)

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+
+        For string-array encoding: _categories_array is directly the array.
+        For nullable-string-array encoding: _categories_array is a Group with "values" key.
+        """
+        if isinstance(self._categories_array, (ZarrArray, H5Array)):
+            return self._categories_array
+        # nullable-string-array encoding: Group with "values" and "mask"
+        return self._categories_array["values"]
+
+    @property
```
ilan-gold:
```suggestion
    @cached_property
```
And then remove the custom caching?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+        categories_array: ZarrArray | H5Array | None = None,
+        *,
+        ordered: bool = False,
+    ):
+        self._categories_array = categories_array
+        self._ordered_flag = bool(ordered)
+        self._cached_n_categories: int | None = None
+        self.__categories: pd.Index | None = (
+            None  # Double underscore to avoid conflicts
+        )
```
ilan-gold: As mentioned in the `def categories` comment, I think you can just cache that property.

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+
+        Examples
+        --------
+        .. code-block:: python
+
+            dtype.head_categories()  # first 5
+            dtype.head_categories(10)  # first 10
+        """
+        # If already fully loaded, slice from cache
+        if self.__categories is not None:
```
ilan-gold:
```suggestion
       "categories" in self.__dict__:
```
If you remove the custom caching (as I assume you should be able to) in `def categories`

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
...
+
+        dtype = col.dtype  # LazyCategoricalDtype
+        dtype.n_categories  # cheap, metadata only
+        # 100000
+        dtype.head_categories(5)  # partial read, first 5
+        # array(['cat_0', 'cat_1', 'cat_2', 'cat_3', 'cat_4'], dtype='<U6')
+        dtype.tail_categories(3)  # partial read, last 3
+        # array(['cat_99997', 'cat_99998', 'cat_99999'], dtype='<U6')
+        dtype.categories  # full load, cached
+        # Index(['cat_0', 'cat_1', ...], dtype='str')
```
ilan-gold:
```suggestion
```

Not sure what this provides over the current description + docs

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,209 @@
         from xarray.core.indexing import ExplicitIndexer
 
 
+class LazyCategoricalDtype(pd.CategoricalDtype):
+    """A CategoricalDtype that lazily loads categories from zarr/h5 storage.
+
+    This dtype provides efficient access to categorical metadata without loading
+    all categories into memory. Categories are loaded lazily on first full access
+    and cached thereafter.
```
ilan-gold:
```suggestion
    This dtype provides efficient access to categorical metadata without loading
    all categories into memory via `head_categories`, `tail_categories` and `n_categories`. Otherwise, `LazyCategoricalDtype.categories` will load the categories into memory.
```

Something along these lines.  Also with intersphinx for the methods once you fix the docs :)

ilan-gold (changes requested):
Thanks! Looking good!

src/anndata/experimental/backed/_lazy_arrays.py
katosh: Just to check that this is what you want: Raw slicing can end up returning encoded byte strings while users might expect to receive the decoded strings.

```python
import h5py
import tempfile
import numpy as np
from anndata._io.specs.registry import read_elem_partial

# Create HDF5 file with string data
with tempfile.NamedTemporaryFile(suffix='.h5') as f:
    with h5py.File(f.name, 'w') as h5:
        # Store strings (HDF5 stores as bytes internally)
        h5.create_dataset('categories', data=['Cat_000', 'Cat_001', 'Cat_002'])

    with h5py.File(f.name, 'r') as h5:
        arr = h5['categories']

        # DIRECT SLICING: Returns bytes
        direct_result = arr[:2]
        print(f"Direct slice: {direct_result}")
        # Output: [b'Cat_000' b'Cat_001']
        print(f"Type: {type(direct_result[0])}")
        # Output: <class 'bytes'>

        # read_elem_partial: Returns decoded strings
        partial_result = read_elem_partial(arr, indices=slice(0, 2))
        print(f"read_elem_partial: {partial_result}")
        # Output: ['Cat_000' 'Cat_001']
        print(f"Type: {type(partial_result[0])}")
        # Output: <class 'str'>
```

**Justification:** `read_elem_partial` handles:
- HDF5 byte-to-string decoding
- Various string encodings (vlen strings, fixed-length)
- Nullable string arrays with masks

src/anndata/experimental/backed/_lazy_arrays.py
katosh: This is needed for the `__eq__` method below. I added it after a CI failure to allow string comparisons.

The merge code doese: https://github.com/scverse/anndata/blob/aa31c44a5e27b5d51b00cddfc0d889d24fb44031/src/anndata/_core/merge.py#L1222-L1225

**CI Failure:** [Run 20816138036](https://github.com/scverse/anndata/actions/runs/20816138036/job/59792169799) failed with exactly this error before `name` was added.

src/anndata/experimental/backed/_lazy_arrays.py
katosh: When we define `__eq__` in a subclass, Python automatically sets `__hash__ = None`, making the subclass unhashable. Since `pd.CategoricalDtype` is hashable by default, we should restore this capability to maintain the contract.

```python
# pd.CategoricalDtype IS hashable:
>>> hash(pd.CategoricalDtype(categories=['a', 'b']))
1532899084736511412

# But subclassing with __eq__ breaks this:
class UnhashableDtype(pd.CategoricalDtype):
    def __eq__(self, other): ...  # Python sets __hash__ = None

>>> hash(UnhashableDtype())
TypeError: unhashable type: 'UnhashableDtype'
```

**Operations that require `__hash__`:**

```python
# 1. COLLECTING UNIQUE DTYPES
unique_dtypes = set()
for col in lazy_adata.obs.columns:
    unique_dtypes.add(lazy_adata.obs[col].dtype)  # Requires __hash__!

# 2. DTYPE -> COLUMN MAPPING (grouping by dtype)
dtype_to_cols = {}
for col in lazy_adata.obs.columns:
    dtype = lazy_adata.obs[col].dtype
    if dtype not in dtype_to_cols:  # Uses __hash__ for dict key lookup!
        dtype_to_cols[dtype] = []
    dtype_to_cols[dtype].append(col)

# 3. @lru_cache DECORATED FUNCTIONS
@lru_cache(maxsize=128)
def process_dtype(dtype):
    return f"Processed: {dtype}"
process_dtype(lazy_dtype)  # Requires __hash__!
```

- Maintains compatibility with `pd.CategoricalDtype` contract
- User code may legitimately use dtypes in sets/dicts
- Without it, `LazyCategoricalDtype` would be less capable than its parent class

src/anndata/experimental/backed/_lazy_arrays.py
katosh: It's true, `None` is NEVER passed in normal anndata usage. However, this mirrors `pd.CategoricalDtype()` which accepts `None` and allows to instantiate the dtype even if no array is present to back it. It is also low cost. Happy to remove it if you prefer though!

katosh:
Thanks for the thorough review! I've implemented most of your suggestions:

**Implemented:**
- `@cached_property` for `categories` - cleaner than manual caching
- `"categories" in self.__dict__` for cache detection
- Removed `_cached_n_categories` double caching - now uses `shape[0]` directly
- Renamed `_categories_array` → `_categories_elem`
- Extracted `_read_partial_categories` helper to deduplicate head/tail logic
- Added `ZarrGroup | H5Group` to type annotation (you were right - if the code handles it, types should reflect that)

**Kept for now with justification:**
- `read_elem_partial` instead of direct slicing - **required for HDF5 byte-to-string decoding** (direct slicing returns `b'Cat_000'` instead of `'Cat_000'`)
- `None` support in type annotation - kept for API completeness/defensive programming, though I confirmed it's never used in practice (even empty categoricals write an empty array, not None)
- `name` property - **essential** for `dtype == "category"` comparison in merge.py (CI failed without it)
- `__hash__` method - **required** for sets/dicts (e.g., collecting unique dtypes, `@lru_cache` functions)

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,192 @@
...
+
+    def __init__(
+        self,
+        categories_elem: ZarrArray | H5Array | ZarrGroup | H5Group | None = None,
+        *,
+        ordered: bool = False,
+    ):
+        # Can be None for edge cases (empty dtype). See test_lazy_categorical_dtype_empty_array.
+        self._categories_elem = categories_elem
+        self._ordered_flag = bool(ordered)
```
ilan-gold, line 76:
```suggestion
        self._ordered_flag = ordered
```
Or?

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,192 @@
...
+            return np.array([])
+
+        # If already fully loaded, slice from cache
+        if "categories" in self.__dict__:
+            return np.asarray(self.categories[:n])
+
+        total = self.n_categories
+        return self._read_partial_categories(0, min(n, total))
+
+    def tail_categories(
```
ilan-gold, line 160: This is still almost idential to `head_categories` - could you refactor the two to share one method parametrized by the arguments to `_read_partial_categories` and `self.categories` (which appears to be the only difference)

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: I think there is no code overriding the existing `name` on `CategoricalDtype` from which we inherit.  I assume [these lines work](https://github.com/scverse/anndata/pull/2288/changes/7edb51062bcd0bd71d77210242337d35fa3c678f#diff-3ad9b1c407ca4160c3ee853931560a3e11e81081e644c314ebb548631981a002R223-R225) whether or not you have this `property` here or not because `self.name` should still be defined.

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: I think I'd rather that as well as removing the `None` argument.

tests/lazy/test_read.py
```diff
@@ -234,6 +247,349 @@ def test_chunks_df(
...
+        X=np.zeros((3, 2)),
+        obs=pd.DataFrame({"cat": pd.Categorical(categories)}),
+    )
+
+    path = tmp_path / "test.zarr"
+    adata.write_zarr(path)
+
+    lazy = read_lazy(path)
+    dtype = lazy.obs["cat"].dtype
+    assert isinstance(dtype, LazyCategoricalDtype)
```
ilan-gold, line 491: I don't think you need to go through `AnnData` for doing most of these tests, `anndata.io.write_elem` can handle writing a categorical and `read_elem` can return the in-memory once while `read_elem_lazy` will give you a `CategoricalArray` (although I think one test that embeds this inside in the anndata object enough and then tests that `read_lazy(path).to_memory() == in_memory_adata` is good).  Then you could reuse the categorical fixture you create :)

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,192 @@
...
+        if self._categories_elem is None:
+            return None
+        arr = self._get_categories_array()
+        if isinstance(arr, ZarrArray):
+            values = arr[...]
+        else:
+            from anndata.io import read_elem
+
+            values = read_elem(self._categories_elem)
+        return pd.Index(values)
```
ilan-gold:
```suggestion
        return pd.Index(read_elem(self._categories_elem))
```

ilan-gold (changes requested):

src/anndata/experimental/backed/_lazy_arrays.py
katosh: Yeah, that is strange. However, we might need `bool()` because HDF5 returns `np.bool_` instead of Python `bool`. While they're functionally equivalent in most contexts, normalizing ensures consistent behavior for hashing and serialization.
```python
import h5py, tempfile
with tempfile.TemporaryDirectory() as tmpdir:
    with h5py.File(f'{tmpdir}/test.h5', 'w') as f:
        f.create_group('cat').attrs['ordered'] = True
    with h5py.File(f'{tmpdir}/test.h5', 'r') as f:
        ordered = f['cat'].attrs['ordered']
        print(f'type: {type(ordered)}')
type: <class 'numpy.bool'>
```

katosh:
Thanks for the thorough second review! I've addressed most of your suggestions. Here's a summary:

## Implemented

- **Removed `name` property** - You were right, it's inherited from `CategoricalDtype` as a class attribute. I had mistakenly thought it might be reset like `__hash__` when defining `__eq__`, but that's not the case.

- **Removed `None` support** - Removed from type annotations and all associated guards. The `__eq__` method now returns `False` when comparing to a `CategoricalDtype` with `None` categories.

- **Simplified `categories` property** - Now just `return pd.Index(read_elem(self._categories_elem))`. You were right that `read_elem` handles both zarr and h5 uniformly.

- **Refactored tests to use `write_elem`/`read_elem_lazy`** - Most unit tests now work at the element level. Added a `_write_categorical_zarr()` helper for creating test fixtures.

## Implemented slightly differently

- **`head_categories`/`tail_categories` refactor** - I am not entirely sure what you mean. I refactored both to use a single `_get_categories_slice` method with a `from_end` to fork between the two cases internally while keeping the public API unchanged. Let me know if you'd prefer a different approach!

- **Integration test** - I avoided the round trip throu Anndata in most test and only mad a single `test_lazy_categorical_roundtrip_via_anndata` integration test which tests the full workflow including `read_lazy(path).to_memory() == original_adata`. It also verifies dtype caching and ordered categoricals through the AnnData path.

## Not yet implemented

- **`bool(ordered)` removal** - I kept `bool(ordered)` because HDF5 returns `np.bool_` instead of Python `bool`:

  ```python
  >>> with h5py.File('test.h5', 'r') as f:
  ...     ordered = f['cat'].attrs['ordered']
  ...     print(type(ordered))
  <class 'numpy.bool'>
  ```

  While `np.bool_` works in most contexts, normalizing to Python `bool` ensures consistent behavior for hashing and serialization. That said, if you'd prefer to remove it and handle `np.bool_` downstream or trust it works fine, I'm happy to change it!

Let me know if you'd like any adjustments to the approach.

src/anndata/experimental/backed/_lazy_arrays.py
ilan-gold: ok!

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,177 @@
...
+                and self._ordered_flag == other._ordered_flag
+            )
+        if not isinstance(other, pd.CategoricalDtype):
+            return False
+        # Compare with regular CategoricalDtype - need to load categories
+        if self.ordered != other.ordered:
+            return False
+        if other.categories is None:
+            return False  # LazyCategoricalDtype always has categories
+        return self.categories.equals(other.categories)
```
ilan-gold: Considering how much more extensive the base implementation, I think we should just get our specialized checks out of the way fast and then fall back to that https://github.com/pandas-dev/pandas/blob/v2.3.3/pandas/core/dtypes/dtypes.py#L401

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +36,177 @@
...
+        np.ndarray or ExtensionArray
+            The last n categories.
+        """
+        return self._get_categories_slice(n, from_end=True)
+
+    def __repr__(self) -> str:
+        if "categories" in self.__dict__:
+            # Fully loaded - use standard repr
+            return f"CategoricalDtype(categories={self.categories!r}, ordered={self.ordered})"
+        return f"LazyCategoricalDtype(n_categories={self.n_categories}, ordered={self.ordered})"
```
ilan-gold: Maybe the repr should always show categories, but just the first `n` for some nice-seeming `n`?

tests/lazy/test_read.py
```diff
@@ -234,6 +247,321 @@ def test_chunks_df(
...
+
+def test_lazy_categorical_dtype_equality(tmp_path: Path):
+    """Test LazyCategoricalDtype equality comparisons."""
+    from anndata.experimental.backed._lazy_arrays import LazyCategoricalDtype
+
+    categories = ["a", "b", "c"]
+    cat = pd.Categorical(categories)
+
+    cat_group = _write_categorical_zarr(tmp_path, cat)
+    lazy_cat = read_elem_lazy(cat_group)
```
ilan-gold: This is closer to what I want but I think the paradigm should be "write once, read many" i.e., write the fixture once (`scope="session"`) and then have a fixture that does `read_elem_lazy` every time.  You probably need one or two different fixtures (maybe for `ordered` and not and one or two other things) but I don't think every test (as appears here) needs its own special underlying categories array written to disc. You even have a few "fixtures" (the `pd.Categorical` at the beginning of each test that I would like to become a proper `pytest.fixture`) that are completely identical.

ilan-gold (changes requested):

ilan-gold:
Every review, fewer comments, getting there, thanks :)

katosh:
Thanks for the review! I am glad to get this polished. Addressed all three points:

**1. `__eq__` simplification** Now defers to `super().__eq__()` for pandas edge cases. Added two fast paths to avoid loading categories:
- Same Python object (`is` check)
- **New:** Same on-disk location check discovered that zarr/h5py arrays already compare equal by location (not content), so we just use `arr1 == arr2` directly. This avoids loading categories when comparing dtypes from the same file opened multiple times.

**2. `__repr__` always shows categories** Truncated for large counts:
```
LazyCategoricalDtype(categories=['a', 'b', 'c'])
LazyCategoricalDtype(categories=['cat_0', 'cat_1', 'cat_2', '...', 'cat_97', 'cat_98', 'cat_99'], n=100)
```
Moved constant to module level as `_N_CATEGORIES_REPR_SHOW`.

**3. Test fixtures** Refactored to session-scoped "write once, read many" pattern with 5 reusable fixtures.

---

**Edit:** Additional testing improvements after further review:

**Verified `arr1 == arr2` location-based comparison behavior:**
- Investigated h5py and zarr source code to confirm equality is location-based, not content-based
- h5py: compares HDF5 object IDs via `self.id == other.id` ([source](https://github.com/h5py/h5py/blob/master/h5py/_hl/base.py))
- zarr 3.x: uses dataclass-generated `__eq__` comparing `StorePath` (URL string comparison)
- Both return `True` for same location (even from different `open()` calls), `False` for different files with same content

**Parametrized all LazyCategoricalDtype tests for both backends:**
- Refactored fixtures with helper functions for writing categorical data to zarr/h5ad
- Created session-scoped path fixtures for each category type and backend
- Created parametrized store fixtures that automatically test both zarr and h5ad

katosh:
@ilan-gold if you like, I could also address #2296 in this PR by setting a default chunk size of 10,000 for category arrays at https://github.com/scverse/anndata/blob/c6f6f54ca10775cb684a928b8fe34ba0b6843834/src/anndata/_io/specs/methods.py#L1107-L1112

by implementing

```python
categories = v.categories.to_numpy()
cat_kwargs = dataset_kwargs
if len(categories) > 10_000 and "chunks" not in dataset_kwargs:
    cat_kwargs = dict(dataset_kwargs, chunks=(10_000,))
_writer.write_elem(g, "categories", categories, dataset_kwargs=cat_kwargs)
```
This would increase the benefit of this PR for zarr stores.

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -20,14 +20,17 @@
     ZarrArray,
 )
 
+# Number of categories to show at head/tail in LazyCategoricalDtype repr
+_N_CATEGORIES_REPR_SHOW = 3
```
ilan-gold: I don't even think `pandas` is this aggressive - It seems they use `10` so let's go with that

src/anndata/experimental/backed/_lazy_arrays.py
```diff
@@ -36,6 +39,186 @@
...
+    def __eq__(self, other) -> bool:
+        if isinstance(other, LazyCategoricalDtype):
+            if self._ordered_flag != other._ordered_flag:
+                return False
+            # Fast path: same Python object
+            if self._categories_elem is other._categories_elem:
+                return True
+            # Fast path: zarr/h5py arrays compare equal by location
+            if self._get_categories_array() == other._get_categories_array():
+                return True
```
ilan-gold:
```suggestion
           has_same_ordering = self._ordered_flag == other._ordered_flag
           are_arrays_equal = (self._categories_elem is other._categories_elem) or (self._get_categories_array() == other._get_categories_array())
           return has_same_ordering and are_arrays_equal
```

ilan-gold (commented):
I will need to look at the tests in a bit but in general they are still a little too repetitive.  What is the difference between small medium large and 50? Why not parametrize by `ordered` and `n_obs`?

src/anndata/experimental/backed/_lazy_arrays.py
katosh: Done. Note that this will add up to a total of 20 previewed categories.

katosh:
Hi @ilan-gold,

Thanks for the continued review feedback! I've addressed your comments about the test fixtures being too repetitive.

### Test fixture consolidation (`ac1cab52`)

Refactored the categorical fixtures from 15 individual fixtures to a config-driven factory pattern:

```python
_CAT_CONFIGS = [
    ("n3", 3, False, ["a", "b", "c"]),      # basic tests, equality, hashing
    ("n100", 100, False, None),              # truncation, n_categories, head/tail
    ("ordered", 3, True, ["low", "medium", "high"]),
]
```

This follows the "write once, read many" pattern you suggested - data is written once per session via `cat_data_paths`, then `_make_cat_fixture()` generates store fixtures that open fresh handles for each test.

I also consolidated `n50` and `n100` into just `n100` since it serves both the head/tail testing and truncation testing use cases.

### Improved equality_no_load test (`edb04fc2`)

Switched from patching `__getitem__` to patching `read_elem`:
- `__getitem__` on zarr/h5py arrays can't be reliably patched (C-level methods)
- `read_elem` is the actual function called to load categories

Also added a positive control within the same test that verifies comparison with `pd.CategoricalDtype` *does* trigger `read_elem`, proving the patch approach works.

### Note on force pushes

I made a few force pushes while iterating on the test improvements - apologies for the noise. The history should be clean now with just the two commits above on top of the previous work.

Let me know if there's anything else you'd like me to address!

Issue #1869, opened 2025-02-21
Accessors for AnnData


srivarra:
### Please describe your wishes and possible alternatives to achieve the desired result.

Hi,

I would like to propose adding built-in support for custom accessors in AnnData. This feature would allow users to register and use dedicated namespaces on AnnData objects without the need for subclassing or mixins, similar to the mechanisms available in [Polars][polars-accessors] and [Xarray][xarray-accessors].

## Motivation
Custom accessors provide a clean and modular way to extend the functionality of a core class. For example, in both Polars and Xarray, users can define custom functionality and then access it through a dedicated namespace (e.g., `df.my_accessor.my_method())`. 

This enables:
- **Enhanced Modularity**: Custom functionality can be isolated in separate modules.
- **Ease of Use**: Users extend AnnData objects with additional methods without modifying or subclassing the core class.
- **Extensibility**: Third-party developers can integrate domain-specific functionality that naturally augments the core AnnData behavior.

## Example

```python
import anndata as ad


@ad.extensions.register_anndata_namespace("transforms")
class Transforms:
    def __init__(self, adata: ad.AnnData):
        self._adata = adata

    def arcsinh(
        self, shift: float, scale: float, layer: str, inplace: bool = False
    ) -> ad.AnnData | None:
        """Apply an arcsinh transformation to the X matrix."""
        import numpy as np

        # Example transformation logic
        transformed = np.arcsinh(self._adata.X / scale) + shift
        return None if inplace else self._adata

```

and you'd use it like so:

```python
adata = ad.AnnData(X=...)
adata.transforms.arcsinh(shift=1, scale=1, layer='arcsinh', inplace=True)
```

## Existing examples out in the scverse

Similar accessor functionality has been implemented in other scverse projects:

- [Spatialdata-Plot][spatialdata-plot-accessors] implements accessors using Xarray’s approach, providing an nice to use API for plotting functionality.

- [holoviz-topics/hv-anndata][holoviz-hv-anndata] uses holoviz's interface solution to extend AnnData.

[polars-accessors]: https://docs.pola.rs/api/python/stable/reference/api.html
[xarray-accessors]: https://docs.xarray.dev/en/latest/internals/extending-xarray.html
[spatialdata-plot-accessors]: https://github.com/scverse/spatialdata-plot/blob/main/src/spatialdata_plot/_accessor.py
[holoviz-hv-anndata]: https://github.com/holoviz-topics/hv-anndata

flying-sheep:
[...]

PR #2290, opened 2026-01-08
feat: accessors

### LINKED ISSUE #2189: Reference paths / accessors

This has been attempted in https://github.com/scverse/anndata/pull/342 already, and would make a lot of things easier and APIs more generic by default:
- Mainly plotting: https://github.com/scverse/scanpy/issues/1500
- But also statistical APIs: https://github.com/scverse/scanpy/issues/1500

I prototyped this functionality with a real-world usecase (reimplementing all Scanpy plotting) here: https://github.com/holoviz-topics/hv-anndata/blob/971789b0e1902ac14ea24b4efcabc894e6d838ec/src/hv_anndata/accessors.py

What we should do differently from the prototype is the holoviews dependency: instead of the `ACCESSOR` constant, anndata should export a factory that takes an `AdPath` subclass to implement something like hv-anndata’s current [Dimension](https://holoviews.org/reference_manual/holoviews.core.dimension.html) AdPath:

```py
import holoviews as hv
from anndata.access import AdAc, AdPath

class HvAdPath(AdPath, Dimension):
    def __init__(self, *, __repr, __apply, __axes, __label, **params: object) -> None: ... 
        AdPath.__init__(self, __repr=__repr, __apply=__apply, __axes=__axes, __label=__label)
        label_kw = {} if __label is None else dict(label=__label)
        Dimension.__init__(self, __repr, **params, **label_kw)

A: AdAc[HvAdPath] = AdAc(HvAdPath)
```

flying-sheep:
- [x] Closes #2189 
- [x] Tests added

# User API design

1. create `AdPath` instances by descending into a central accessor constant `A`: https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/tests/test_accessors.py#L25-L42
2. Inspect `AdPath` instance, e.g. to figure out which axes the resulting vector spans: https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/tests/test_accessors.py#L101-L107
3. Call `AdPath` instance to extract a vector (see 1. for examples)

# subclassing

… is a direct goal of this, as people should be able to use `AdPath` subclasses. This means that

1. the `AdPath` API is minimal and inspection can be done through `.acc` (could be even more minimal by just putting it all into a container?): https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/src/anndata/acc/__init__.py#L56-L57 https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/src/anndata/acc/__init__.py#L63-L64
4. It’s trivial to create a new `AdAcc` constant that produces your own `AdPath` subclass: https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/src/anndata/acc/__init__.py#L415
5. The data flow is easy to understand:
   1. User uses `VecAcc.__getitem__` to get an `AdPath` or a list of them. In that process,
      1. `__getitem__` calls `process_idx` which verifies and simplifies the index
      2. `axes`, `__repr__`, and `idx_repr` get called too to validate things work
   2. The public API of `AdPath` can be used, which basically just delegates to `VecAcc`’s `axes`, `__repr__`, `idx_repr`, and `__call__`

So in the end everything except for `__call__` is validated, i.e. `VecAcc.__getitem__` raises exceptions on misuse

https://github.com/scverse/anndata/blob/61fc92068b51c06dd0d240026493c9bedb6bc9c5/src/anndata/acc/__init__.py#L101-L119

# TODO:
- [x] serialization
- [x] docs (especially data flow)
  - nomenclature: “vector” vs “array”, “accessor”, “vector accessor”, “reference”, “path”, … get it straight!
    Especially since we already call these accessors! #1869 
- [x] Array types: fix typing and test with all array types we support
- [x] improve `__contains__` so it doesn’t try to access the array
- [x] add JSON schema
  - [x] also add it to docs (linked from [`from_json`](https://icb-anndata--2290.com.readthedocs.build/en/2290/generated/anndata.acc.AdAcc.html#anndata.acc.AdAcc.from_json)/`to_json`)
  - [x] and test it
- [x] terminology: axis → dim
- [ ] (maybe) runtime autocompletion (`__dir__`)


src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,605 @@
...
+        return self.__repr
+
+    def __call__(self, adata: AnnData) -> Array:
+        """Retrieve referenced array from AnnData."""
+        return self.acc.get(adata, self.idx)
+
+
+@dataclass(frozen=True)
+class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
+    """Abstract base class for reference accessors.
```
ilan-gold: I think a terminology page where you explain these concepts (either centralized on `acc` or within each class) would be good explaining full name i.e., "reference" for `Ref` and its meanings (and hten how they are composed)

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,605 @@
...
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "layers", LayerMapAcc(ref_class=self.ref_class))
+        object.__setattr__(self, "obs", MetaAcc("obs", ref_class=self.ref_class))
+        object.__setattr__(self, "var", MetaAcc("var", ref_class=self.ref_class))
+        object.__setattr__(self, "obsm", MultiMapAcc("obs", ref_class=self.ref_class))
+        object.__setattr__(self, "varm", MultiMapAcc("var", ref_class=self.ref_class))
+        object.__setattr__(self, "obsp", GraphMapAcc("obs", ref_class=self.ref_class))
+        object.__setattr__(self, "varp", GraphMapAcc("var", ref_class=self.ref_class))
+
+    def to_json(self, ref: R) -> list[str | int | None]:
```
ilan-gold, line 643: Why have a `to_json` method? Genuinely wondering about the use-case here.

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,605 @@
...
+@dataclass(frozen=True)
+class MetaAcc[R: AdRef[str | None]](RefAcc[R, str | None]):
+    r"""Reference accessor for arrays from metadata containers (`A.`\ :attr:`~AdAcc.obs`/`A.`\ :attr:`~AdAcc.var`).
+
+    Examples
+    --------
+    You can refer to columns or the index in the way you would access them
+    in a :class:`~pandas.DataFrame`:
+
+    >>> from anndata.acc import A
```
ilan-gold: I would add a line
```suggestion
    >>> from anndata.acc import A, MetaAcc
    >>> isinstance(A.obs, MetaAcc) 
```
I think it's implicit in the description, but I would find this helpful (in all of the classes)

ilan-gold (commented):
Mentioned on slack, but I think the landing page needs to make clear `A.obsm["pca"](adata)` is possible!

I have to run now but I have a broader question - is there a desire to do some form of validation as well? Like `A.obs["x"].validate(adata)` or similar.  And relatedly asserting types might be in-scope like `A.obsm["X_pca"].is(np.ndarray).validate(adata)` or similar

Super nice! Very exciting

src/anndata/_core/anndata.py
```diff
@@ -1840,9 +1841,13 @@ def _check_uniqueness(self) -> None:
         if self.var.index[~self.var.index.isna()].has_duplicates:
             utils.warn_names_duplicates("var")
 
-    def __contains__(self, key: Any) -> NoReturn:
-        msg = "AnnData has no attribute __contains__, don’t check `in adata`."
-        raise AttributeError(msg)
+    def __contains__(self, key: AdRef) -> bool:
+        """Check if array is in AnnData."""
+        try:
+            key(self)
```
ilan-gold: Ok so to my above point, a `validate` method would be different that just using this method.  For example, if `self` is backed and a view, doing `key(self)` could read data into memory as far as I understand (the same problem appears for other views as well i.e., doing the subetting operation, it just has lesser performance impacts).  This seems bad!

docs/accessors.rst
```diff
@@ -0,0 +1,122 @@
...
+    {'var'}
+    >>> pc0.acc.k
+    'pca'
+
+#.  convenient
+
+    Want to reference multiple vectors from the same object?
+    Pass a list of indices to the vector accessor:
+
+    >>> A.obsp["connectivities"][:, ["cell0", "cell1"]]
```
ilan-gold, line 64: I'm interested in loosening the index constraint at least to taking `pandas.Index` objects (I think `Idx2DList` disallows this if I'm not mistaken).  Or do you think people really shouldn't be doing this?

src/anndata/acc/__init__.py
flying-sheep: … you asked for serialization! This creates an easily serializable object.

Did I misunderstand what you meant with serialization support?

src/anndata/acc/__init__.py
flying-sheep: Hmm, then it’s not clear enough. I was going for a few introductory paragraphs that explain them (maybe add some **bold** text), but apparently that’s not clear enough

src/anndata/acc/__init__.py
flying-sheep: yeah! I’m already doing that one in AdRef’s docs but will add more.

src/anndata/acc/__init__.py
ilan-gold: I was thinking more "schema-validateable in other contexts/languages" and less "can we serialize."  I think I had mentioned this might be too big of a lift for this PR

docs/accessors.rst
flying-sheep: I think that makes sense. I didn’t want to allow tuples intentionally to avoid ambiguities here, but string indices specifically makes sense to me!

src/anndata/_core/anndata.py
flying-sheep: I think a better option would just be implementing `__contains__` better.

No reason to have `__contains__` and `validate` if one is just a worse version of the other

src/anndata/acc/__init__.py
flying-sheep: well, this would define the “reference path” part of the schema, no?

docs/accessors.rst
ilan-gold: Right, tuples definitely seems bad, agreed.

src/anndata/acc/__init__.py
ilan-gold: > well, this would define the “reference path” part of the schema, no?

Ok! Is it worth schema-tizing this? https://json-schema.org/understanding-json-schema/reference/array#tupleValidation

flying-sheep: totally! we can
1. add the schema and test that it works with the path examples
2. simply publish the schema on readthedocs

ilan-gold: Yes!

flying-sheep:
@ilan-gold I didn’t update the docs since your last review, but the rest should be finished now!

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+
+class AdRef[I: Hashable]:
+    """A reference to a 1D or 2D array along one or two dimensions of an AnnData object.
+
+    Examples
+    --------
+    Use :data:`A` in the same way you would an :class:`~anndata.AnnData` object if you wanted to access a 1D or 2D array along the whole dimension.
+    You can then introspect the reference:
+
+    >>> from anndata.acc import A
```
ilan-gold, line 68: It's not immediately clear from the docstring what this class refers to - I would guess (it's been a week since I looked at the PR so judging by the docstring / memory) `A.obs["x"]`

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+    >>> A.obs.index.dims
+    {'obs'}
+    >>> A.obs["x"].idx
+    'x'
+    >>> type(A.var["y"].acc)
+    <class 'anndata.acc.MetaAcc'>
```
ilan-gold:
```suggestion
    >>> type(A.var["y"].acc)
    <class 'anndata.acc.MetaAcc'>
     >>> type(A.var["y"])
     <class 'anndata.acc.AdRef'>
```

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+        """Get the referenced array from the AnnData object."""
+
+    def _maybe_flatten(self, idx: I, a: Array) -> Array:
+        if len(self.dims(idx)) != 1:
+            return a
+        if isinstance(a, DaskArray):
+            a = a.map_blocks(lambda x: self._maybe_flatten(idx, x))
+        if isinstance(a, (sp.sparray, sp.spmatrix)):
+            a = a.toarray()
+        return a.flatten()
```
ilan-gold: array-api now - don't see any entry for `flatten` or `ravel` specifically so probably [`reshape`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.reshape.html#reshape) is the way to go :)

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+                raise TypeError(msg)
+
+    def __repr__(self) -> str:
+        return f"A.layers[{self.k!r}]"
+
+    def idx_repr(self, idx: Idx2D) -> str:
+        return f"[{idx[0]!r}, {idx[1]!r}]"
+
+    def isin(self, adata: AnnData, idx: Idx2D | None = None) -> bool:
+        if adata.X is None if self.k is None else self.k not in adata.layers:
```
ilan-gold, line 250: Is this in preparation for `X` - `Layers` unification i.e., returning `False` when `k is None` and `adata.X is None`?

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+
+
+type Axes = Collection[Literal["obs", "var"]]
+type Array = _InMemoryArray
+
+type Idx2D = tuple[str | slice, slice] | tuple[slice, str | slice]
+"""Index along the full length of one or both AnnData dimensions, resulting in a 1D or 2D array.
+
+E.g. `a[:, 5]`, `a[18, :]`, or `a[:, :]`
+"""
```
ilan-gold, line 34:
```suggestion
type Idx2D = tuple[str | slice, slice] | tuple[slice, str | slice]
"""Index along the full length of one or both AnnData dimensions, resulting in a 1D or 2D array.

E.g. `a[:, "c"]`, `a["b", :]`, or `a[:, :]`
"""
```
I would think these should be strings?

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+
+    def __repr__(self) -> str:
+        return f"A.{self.dim}m[{self.k!r}]"
+
+    def idx_repr(self, i: int) -> str:
+        return f"[:, {i!r}]"
+
+    def isin(self, adata: AnnData, idx: int | None = None) -> bool:
+        m: AxisArrays = getattr(adata, f"{self.dim}m")
+        if (arr := m.get(self.k)) is None:
```
ilan-gold:
```suggestion
        if self.k not in m:
```
Genuinely from a python perspective why not?

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,715 @@
...
+                msg = f"Invalid index: {idx}"
+                raise TypeError(msg)
+
+    def __repr__(self) -> str:
+        return f"A.layers[{self.k!r}]"
+
+    def idx_repr(self, idx: Idx2D) -> str:
+        return f"[{idx[0]!r}, {idx[1]!r}]"
+
+    def isin(self, adata: AnnData, idx: Idx2D | None = None) -> bool:
```
ilan-gold, line 249: Why only `Idx2D` here? Why not allow multiple labels on the "off-axis"?

src/anndata/acc/__init__.py
flying-sheep: yes!

src/anndata/acc/__init__.py
flying-sheep: yes and no, `self.k` being None already means `X` for accessors, so this logic is necessary for things functioning right now.

But of course I designed it this way in preparation of the unification.

src/anndata/acc/__init__.py
flying-sheep: This is intended as less of a public API and more to dispatch from `AnnData.__contains__`. (Of course it’s intended to be overridden if one wants.)

src/anndata/acc/__init__.py
flying-sheep: because I use `arr` farther down, but sure, I could also change that to `m[self.k].shape` below instead

src/anndata/_core/anndata.py
```diff
@@ -1875,9 +1835,20 @@ def _check_uniqueness(self) -> None:
...
-    def __contains__(self, key: Any) -> NoReturn:
-        msg = "AnnData has no attribute __contains__, don’t check `in adata`."
-        raise AttributeError(msg)
+    def __contains__(self, key: AdRef | RefAcc | MapAcc) -> bool:
+        """Check if array is in AnnData."""
+        from ..acc import AdRef, MapAcc, RefAcc
+
+        match key:
+            case MapAcc():
+                return True  # .layers and .{obs,var}{m,p} always exist, they might just be empty
```
ilan-gold: Might be worth asking the community or some friends what they think here.  While it's true they exist, we don't show them in reprs for example so false might not be a bad option in the case of no keys

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,746 @@
...
+    You can then introspect the reference:
+
+    >>> from anndata.acc import A
+    >>> ref = A.obs["x"]
+    >>> ref.dims
+    {'obs'}
+    >>> ref.idx
+    'x'
+    >>> type(ref.acc)
+    <class 'anndata.acc.MetaAcc'>
```
ilan-gold:
```suggestion
    >>> type(ref)
    <class 'anndata.acc.AdRef'>
    >>> ref.acc
    A.obs
```

This is what I was getting at i my last comment.  This docstring is for `ref` not `ref.acc` but from the snippet here, it's not immediately clear what is the `AdRef`.

docs/accessors.rst
```diff
@@ -0,0 +1,212 @@
+.. _accessors:
+
+Accessors and paths
```
ilan-gold, line 3: I stand by my previous comment.  We need a terminology page here. What is an accessor? What is a reference? What is a reference accessor i.e., how do these terms compose?

ilan-gold: And then of course, how do the classes map on to these definitions

docs/accessors.rst
```diff
@@ -0,0 +1,212 @@
...
+    :template: class-minimal
+
+    AdRef
+
+.. _reference accessors:
+
+Reference accessors
+~~~~~~~~~~~~~~~~~~~
+
+The following :class:`!RefAcc` subclasses can be accessed using :attr:`AdRef.acc`,
```
ilan-gold:
```suggestion
The following :class:`!RefAcc` subclasses correspond to the various key-value stores that anndata exposes and can be accessed using :attr:`AdRef.acc`,
```
Something to make this concept more concrete

docs/accessors.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. module:: anndata.acc
+
+:mod:`!anndata.acc` creates references to 1D and 2D arrays in an :class:`~anndata.AnnData` object.
+You can use these to drive e.g. plotting or validation code.
+For these purposes, they are
+
+#.  easy to create:
+
+    The central :attr:`A` object allows you to create
```
ilan-gold:
```suggestion
    The central :attr:`A` object is an accessor for the whole :class:`~anndata.AnnData` object and allows you to create
```
This sort of language is what I'm getting at with https://github.com/scverse/anndata/pull/2290/changes#r2793004645

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,746 @@
...
+    """
+
+    acc: RefAcc[Self, I]
+    """The accessor containing information about this array.
+
+    See :ref:`here <reference accessors>` for all possible types this can assume.
+    """
+
+    idx: I
+    """The index specifying the array."""
```
ilan-gold:
```suggestion
    """The index that materializes the :class:`RefAcc` specifying the array path as an array i.e., a key of a `obs` or an index into an array that makes it one or two dimensional."""
```

ilan-gold: Index is a weird term considering that `x` in the above example is a column in dataframe-speak, so I think clarifying this is important.  You can also highlight that in the example, the above is just a suggestion

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,746 @@
...
+    @abc.abstractmethod
+    def get(self, adata: AnnData, idx: I, /) -> InMemoryArray:
+        """Get the referenced array from the AnnData object."""
+
+    def _maybe_flatten(self, idx: I, a: InMemoryArray) -> InMemoryArray:
+        if len(self.dims(idx)) != 1:
+            return a
+        if isinstance(a, DaskArray):
+            a = a.map_blocks(lambda x: self._maybe_flatten(idx, x))
+        if isinstance(a, sp.sparray | sp.spmatrix | ArrayView):
```
ilan-gold: Missing `cupy` judging by CI

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,746 @@
...
+    --------
+    >>> from anndata.acc import A, LayerAcc
+    >>> assert isinstance(A.layers["counts"], LayerAcc)
+    >>> A.layers["counts"]["cell-1", :]
+    A.layers['counts']['cell-1', :]
+
+    """
+
+    k: str | None
+    """Key this accessor refers to, e.g. `A.layers['counts'].k == 'counts'`."""
```
ilan-gold, line 210:
```suggestion
    """Key this accessor refers to, e.g. `A.layers['counts'].k == 'counts'`. `None` is used by the subclass :class:`AdAcc` for as a identifier for`X`"""
```

ilan-gold: Or something like this, I think it is a bit strange that `A.X` doesn't exist

docs/accessors.rst
```diff
@@ -0,0 +1,212 @@
...
+    :class:`AdRef` objects that reference arrays
+    along one or two dimensions of an :class:`~anndata.AnnData` object:
+
+    >>> from anndata.acc import A
+    >>> A[:, "gene-3"]  # reference to `adata[:, "gene-3"].X` as 1D vector
```
ilan-gold: I think `A.X` should exist if it's not too much of a break architecturally.  I'm open to why it shouldn't exist though, but I think if we're trying to mimic anndata, I could see a v2 type feature being anndata-style subsetting syntax i.e.,:

```python
gene_3_ref = A[:, "gene-3"]
gene_3_ref.var["ensembl_id"]
gene_3_ref.layers["spliced"]
```
all being valid things to do, unless you really are opposed (would like to understand why).

docs/accessors.rst
```diff
@@ -0,0 +1,212 @@
...
+
+There are three layers of extensibility:
+
+#.  subclassing :class:`RefAcc` and creating a new :class:`AdRef` instance for creating them:
+
+    ..  code-block:: python
+
+        import my_plotting_library as pl
+
+        class AdDim(RefAcc, pl.Dimension): ...
```
ilan-gold: Might be worth explaining what a `Dimension` is

src/anndata/acc/hv.py
ilan-gold, line 1: Should this file's functionality be documented?

src/anndata/acc/_parse_str.py
```diff
@@ -0,0 +1,81 @@
...
+    if not strict:
+        try:
+            parse(a, spec)
+        except ValueError:
+            return None
+
+    if spec.startswith("X["):
+        return _parse_path_2d(lambda _: a, spec)
+    if "." not in spec:
+        msg = f"Cannot parse accessor {spec!r}"
```
ilan-gold:
```suggestion
        msg = f"Cannot parse accessor {spec!r} that isn't period-separated i.e., `A.obs.x` for accessing the 1d array at column `x` of `obs`."
```

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,746 @@
...
+            case AdRef():
+                return self.acc == value.acc and self.idx == value.idx
+            case str():
+                # TODO: more elegant: access `A` through `acc`?
+                a = (
+                    A
+                    if type(self).__eq__ is AdRef.__eq__
+                    else AdAcc(ref_class=type(self))
+                )
+                return self == a.resolve(value, strict=False)
```
ilan-gold: why `strict=False`? Given that

```python
A.obs["x"] == "A.obs.x"
```
is `False`, I could see someone making that error and maybe wanting to know.

flying-sheep: Good point, I’ll just return NotImplemented instead, which will raise TypeError if no side implements it.

src/anndata/_core/anndata.py
flying-sheep: Yeah, true. I guess the other behavior might be useful.

I’ll switch to it and document that.

docs/accessors.rst
flying-sheep: A whole page? Not a section? Where should it go?

Do you mean something like https://docs.python.org/3/glossary.html

docs/accessors.rst
flying-sheep: `A.X[...]` instead of `A[...]` makes sense, I wasn’t sure about that one from the start.

I also thing that pre-slicing and 0D references should be out of scope for this PR.

src/anndata/acc/__init__.py
flying-sheep: yeah, we don’t use “path” as a term at all, so I’ll modify that.

src/anndata/acc/__init__.py
flying-sheep: I’ll just change it to `A.X` and not make AdAcc an accessor for X.

1. remove superclass and `k` override
2. add X attribute
3. change docs, tests
4. change docs for `A[…]`

docs/accessors.rst
flying-sheep: I reworked this example to be an actually functional `matplotlib` demo

src/anndata/acc/hv.py
flying-sheep: Hmm, without the AnndataInterface from `hv-anndata`, it’s not functional, so I think I’ll just remove it again.

docs/accessors.rst
ilan-gold: Sorry yes, not a whole page, glossary is great

docs/accessors.rst
ilan-gold: > I also thing that pre-slicing and 0D references should be out of scope for this PR.

Agreed, was more of a counterexample

src/anndata/acc/hv.py
flying-sheep: https://github.com/holoviz-topics/hv-anndata/issues/127

src/anndata/acc/_parse_str.py
flying-sheep: done. this should be `e.g.` though, not `i.e.` …

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+        return idx in attr
+
+    def get(
+        self, adata: AnnData, k: str | None, /
+    ) -> pd.api.extensions.ExtensionArray | XVariable:
+        match getattr(adata, self.dim), k:
+            case pd.DataFrame() as df, None:
+                return df.index.array
+            case Dataset2D() as ds, None:
+                return ds.xr_index.variable  # TODO: return cached index instead?
```
flying-sheep: @ilan-gold what do you think

ilan-gold: `to_pandas().array` or something might be a bit more consistent?

ilan-gold: @flying-sheep I think 
```suggestion
                return ds.true_xr_index.variable
```
I'm ambivalent about returning a variable or not.  The index in `Dataset2D` is special in so far as all "indexes" (or `coords` in xarray-parlance) are in-memory.  If the goal is to make plotting nice, I think maybe
```suggestion
                return ds.true_index.array
```
makes the most sense as its guaranateed to be a pandas array and consists of hte "true" labels for the dataset.

flying-sheep: Pandas *array*, not `pd.Index`? But your first suggestion is `ds.true_index.variable`, a pandas array doesn’t have `.variable` right?

there’s still this a few lines later, what do you think about that?

```py
            case Dataset2D() as ds, k:
                return ds[k].variable
```

docs/accessors.rst
```diff
@@ -0,0 +1,231 @@
+.. _accessors:
+
+Accessors and paths
+===================
+
+.. module:: anndata.acc
+
+:mod:`!anndata.acc` provides :term:`accessor`\ s that create :term:`reference`\ s
+to 1D and 2D arrays in :class:`~anndata.AnnData` objects.
```
ilan-gold:
```suggestion
to axis-aligned 1D and 2D arrays in :class:`~anndata.AnnData` objects.
```
I think this is crucial because things like `A.obsm["cell 5", :]` actually don't work. Maybe there's a better wording for this, but it's not just 1d or 2d arrays.

docs/accessors.rst
```diff
@@ -0,0 +1,231 @@
...
+    The central :attr:`A` object is an :term:`accessor` for the whole :class:`~anndata.AnnData` object,
+    and allows you to create :class:`AdRef` objects,
+    which are :term:`reference`\ s to arrays spanning one or two dimensions
+    of an :class:`~anndata.AnnData` object (without being bound to a specific object):
+
+    >>> from anndata.acc import A
+    >>> A.X[:, "gene-3"]  # reference to `adata[:, "gene-3"].X` as 1D vector
+    A.X[:, 'gene-3']
+    >>> type(A.X[:, "gene-3"])
+    <class 'anndata.acc.AdRef'>
```
ilan-gold, line 24:
```suggestion
    >>> type(A.X[:, "gene-3"])
    <class 'anndata.acc.AdRef'>
    >>> type(A.obsm["cell 5", :]) # Doesn't work because it does not span `obs` or `var`
```

docs/accessors.rst
```diff
@@ -0,0 +1,231 @@
...
+#.  extensible: see `extending accessors`_.
+
+API & Glossary
+--------------
+
+The central :term:`accessor` is :data:`!A`:
+
+.. autodata:: A
+
+See :class:`AdAcc` for examples of how to use it to create :term:`reference`\ s (:attr:`AdRef`\ s).
```
ilan-gold:
```suggestion
See :class:`AdAcc` for examples of how to use it to create :term:`reference`\ s (i.e., :attr:`AdRef`\ s).
```

docs/accessors.rst
```diff
@@ -0,0 +1,231 @@
...
+                - `A.X[:, :]`, `A.layers["c"][:, "g0"]`
+            -   - :class:`MetaAcc`
+                - :attr:`MetaAcc.dim`
+                - `A.obs["a"]`, `A.var["b"]`
+            -   - :class:`MultiAcc`
+                - :attr:`MultiAcc.dim`, :attr:`MultiAcc.k`
+                - `A.obsm["d"][:, 2]`
+            -   - :class:`GraphAcc`
+                - :attr:`GraphAcc.dim`, :attr:`GraphAcc.k`
+                - `A.obsp["e"][:, "c1"]`, `A.vbsp["e"]["g0", :]`
```
ilan-gold, line 136: These are examples of AdRef i.e., `isinstance(A.obsm["d"][:, 2], AdRef)`. I would think the examples of `reference accessor` header would be a reference accessor?

docs/accessors.rst
```diff
@@ -0,0 +1,231 @@
...
+    AdRef
+
+..  glossary::
+
+    reference
+        An instance of :class:`AdRef`.
+        References a 1D or 2D array in :class:`~anndata.AnnData` objects.
+        It is independent of individual objects and can be inspected,
+        checked for equality, used as mapping keys, or applied to concrete objects,
+        e.g. via `ref in adata` or `adata[ref]`.
```
ilan-gold, line 92:
```suggestion
        e.g. via `ref in adata` or `adata[ref]`.
        An example of this would be `A.obsm["d"][:, 2]` but not `A.obsm["d"]`, which is a :term:`reference accessor`.
```

docs/accessors.rst
ilan-gold: Or maybe make this "Example Usage" and then give a concrete example of a reference accessor in the description  i.e., `A.obs` or so

src/anndata/acc/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+
+    def dims(self, idx: Idx2D, /) -> set[Literal["obs", "var"]]:
+        for dim_idx in idx:
+            if isinstance(dim_idx, str):
+                continue
+            if isinstance(dim_idx, slice) and dim_idx == slice(None):
+                continue
+            msg = f"Unsupported dimension index {dim_idx!r} in index {idx!r} (not `:` or a string)"
+            raise ValueError(msg)
+        match idx:
```
ilan-gold, line 232: Not sure what you want to do about this, but figured I'd note the current state doesn't support the shortened syntax `A.X["cell 0"]` (unlike `adata[a_1d_index]` which works IIRC).  Not sure if this is by-design or not. I don't see this covered in `test_invalid`

src/anndata/acc/_parse_str.py
```diff
@@ -0,0 +1,84 @@
...
+            parse(a, spec)
+        except ValueError:
+            return None
+
+    if spec.startswith("X["):
+        return _parse_path_2d(lambda _: a.X, spec)
+    if "." not in spec:
+        msg = (
+            f"Cannot parse accessor {spec!r} that isn’t period-separated, "
+            "e.g. `A.obs.x` for accessing the 1d array at column `x` of `obs`."
```
ilan-gold:
```suggestion
            "e.g. `obs.x` for accessing the 1d array at column `x` of `obs`."
```
`A.` is not supported!

docs/accessors.rst
flying-sheep: can’t do things that don’t work in doctests.

ilan-gold: I wasn't sure about this.  I do think this is worth documenting somewhere though, doesn't have to be here.

ilan-gold (approved):

patcon:
Yay! I thiiiiink this is cool and exciting, but it's written at a deeper level of abstraction than I can make easy sense of, and that makes me unsure.. :)

is there a place where I can read what the user level interface will be for the thing this is working toward? I'd love to provide comment on the user level API 🙏🏻

Maybe to cut to the chase: how do you envision this will allow someone to color a `sc.pl.embedding()` by the third principle component? Or draw 4 plots colored by Leiden, and the first 3 principal components?

thanks! Also pls feel welcome to direct me elsewhere

flying-sheep:
Don’t worry! While there are quite some possible usage examples strewn throughout this PR, of course the anndata-plot/hv-anndata/scanpy APIs will be documented with more focused examples.

There is currently a prototype for the new scanpy plotting [here](https://hv-anndata.readthedocs.io/en/latest/api/hv_anndata.plotting.scanpy.html), but beware as it doesn’t use the final accessor API as defined in this PR.

So to answer your questions:

> color a `sc.pl.embedding()` by the third principle component

probably one of these

```py
import scanpy as sc
from hv_anndata import A  # maybe this will also live on `sc`, idk yet

sc.pl.scatter(adata, A.obsm["umap"], color=A.obsm["pca"][:, 2]).opts(cmap="tab10")
# or more explicitly
sc.pl.scatter(adata, A.obsm["umap"][:, [0, 1]], color=A.obsm["pca"][:, 2]).opts(cmap="tab10")
```

> 4 plots colored by Leiden, and the first 3 principal components

probably just replace the `color` argument above with

```py
color=[A.obs["leiden"], *A.obsm["pca"][:, [0, 1, 2]]]
```

ilan-gold:
@patcon Check out our latest docs: https://anndata.readthedocs.io/en/latest/accessors.html which point at the `main` branch

patcon:
Nice! Thanks! I'll probably add my own syntactic sugar, but this will make it really simple :)

```py
sc.pl.embedding(adata, basis="umap", color="X_pca[2]")

sc.pl.emvedding(adata, basis="umap", color=["leiden", "X_pca[0, 1, 2]"])
```

I have found it quite easy to wrap scanpy and have the scripts be quite legible to new coders at a weekly event I host, especially with some visuals of the data structure. The above was what I felt seemed most legible -- but I can always do a thin wrapper with the excellent code you've created 🙏🏻

flying-sheep:
Don’t make your own, there is already a shortcut syntax: https://anndata.readthedocs.io/en/latest/generated/anndata.acc.AdAcc.html#anndata.acc.AdAcc.resolve

If you want to extend it or think it could be better, please present improvements to the design in an issue before we release 0.13, then we can still change it.

PR #2298, opened 2026-01-16
docs: Sphinx 9 and fix member skipping

### LINKED ISSUE #14262: `typing.Protocol` support
**Is your feature request related to a problem? Please describe.**
I noticed that autodoc doesn’t include e.g. `__call__` unless I manually tell it to.

**Describe the solution you'd like**
The autodoc rendering of `Protocol`s shouldn’t be configurable as all their members are by definition public. A class inheriting directly from `Protocol` should have all its members visible in the docs.

Since protocols can also inherit from others, maybe it should be possible to not include members that are part of a superprotocol’s members.

**Describe alternatives you've considered**
On top of the above, we could do the full treatment: A `py:protocol` role, a `.. autoprotocol::` directive, a autosummary template, …

In that case of course `py:class` needs some fallback code to be able to link to `Protocol`s that used to be documented as classes.

**Additional context**
Since Python 3.13, there are [`typing.get_protocol_members`](https://docs.python.org/3/library/typing.html#typing.get_protocol_members) and [`typing.is_protocol`](https://docs.python.org/3/library/typing.html#typing.is_protocol).

Combined, it’s pretty easy to retrieve the set of protocol members:

```py
always_include_member = is_protocol(cls) and member_name in get_protocol_members(cls)
```

`sphinx-toolbox` has this as an extension, but it’s incompatible with Sphinx 9: https://sphinx-toolbox.readthedocs.io/en/latest/extensions/more_autodoc/autoprotocol.html

flying-sheep:
This makes use of the `type` support in Sphinx 9.

Merging it as-is would lose us support for the `protocol` document type, so protocols are `class`es in our intersphinx inventory and it says **class** in front of their names:

- https://anndata.readthedocs.io/en/latest/generated/anndata.experimental.Read.html
- https://icb-anndata--2298.com.readthedocs.build/en/2298/generated/anndata.experimental.Read.html

Since I like the way `autoprotocol` inlines things (which we don’t do because of our autosummary template), I added a template that we have to manually use for protocols, ABCs and small classes until this is fixed: https://github.com/sphinx-doc/sphinx/issues/14262

To fix protocols, this uses the data autosummary gives us for skipping members and updates our member skipping logic to always include not only `abstractmethod`s but also protocol members

Works great for everything …

- https://icb-anndata--2298.com.readthedocs.build/en/2298/generated/anndata.experimental.AnnLoader.html?readthedocs-diff=true&readthedocs-diff-chunk=1
- https://icb-anndata--2298.com.readthedocs.build/en/2298/generated/anndata.experimental.backed.MaskedArray.html?readthedocs-diff=true&readthedocs-diff-chunk=1

… except where we were relying on a bug, such as in `CSCDataset|CSRDataset`, where `__getitem__` and `to_memory` *are* inherited from a (private) superclass, so I circumvented that by assigning them directly to the subclasses. AND this PR fixes the missing protocol members `shape`, `dtype`, and `backend`, so that’s good!

- https://icb-anndata--2298.com.readthedocs.build/en/2298/generated/anndata.abc.CSCDataset.html?readthedocs-diff=true&readthedocs-diff-chunk=1


docs/conf.py
```diff
@@ -168,19 +169,25 @@ def res(
...
-    "types.EllipsisType": ("py:data", "types.EllipsisType"),
-    "pathlib._local.Path": "pathlib.Path",
 }
 autodoc_type_aliases = dict(
     NDArray=":data:`~numpy.typing.NDArray`",
     AxisStorable=":data:`~anndata.typing.AxisStorable`",
+    # https://github.com/python/cpython/issues/124089
+    # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/580
+    K=":class:`zarr.Array` | :class:`h5py.Dataset`",
+    S=":class:`anndata.experimental.StorageType`",
```
ilan-gold, line 191: Maybe worth linking to where these are used

ilan-gold (commented):
Is there no way to force https://github.com/scverse/anndata/pull/2298/changes#diff-b8b6e67f064e921ecb963b837406d9e056dedcb77f1d9f03cbc1d32e7867e9a6 to potetially be protocol instead of class when applicable (even at the expense of duplicating the entire rst file theoretically).  I'm wondering how `sphinx` managed to do this originally when `toolbox` existed and if we can replicate it.

And if not, maybe it's worth manually annotating protocols as such in the docstring?

Other than that, things look good!

flying-sheep:
> Is there no way to force [#2298 (changes)](https://github.com/scverse/anndata/pull/2298/changes#diff-b8b6e67f064e921ecb963b837406d9e056dedcb77f1d9f03cbc1d32e7867e9a6) to potetially be protocol instead of class when applicable (even at the expense of duplicating the entire rst file theoretically). I'm wondering how sphinx managed to do this originally when toolbox existed and if we can replicate it.

Unfortunately Sphinx 9 completely refactored autodoc, so there’s no non-legacy way to do this that isn’t a huge undertaking. I’d rather wait for Sphinx to add first-class protocol support: https://github.com/sphinx-doc/sphinx/issues/14262

PR #2299, opened 2026-01-19
fix: merge indexes when `load_annotation_index=False`

ilan-gold:
Fixes issue found in: https://github.com/scverse/annbatch/blob/627eb08d699b9cb07ab24fa67775e1e794c07245/src/annbatch/io.py#L185-L190

If you don't load the annotation index and then try to merge, you'll merge on string-ified range indexes - bad!! So for correctness the indexes need to be loaded into memory. I implemented a further fix in the `.index` (aside from the change in `merge.py`) setter because the previous behavior didn't account for the `name` change properly and would error out when setting `.index` to an existing column in `Dataset2D`.

- [ ] Closes #
- [x] Tests added


tests/test_xarray.py
```diff
@@ -73,40 +74,66 @@ def test_true_index_dim_column_subset(dataset2d, df):
...
+        subtests.test("error"),
+        pytest.raises(ValueError, match=r"Unknown variable `test`\."),
+    ):
         dataset2d.true_index_dim = "test"
 
-    dataset2d.true_index_dim = None
-    assert dataset2d.true_index_dim == dataset2d.index_dim
+    with subtests.test("unset"):
+        dataset2d.true_index_dim = None
+        assert dataset2d.true_index_dim == dataset2d.index_dim
```
flying-sheep: Any reason why this isn’t 3 tests? I mean, with subtests, it’s OK, but there’s no expensive data setup, so it could just be dedented and turned into 3.

tests/lazy/conftest.py
```diff
@@ -203,9 +203,12 @@ def stores_for_concat(
 
 @pytest.fixture
 def lazy_adatas_for_concat(
-    stores_for_concat,
+    stores_for_concat: list[AccessTrackingStore], *, load_annotation_index: bool
 ) -> list[AnnData]:
-    return [read_lazy(store) for store in stores_for_concat]
+    return [
+        read_lazy(store, load_annotation_index=load_annotation_index)
```
flying-sheep, line 209: Can you explain what this changes? Adds test cases for `load_annotation_index=not the_default`?

src/anndata/_core/xarray.py
```diff
@@ -154,12 +157,24 @@ def index(self) -> pd.Index:
     @index.setter
     def index(self, val) -> None:
```
flying-sheep: could you add a type here? what’s the else branch dealing with?

flying-sheep (approved):
I can’t say I fully understand it, but the parts I do understand look good!

tests/lazy/conftest.py
ilan-gold: Exactly!

tests/test_xarray.py
ilan-gold: I forgot that subtests have landed! Yes that makes things so much better, they felt just different enough to not require it, but similar enough that I considered parametrizing.  subtests FTW

src/anndata/_core/xarray.py
ilan-gold: Ok, it's `Any` in the underlying code TBF: https://github.com/pydata/xarray/blob/9fe835f68dc588e178ae2222fc34a936428dbc3c/xarray/core/coordinates.py#L615-L616

tests/test_xarray.py
ilan-gold: I will make this three separate tests though

src/anndata/_core/xarray.py
flying-sheep: That doesn’t mean we should!

`Any` is always bad as parameter signature. Should have been `pd.Index | XDataArray | object`, that way we actually have type checking active in that function.

PR #2306, opened 2026-01-21
chore: add warning for `zarr` format v2

ilan-gold:
- [ ] Begins migration to zarr v3 writing by default (will follow this up with a PR to remove zarr v2 the package as well)
- [x] Tests added

### DIFF
--- a/src/anndata/_io/zarr.py
+++ b/src/anndata/_io/zarr.py
@@ -148,8 +148,12 @@ def read_dataframe(group: zarr.Group | zarr.Array) -> pd.DataFrame:
 def open_write_group(
     store: StoreLike, *, mode: AccessModeLiteral = "w", **kwargs
 ) -> zarr.Group:
-    if not is_zarr_v2() and "zarr_format" not in kwargs:
-        kwargs["zarr_format"] = settings.zarr_write_format
+    if "zarr_format" not in kwargs:
+        if settings.zarr_write_format == 2 or is_zarr_v2():
+            msg = "Writing zarr v2 data will no longer be the default in the next minor release. v3 data will be written by default. If you are explicitly setting this configuration, consider migrating to zarr v3."
+            warn(msg, UserWarning)
+        if not is_zarr_v2():
+            kwargs["zarr_format"] = settings.zarr_write_format
     return zarr.open_group(store, mode=mode, **kwargs)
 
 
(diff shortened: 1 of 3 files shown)


src/anndata/_io/zarr.py
```diff
@@ -148,8 +148,12 @@ def read_dataframe(group: zarr.Group | zarr.Array) -> pd.DataFrame:
 def open_write_group(
     store: StoreLike, *, mode: AccessModeLiteral = "w", **kwargs
 ) -> zarr.Group:
-    if not is_zarr_v2() and "zarr_format" not in kwargs:
-        kwargs["zarr_format"] = settings.zarr_write_format
+    if "zarr_format" not in kwargs:
+        if settings.zarr_write_format == 2 or is_zarr_v2():
+            msg = "Writing zarr v2 data will no longer be the default in the next minor release. v3 data will be written by default. If you are explicitly setting this configuration, consider migrating to the zarr v3 file format."
+            warn(msg, UserWarning)
```
flying-sheep, line 154: Why `UserWarning`?

src/anndata/_io/zarr.py
```diff
@@ -148,8 +148,12 @@ def read_dataframe(group: zarr.Group | zarr.Array) -> pd.DataFrame:
...
+        if not is_zarr_v2():
+            kwargs["zarr_format"] = settings.zarr_write_format
```
flying-sheep, line 156: what does this change?

src/anndata/_io/zarr.py
ilan-gold: Something more appropriate? We're no deprecating anything.

src/anndata/_io/zarr.py
ilan-gold: Nothing? Could you be more specific?

flying-sheep: I’m curious why you added these two lines!

src/anndata/_io/zarr.py
flying-sheep: Good point! Another thing: So since `2` is the default setting, this warning will be shown even if they manually set it.

Would it make sense to make the default `None` which means “2 with warning”?

ilan-gold: I don't think anyone should be writing zarr v2 data anymore so I'd rather nudge them in general - OMENGFF isn't even allowing writing zarr v2 on their latest version

src/anndata/_io/zarr.py
ilan-gold: Did I add these two lines? They look the same as before

flying-sheep: oh yeah wow, no idea what I saw yesterday.

flying-sheep (approved):

PR #2307, opened 2026-01-21
fix: empty `Dataset2D` warning + columns setting implementation

ilan-gold:
Setting the columns (which occurs with `pandas_as_str` in our codebase) was unimplemented for `Dataset2D` (thus causing a `AttributeError`), a fact I discovered because I never implemented a test for setting `.obs` to an empty `Dataset2D` (i.e., no columns) and ran some tests on it. This implements that test and catches the warning.

Ideally, the warning would not be raised if that caller is `anndata` itself but I am not sure doing stack inspection is really worth it.  Happy to implement though!

- [x] Discovered in https://github.com/scverse/annbatch/actions/runs/21143427459/job/60803045742?pr=117
- [x] Tests added

### DIFF
--- a/src/anndata/_core/xarray.py
+++ b/src/anndata/_core/xarray.py
@@ -286,15 +286,13 @@ class Dataset2D:
 
     @columns.setter
     def columns(self, val) -> None:
+        if len(self.columns.symmetric_difference(val)) > 0:
+            msg = "Trying to rename the keys of the mapping with new names - please use a different API to rename the keys of the underlying dataset mapping."
+            raise ValueError(msg)
         warn(
             "Setting columns on `Dataset2D` has no effect because the underlying data structure has no apparent ordering on its keys",
             UserWarning,
         )
-        if len(self.columns.symmetric_difference(val)) > 0:
-            msg = (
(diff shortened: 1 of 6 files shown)


src/anndata/_core/xarray.py
```diff
@@ -282,6 +284,16 @@ def columns(self) -> pd.Index:
             columns.discard(index_key)
         return pd.Index(columns)
 
+    @columns.setter
+    def columns(self, val) -> None:
+        if len(self.columns.symmetric_difference(val)) > 0:
+            msg = "Trying to rename the keys of the mapping with new names - please use a different API to rename the keys of the underlying dataset mapping."
+            raise ValueError(msg)
+        warn(
+            "Setting columns on `Dataset2D` has no effect because the underlying data structure has no apparent ordering on its keys",
```
flying-sheep: Maybe reword this so it’s clearer that we interpret this action as the user trying to reorder columns?

flying-sheep (approved):

src/anndata/_core/xarray.py
ilan-gold: Good call!

Issue #2238, opened 2025-12-03
Handle new zarr dtypes (both custom and datetime)


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

At the moment, everything that is in a `numpy.ndarray` is treated identically when writing to disk (with some exceptions like `rec-array` and `string-array`): 

https://github.com/scverse/anndata/blob/main/src/anndata/_io/specs/methods.py#L414

However, zarr currently has support for totally custom data types.  At the moment, I don't think this is too much of a problem (in theory, you could put something custom or a datetime in `obsm` or `X` but I don't think this is really something to be too worried about at the moment). 

Beyond those pathological cases, I think the one place this sort of thing could in theory happen would be via `obs` with datetimes

```python
import anndata as ad
import pandas as pd

arr = pd.date_range("2018-01-01", periods=5, freq="h").to_numpy()
ts = pd.Series(arr, index=range(len(arr)))
df = pd.DataFrame({"dt": ts})
 ad.AnnData(obs=df).write_zarr("foo.zarr")
```
although this errors out with

```
IORegistryError: No method registered for writing <class 'pandas.core.arrays.datetimes.DatetimeArray'> into <class 'zarr.core.group.Group'>
Error raised while writing key 'dt' of <class 'zarr.core.group.Group'> to /obs
```

so it's possible that one can not do this directly although with `write_elem` I think it could be done, even for custom dtypes:

```python
import zarr
z = zarr.open("hooray2.zarr")
ad.io.write_elem(z, "arr", ts.values)
```
which will write out a zarr dtype of 
```json
"data_type": {
    "name": "numpy.datetime64",
    "configuration": {
      "unit": "ns",
      "scale_factor": 1
    }
  },
```

So I think this sort of thing is instructive as the goal we want - we _should_ be able to write out `obs` with datetime because it is an officially supported zarr dtype, and we should do it with confidence!

More pressingly, https://hackmd.io/@zarr/SyFKW_A6xg?utm_source=preview-mode&utm_medium=rec will come at some point so we should probably begin future proofing for arrow-in-zarr.  It will be implemented via https://zarr.readthedocs.io/en/stable/user-guide/data_types/#integral

Some action items:

- [ ] Add `rec-array` to the spec (it isn't there)
- [ ] Plan the next step after some discussion. Some possibilities:
1. Figure out a mechanism for how to handle custom zarr dtypes generically
2. Decide that we will go one by one i.e., we handle only certain ones
[...]

Issue #1708, opened 2024-10-09
IO support for pandas datetime format


le-ander:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of anndata.
- [ ] (optional) I have confirmed this bug exists on the master branch of anndata.

### Report

Code:

```python
adata.write("file.h5ad")
```

Traceback:

```pytb
---------------------------------------------------------------------------
IORegistryError                           Traceback (most recent call last)
Cell In[13], line 1
----> 1 adata.write("data[/microbiome_layered.h5ad](http://localhost:8888/microbiome_layered.h5ad)")

File [~/.pyenv/versions/3.12.3/envs/ehrapy-3.12.3/lib/python3.12/site-packages/anndata/_core/anndata.py:1863](http://localhost:8888/lab/tree/work/projects/hackathon_ehrapy/~/.pyenv/versions/3.12.3/envs/ehrapy-3.12.3/lib/python3.12/site-packages/anndata/_core/anndata.py#line=1862), in AnnData.write_h5ad(self, filename, compression, compression_opts, as_dense)
   1860 if filename is None:
   1861     filename = self.filename
-> 1863 write_h5ad(
   1864     Path(filename),
   1865     self,
   1866     compression=compression,
   1867     compression_opts=compression_opts,
   1868     as_dense=as_dense,
   1869 )
   1871 if self.isbacked:
   1872     self.file.filename = filename

File [~/.pyenv/versions/3.12.3/envs/ehrapy-3.12.3/lib/python3.12/site-packages/anndata/_io/h5ad.py:101](http://localhost:8888/lab/tree/work/projects/hackathon_ehrapy/~/.pyenv/versions/3.12.3/envs/ehrapy-3.12.3/lib/python3.12/site-packages/anndata/_io/h5ad.py#line=100), in write_h5ad(filepath, adata, as_dense, dataset_kwargs, **kwargs)
     99 elif adata.raw is not None:
    100     write_elem(f, "raw", adata.raw, dataset_kwargs=dataset_kwargs)
--> 101 write_elem(f, "obs", adata.obs, dataset_kwargs=dataset_kwargs)
    102 write_elem(f, "var", adata.var, dataset_kwargs=dataset_kwargs)
    103 write_elem(f, "obsm", dict(adata.obsm), dataset_kwargs=dataset_kwargs)

[...]

PR #2309, opened 2026-01-22
chore: remove zarr v2 the package + default to v3 format

ilan-gold:
- [x] In preparation for #2238 and #1708
- [ ] Tests added


src/anndata/_io/specs/methods.py
```diff
@@ -600,44 +592,25 @@ def write_vlen_string_array_zarr(
...
-            dtype=object,
-            object_codec=numcodecs.VLenUTF8(),
-            **dataset_kwargs,
-        )
-        f[k][:] = elem
-    else:
-        from numcodecs import VLenUTF8
-        from zarr.core.dtype import VariableLengthUTF8
+    from numcodecs import VLenUTF8
+    from zarr.core.dtype import VariableLengthUTF8
```
flying-sheep: these should go to toplevel now that they’re always there. We import zarr unconditionally elsewhere right?

src/anndata/tests/helpers.py
```diff
@@ -1143,6 +1138,15 @@ def __init__(self, *args, **kwargs):
 
         self._read_only = True
```
flying-sheep: wait, what if `bool(kwargs.get("read_only")) is True`?

tests/test_backed_sparse.py
```diff
@@ -394,18 +394,13 @@ def test_lazy_array_cache(
...
-        case False, True:
-            assert store.get_access_count("X/indptr") == 5
-        case True, False:
-            assert store.get_access_count("X/indptr") == 4
-        case False, False:
-            assert store.get_access_count("X/indptr") == 7
+    if should_cache_indptr:
+        assert store.get_access_count("X/indptr") == 4
+    else:
+        assert store.get_access_count("X/indptr") == 7
```
flying-sheep:
```suggestion
    c_expected = 4 if should_cache_indptr else 7
    assert store.get_access_count("X/indptr") == c_expected
```

flying-sheep (approved):
beautiful, love this kind of PRs!

one thing about `read_only` I didn’t understand, but that was in place before this PR.

PR #2325, opened 2026-01-30
docs: clean up typing module

flying-sheep:
This PR makes our usage of `type`s clearer in our code:
- `type Public = ...`
- `_private: TypeAlias = ...`

If I figure out in the future how to make our docs build expand non-public types reliably (theislab/scanpydoc#254), we can make them all `type`s.

It also moves all the type aliases into one page, and exports Index1D so things are actually readable: https://icb-anndata--2325.com.readthedocs.build/en/2325/typing.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,8 @@ __pycache__/
 .ipynb_checkpoints
 
 # docs
-/docs/generated/
+/docs/generated/*
+!/docs/generated/anndata.typing.Index.rst
 /docs/_build/
 
 # IDEs
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,7 @@ __pycache__/
 .ipynb_checkpoints
 
 # docs
-/docs/generated/*
-!/docs/generated/anndata.typing.Index.rst
+/docs/generated/
 /docs/_build/
 
 # IDEs
--- a/docs/api.md
+++ b/docs/api.md
@@ -253,12 +253,17 @@ Types used by the former:
    abc.CSCDataset
 ```
 
-.. these are types, not classes
+<!-- these are types, not classes, so don’t use the above template -->
 
 ```{eval-rst}
-.. autosummary::
-   :toctree: generated/
+.. toctree::
+   :hidden:
 
+   typing
+
+.. autosummary::
+
+   typing.Index1D
    typing.Index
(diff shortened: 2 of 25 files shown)


docs/api.md
```diff
@@ -253,12 +253,17 @@ Types used by the former:
...
 
-.. these are types, not classes
+<!-- these are types, not classes, so don’t use the above template -->
 
 ```{eval-rst}
+.. toctree::
+   :hidden:
+
+   typing
+
```
flying-sheep, line 263: alternative to this would be to move the contents of `docs/typing.md` into the module docstring, and do this instead, but that might mean adapting the module template.


```suggestion
.. autosummary::
   :toctree: generated/

   typing
```

docs/conf.py
```diff
@@ -145,9 +145,10 @@ def res(
         for kind in ["", "View"]
     },
     # Can’t use `set_module` for `type`s. When moving out of .experimental, define in actual location.
-    "anndata.compat.Index": "anndata.typing.Index",
     "anndata._types.StorageType": "anndata.experimental.StorageType",
     # https://github.com/theislab/scanpydoc/issues/254
+    "anndata.typing.Index1D": "anndata.typing.Index1D",
+    "anndata.typing.Index": "anndata.typing.Index",
```
ilan-gold, line 151: Both sides of this look identical, no?

.gitignore
```diff
@@ -24,7 +24,8 @@ __pycache__/
 .ipynb_checkpoints
 
 # docs
-/docs/generated/
+/docs/generated/*
+!/docs/generated/anndata.typing.Index.rst
```
ilan-gold: Intentional?

docs/api.md
ilan-gold: I think looks good here, and this feels somehow simpler

ilan-gold (approved):

docs/conf.py
flying-sheep: yeah, see the link above that block of declarations.

.gitignore
flying-sheep: ah that was a previous attempt!

flying-sheep:
backporting this would require python 3.12, so no.

PR #2327, opened 2026-02-04
chore: add `anndata.settings.copy_on_write_X`

ilan-gold:
`copy_on_write_X` will be `False` and then the setting will be removed in the next minor release (i.e., `True`).
 
- [x] Grossly simplifies https://github.com/scverse/anndata/pull/1707 because we will remove the implicit modification of `X`
- [x] Tests added

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -587,13 +587,10 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):  # noqa: PLW1641
         elif self.is_view and self._adata_ref.X is None:
             X = None
         elif self.is_view:
-            if self._X is not None:
-                X = self._X
-            else:
-                X = as_view(
-                    _subset(self._adata_ref.X, (self._oidx, self._vidx)),
-                    ElementRef(self, "X"),
-                )
+            X = as_view(
+                _subset(self._adata_ref.X, (self._oidx, self._vidx)),
+                ElementRef(self, "X"),
+            )
         else:
             X = self._X
         return X
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -625,7 +625,7 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):  # noqa: PLW1641
                 self._init_as_actual(new)
                 return None
             else:
-                msg = "Setting element `.X` of view of `AnnData` object will obey copy-on-write semantics in the next minor release."
+                msg = "Setting element `.X` of view of `AnnData` object will obey copy-on-write semantics in the next minor release. "
                 "In other words, this subset of your original `AnnData` will be copied-in-place and initialized with the value passed into this setter. "
                 "Set `anndata.settings.copy_on_write_X = True` to begin opting in to this behavior."
                 warn(msg, FutureWarning)
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -625,7 +625,9 @@ class AnnData(metaclass=utils.DeprecationMixinMeta):  # noqa: PLW1641
             if settings.copy_on_write_X:
                 msg = "Setting element `.X` of view, initializing view as actual."
                 warn(msg, ImplicitModificationWarning)
-                self._X = value
+                new = self.copy()
+                new._X = value
+                self._init_as_actual(new)
                 return None
             else:
                 msg = "Setting element `.X` of view will obey copy-on-write semantics in the next minor release. Set `anndata.settings.copy_on_write_X = True` to begin opting in to this behavior."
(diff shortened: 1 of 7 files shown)


src/anndata/_core/anndata.py
```diff
@@ -602,7 +602,31 @@ def X(self) -> _XDataType | None:
...
+            raise ValueError(msg)
+        if self._is_view:
+            if settings.copy_on_write_X:
+                msg = "Setting element `.X` of view, initializing view as actual."
+                warn(msg, ImplicitModificationWarning)
+                new = self._mutated_copy(X=value)
+                self._init_as_actual(new)
+                return None
+            else:
+                msg = "Setting element `.X` of view will obey copy-on-write semantics in the next minor release. Set `anndata.settings.copy_on_write_X = True` to begin opting in to this behavior."
```
Zethson: > will obey copy-on-write semantics in the next minor release.

I don't know whether people with less experience will understand this explanation. Could we make it clearer for them?

ilan-gold: https://github.com/scverse/anndata/pull/2327/commits/d76a96e91a030cf4a6b3653d15e5fc8fc18c9a9e

src/anndata/_core/anndata.py
```diff
@@ -624,65 +648,51 @@ def X(self, value: _XDataType | None):  # noqa: PLR0912
             oidx, vidx = np.ix_(self._oidx, self._vidx)
         else:
             oidx, vidx = self._oidx, self._vidx
-        if (
-            np.isscalar(value)
-            or (hasattr(value, "shape") and (self.shape == value.shape))
-            or (self.n_vars == 1 and self.n_obs == len(value))
-            or (self.n_obs == 1 and self.n_vars == len(value))
-        ):
```
ilan-gold, line 632: The logic below here should be identical except that I lifted this condition up into `can_set_direct_if_not_none` so its `else` is now just guard clause at the top.

ilan-gold:
Tested locally against scanpy and getting no warnings.  I also `ctrl+F`'ed the codebase and don't see anywhere this might be triggered anyway.

tests/test_x.py
```diff
@@ -50,17 +50,35 @@ def test_repeat_indices_view():
...
-    if isinstance(orig_X, np.ndarray) and sparse.issparse(to_assign):
-        # https://github.com/scverse/anndata/issues/500
-        pytest.xfail("Cannot set a dense array with a sparse array")
-    view = adata[:9, :9]
-    view.X = to_assign
-    np.testing.assert_equal(asarray(view.X), np.ones((9, 9)))
-    assert isinstance(view.X, type(orig_X))
+@pytest.mark.parametrize("copy_on_write_X", [True, False], ids=["CoW", "update"])
+def test_setter_view(orig_array_type, new_array_type, *, copy_on_write_X: bool):
+    with ad.settings.override(copy_on_write_X=copy_on_write_X):
```
flying-sheep: We don’t reset settings after tests?

If so we should start to do it. If you don’t want to do that in this PR, instead
1. If I didn’t do it yet, create a global `exit_stack: ExitStack` fixture that does nothing other than `with ExitStack() as es: yield es`
2. use it here, and just do `exit_stack.enter_context(ad.settings.override(...))`

tests/test_x.py
```diff
@@ -50,17 +50,35 @@ def test_repeat_indices_view():
...
+        to_assign = new_array_type(np.ones((9, 9)))
+        if not copy_on_write_X:
+            expected_X[:9, :9] = asarray(to_assign)
+        if (
+            not copy_on_write_X
+            and isinstance(orig_X, np.ndarray)
+            and sparse.issparse(to_assign)
+        ):
+            # https://github.com/scverse/anndata/issues/500
+            pytest.xfail("Cannot set a dense array with a sparse array")
```
flying-sheep: please never use `pytest.xfail` outside of an `except:` block. use `request.addmarker` (or so) instead.

src/anndata/_settings.py
```diff
@@ -526,5 +526,14 @@ def validate_sparse_settings(val: Any, settings: SettingsManager) -> None:
 )
 
 
+settings.register(
+    "copy_on_write_X",
+    default_value=False,
+    description="Whether to copy-on-write X. Currently `my_adata_view[subset].X = value` will write back to the original AnnData object at the `subset` location. `X` is the only element where this behavior is implemented though.",
```
flying-sheep:
```suggestion
    description=(
        "Whether to copy-on-write X. "
        "Currently `my_adata_view[subset].X = value` will write back to the original AnnData object at the `subset` location. "
        "`X` is the only element where this behavior is implemented though."
    ),
```

src/anndata/_core/anndata.py
flying-sheep: Great! With the “hide whitespace changes” view setting, I can see that you indeed changed nothing (I checked because you said “should”)

src/anndata/_core/anndata.py
```diff
@@ -602,7 +602,32 @@ def X(self) -> _XDataType | None:
         #     return X
 
     @X.setter
-    def X(self, value: _XDataType | None):  # noqa: PLR0912
+    def X(self, value: _XDataType | None):  # noqa: PLR0912, PLR0915
```
flying-sheep: Please don’t add ignores for complexity warnings.

The idea was that we add the ignores for existing code, but slowly reduce them and never add any.

Their point is so we make things less complex when refactoring, not more.

flying-sheep (changes requested):
Awesome, very few code changes necessary!

Do please reduce complexity instead of adding more ignores for complexity warnings though.

src/anndata/_core/anndata.py
ilan-gold: > Their point is so we make things less complex when refactoring, not more.

The idea is that this will only exist for about 5 minutes on the `main` branch because I'm going to immediately remove the non-CoW stuff. I'll split it up though ok i.e., backport this, release, and then remove on `main` for `0.13`

flying-sheep (approved):

PR #2328, opened 2026-02-04
feat: `DataFrame` API for custom `obs` and `var`

### LINKED ISSUE #355: Native support for cuPy/cuDF backed Anndata
This would be immensely useful for the GPU data science community as it would start to enable pipelines fully on GPU.

### LINKED ISSUE #2043: `DataFrame` API for `obs` and `var` keys

ilan-gold:
cc @srivarra @Intron7 

possible TODOs:

- [ ] `concat` and `merge` API (or disallow and keep pandas/xarray special-cased, would require `equal` implementation as well)
- [ ] adopt `__dataframe__` protocol in a meaningful way ideally for writing, but potentially as a solution to the above to cases (i.e., drop into `pandas` and leave the data there): https://data-apis.org/dataframe-protocol/latest/
- [ ] device-specific handling/indexing (see #2071  ) for handling `cuDF` or GPU-based `polars`
- [ ] Remove `_gen_dataframe` special casing
- [ ] Determine if we need a `to_numpy` method

This PR is somewhat self-testing because I will have in theory removed all references to both `pandas.Dataframe` and `XDataset`/`Dataset2D` everywhere except concatenation internally.

One downside of this approach is that it puts us somewhat on the hook for `pandas.DataFrame` API but the flipside is that if we choose to switch to another library, we could do that relatively cleanly now/more easily

- [x] closes #2043, closes #355
- [x] Tests added


src/anndata/_core/xarray.py
```diff
@@ -33,16 +39,14 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
     return wrapper
 
 
-class Dataset2D(Mapping[Hashable, XDataArray | Self]):
+class Dataset2D:
```
srivarra, line 42: > adopt `__dataframe__` protocol in a meaningful way ideally for writing, but potentially as a solution to the above to cases

So will `Dataset2D` eventually support the `__dataframe__` protocol?

ilan-gold: > So will Dataset2D eventually support the __dataframe__ protocol?

Mmmm depends if I actually implement its use here unless you have a use-case (you probably do?).  In theory, I don't see why not but I haven't actually investigated what would go into that.  There [hasn't even been activity in the repo](https://github.com/data-apis/dataframe-api/tree/main) that hosts the protocols for like 2 years but it is documented on the arrow website: https://arrow.apache.org/docs/python/interchange_protocol.html

srivarra: It looks like in DataFrame land people have been using [PyCapsule for an interchange](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) as of late, here it is in [pandas](https://pandas.pydata.org/docs/reference/api/pandas.api.interchange.from_dataframe.html)

Specifically Pandas seem to be fully onboard with the Arrow PyCapsule interface.

> For new development, we highly recommend using the Arrow C Data Interface alongside the Arrow PyCapsule Interface instead of the interchange protocol. From pandas 3.0 onwards, from_dataframe uses the PyCapsule Interface, only falling back to the interchange protocol if that fails.

> From pandas 4.0 onwards, that fallback will no longer be available and only the PyCapsule Interface will be used.

I guess with this now, do you think `__dataframe__` is still worth your time and effort, if you'll have to add some backwards compatibility or special handling come pandas 4?

https://github.com/data-apis/dataframe-api/issues/363

ilan-gold: > I guess with this now, do you think __dataframe__ is still worth your time and effort, if you'll have to add some backwards compatibility or special handling come pandas 4?

I'm definitely not tied to the idea, so if that is the better interchange, we should use that instead.  Thanks for pointing this out!

srivarra:
@ilan-gold Could I try out the Arrow PyCapsule interchange here?

ilan-gold:
> @ilan-gold Could I try out the Arrow PyCapsule interchange here?

Go for it!

PR #2335, opened 2026-02-09
docs: link Array API

flying-sheep:
and some random fixes I made in the accessors PR that should be independent

### DIFF
--- a/src/anndata/types.py
+++ b/src/anndata/types.py
@@ -6,7 +6,7 @@ if TYPE_CHECKING:
     from enum import Enum
     from typing import Any, Literal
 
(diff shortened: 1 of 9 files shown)


src/anndata/types.py
```diff
@@ -35,7 +36,7 @@ def __array_namespace__(
         self,
         *,
         api_version: Literal["2021.12", "2022.12", "2023.12", "2024.12"] | None = None,
-    ) -> ModuleType: ...
+    ) -> ArrayNamespace: ...
```
ilan-gold, line 39: Absolutely great call!

PR #2338, opened 2026-02-10
feat!: CoW `X` by default when setting i.e., `subset.X = ...` does not update parent of `subset`

ilan-gold:
- [x] Next step in https://github.com/scverse/anndata/pull/1707#issuecomment-3847968053
- [x] Tests added

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -147,6 +147,7 @@ def test_views():
     assert adata[:, 0].X.tolist() == np.reshape([1, 4, 7], (3, 1)).tolist()
     with (
         pytest.warns(ImplicitModificationWarning, match=r"initializing view as actual"),
+        pytest.warns(FutureWarning, match=r"Automatic reshaping"),
     ):
         adata[:2, 0].X = [0, 0]
 
@@ -444,8 +445,11 @@ def test_set_scalar_subset_X(matrix_type, subset_func):
     subset_idx = subset_func(adata.obs_names)
 
     adata_subset = adata[subset_idx, :]
-
-    adata_subset.X = 1
+    with (
+        pytest.warns(ImplicitModificationWarning, match=r"initializing view as actual"),
+        pytest.warns(FutureWarning, match=r"The ability to set X with a scalar value"),
+    ):
+        adata_subset.X = 1
 
     # Subset materializes on write
     assert not adata_subset.is_view
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -34,6 +34,7 @@ from anndata._core.views import (
     SparseCSRArrayView,
     SparseCSRMatrixView,
 )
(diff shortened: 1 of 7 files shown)


ilan-gold:
@Intron7 Can you run this branch in your tests perhaps? I could see you explicitly or implicitly relying on some sort of behavior here that we are not accounting for.

tests/test_views.py
```diff
@@ -148,7 +147,7 @@ def test_views():
 
     adata[:2, 0].X = [0, 0]
 
-    assert adata[:, 0].X.tolist() == np.reshape([0, 0, 7], (3, 1)).tolist()
+    assert adata[:, 0].X.tolist() == np.reshape([1, 4, 7], (3, 1)).tolist()
 
     adata_subset = adata[:2, [0, 1]]
```
flying-sheep, line 156: check here that the view has the changed values?

tests/test_x.py
```diff
@@ -46,54 +43,23 @@ def test_setter_singular_dim(shape, orig_array_type, new_array_type):
...
+        pytest.warns(ImplicitModificationWarning, match=r"initializing view as actual"),
     ):
         view.X = to_assign
     assert_equal(view.X, to_assign)
-    assert isinstance(view.X, type(to_assign) if copy_on_write_X else type(orig_X))
+    assert isinstance(view.X, type(to_assign))
     assert_equal(adata.X, expected_X)
     # If cow, then not a view and if not cow, it is a view
-    assert view.is_view != copy_on_write_X
+    assert not view.is_view
```
flying-sheep: well the way this reads that would little sense if I didn’t know what’s going on. maybe move this check up and rename the variable where it stops being a view?

flying-sheep (commented):
> [!IMPORTANT]
> Can we get rid of ArrayView in this PR?

ilan-gold:
> Can we get rid of ArrayView in this PR?

No, I should clarify that this is only making `CoW` `subset.X = ...` which previously would have updated the actual anndata.  This is distinct from `X = subset.X` and then `X[another_subset] = ...` which is CoW.  I will update hte PR title.  My zulip freakout was about "do we need views at all", which has to do with the later point.

tests/test_views.py
ilan-gold: But it's checked below? The oprevious lines I changed i.e., `assert adata[:, 0].X.tolist() == np.reshape([1, 4, 7], (3, 1)).tolist()` are specifically checking that the parent didn't change despite the update

flying-sheep: I mean that `adata[:2, 0].X = [0, 0]` creates a view, then actualizes it and sets the new non-view AnnData’s values. But I guess that’s tested elsewhere already.

maybe at least check that that line specifically raises the “view as actual” warning

flying-sheep (approved):

PR #2340, opened 2026-02-11
fix: scalar setting/`reshape` on CoW `.X`

ilan-gold:
@flying-sheep I'm ready to merge #2338 but noticed I hadn't backported this feature with the `CoW` branch.  I'm actually not even 100% sure we should support these shortcuts.  WDYT? I'll hold off on merging the other PR until I hear back.

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -24,6 +24,7 @@ from anndata._core.views import (
     SparseCSRArrayView,
     SparseCSRMatrixView,
 )
+from anndata._warnings import ImplicitModificationWarning
 from anndata.compat import NUMPY_2, CSArray, CupyCSCMatrix, DaskArray
 from anndata.tests.helpers import (
     BASE_MATRIX_PARAMS,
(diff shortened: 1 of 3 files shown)


flying-sheep (approved):

flying-sheep:
> I'm actually not even 100% sure we should support these shortcuts

Hmm, maybe only if X is not `None` so we don’t have to guess the container data type?

Anyway, I don’t have a strong opinion here

ilan-gold:
> Hmm, maybe only if X is not None so we don’t have to guess the container data type?

That's a good point.  But then what if it's sparse? We do a dense-sparse matrix? I think I will add a warning that setting via scalars (and reshaping) will be deprecated at some point and for now it is `np.full` when scalar

tests/test_views.py
```diff
@@ -119,21 +120,43 @@ def mapping_name(request):
...
-    assert adata[:, 0].X.tolist() == np.reshape([0, 0, 7], (3, 1)).tolist()
+    ctx_managers = (
+        (
+            pytest.warns(
+                ImplicitModificationWarning, match=r"initializing view as actual"
+            ),
+            pytest.warns(FutureWarning, match=r"Automatic reshaping"),
+        )
+        if copy_on_write_X
+        else (nullcontext(),)
```
flying-sheep:
```suggestion
        else ()
```

tests/test_views.py
```diff
@@ -385,18 +408,37 @@ def test_not_set_subset_X_dask(matrix_type_no_gpu, subset_func):
...
+        (
+            pytest.warns(
+                ImplicitModificationWarning, match=r"initializing view as actual"
+            ),
+            pytest.warns(
+                FutureWarning, match=r"The ability to set X with a scalar value"
+            ),
+        )
+        if copy_on_write_X
+        else (nullcontext(),)
```
flying-sheep:
```suggestion
        else ()
```

ilan-gold: Ah great point yes!

PR #2342, opened 2026-02-11
fix: `ad.settings.override` order for sharding

### LINKED ISSUE #2341: Saving AnnData with sharding

### Report

`zarr_write_format` and `auto_shard_zarr_v3` cannot be set at the same time in a context manager.

Code:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///

import anndata
# your reproducer code

import numpy as np

adata = anndata.AnnData(np.random.randn(10, 5))

with anndata.settings.override(**{"zarr_write_format": 3, "auto_shard_zarr_v3": True}):
    adata.write_zarr("dummy.zarr")
```

Traceback:

```pytb
.conda/envs/fibrysis-py312/lib/python3.12/site-packages/zarr/core/array.py:4530: ZarrUserWarning: Automatic shard shape inference is experimental and may change without notice.
  shard_shape_parsed, chunk_shape_parsed = _auto_partition(
.conda/envs/fibrysis-py312/lib/python3.12/site-packages/zarr/api/asynchronous.py:244: ZarrUserWarning: Consolidated metadata is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.
  warnings.warn(
  File ".conda/envs/fibrysis-py312/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File ".conda/envs/fibrysis-py312/lib/python3.12/site-packages/anndata/_settings.py", line 375, in override
    setattr(self, attr, value)
  File ".conda/envs/fibrysis-py312/lib/python3.12/site-packages/anndata/_settings.py", line 310, in __setattr__
    registered_option.validate(val, self)
  File ".conda/envs/fibrysis-py312/lib/python3.12/site-packages/anndata/_settings.py", line 450, in validate_zarr_write_format
    raise ValueError(msg)
ValueError: Cannot set `zarr_write_format` to 2 with autosharding on.  Please set to `False` `anndata.settings.auto_shard_zarr_v3`
```

### Versions

numpy   2.3.5
anndata 0.12.10
----    ----
session-info2   0.2.3
pyarrow 22.0.0
natsort 8.4.0
scanpy  1.11.5
pytz    2025.2
six     1.17.0
crc32c  2.8
donfig  0.8.1.post1
setuptools      80.9.0
h5py    3.15.1
packaging       25.0
python-dateutil 2.9.0.post0
PyYAML  6.0.3
scipy   1.16.3
charset-normalizer      3.4.4
typing_extensions       4.15.0
pandas  2.3.3
zarr    3.1.3
legacy-api-wrap 1.5
numcodecs       0.16.3
----    ----
Python  3.12.12 | packaged by conda-forge | (main, Oct 22 2025, 23:25:55) [GCC 14.3.0]
OS      Linux-4.18.0-425.19.2.el8_7.x86_64-x86_64-with-glibc2.28
Updated 2026-02-11 16:53

ilan-gold:
- [x] Closes #2341
- [x] Tests added


docs/release-notes/2087.feat.md
flying-sheep, line 1: do we not validate these in a GitHub workflow or is that only active for scanpy?

no need to change that in this PR, just wondering.

src/anndata/_settings.py
```diff
@@ -374,9 +374,9 @@ def override(self, **overrides):
                     setattr(self, k, overrides.get(k))
             yield None
         finally:
-            # TODO: does the order need to be preserved when restoring?
-            for attr, value in restore.items():
-                setattr(self, attr, value)
+            for k in reversed(self._config.keys()):
```
flying-sheep, line 380: OK, I don’t quite get it.

I assume there are dependencies between the config options, but they are only expressed in the order. So reversing the order causes this to work?

No matter if I’m right or wrong, please explain in a comment why that’s the order we want.

If I’m right, we should eventually make the dependencies explicit rather than implicit.

flying-sheep (changes requested):

docs/release-notes/2087.feat.md
ilan-gold: I guess not!

src/anndata/_settings.py
ilan-gold: I will add a comment, should have assumed you would ask about this...

flying-sheep: Thank you!

flying-sheep (approved):

docs/release-notes/0.12.4.md
```diff
@@ -1,4 +1,6 @@
 (v0.12.4)=
 ### 0.12.4 {small}`2025-10-27`
 
-No significant changes.
+#### Bug fixes
+
+- {func}`dask.array.store` was producing corrupted data with zarr v3 + distributed scheduler + a lock (which we used internally): see {ref}`dask/dask#12109`. Thus dense arrays were potentially being stored with corrupted data. The solution is to remove the lock for newer versions of dask but without the lock in older versions, it is impossible to store the data. Thus versions of dask older than `2025.4.0` will not be supported for writing dense data. {user}`ilan-gold`
```
flying-sheep: This should fix the docs, right:

```suggestion
- {func}`dask.array.store` was producing corrupted data with zarr v3 + distributed scheduler + a lock (which we used internally): see {issue}`dask/dask#12109`. Thus dense arrays were potentially being stored with corrupted data. The solution is to remove the lock for newer versions of dask but without the lock in older versions, it is impossible to store the data. Thus versions of dask older than `2025.4.0` will not be supported for writing dense data. {user}`ilan-gold`
```

PR #2352, opened 2026-02-23
ci: CuPy 14 + Python 3.14

flying-sheep:
- [ ] Closes #
- [x] Tests added
- [x] Release note not necessary because: CI

gets rid of yq and just uses the pinned version of some package that does the exact thing we want.

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -62,13 +62,13 @@ jobs:
           import os
           import check_python_versions.sources.pyproject as cpv
           *_, max_ver = cpv.get_supported_python_versions()
-          with open(os.environ["GITHUB_ENV"], "a") as f:
-            print(f"max_python_version={max_ver}", file=f)
+          with open(os.environ["GITHUB_ENV"], "a") as env:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-gpu.yml
```diff
@@ -52,24 +52,19 @@ jobs:
...
-          classifiers=$(yq .project.classifiers pyproject.toml -oy  | grep --only-matching --perl-regexp '(?<=Python :: )(\d\.\d+)')
-          max_version=$(echo "$classifiers" | sort -V | tail -1)
-          echo "max_python_version=$max_version" >> $GITHUB_ENV
-
       - name: Install UV
         uses: astral-sh/setup-uv@v6 # TODO: upgrade once cirun image supports node 24
         with:
           enable-cache: true
-          # Any Cuda 14+ will support Python 3.14: https://github.com/cupy/cupy/issues/9346
-          python-version: '3.13'  # ${{ env.max_python_version }}
```
flying-sheep, line 72: all `python-version` does is setting `UV_PYTHON` so we can just do that manually after installing uv

ilan-gold (approved):

PR #2355, opened 2026-02-27
fix: virtualenv pin

ilan-gold:
- [ ] Closes #
- [ ] Tests added


ilan-gold:
Ah this doesn't work! Right!

flying-sheep:
The way Tim did it works a bit more generally than CLI arguments though, as you only need to set the env variable once per workflow: https://github.com/scverse/squidpy/pull/1129

PR #2360, opened 2026-03-09
Replace `use_cuda` with `device` parameter in AnnLoader

LiudengZhang:
## Summary

- Replace `use_cuda: bool` with `device: str = "cpu"` in `default_converter` and `AnnLoader.__init__`, addressing the TODO comment in the code (`# maybe replace use_cuda with explicit device option`).
- `use_cuda` is kept for backward compatibility but emits a `FutureWarning` when explicitly passed. Passing both `use_cuda` and `device` raises a `ValueError`.
- `pin_memory` logic is now correctly guarded to only apply on CPU (it is a CPU-only operation).

## Motivation

The existing `use_cuda` boolean is limiting — it only supports CPU or CUDA. The new `device` parameter accepts any PyTorch device string (e.g., `"cpu"`, `"cuda"`, `"cuda:1"`, `"mps"`), making the API more flexible and consistent with PyTorch conventions.

## Test plan

- [x] `test_annloader_default_device` — default `device="cpu"` produces CPU tensors
- [x] `test_annloader_explicit_cpu_device` — explicit `device="cpu"` works
- [x] `test_annloader_use_cuda_deprecation_warning` — `use_cuda` emits `FutureWarning`
- [x] `test_annloader_use_cuda_and_device_conflict` — both params raises `ValueError`

### DIFF
--- a/src/anndata/experimental/pytorch/_annloader.py
+++ b/src/anndata/experimental/pytorch/_annloader.py
@@ -71,7 +71,9 @@ class BatchIndexSampler(Sampler):
         return length
 
 
-def default_converter(arr: Array, *, device: Literal["cpu", "cuda", "mps"] = "cpu", pin_memory: bool):
+def default_converter(
+    arr: Array, *, device: Literal["cpu", "cuda", "mps"] = "cpu", pin_memory: bool
+):
     if isinstance(arr, torch.Tensor):
         arr = arr.to(device)
         if device == "cpu" and pin_memory:
--- a/src/anndata/experimental/pytorch/_annloader.py
+++ b/src/anndata/experimental/pytorch/_annloader.py
@@ -11,6 +11,7 @@ import numpy as np
 from scipy.sparse import issparse
 
 from ..._core.anndata import AnnData
+from ..._warnings import warn
 from ...compat import old_positionals
 from ..multi_files._anncollection import AnnCollection, _ConcatViewMixin
 
@@ -26,6 +27,8 @@ if TYPE_CHECKING:
     from scipy.sparse import spmatrix
 
     type Array = torch.Tensor | np.ndarray | spmatrix
+
+_UNSET = object()
 
 
 # Custom sampler to get proper batches instead of joined separate indices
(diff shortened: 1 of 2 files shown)


src/anndata/experimental/pytorch/_annloader.py
```diff
@@ -154,9 +156,23 @@ def __init__(
         batch_size: int = 1,
         shuffle: bool = False,
         use_default_converter: bool = True,
-        use_cuda: bool = False,
+        device: str = "cpu",
+        use_cuda: bool = _UNSET,
         **kwargs,
     ):
+        if use_cuda is not _UNSET:
```
flying-sheep: we have a standard `Empty.TOKEN` object for this: https://github.com/scverse/anndata/blob/a0c428379167741833eae806b18e7bda4af2b997/src/anndata/compat/__init__.py#L101-L102

src/anndata/experimental/pytorch/_annloader.py
```diff
@@ -154,9 +156,23 @@ def __init__(
         batch_size: int = 1,
         shuffle: bool = False,
         use_default_converter: bool = True,
-        use_cuda: bool = False,
+        device: str = "cpu",
```
flying-sheep: please use a `Literal` type enumerating all valid values. (maybe imported from torch if it exists there)

flying-sheep (changes requested):
thanks! Some small notes from me

src/anndata/experimental/pytorch/_annloader.py
LiudengZhang: Good call — switched to `Empty.TOKEN` in 6f01e74.

src/anndata/experimental/pytorch/_annloader.py
LiudengZhang: Done — used `Literal["cpu", "cuda", "mps"]` since torch doesn't export a device string literal. Fixed in 6f01e74.

LiudengZhang:
Hi @flying-sheep, I've addressed both points (switched to Empty.TOKEN and used Literal type) in 6f01e74. Let me know if there's anything else\!

PR #2362, opened 2026-03-10
chore: add sparse index dtype checks to assert_equal

### LINKED ISSUE #860: assert_equal fails to detect differences in dtype of sparse_matrix dtype
`assert_equal` fails to detect the differences in `dtype` of `sparse_matrix` indexing `indptr` and `indices`

code to reproduce:
```
import anndata as ad, pandas as pd, numpy as np
from scipy import sparse
from anndata.tests.helpers import assert_equal

x = sparse.csr_matrix(np.array([[0, 1], [2, 3]]), dtype=np.float32)
x.indptr = x.indptr.astype(np.int64)
x.indices = x.indices.astype(np.int64)

a = ad.AnnData(x)
print('a indptr', a.X.indptr.dtype)
print('a indices', a.X.indices.dtype)

y = sparse.csr_matrix(np.array([[0, 1], [2, 3]]), dtype=np.float32)

b = ad.AnnData(y)
print('b indptr', b.X.indptr.dtype)
print('b indices', b.X.indices.dtype)

assert_equal(a, b)
```

output:
```
a indptr int64
a indices int64
b indptr int32
b indices int32
```

LiudengZhang:
## Summary
- Enhanced `assert_equal` for sparse matrices to check `indptr` and `indices` dtype consistency when `exact=True`
- Previously, sparse matrices were converted to dense via `asarray()` before comparison, silently losing index dtype information (e.g. int32 vs int64 mismatches went undetected)
- Added test verifying the new check catches dtype mismatches

## Changes
- `src/anndata/tests/helpers.py`: Added `indptr`/`indices` dtype assertions in `assert_equal_sparse` before dense conversion, gated on `exact=True` and both operands being sparse
- `tests/test_helpers.py`: Added `test_assert_equal_sparse_index_dtype` — creates CSR matrices with identical values but different index dtypes, verifies exact comparison catches the mismatch

## Test plan
- [x] `pytest tests/test_helpers.py::test_assert_equal_sparse_index_dtype` — passes
- [x] All non-awkward tests in `test_helpers.py` pass (awkward failures are pre-existing env issue)

Fixes #860

### DIFF
--- a/src/anndata/tests/helpers.py
+++ b/src/anndata/tests/helpers.py
@@ -688,6 +688,13 @@ def assert_equal_sparse(
     exact: bool = False,
     elem_name: str | None = None,
 ):
+    if exact and sparse.issparse(b) and hasattr(a, "indptr") and hasattr(b, "indptr"):
+        assert a.indptr.dtype == b.indptr.dtype, (
+            f"{elem_name}: indptr dtype mismatch: {a.indptr.dtype} vs {b.indptr.dtype}"
+        )
+        assert a.indices.dtype == b.indices.dtype, (
+            f"{elem_name}: indices dtype mismatch: {a.indices.dtype} vs {b.indices.dtype}"
+        )
     a = asarray(a)
     assert_equal(b, a, exact=exact, elem_name=elem_name)
 
--- a/src/anndata/tests/helpers.py
+++ b/src/anndata/tests/helpers.py
@@ -690,10 +690,10 @@ def assert_equal_sparse(
 ):
     if exact and sparse.issparse(b) and hasattr(a, "indptr") and hasattr(b, "indptr"):
         assert a.indptr.dtype == b.indptr.dtype, (
-            f"{elem_name}: indptr dtype mismatch: {a.indptr.dtype} vs {b.indptr.dtype}"
+            f"{elem_name}: indptr dtype mismatch"
         )
         assert a.indices.dtype == b.indices.dtype, (
-            f"{elem_name}: indices dtype mismatch: {a.indices.dtype} vs {b.indices.dtype}"
+            f"{elem_name}: indices dtype mismatch"
         )
     a = asarray(a)
     assert_equal(b, a, exact=exact, elem_name=elem_name)
--- a/src/anndata/tests/helpers.py
+++ b/src/anndata/tests/helpers.py
@@ -689,9 +689,7 @@ def assert_equal_sparse(
     elem_name: str | None = None,
 ):
     if exact and sparse.issparse(b) and hasattr(a, "indptr") and hasattr(b, "indptr"):
-        assert a.indptr.dtype == b.indptr.dtype, (
-            f"{elem_name}: indptr dtype mismatch"
-        )
+        assert a.indptr.dtype == b.indptr.dtype, f"{elem_name}: indptr dtype mismatch"
         assert a.indices.dtype == b.indices.dtype, (
             f"{elem_name}: indices dtype mismatch"
         )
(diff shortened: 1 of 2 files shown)


ilan-gold:
Please update your PR title to be `chore: ...` :)

src/anndata/tests/helpers.py
```diff
@@ -688,6 +688,13 @@ def assert_equal_sparse(
     exact: bool = False,
     elem_name: str | None = None,
 ):
+    if exact and sparse.issparse(b) and hasattr(a, "indptr") and hasattr(b, "indptr"):
+        assert a.indptr.dtype == b.indptr.dtype, (
+            f"{elem_name}: indptr dtype mismatch: {a.indptr.dtype} vs {b.indptr.dtype}"
+        )
+        assert a.indices.dtype == b.indices.dtype, (
+            f"{elem_name}: indices dtype mismatch: {a.indices.dtype} vs {b.indices.dtype}"
+        )
```
ilan-gold:
```suggestion
        assert a.indptr.dtype == b.indptr.dtype, (
            f"{elem_name}: indptr dtype mismatch"
        )
        assert a.indices.dtype == b.indices.dtype, (
            f"{elem_name}: indices dtype mismatch"
        )
```
I think the mismatch's contents are already reported by `assert`

tests/test_helpers.py
```diff
@@ -248,6 +248,21 @@ def test_assert_equal_dask_arrays():
     assert_equal(c, d)
 
 
+def test_assert_equal_sparse_index_dtype():
+    """assert_equal(exact=True) should detect indptr/indices dtype mismatches."""
+    a = sparse.random(10, 10, format="csr", density=0.3)
+    b = a.copy()
+    b.indptr = b.indptr.astype(np.int64)
+    b.indices = b.indices.astype(np.int64)
```
ilan-gold: Please parametrize this test to change each of these one-at-a-time and then have the `match` check that `indices` or `indptr` is in the message accordingly

src/anndata/tests/helpers.py
LiudengZhang: Done — simplified the assert messages and included `elem_name` for context. Updated in 6cb2b354.

tests/test_helpers.py
LiudengZhang: Done — parametrized with `@pytest.mark.parametrize("attr", ["indices", "indptr"])`, each tested one-at-a-time with `match=attr`. Updated in 6cb2b354.

tests/test_helpers.py
```diff
@@ -248,6 +248,21 @@ def test_assert_equal_dask_arrays():
     assert_equal(c, d)
 
 
+@pytest.mark.parametrize("attr", ["indices", "indptr"])
+def test_assert_equal_sparse_index_dtype(attr):
+    """assert_equal(exact=True) should detect indptr/indices dtype mismatches."""
+    a = sparse.csr_matrix(np.eye(3))
+    b = sparse.csr_matrix(np.eye(3))
+    setattr(b, attr, getattr(b, attr).astype(np.int32))
```
ilan-gold: https://github.com/scverse/anndata/pull/2362/changes/6cb2b35423e32f81ec06bea80cfdbf2ed59d119c#diff-08a6944653efc734f4c8cb856fc854b9f21d655ca69796af2b170c53bff7919eL255-L256 this si what you want?
```suggestion
    setattr(b, attr, getattr(b, attr).astype(np.int64))
```

LiudengZhang: Good catch — updated to int64. The default scipy index dtype is int32, so int32 was a no-op.

PR #2366, opened 2026-03-16
Enable write to existing zarr store

joshua-gould:
Added ability for write_zarr to accept a zarr group


ilan-gold (commented):
@joshua-gould Awesome feature - could you look at https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_io/specs/registry.py#L389-L397.  This is just some old behavior that was never updated but I _think_ this should be relaxed to just clear out all subgroups/keys instead of blindly clearing the entire store.  I'm concerned about writing to an existing group but then you end up deleting _all_ of its contents even those in "parent" groups

Other than that, this looks good to me!

PR #2367, opened 2026-03-18
fix!: remove `__delitem__` and `__setitem__` from the `AnnData` object

ilan-gold:
Towards making `backed` mode simpler, we remove these two methods

- [ ] See https://github.com/scverse/anndata/issues/2369
- [ ] Tests added

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -63,7 +63,7 @@ jobs:
         id: changes
         with:
           filters: | # this is intentionally a string
-            relnotes: 'docs/release-notes/${{ github.event.pull_request.number }}.${{ (contains(needs.check-milestone.outputs.type, '!') && 'breaking') || needs.check-milestone.outputs.type }}.md'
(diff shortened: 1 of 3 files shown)


.github/workflows/check-pr.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         id: changes
         with:
           filters: | # this is intentionally a string
-            relnotes: 'docs/release-notes/${{ github.event.pull_request.number }}.${{ needs.check-milestone.outputs.type }}.md'
+            relnotes: 'docs/release-notes/${{ github.event.pull_request.number }}.${{ (contains(github.event.pull_request.title, '!') && 'breaking') || needs.check-milestone.outputs.type }}.md'
```
ilan-gold, line 66: No better way to test/figure this out than a real PR

flying-sheep: I think this is basically the logic yeah!

flying-sheep (approved):
good call!

This gets rid of `del adata[cells, :]` in favor of `adata = adata[~adata.obs_names.isin(cells), :].copy()`, which is less elegant, but possibly niche enough that facilitating it isn’t necessary.

Issue #2369, opened 2026-03-18
Simplify `backed` internal checking


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

Right now `AnnData.isbacked` (i.e., `X is None` and `AnnData.filename` is present) is used for checking for a variety of different behaviors/conditions aside from just getting `X`:

1. It prevents [making a view-of-a-view](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L281-L286) when backed mode is present
2. [Custom repr](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L548)
3. [Preventing string->categorical conversion when `AnnData.is_view`](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L1153-L1156) this is meant to prevent accidentally reading a backed array into memory, I guess, since it will materialize `X`
4. [`AnnData.T`](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L1234-L1243) although this is actually just broken in general - you can't tranpose something backed and will get an `AttributeError` about `.T` on `h5py.Dataset` - again ?????
5. [`AnnData.concatenate`](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L1679-L1681): https://github.com/scverse/anndata/pull/2370
6. [A bunch of stuff in `Raw`](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/raw.py)
7. `AnnData.__{del,set}item__`: https://github.com/scverse/anndata/pull/2367
8. [Making a `copy` of `AnnData`](https://github.com/scverse/anndata/blob/ef466f23c193a41b4a60ab84bdbc5ce91c488346/src/anndata/_core/anndata.py#L1436-L1447)

I think these cases can succinctly be broken down into two categories:

- [ ] **Broken/unnecessary behavior** I think 5 (because it is deprecated) and 2 fall into this category.  To solve this I would propose *just remove the behavior*
[...]

PR #2370, opened 2026-03-18
chore!: remove `AnnData.concatenate`

ilan-gold:
- [x] Towards https://github.com/scverse/anndata/issues/2369
- [x] Tests added

### DIFF
--- a/docs/release-notes/0.9.0.md
+++ b/docs/release-notes/0.9.0.md
@@ -39,7 +39,7 @@
 
 #### Deprecations
 
-- {meth}`AnnData.concatenate() <anndata.AnnData.concatenate>` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
+- `AnnData.concatenate() <anndata.AnnData.concatenate>` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
 
 #### Bug fixes
 
--- a/docs/release-notes/0.9.0.md
+++ b/docs/release-notes/0.9.0.md
@@ -39,7 +39,7 @@
 
 #### Deprecations
 
-- `AnnData.concatenate() <anndata.AnnData.concatenate>` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
+- `AnnData.concatenate()` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
 
 #### Bug fixes
 
--- a/tests/test_concatenate.py
+++ b/tests/test_concatenate.py
@@ -48,10 +48,6 @@ if TYPE_CHECKING:
 
     from anndata._types import Join_T
 
-mark_legacy_concatenate = pytest.mark.filterwarnings(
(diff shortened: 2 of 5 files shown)


docs/release-notes/0.9.0.md
```diff
@@ -39,7 +39,7 @@
 
 #### Deprecations
 
-- {meth}`AnnData.concatenate() <anndata.AnnData.concatenate>` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
+- `AnnData.concatenate() <anndata.AnnData.concatenate>` is now deprecated in favour of {func}`anndata.concat` {pr}`845` {user}`ivirshup`
```
flying-sheep: broken syntax now

tests/test_concatenate.py
```diff
@@ -360,24 +322,23 @@ def test_concatenate_dense():
...
+            [1.0, 2.0],
+            [2.0, 1.0],
+            [np.nan, 0.0],
         ])
     )
     assert np.array_equal(var_ma.mask, var_ma_ref.mask)
     assert np.allclose(var_ma.compressed(), var_ma_ref.compressed())
 
 
-@mark_legacy_concatenate
```
flying-sheep, line 373: The possibly redundant ones start here I think

tests/test_concatenate.py
```diff
@@ -612,8 +568,42 @@ def get_obs_els(adata):
...
+            [
+                "annoA",
+                "annoB",
+                "annoC",
+                "annoE",
+            ],
+        ),
+        ("only", ["annoE"]),
+        (None, []),
+    ],
```
flying-sheep, line 579: can you make that more compact using the magic comma please?

flying-sheep (commented):
looks great, but please go through all the tests you now switched from `concatenate` to concat and delete all the duplicate ones if any exist – I assume we test `concat` elsewhere and some of the changed tests now test the exact same code paths as previously existing ones.

tests/test_concatenate.py
ilan-gold: Weirdly enough, this one is simultaneously testing two things not currently tested - how does `join` affect `obs` columns and `X` when `X` is dense.  There is a similar test for sparse, `test_concatenate_sparse` which doesn't do the `obs`. Should they be refactored for parametrization? Maybe, I'd need to look closer to see if that's feasible.

That's why I did refactor the other test where the big diff is - the behavior literally changed and would not have been tested otherwise.

flying-sheep: yeah, I just mean that there are a bunch of tests where you remove `@mark_legacy_concatenate` and replace `concatenate` with `concat`. Are there no pre-existing `concat` tests that cover what these ones do?

ilan-gold: > Are there no pre-existing concat tests that cover what these ones do?

That is what I'm saying.  For example if you search `join` you will find no test that checks the `obs` column/key behavior AFAIK.  The `obs` behavior is also very poorly specified ATM (sorry thought there was an issue for this):  https://github.com/scverse/anndata/issues/2374 So I just figured I'd leave this test alone for now

flying-sheep (approved):

PR #2236, opened 2025-11-29
feat: Add HTML representation

katosh:
# Rich HTML representation for AnnData

- [x] Closes #675
- [x] Tests added
- [ ] Release note not necessary because:

## Summary

Implements rich HTML representation (`_repr_html_`) for AnnData objects in Jupyter notebooks. Builds on previous draft PRs (#784, #694, #521, #346) with a complete, production-ready implementation.

**[Live Demo](https://htmlpreview.github.io/?https://gist.githubusercontent.com/katosh/4a2399d1472c733b041ef8dfd5b489b9/raw/repr_html_visual_test.html)** | **[Reviewer's Guide](https://gist.github.com/katosh/26cac20f9792ddf27e4dd628de8f5685)** *(technical details, design decisions, extensibility examples)*

## Screenshot

screenshot2

### Interactive Display
- **Foldable sections** with auto-collapse for large datasets
- **Search/filter** with regex and case-sensitive toggles
- **Copy-to-clipboard** for field names
- **Nested AnnData** expansion with configurable depth
- **`.raw` section** showing unprocessed data (#349)

### Visual Indicators
- **Category colors** from `uns` palettes (e.g., `cell_type_colors`)
- **Type badges** for views, backed mode, sparse matrices, Dask arrays
- **Serialization warnings** for data that won't write to H5AD/Zarr
- **Value previews** for simple `uns` values
- **README support** via modal (renders markdown from `uns["README"]`)
- **Memory info** in footer

### Serialization Warnings

Proactively warns about data that won't serialize:

| Level | Issue | Related |
|-------|-------|---------|
| 🔴 Error | datetime64/timedelta64 in obs/var | #455, #2238 |
| 🔴 Error | Non-string keys | #321 |
| 🔴 Error | Object columns with dicts/lists/custom objects | #1923, #567, #636 |
| 🔴 Error | Non-serializable types in uns | |
| 🟡 Warning | Keys with `/` (deprecated) | #1447, #2099 |
| 🟡 Warning | String→categorical auto-conversion | #534, #926 |

### Compatibility
- **Dark mode** auto-detection (Jupyter Lab/VS Code, Furo/sphinx-book-theme)
- **No-JS fallback** with graceful degradation
- **JupyterLab safe** - CSS scoped to `.anndata-repr` prevents style conflicts
- **Lazy-loading safe** - configurable partial loading for `read_lazy()` (categories, colors)
- **Zero dependencies** added

## Extensibility

[...]

PR #2372, opened 2026-03-19
feat: `AnnData.unwriteable` based on `AnnData._reduce` + `iter_outer` + refactorings of other relevant functions

ilan-gold:
maybe TODOs:
- [x] `fold`? `reduce`?
- [ ] Implement this inside `write_{zarr,h5ad}` to prevent partial stores from being written?

We introduce `AnnData.fold` and `AnnData.can_write` for "crawling" an `AnnData` and accumulating, implementing a `can_write` to tell if an `AnnData` can be written to a given format.

I'm also open to `reduce` as the name instead of `fold`.

For #2236, I think we could implement this more complexly by building on `fold` and accumulating a `set` of `AdRef` (or similar) objects pointing at what can't be written. Open to implement this! I reimplemented `__sizeof__` to stress-test using reference-accessors and it went ok, but it's not amazingly smooth (like `uns` and trying to figure out if you are in a "top-level" elem like `obs` or `varm`)

If https://github.com/zarr-developers/zarr-python/pull/3613 ever happens (I hope so) or we do https://github.com/scverse/anndata/issues/2238, this will be crucial to making things smoother with hdf5 for preventing incompatibilities.

- [ ] Towards #2369 and moving towards something raised in #2236
- [ ] Tests added


src/anndata/_core/anndata.py
```diff
@@ -1446,6 +1453,75 @@ def copy(self, filename: PathLike[str] | str | None = None) -> AnnData:
...
+        from anndata._io.specs.registry import _REGISTRY
+
+        writeable_elems = {
+            src_type
+            for (dest_type, src_type, __) in _REGISTRY.write
+            if store_type is None or store_type in dest_type.__module__
+        }
+
+        def predicate(x: RWAble, *, accumulate: bool, ref_acc: AdRef | RefAcc | None):
+            if isinstance(ref_acc, AdRef) or ref_acc is None:
```
ilan-gold: Here and above `uns` is `ref_acc is None` but I feel like there should be a better way to describe this!

ilan-gold:
Just looking for a first-pass vibe check on this!

src/anndata/_core/anndata.py
```diff
@@ -1446,6 +1453,75 @@ def copy(self, filename: PathLike[str] | str | None = None) -> AnnData:
...
+        ]:
+            attr = getattr(self, attr_name)
+            acc = getattr(A, attr_name)
+            if attr_name != "X":
+                for elem_name in attr:
+                    ref = acc[elem_name]
+                    accumulate = func(
+                        attr[elem_name], accumulate=accumulate, ref_acc=ref
+                    )
+            accumulate = func(attr, accumulate=accumulate, ref_acc=acc)
```
ilan-gold: So we could make BFS vs. DFS an option for iteration order - not really sure who would want BFS but I'm open to it. I think the bigger thing is that this is only DFS-like where the parent node is visited last instead of first (helpful for `__sizeof__`) i.e., this call goes before `for_elem_name in attr`

src/anndata/_core/anndata.py
```diff
@@ -513,36 +517,57 @@ def _init_as_actual(  # noqa: PLR0912, PLR0913, PLR0915
...
-        return sum(sizes.values())
+                # an element of uns
+                or (ref_acc is None and X is not self.uns)
+            ):
+                if is_ad_ref:
+                    key = type(ref_acc.acc)
+                elif is_ref_acc:
+                    key = ref_acc.parent_type
+                else:
+                    key = None
```
ilan-gold: this was discussed during the PR for reference accessors but I think it should be easier to traverse up the tree.  open to ideas!

src/anndata/_core/anndata.py
```diff
@@ -1446,6 +1471,89 @@ def copy(self, filename: PathLike[str] | str | None = None) -> AnnData:
...
             return read_h5ad(filename, backed=mode)
 
+    def fold[T](
+        self,
+        func: FoldFunc[T],
+        *,
+        init: T,
+        order: Literal["DFS-pre", "DFS-post"] = "DFS-post",
+    ) -> T:
+        """Accumulate a value starting from init by iterating over the "elems"/leaf nodes of the AnnData object.
```
flying-sheep: Please explain what this function passes to `ref_acc` for which elements.

Either
- link explicitly to `FoldFunc` (additionally to the typehints auto link) and explain it there, or
- explain it here and direct people from the `FoldFunc` docs to here.

flying-sheep (commented):
Very exciting! I think this mainly needs some serious documenting and examples, not much else.

For examples, maybe first something that just handles one kind of accessor/reference first to show how one can short-circuit by returning the accumulator early.

Regarding naming: the canonical name in the stdlib is `reduce`, it even used to be in `builtins`, but maybe the scientific python world deviates for no good reason again. It’s also fair if you have other objections to `reduce`!

src/anndata/_core/anndata.py
flying-sheep: right, let’s talk about this!

src/anndata/_core/anndata.py
```diff
@@ -513,36 +517,57 @@ def _init_as_actual(  # noqa: PLR0912, PLR0913, PLR0915
...
-                    f"Size of {attr.replace('_', '.'):<7}: {tqdm.format_sizeof(s, 'B')}"
+            if is_elem := (
+                (
+                    # an array of some sort i.e., from AdRef (from obs/var) or a reference to one
+                    (is_ad_ref := isinstance(ref_acc, AdRef))
+                    or (
+                        is_ref_acc := isinstance(
+                            ref_acc, LayerAcc | MultiAcc | GraphAcc
+                        )
+                    )
```
flying-sheep: this is

```py
isinstance(ref_acc, AdRef | RefAcc) and not isinstance(ref_acc, MetaAcc)
```

because you want to handle each homogeneous array.

I think there are some ways to make this more ergonomic, e.g.:
- we can add more marker superclasses that handle this use case, or
- we could change the implementation of fold/reduce so it descends into things and returning or raising some marker would prevent descent. that might help handling things as early as possible and therefore having less complicated logic.

katosh:
Hi @ilan-gold, interesting direction! The HTML repr PR (#2236) intersects with `fold` in a couple of ways, so I wanted to share some observations from the implementation side.

## HTML repr (#2236): traversal overlap

The `_repr_html_` traverses the same element tree that `fold` formalizes. The core loop lives in [`html.py:_render_all_sections`](https://github.com/scverse/anndata/blob/7eba3a49/src/anndata/_repr/html.py#L331-L361) and iterates over a [`SECTION_ORDER`](https://github.com/scverse/anndata/blob/7eba3a49/src/anndata/_repr/__init__.py#L360-L371) tuple:

```python
SECTION_ORDER = ("X", "obs", "var", "uns", "obsm", "varm", "layers", "obsp", "varp", "raw")

for section in SECTION_ORDER:
    parts.append(_render_section(adata, section, context))
```

This is structurally very similar to `fold`'s inner loop. A few differences worth noting:

1. **`raw` is missing from `fold`.** The repr includes `.raw` as a collapsible section (X, var, varm). Would `fold` eventually cover `raw`?

2. **Section-level grouping.** The repr builds a tree (section headers, then entries, then nested content). Since entries within a section are visited contiguously, the callback could track state in the accumulator to detect section transitions: in DFS-pre mode the parent visit signals "entering," and the next section's parent visit implicitly signals "exiting" the previous one. This is workable, but it would be smoother if `fold` supported explicit enter/exit events for sections, so consumers don't need to roll their own state tracking for something that `fold` already knows internally.

3. **Custom/unknown sections.** The repr also detects mapping-like attributes *not* in the standard list (for ecosystem packages like TreeData that add `.obst`/`.vart`). `fold` currently hardcodes the same attribute list. Worth considering whether it should be extensible or whether that's out of scope.

4. **Nested AnnData in `uns`.** The repr handles AnnData objects found inside `uns` values, with depth tracking to prevent infinite recursion. `fold` doesn't recurse into nested AnnData within `uns`, so that use case wouldn't be covered.

On the positive side, the per-column iteration of obs/var via `MetaAcc` aligns well with how the repr already iterates `df.columns` to render each column individually.

The repr would naturally use DFS-pre order (render section header first, then its entries), so it's good that option exists. If `fold` stabilizes, I could likely use it as the backbone of the repr's traversal by accumulating a `list[RenderedSection]`. The `SECTION_ORDER` loop would become a `fold` call. The main prerequisite would be that `fold` supports distinguishing "entering a section" from "visiting a leaf within it."

## Serialization checks: `can_write` overlap

I understand `can_write` has a different primary goal: preventing partial writes to zarr/h5 stores by checking upfront whether the whole object is writable. That's valuable on its own. But there's a nice overlap: the repr already performs the same registry lookups at element granularity ([`utils.py:is_serializable`](https://github.com/scverse/anndata/blob/7eba3a49/src/anndata/_repr/utils.py#L67-L110), [`formatters.py:_check_series_serializability`](https://github.com/scverse/anndata/blob/7eba3a49/src/anndata/_repr/formatters.py#L128-L167)) to show warning icons on entries that can't be written. These checks:

- Query `_REGISTRY.write` to test if a type has a registered writer (same approach as `can_write`'s predicate)
- Additionally check Series backing arrays, key name validity (slashes, non-string keys), and nested container contents
- Return `(is_serializable, reason)` tuples so the repr can show *why* something can't be written

If the underlying element-level check were factored into a shared utility, something like `can_write_elem(obj, store_type=) -> tuple[bool, str]`, both use cases could build on it: `can_write` would `fold` with `all(...)`, and the repr would call it per-entry to get the reason string for its warning icon. I'd be happy to contribute to making that work if it's a direction you'd consider.

Happy to discuss further, just wanted to flag the touchpoints while the API is still taking shape.

ilan-gold:
> raw is missing from fold. The repr includes .raw as a collapsible section (X, var, varm). Would fold eventually cover raw?

Great catch! I think we currently don't have a path accessor for it, so I think that's the same thing as `uns`

> Nested AnnData in uns. The repr handles AnnData objects found inside uns values, with depth tracking to prevent infinite recursion. fold doesn't recurse into nested AnnData within uns, so that use case wouldn't be covered.

Phil brought this up in a call on Friday as well, not sure (again) what is to be done

> The main prerequisite would be that fold supports distinguishing "entering a section" from "visiting a leaf within it."

I'm not sure I follow this line of thinking fully - why can't you distinguish between these two now? If you are doing `pre` order, you know you enter a section when you see the reference to `obs` or `varp` etc. No? Why do you need an exit signal as well?

> If the underlying element-level check were factored into a shared utility, something like can_write_elem(obj, store_type=) -> tuple[bool, str], both use cases could build on it: can_write would fold with all(...), and the repr would call it per-entry to get the reason string for its warning icon. I'd be happy to contribute to making that work if it's a direction you'd consider.

That could make sense.  Once you merge this into your PR we can refactor this if needed + work out more kinks

src/anndata/_core/anndata.py
```diff
@@ -513,36 +517,57 @@ def _init_as_actual(  # noqa: PLR0912, PLR0913, PLR0915
...
-            elif isinstance(X, BaseCompressedSparseDataset) and with_disk:
-                return cs_to_bytes(X._to_backed())
-            elif issparse(X):
-                return cs_to_bytes(X)
-            else:
-                return X.__sizeof__()
-
-        sizes = {}
-        attrs = ["X", "_obs", "_var"]
-        attrs_multi = ["_uns", "_obsm", "_varm", "varp", "_obsp", "_layers"]
```
ilan-gold, line 530: FWIW this never handled `raw` anyway

ilan-gold:
See the `reduce` docs for more info but we can now visit + introspect said visits to `.raw` and `.uns`.  Aside from that, things should be clearer now:

1. `__sizeof__` now actually visits `raw` and does more than `__sizeof__` on `uns` (which as I understand, doesn't actually visit the values of that `dict`)
2. `fold` -> `reduce`
3. Docs added with clear explanation of how the accumulation can be used for introspection
4. Tests added that explicitly check the traversal in `show_stratified` of all elements

katosh:
Hi @ilan-gold, thanks for the updates! The rename to `reduce`, the `raw`/`uns` support, the docs, and the traversal tests are all great improvements. Really nice to see `__sizeof__` now covering `raw` properly too.

I was curious how `reduce` would work as a traversal backbone for the HTML repr, so I ran an experiment: merged `ig/fold_can_write` into the repr branch and rewrote `_render_all_sections` to use `adata.reduce(DFS-pre)`. The experiment lives here: [settylab/anndata#6](https://github.com/settylab/anndata/pull/6) (the PR description has a guided walkthrough, and the code has `PAIN POINT` comments at each friction point).

Before getting into findings: is `reduce` intended as a general-purpose traversal tool that consumers like the repr could build on? Or is it primarily an internal utility for `__sizeof__` and `can_write`, where the callback cares about individual values but not about the tree structure? The answer changes how much the points below matter.

**What worked well:**

The shared IO registry utilities are a clear win. I extracted `IORegistry.get_writeable_types()` and `IORegistry.has_spec()` from the `can_write` logic, and both `can_write` and the repr's serializability checks now use them instead of each rolling their own `_REGISTRY.get_spec()` try/except. This simplification is independent of `reduce` itself.

**What I ran into:**

1. **Accumulator complexity (the enter/exit question).** Without enter/exit signals, the callback must carry a 4-tuple through every call to track which section is "open." The "close section" logic appears in two places, and section entry is detected via isinstance checks on the accessor hierarchy. These checks are fragile: for example, `LayerAcc` is used for both X (`k=None`) and individual layer entries (`k='counts'`), so a naive isinstance match treats each layer as its own section. The initial implementation had exactly this bug. All of this re-derives structure that `reduce` already knows internally.

   **Without** enter/exit events (what the experiment implements):

   ```python
   def render_html(elem, *, accumulate, ref_acc):
       sections, current_rows, current_ref, current_elem = accumulate
   
       # Detect section entry by isinstance-checking the accessor hierarchy.
       # Also implicitly closes the *previous* section, since there's no exit signal.
       if isinstance(ref_acc, (AdAcc, MapAcc)):
           if current_rows is not None:
               # Wrap up the previous section (need current_elem for entry count, collapse, etc.)
               sections.append(render_section(current_ref, "\n".join(current_rows), n_items=len(current_elem), ...))
           current_ref = ref_acc
           current_elem = elem
           current_rows = []
       elif isinstance(ref_acc, (AdRef, RefAcc)):
           # Leaf: format and collect
           output = formatter_registry.format_value(elem, context)
           current_rows.append(_render_entry_row(ref_acc.k, output))
   
       return sections, current_rows, current_ref, current_elem
   
   sections, leftover_rows, last_ref, last_elem = adata.reduce(
       render_html, init=([], None, None, None), order="DFS-pre",
   )
   # The *last* section never got an "enter-next" to trigger its wrap-up,
   # so we must duplicate the render_section() call here:
   if leftover_rows is not None:
       sections.append(render_section(last_ref, "\n".join(leftover_rows), n_items=len(last_elem), ...))
   ```

   **With** optional `on_enter`/`on_exit` hooks (hypothetical):

   ```python
   def on_enter(elem, *, accumulate, ref_acc):
       accumulate.append(SectionBuilder(ref_acc=ref_acc, elem=elem, rows=[]))
       return accumulate
   
   def on_exit(elem, *, accumulate, ref_acc):
       sb = accumulate[-1]
       accumulate[-1] = render_section(sb.ref_acc, "\n".join(sb.rows), n_items=len(sb.elem), ...)
       return accumulate
   
   def visit_leaf(elem, *, accumulate, ref_acc):
       output = formatter_registry.format_value(elem, context)
       accumulate[-1].rows.append(_render_entry_row(ref_acc.k, output))
       return accumulate
   
   sections = adata.reduce(
       visit_leaf, init=[],
       on_enter=on_enter, on_exit=on_exit,
   )
   ```

   The accumulator becomes a plain `list`, each concern is its own function, and the `ReduceFunc` protocol stays unchanged. Consumers that don't need hooks simply don't pass them. See the [PR diff](https://github.com/settylab/anndata/pull/6/files) for the full implementation, specifically `_finalize_section` and `_is_section_visit`.

2. **Metadata mismatch.** For obs/var, `len(elem)` returns the number of rows, not columns. The callback needs a per-type branch to get the correct entry count. The original renderer just called `len(df.columns)`.

3. **Opaque uns/raw.** `reduce` doesn't descend into these, so the callback falls back to the same dedicated renderers the old code used (`_render_uns_section`, `_render_raw_section`). For the repr, these are the most complex sections (nested AnnData, type hints, depth tracking), so `reduce` doesn't help where it would help most.

4. **Hardcoded attribute list.** Extension packages (TreeData adding `.obst`/`.vart`) are invisible to `reduce`. Custom sections and unknown section detection must be handled entirely outside the `reduce` call. For `can_write` this is a correctness concern: it could return `True` when a custom section contains non-serializable data. Having extensions override `reduce` wouldn't compose well either (last override wins), so some form of discovery mechanism would be needed.

5. **Robustness regression.** This is probably the most important point. The current per-section rendering is deliberately defensive: each section is rendered independently, and if one section fails, the rest still render fine. With `reduce` as the backbone, a single problematic element can crash the entire traversal. For example, a non-string DataFrame column name (e.g., a tuple key in obs) causes the accessor system to throw, which aborts `reduce` entirely. The repr goes from showing all other sections correctly with just one error indicator, to showing nothing but "reduce failed."

   Visual test comparison (test 23 "Serialization warnings"):

   - [**Original** (per-section rendering)](https://htmlpreview.github.io/?https://gist.githubusercontent.com/katosh/4a2399d1472c733b041ef8dfd5b489b9/raw/repr_html_visual_test.html) — tuple column name handled gracefully, other sections render fine
   - [**With reduce**](https://htmlpreview.github.io/?https://gist.github.com/katosh/4ec6336095b3deb73e5b249ce706eaf8/raw/repr_html_visual_test.html) — entire AnnData shows "reduce failed" because the accessor system crashes on the tuple key

   <details><summary>Screenshots (test 23)</summary>


   **Original — per-section rendering:**
   image
   image


   **With reduce:**
   image
   image

   </details>

**`can_write_elem` shared utility:**

Sounds like a good plan. Both `can_write` and the repr's per-element warning icons ultimately query `_REGISTRY` for writers, so extracting a shared `has_writer(obj, store_type=)` utility should be straightforward once both PRs are closer to merging.

**Nested AnnData in `uns`:**

Understood that this is still an open question. For reference, the HTML repr already handles this case with a simple depth counter: when it encounters an `AnnData` inside `uns`, it calls `generate_repr_html` recursively with `depth + 1` and stops at `max_depth` ([html.py#L273](https://github.com/settylab/anndata/blob/0200ac59/src/anndata/_repr/html.py#L273)). If `reduce` eventually wants to descend into nested AnnData, that pattern could serve as a reference. No rush on this though, perfectly fine to scope it out for now.

**Net assessment:**

For `__sizeof__` and `can_write`, where the callback accumulates a scalar and doesn't care about tree structure, `reduce` works well. For the repr, where the callback needs to produce grouped, ordered, nested HTML with section headers, entry counts, truncation, and custom sections, `reduce` makes the code harder to follow than the original per-section loop. The shared registry utilities are the real win, and they don't depend on `reduce`.

Happy to walk through the code in more detail. The experiment branch has `PAIN POINT` comments at each friction point.

src/anndata/_core/anndata.py
```diff
@@ -1417,6 +1442,114 @@ def copy(self, filename: PathLike[str] | str | None = None) -> AnnData:
...
+        self,
+        func: ReduceFunc[T],
+        *,
+        init: T,
+        order: Literal["DFS-pre", "DFS-post"] = "DFS-post",
+    ) -> T:
+        """Accumulate a value starting from init by iterating over the "elems"/leaf nodes of the AnnData object.
+
+        All visits inside the user-defined `func` (see :func:`types.ReduceFunc`) are distinguishable via the `ref_acc` + `elem` args.
+        Visits to {attr}`~AnnData.raw` pass `ref_acc is None` and `isinstance(elem, Raw)` to the :func:`types.ReduceFunc`.
```
flying-sheep: these links are broken, check the docs preview! https://icb-anndata--2372.com.readthedocs.build/en/2372/generated/anndata.AnnData.reduce.html

probablyn needs `anndata.` in there

docs/api.md
```diff
@@ -264,6 +264,13 @@ Types used by the former:
    abc.CSCDataset
 ```
 
+```{eval-rst}
+.. autosummary::
+   :toctree: generated/
+
+   types.ReduceFunc
```
flying-sheep: use the [class-minimal.rst](https://github.com/scverse/anndata/blob/main/docs/_templates/autosummary/class-minimal.rst) template to get inline `__call__`

src/anndata/_core/anndata.py
```diff
@@ -512,36 +516,57 @@ def _init_as_actual(  # noqa: PLR0912, PLR0913, PLR0915
...
+                accumulate[Raw] += get_size(X.X)
+                accumulate[Raw] += get_size(X.var)
+                for key in X.varm:
+                    accumulate[Raw] += get_size(X.varm[key])
+            elif ref_acc is None:  # "None but not Raw" is uns
+                accumulate[None] = get_size(self.uns)
+            if is_elem := (
+                # an array of some sort i.e., from AdRef (from obs/var) or a reference to one
+                (is_ad_ref := isinstance(ref_acc, AdRef))
+                or isinstance(ref_acc, LayerAcc | MultiAcc | GraphAcc)
```
flying-sheep: so `RefAcc` without `MetaAcc`, or “child of a MapAcc”.

I wonder if this use case would be better handled with a control flow API. See main review comment.

flying-sheep (commented):
OK so I brought it up before, but I maybe there’s an alternative design that would work better:

Basically when using DFS-pre, the user would be able to skip descending, here’s a little doodle how that could look with an iterator API, but it could also be done with return values of the current fold function: https://gist.github.com/flying-sheep/b4274c148a08b7f037c732bec4a163a5

katosh:
Building on @flying-sheep's [iterator + descend control sketch](https://gist.github.com/flying-sheep/b4274c148a08b7f037c732bec4a163a5) and the [reduce experiment](https://github.com/settylab/anndata/pull/6):

`reduce` is a nice formalization of the section-by-section traversal that already lives in many places across anndata. It works well for scalar accumulators like `__sizeof__` and `can_write`. For tree-aware consumers like the repr, each one needs such different structure (error isolation, traversal depth, section metadata, leaf semantics) that a general traversal risks either under-serving complex consumers or over-burdening simple ones. Independently of how the iteration itself works, I think there's an opportunity to complement `reduce` with a shared notion of *which sections exist*.

What *would* help all these consumers in my opinion is **shared metadata infrastructure**:

1. **Section registry** so the section list isn't hardcoded everywhere. I counted at least nine places on `main` that each maintain their own list:

   <details><summary>Hardcoded section lists on <code>main</code></summary>


   - [`write_h5ad`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_io/h5ad.py#L96-L103)
   - [`write_anndata`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_io/specs/methods.py#L289-L299)
   - [`read_anndata`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_io/specs/methods.py#L310-L322)
   - [`read_h5ad_backed`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_io/h5ad.py#L176-L177)
   - [`__sizeof__`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_core/anndata.py#L529-L530)
   - [`_gen_repr`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_core/anndata.py#L549-L558)
   - [`_mutated_copy`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_core/anndata.py#L1329)
   - [`to_memory`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_core/anndata.py#L1367-L1377)
   - [`concat`](https://github.com/scverse/anndata/blob/e49e1272/src/anndata/_core/merge.py#L1808-L1819)

   </details>

   For packages that build on anndata like TreeData, this means reimplementing every one of these functions just to add `.obst`/`.vart` to the list. TreeData for example [rewrites the entire write traversal](https://github.com/YosefLab/treedata/blob/main/src/treedata/_core/write.py#L73-L90), plus `__repr__`, `copy`, subsetting, and anything else that iterates sections. A section registry would let extensions register once and be discovered everywhere, and `can_write` wouldn't silently skip extension sections.

2. **Shared IO registry utilities** (`get_writeable_types`, `has_spec`). I extracted these in the experiment branch; both `can_write` and the repr's serializability warnings benefit. `can_write` is fundamentally an IO registry question ("does `write_elem` have a writer for each element?"), not a traversal question.

ilan-gold:
I like controlling descent but I don't know if introducing a class makes sense - it _feels_ like this should be accomplishable with the current crop of tools.

> I wonder if this use case would be better handled with a control flow API. See main review comment.

So what I don't get is what we save on "control flow" by using your gist, specifically https://gist.github.com/flying-sheep/b4274c148a08b7f037c732bec4a163a5#file-hierarchical-iter-py-L34-L35 looks not materially diffferent from the checks I ahve here.  Cleaner? Yes.  But implementers are still responsible for some form control flow either way (I assume) because, at least in your case, they would set `descend=True` when they have a "matching element."  

Is the point of what you're proposing giving a cleaner way of preventing descent i.e., making control flow easier in the case where there is _no_ descent needed? That I could see because you just mark each `obs` `var` `obsp` etc. element as `descend=False` somehow and then you _only_ see those parent elements.

> For packages that build on anndata like TreeData, this means reimplementing every one of these functions just to add .obst/.vart to the list. TreeData for example [rewrites the entire write traversal](https://github.com/YosefLab/treedata/blob/main/src/treedata/_core/write.py#L73-L90), plus __repr__, copy, subsetting, and anything else that iterates sections. A section registry would let extensions register once and be discovered everywhere, and can_write wouldn't silently skip extension sections.

@katosh This sounds like a use-case of what Phil is describing i.e., write methods use `descend=False` on all the "parent elements" so I think it is the correct way to go about things.  What I'm allergic to is introducing a new class but let's see.

ilan-gold:
1. `iter` (which should probably be renamed to `items` becuase it gives key-value pairs) is now reused and forms the backbone of `reduce`
2. `reduce` looks kind of silly now, not sure there's much sense in it...
3. I'm not sure either of these functions really needs to be public at the moment unless people come asking for them.  Even for @katosh I am not sure these need to be public.  This would save us from various painful questions about their APIs

katosh:
Thanks for the update @ilan-gold! Agreed on all three points.

`iter`/`items` as the backbone makes a lot of sense. The HTML repr's section loop would switch to it cleanly. Keeping things private for now also sounds right, the repr doesn't need a public API, just a canonical section list to iterate. Happy to adapt once you and @flying-sheep settle on the final shape.

flying-sheep:
> [`.iter()`] should probably be renamed to `items` because it gives key-value pairs

Python has `__iter__`, and `dict(thing)` expects `thing` to have either `keys` or `items` I think, so we I should probably make that work.

About the rest: yeah, if `reduce` doesn’t descend, it’s not much use I think. If it would descend into `AnnData`s in `.uns` it would make more sense, but that’d need us to define an `.acc` API for it.

ilan-gold:
> Python has __iter__, and dict(thing) expects thing to have either keys or items I think, so we I should probably make that work.

My thinking here went, more generally, we could come up with an informative name that maybe itsn't `items` or `iter` but that doesn't reserve `__iter__` for this specific use-case since `__iter__` could also (from a user perspective) be iteration through rows on an `AnnData` (like nd-array libraries).

ilan-gold:
So I've made `iter_outer` i.e., what was `AnnData.iter` a separate function that isn't exported and I also made `reduce` private.  So we are only exporting now `can_write`.  This should alleviate any concerns in this PR about APIs

docs/concatenation.rst
```diff
@@ -26,7 +26,6 @@ Let's start off with an example:
     AnnData object with n_obs × n_vars = 700 × 765
         obs: 'bulk_labels', 'n_genes', 'percent_mito', 'n_counts', 'S_score', 'G2M_score', 'phase', 'louvain'
         var: 'n_counts', 'means', 'dispersions', 'dispersions_norm', 'highly_variable'
-        uns: 'bulk_labels_colors', 'louvain', 'louvain_colors', 'neighbors', 'pca', 'rank_genes_groups'
```
flying-sheep, line 29: why is this gone?

src/anndata/_core/anndata.py
```diff
@@ -1417,6 +1412,86 @@ def copy(self, filename: PathLike[str] | str | None = None) -> AnnData:
...
+            if attr_name in (
+                "obs",
+                "obsm",
+                "varm",
+                "var",
+                "layers",
+                "varp",
+                "obsp",
+                "uns",
+            ) or isinstance(elem, pd.DataFrame | XDataset | MutableMapping):
```
flying-sheep: use sets for `in multiple_literals` checks.

flying-sheep (commented):
Some nice gains, great!

I wonder if a more granular `can_write` would be helpful for people that shows them which arrays/columns block the writeability.

docs/concatenation.rst
ilan-gold: It got moved to the end, absorbed by `obsp: ...`

ilan-gold:
> I wonder if a more granular can_write would be helpful for people that shows them which arrays/columns block the writeability.

That would be nice but feels downstream of "we might need better ref/acc objects to describe the _whole_ anndata" which I think we have realized in the course of this PR is too big of a lift ATM.

I think keeping the `reduce` paradigm around is a good step towards this and other goals that might involve per-array iteration.

docs/concatenation.rst
flying-sheep: hmm, maybe replace `obsp: ...` with `...` then to make that clear

flying-sheep (commented):
Looks great!

I’m ill, so I won’t check for myself, but is `can_write` complete? Like will it actually handle all non-bug cases where an anndata object fails to write, like a nested anndata object in uns that contains something unwriteable? Or some unhandled dtype in a categorical column or so?

If yes, I just have one ask: I think we should add a `note` or so to `can_write` that the return type isn’t guaranteed to stay the same, but the coercion to bool is stable: `bool(adata.can_write(...)` / `if adata.can_write(...):`. That way we can make it return a structure later that has info about unwriteable bits or so.

ilan-gold:
> If yes, I just have one ask: I think we should add a note or so to can_write that the return type isn’t guaranteed to stay the same, but the coercion to bool is stable: bool(adata.can_write(...) / if adata.can_write(...):. That way we can make it return a structure later that has info about unwriteable bits or so.

I wonder if this is too strong.  For example:


```python
from anndata.acc import A

assert bool(A.obs)
```

We would then need special classes whose `__bool__` is overriden?

flying-sheep:
> We would then need special classes whose `__bool__` is overriden?

yeah, the only useful alternative could be renaming it to `unwriteable` and having it return a list of unwriteable things in the future (a empty/falsy list if all is writeable)

do you have a different idea?

ilan-gold:
No and I that's even better given the current state of things.  You would want to use this in a guard clause anyway so

```python
if adata.unwriteable():
```
is nicer than
```python
if not adata.can_write()
```

flying-sheep (approved):
Yay! Thank you for your patience!

Issue #2373, opened 2026-03-22
Lenient write


flying-sheep:
### Please describe your wishes and possible alternatives to achieve the desired result.

We could add a `strict: bool = True` parameter to write.

When `False`, it would return a `dict[AdRef, ad.WriteError]` (or however we call that exception).

Wouldn’t be trivial, since we’d have to front-load some of the checks that currently happen while writing.

Issue #2241, opened 2025-12-08
i/o of different index types


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

Somewhat related to #777, but more broadly, `pandas` has a rich ecosystem of extension arrays that can be baked into indexes (see https://xarray-indexes.readthedocs.io/ for example).

Our current spec is ambiguous as to what a valid on-disk index is: https://anndata.readthedocs.io/en/latest/fileformat-prose.html:

```
The group MUST contain an array for the index
```

But is this an hdf5/zarr array or one of our serializable arrays?

Also related: #2238 would allow writing anything that is arrow-based

So there are a few TODOs here that probably have some overlap so it's not clear if they should be separate issues

- [ ] Serializing from #777 probably should be `None` which means we should **bump the spec** to allow for `None` in index
- [ ] It's possible that different non-standard/explicit indices (a genome or genome locations of some sort that isn't loaded into memory? cross-product index i.e., for pixels like from @maltekuehl's https://github.com/complextissue/spatiomic?) will want to make it explicit that they have different semantics but still serialize as `None`.  For example, a cross product could only store the min/max/step size per dimension.  We should create a specific namespace for this in `attrs` that is customizeable to allow for users to create their own custom non-standard/explicit indexing strategies.   The idea is that even if this user-defined metadata is unusable for some reason, #777 would step in as a fallback
- [ ] Clarify what is meant by "array" in a **spec bump** for the index of a dataframe (maybe one spec bump for all three points?).  This could allow for e.g., categorical indices more clearly, for example

Issue #777, opened 2022-05-31
Feature request - var_names/obs_names as fixed-sized types (integer or bytes)


ilibarra:
Following issue #35, using integers in obs_names/var_names is allowed, yet slicing the objects is not possible after that. Could solutions for this particular exceptional case be discussed/added to the codebase?

Broadly speaking, if this is solved, it would help in integer- and bit-based representation of biological sequences as _k_-mers, and would play a role not only in sequence-based analyses of genomics data but additionally proteomics, RNA biology, etc.

Others who are interested in this and could maybe join in the discussion about memory and implementation considerations are @gtca @olgabot. Please tag others.

Thank you.

ivirshup:
I'm all for this, but think it requires a new API for indexing.

How do we make sure the result of `adata[[1,2,3]]` is unambiguous if the labels are integer valued? `xarray` and `pandas` support this with indexing attributes: `df.iloc[...]` vs `df.loc[...]` for pandas and `da[...]` vs `da.loc[...]` for xarray.

The hard part here is backwards compatibility.

gtca:
We would also need this for single nucleotide resolution data. Tagging @mffrank here.

@ivirshup, maybe one way to do that would be to only allow a range index, this way e.g. `adata[np.array(0, 2)]` will mean getting the first and the third `obs` irrespective of whether you think those are names or indices.
As range indices were not possible before, that should be ok with backwards compatibility.

ivirshup:
@LucaMarconato, this is relevant for speeding up reading of points for FISH-like data

ivirshup:
My current thinking here is just to copy the API of xarray as much as possible.

The idea is to eventually move all label based indexing to a `.loc` attribute, and keep `[]` for positional.

So, we deprecate indexing with `[]` when it's label based. However, this would not yet allow label based indexing by integers, as any current calls to `[]` that used positional indexing but was passed an AnnData with integer indexes would have incorrect behavior.

The most important goal is to do this in a way that causes the fewest bugs possible. However, we would also like to quickly get to allowing integer based indexes.

### Solution 1: time

We could have a significant period of time where using labels when indexing with `[]` throws an error. This should give downstream packages enough time to update.

Then we start allowing integer indices, assuming all label based indexing code is using `.loc`.

### Solution 2: `iloc`

At the same time as `loc` we could introduce `iloc` for positional indexing. This would get us to integer indices much faster.

[...]

PR #2379, opened 2026-03-31
feat: custom in-memory indexes

ilan-gold:
From the proteomics hackathon, we definitely need to define some parameters and work out kinks:

1. Writeability: https://github.com/scverse/anndata/pull/2372 + https://github.com/scverse/anndata/issues/2373 + https://github.com/scverse/anndata/issues/2238.  Conretely, #2372 needs to go in first to prevent people from even trying to write these i.e., when there is a `MultiIndex`
2. What exactly are we changing? In other words, what does this setting do exactly? I think the two tests are clear about this, but there is some subtle stuff like you used to be able to do `adata.obs_names = non_string_index` and this evidently worked? This will be essential for documentation
3. Do we need any guardrails? No arrow integer types? What sort of filtering can/should we do here?
4. In the case of no index provided on declaration, should we make new decisions? https://github.com/scverse/pandas-uuid maybe?

- [x] Towards https://github.com/scverse/anndata/issues/2241 and https://github.com/scverse/anndata/issues/777
- [x] Tests added


ilan-gold:
Just looking for preliminary thoughts here @flying-sheep

src/anndata/_core/aligned_df.py
```diff
@@ -78,17 +79,18 @@ def _gen_dataframe_df(
...
 ):
-    if isinstance(anno.index, pd.MultiIndex):
-        msg = (
-            "pandas.MultiIndex not supported as index for obs or var on declaration.\n\
+    if isinstance(anno.index, pd.MultiIndex) and settings.restrict_index_types:
+        msg = "pandas.MultiIndex not supported as index for obs or var on declaration.\n\
             You can set `obs_names` manually although most operations after will error or convert to str.\n\
-            This behavior will likely be clarified in a future breaking release."
-        )
+            You can also opt out of `settings.restrict_index_types` which will allow pandas.MultiIndex."
```
flying-sheep: the way it is currently is that the second and third line are indented by whatever the code indent level is, which I guess isn’t intentional.

```suggestion
        msg = (
            "pandas.MultiIndex not supported as index for obs or var on declaration.\n"
            "You can set `obs_names` manually although most operations after will error or convert to str.\n"
            "You can also opt out of `settings.restrict_index_types` which will allow pandas.MultiIndex."
        )
```

src/anndata/_settings.py
```diff
@@ -527,6 +527,17 @@ def validate_sparse_settings(val: Any, settings: SettingsManager) -> None:
     get_from_env=check_and_get_bool,
 )
 
+settings.register(
+    "restrict_index_types",
+    default_value=True,
+    description=(
+        "Whether to force coercion to a string index upon declaration of the `AnnData` object or setting `obs`/`var`. "
+        "Setting this to `False` will also allow `MultiIndex` indexes upon declaration/setting."
```
flying-sheep: we should clarify that integer indices are always converted to string.

src/anndata/_core/index.py
```diff
@@ -207,13 +216,16 @@ def _from_array(
 
 def _normalize_index(
     indexer: Index1D[IndexManager], index: pd.Index
-) -> _Index1DNorm | int | np.integer | SupportsArrayApi:
+) -> _Index1DNorm | int | np.integer | SupportsArrayApi | pd.MultiIndex:
     # TODO: why is this here? All tests pass without it and it seems at the minimum not strict enough.
```
flying-sheep: maybe also a good opportunity to address this comment

tests/test_annot.py
```diff
@@ -73,24 +78,120 @@ def test_error_col_multiindex():
...
+    ):
+        assert_equal(
+            adata[pd.Index(pd.arrays.ArrowExtensionArray(arr))[0:2]],
+            adata[0:2],
+        )
+
+
+def test_arrow_index_write():
+    ad.settings.restrict_index_types = False
+    # See: https://github.com/pandas-dev/pandas/issues/64889 for why we can't just use dicts/lists
```
flying-sheep: use dicts/lists where?

src/anndata/_core/index.py
```diff
@@ -156,16 +159,22 @@ def _from_sequence(
...
     index: pd.Index,
 ) -> SupportsArrayApi:
     # convert to the 1D if it's accidentally 2D column/row vector
     # convert sparse into dense arrays if needed
     use_xp = has_xp(indexer)
-    is_pandas = isinstance(indexer, pd.api.extensions.ExtensionArray)
+    is_pandas = isinstance(
+        indexer,
+        pd.api.extensions.ExtensionArray | pd.MultiIndex,
+    )
```
flying-sheep, line 178: bit unclear why exactly these are handled here and not `pd.Index` … I assume because `pd.Index` is turned into an ExtensionArray before it lands here?

flying-sheep (commented):
direction looks great!

src/anndata/_core/index.py
ilan-gold: Yup! I'll add a comment pointing back at the function where this happens!

tests/test_annot.py
ilan-gold: Will add comment, but in the arrow extension array itself

src/anndata/_core/index.py
ilan-gold: This looks like old cruft from a time we didn't enforce string `index` objects (note that `index` here is the underlying index of the anndata axis, which we pre-emptively and aggressively convert to a string via `pandas_as_str` upon initialization)

ilan-gold:
> What exactly are we changing? In other words, what does this setting do exactly? I think the two tests are clear about this, but there is some subtle stuff like you used to be able to do adata.obs_names = non_string_index and this evidently worked? This will be essential for documentation

This is still an outstanding question to me but on second thought, nothing seems to have actually changed.  This was always allowed before, but maybe this PR should disallow it if the restriction setting is on? There's a warning about this but that's it. So maybe a test would be good

UPDATE: Ok I thought of that and https://github.com/scverse/anndata/pull/2379/changes#diff-93807bfddb872b9cf4ea7586b7ee8feb73bb9cab28969a82b4b6974001ff3462R749-R753 is what we do i.e., leave the old mechanism in place, which is probably the least disruptive thing to do

src/anndata/_settings.py
```diff
@@ -43,6 +43,12 @@ class Settings(
 
     auto_shard_zarr_v3: bool | None = True
     """Whether or not to use zarr's auto computation of sharding for v3.  For v2 this setting will be ignored. The setting will apply to all calls to anndata's writing mechanism (write_zarr / write_elem) and will **not** override any user-defined kwargs for shards."""
+    restrict_index_types: bool = True
```
flying-sheep, line 47: missing empty line

```suggestion

    restrict_index_types: bool = True
```

docs/release-notes/2379.feat.md
```diff
@@ -0,0 +1,4 @@
+Create a new setting {attr}`anndata.settings.restrict_index_types` to allow for custom in-memory indexes (i.e., for {attr}`~anndata.AnnData.obs` and {attr}`~anndata.AnnData.var`) to be passed to an {class}`~anndata.AnnData` objects without being converted to a string type.
```
ilan-gold:
```suggestion
Create a new setting {attr}`anndata.settings.restrict_index_types` to allow for custom in-memory indexes (i.e., for {attr}`~anndata.AnnData.obs` and {attr}`~anndata.AnnData.var`) to be passed to an {class}`~anndata.AnnData` object without being converted to a string type.
```

flying-sheep (approved):
Awesome! Very nice to see so few changes necessary, we really improved this codebase!

/edit: GitHub however did not, I had to delete and recreate the comment below, since GitHub inexplicably moved it to a different line, breaking the diff.

src/anndata/_settings.py
```diff
@@ -44,5 +44,12 @@ class Settings(
     auto_shard_zarr_v3: bool | None = True
     """Whether or not to use zarr's auto computation of sharding for v3.  For v2 this setting will be ignored. The setting will apply to all calls to anndata's writing mechanism (write_zarr / write_elem) and will **not** override any user-defined kwargs for shards."""
 
+    restrict_index_types: bool = True
+    """
+    Whether to force coercion to a string index upon declaration of the `AnnData` object or setting `obs`/`var`.
+    "Setting this to `False` will also allow `MultiIndex` indexes upon declaration/setting.
+    "Integer indices i.e., those caught by :func:`pandas.api.types.is_integer_dtype` will still be converted to strings.
```
flying-sheep:
```suggestion
    "Setting this to `False` will e.g. also allow `MultiIndex` indexes upon declaration/setting.
    "Only integer indices i.e., those caught by :func:`pandas.api.types.is_integer_dtype` will always be converted to strings.
```

PR #2387, opened 2026-04-14
chore: use `scverse-misc`

ilan-gold:
- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: dev change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ dependencies = [
     "legacy-api-wrap",
     "zarr >=3.1",
     "typing-extensions; python_version<'3.13'",
+    "scverse-misc>=0.0.3",
 ]
 dynamic = [ "version" ]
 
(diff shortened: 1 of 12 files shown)


flying-sheep (approved):
Awesome!

We do have to take to keep the mission of `scverse-misc` focused though, it would be unfortunate if it sprawled to become an unmaintainable mess.

> Miscellaneous utility code used by scverse packages

is a bit too broad.

I’ll file an issue!

flying-sheep (changes requested):
/edit: I thought it emitted `DeprecationWarning`s, but it actually does emit `FutureWarning`s as we want it to.

flying-sheep (approved):

ilan-gold:
> We do have to take to keep the mission of scverse-misc focused though, it would be unfortunate if it sprawled to become an unmaintainable mess.

Agreed I already asked that some `pandas` code be removed.  A smell-test is that it shouldn't have dependencies like `pandas` unless there is an amazing reason.  Super light, focused, python.

PR #2390, opened 2026-04-15
Backport PR #2387: chore: use `scverse-misc`

ilan-gold:
- [ ] Closes #
- [ ] Tests added


ilan-gold:
@flying-sheep The docs are failing because `version-deprecated`, which is used by `scverse-misc` as the directive, is `sphinx>=9` specific: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-version-deprecated

Do you think we should change `scverse-misc`? Probably and make it dynamic based on detected sphinx version? cc @ilia-kats

flying-sheep:
I think we just forgot to backport #2298, so let’s do that.

I don’t think there’s a use case for building the docs with an old environment.

PR #2395, opened 2026-04-20
perf: fast path for unchunked minor axis `dask` sparse reindexing when `concat`ing

ilan-gold:
Someone tried annbatch shuffling-ing the whole cellxgene census which raised this bug where our outer-join, while correct, was effectively a densification operation for CSR matrices.  Why? The previous iteration basically did this for every block

```python
import scipy.sparse as sp

mat = sp.random(10, 20, format="csr")
mat[0, 15] = 0
assert 0 in mat.data
```
which means 0's are actually stored.  After fixing this, it took sub-7 hours, which is basically what we'd expect given our benchmark in our paper. The speedup / memory gains are confirmed on the benchmark (see below for results).

- [ ] Closes #
- [ ] Tests added

### DIFF
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -595,7 +595,12 @@ class Reindexer:
             and is_outer
         ):
             return el.map_blocks(
-                partial(self._apply_to_sparse, axis=axis, fill_value=fill_value),
+                partial(
+                    self._apply_to_sparse,
+                    axis=axis,
+                    fill_value=fill_value,
+                    keep_format=True,
+                ),
                 chunks=(el.chunks[0], len(self.new_idx))
                 if minor_axis == 1
                 else (len(self.new_idx), el.chunks[1]),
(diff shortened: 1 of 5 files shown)


src/anndata/_core/merge.py
```diff
@@ -583,6 +583,22 @@ def _apply_to_df_like(self, el: pd.DataFrame | Dataset2D, *, axis, fill_value=No
     def _apply_to_dask_array(self, el: DaskArray, *, axis, fill_value=None):
         import dask.array as da
 
+        indexer = self.idx
+        is_outer = any(indexer == -1)
+        # Fast path for the majority of sparse matrices whose minor-axis is unchunked and is being reindexed.
+        # This prevents 0's from being stored explicitly in the sparse matrices when outer joining, for example (see below).
+        if (
+            is_sparse_sub := isinstance(el._meta, CSArray | CSMatrix)
+            and el._meta.format == "csr"
```
flying-sheep: this could be `CSRBase` or `CSRArray | CSRMatrix`, no?

otherwise you could generalize this to work on both CSR+ax1 and CSC+ax0

flying-sheep (approved):
Looks great! One nitpick, otherwise let’s go!

src/anndata/_core/merge.py
ilan-gold: Ah yes, as you can see from the name changes to this PR, I lost track of this a bit when I got confused about something!

PR #2406, opened 2026-04-22
fix: disallow `copy` for "backed" objects

ilan-gold:
You could argue that taking a `filename` keyword should trigger a copy even in this case, but the problem is how we "reconstitute the object." We would have to track _which_ elements were "backed" and reopen them identically, non-trivial.  Furthermore, this function wraps such trivial functionality for `backed` mode (as it exists now), that I would argue we shouldn't be encoding this behavior anyway.

In any case, this is a bug since you can't copy zarr/h5py objects.

- [x] Continuing with #2369
- [x] Tests added

### DIFF
--- a/docs/release-notes/2406.fix.md
+++ b/docs/release-notes/2406.fix.md
@@ -1,1 +1,1 @@
-Fix {meth}`anndata.AnnData.copy` so that it provides an informative error when trying to `copy` and object that contains {class}`h5py.Dataset`, {class}`zarr.Array` ,{class}`anndata.abc.CSRDataset`, or {class}`anndata.abc.CSCDataset` {user}`ilan-gold`
+Fix {meth}`anndata.AnnData.copy` so that it provides an informative error when trying to `copy` and object that contains {class}`h5py.Dataset`, {class}`zarr.Array`, {class}`anndata.abc.CSRDataset`, or {class}`anndata.abc.CSCDataset` {user}`ilan-gold`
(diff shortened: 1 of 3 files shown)


docs/release-notes/2406.fix.md
```diff
@@ -0,0 +1 @@
+Fix {meth}`anndata.AnnData.copy` so that it provides an informative error when trying to `copy` and object that contains {class}`h5py.Dataset`, {class}`zarr.Array` ,{class}`anndata.abc.CSRDataset`, or {class}`anndata.abc.CSCDataset` {user}`ilan-gold`
```
flying-sheep:
```suggestion
Fix {meth}`anndata.AnnData.copy` so that it provides an informative error when trying to `copy` and object that contains {class}`h5py.Dataset`, {class}`zarr.Array`, {class}`anndata.abc.CSRDataset`, or {class}`anndata.abc.CSCDataset` {user}`ilan-gold`
```

flying-sheep (approved):
Makes sense! I guess we could technically copy such an anndata object if all arrays are part of the same store/file, but that would be a whole project, and IDK if it’s worth spending time on that if “copy the file/directory, then reopen” is a two line workaround.

ilan-gold:
> Makes sense! I guess we could technically copy such an anndata object if all arrays are part of the same store/file, but that would be a whole project, and IDK if it’s worth spending time on that if “copy the file/directory, then reopen” is a two line workaround.

Exactly was my thinking! You're right - the arrays could even come from different places

PR #2407, opened 2026-04-22
refactor: allow multi-indexed backed mode

ilan-gold:
This one I don't fully get.  All of our views are lazy so why would multiple indexing be bad? I even added a test to check specifically that we're not doing "too many" disk accesses if we index multiple times but don't "access" `X`.  Am I missing something? @ivirshup perhaps? Maybe this was for writing back to the store?

Does this need more tests?

- [x] For #2369 part 1
- [x] Tests added

- [x] Release note not necessary because: I don't think this needs a note given that the operation works on "normal" non-backed zarr/hdf5 objects (see the altered test in `test_backed_sparse.py` for evidence of this fact)

### DIFF
--- a/src/anndata/_core/anndata.py
+++ b/src/anndata/_core/anndata.py
@@ -283,12 +283,6 @@ class AnnData:  # noqa: PLW1641
         oidx: _Index1DNorm | int | np.integer,
         vidx: _Index1DNorm | int | np.integer,
     ):
-        if adata_ref.isbacked and adata_ref.is_view:
-            msg = (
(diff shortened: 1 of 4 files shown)


flying-sheep (commented):
That was introduced in 2b622f4518d670c3cddd3a861a1a718d13575c15

Maybe back then it didn’t work due to some implementation detail, but a more recent change made it work?

We have `test_assign_x_subset` that tests that writethrough works for views of backed adata. Maybe we should just test that writethrough works for views of views and call it a day?

ilan-gold:
Works!

flying-sheep (approved):

PR #2413, opened 2026-04-28
chore: add zizmor

ilan-gold:
- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: dev


flying-sheep:
https://github.com/dependabot/dependabot-core/pull/14349

.github/workflows/test-gpu.yml
```diff
@@ -42,18 +46,21 @@ jobs:
     # Setting a timeout of 30 minutes, as the AWS costs money
     # At time of writing, a typical run takes about 5 minutes
     timeout-minutes: 30
+    permissions:
+      pull-requests: write
 
     steps:
-      - uses: actions/checkout@v4 # TODO: upgrade once cirun image supports node 24
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
flying-sheep, line 53: Since we rely on dependabot more, I think we need the up-to-date AMI for GPU tests, and then update all the actions here to the newest version

.github/dependabot.yml
```diff
@@ -7,3 +7,5 @@ updates:
     labels:
       - no milestone
       - skip-gpu-ci
+    cooldown:
+      default-days: 7
```
flying-sheep, line 11: what’s that for? what does it do that `schedule: { interval: weekly }` doesn’t?

ilan-gold: https://docs.zizmor.sh/audits/#dependabot-cooldown I think it prevents pulling in a fresh, potentially poisoned, version.

.github/workflows/test-gpu.yml
ilan-gold: tried my hand at it: https://github.com/scverse/gpu-ci-config/issues/3

flying-sheep: OK, I hope that’ll bear fruit soon.

The update removed the existing comments, please add them back until we can actually update

.github/workflows/codespell.yml
flying-sheep, line 1: why does this job have the `contents: read` permission? others don’t need it for `actions/checkout`.

.github/workflows/test-cpu.yml
```diff
@@ -99,14 +102,17 @@ jobs:
 
   build:
     runs-on: ubuntu-24.04
+    permissions:
+      id-token: write
```
flying-sheep: Where does this come from? It wasn’t there before.

Also all the other `id-token: write` permissions have a comment explaining which step needs it.

.github/workflows/test-gpu.yml
```diff
@@ -30,8 +32,10 @@ jobs:
     name: "Triage: Check if GPU tests are allowed to run"
     if: ${{ !(github.event.pull_request.user.login == 'pre-commit-ci[bot]' || contains(github.event.pull_request.labels.*.name, 'skip-gpu-ci')) }}
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
```
flying-sheep: I doubt that this is necessary, we’re just accessing `github.event.pull_request` and not doing any API calls

.github/workflows/test-gpu.yml
```diff
@@ -42,18 +45,22 @@ jobs:
...
           fetch-depth: 0
           filter: blob:none
+          persist-credentials: false
 
       - name: Nvidia SMI sanity check
         run: nvidia-smi
 
       - name: Install UV
-        uses: astral-sh/setup-uv@v6 # TODO: upgrade once cirun image supports node 24
+        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0
```
ilan-gold, line 64:
```suggestion
        # TODO: upgrade once cirun image supports node 24
        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0
```

ilan-gold:
cc @Intron7 could you have a look at the failing GPU tests

flying-sheep (approved):

PR #2416, opened 2026-05-03
feat: `aligned_axis_key_join` for on-axis key joining

Ekin-Kahraman:
Closes #2374. Adds `aligned_axis_key_join` to `concat()` for controlling on-axis annotation-key joining (obs columns, obsm/obsp keys, symmetric over var, and layers keys) independently of the off-axis index `join`.

## API

- New parameter: `aligned_axis_key_join: Literal["inner", "outer"] | None = None`
- Default `None` falls back to `join`, preserving existing behaviour
- Validates against `"inner"`, `"outer"`, or `None` only

## Scope

- ✅ obs columns
- ✅ obsm keys (the on-axis aligned mapping)
- ✅ obsp keys (pairwise mapping)
- ✅ Symmetric var/varm/varp when `axis="var"`
- ✅ Layers keys (off-axis content alignment of each kept layer still follows `join`)
- ✅ Forwards to `raw` recursive concat
- ⏭ `concat_on_disk` skipped (per #2374 thread)

## Implementation

The existing `inner_concat_aligned_mapping` and `outer_concat_aligned_mapping` helpers now take a `keys=` parameter (defaulting to `intersect_keys` / `union_keys` respectively, so existing callers are unchanged). `inner_concat_aligned_mapping` additionally takes `fill_value` and handles the missing-key case for the outer-key + inner-content combination, honouring caller-provided `reindexers` so layers stay aligned to X's alt-axis.

## Tests (15 under `aligned_axis_key_join` block)

- Default fallback to `join` for both axes (parametrised obs/var)
- Outer key join with inner content join, and reverse
- Layers follow `aligned_axis_key_join` on both axes (keys union under outer + inner content; keys intersect under inner + outer content), with shape and content spot-checks
- Alt-axis mappings (varm) unaffected when concatenating along obs across all `merge` strategies
- obsp pairwise paths
- DataFrame-content obsm following `join`
- Inner content with missing keys (DataFrame + ndarray branches)
- Invalid value raises `ValueError`
- Awkward arrays + missing-key + inner-content raises `NotImplementedError` with a clear message
- Direct reproduction of the #2374 example


Ekin-Kahraman:
Quick triage note for the failing checks:

- **Triage labels / GPU**: maintainer-gated, can't self-add (`gh pr edit --add-label` returns no permission)
- **`test (hatch-test.stable, 3.14, --strict-warnings, not zarr_io)`**: failure reproduces locally when running this test against upstream `main`'s `merge.py` (no PR changes), so it appears pre-existing rather than introduced here. Other strict-warnings matrices look like fail-fast cancellations off the same root.
- **`benchmark (3.12, ubuntu-latest)`**: exits 2 after a flood of `SparseEfficiencyWarning` from the existing `concat` benchmark; same concern about whether it's PR-introduced. Happy to dig if you suspect it is.
- **`codecov/patch`**: `codecov/project` is passing, so overall coverage held; patch fail is a soft signal.

Happy to add more coverage or address anything once you've had a look.

ilan-gold (commented):
Two things that stick out to me about this:

1. I think I didn't fully consider things - I suppose `layers` are aligned indeed (along both axes actually) and would benefit from control over key-based joins.  So they should probably be handled here as well identically. In other words, `aligned_join == "inner"` does an inner join of the keys using `inner_concat_aligned_mapping` and vice-versa for `outer`.
2. With or without the above, why can't we just reuse the current functions that are renamed to `concat_aligned_mapping` for inner/outer joining?

Ekin-Kahraman:
Both points addressed, main merged in. Helpers reused — `inner`/`outer` aligned-mapping helpers take a `keys=` argument now and the split helper is gone. Layers also respect `aligned_axis_key_join`; the only subtlety was making the missing-key inner path honour the existing reindexers so layers stay aligned to X.

Ekin-Kahraman:
Rebased on main to pick up #1707. Two small conflicts in `merge.py`: kept your new `Default("inner")` signature alongside `aligned_axis_key_join`, and dropped the old `X = concat_Xs(...)` line since X lives in layers now.

Two of my layer-key tests broke because `sorted()` doesn't like `None` and strings mixed. Switched them to set comparison and included the `None` (X) key in the expected sets.

CI green at `dac79dd7`. Ready for another look.

src/anndata/_core/merge.py
```diff
@@ -1642,6 +1778,16 @@ def concat(  # noqa: PLR0912, PLR0913, PLR0915
     merge = resolve_merge_strategy(merge)
     uns_merge = resolve_merge_strategy(uns_merge)
 
+    if aligned_axis_key_join is not None and aligned_axis_key_join not in (
```
ilan-gold: We have a `JoinT: Literal` you can use to get the values using `typing.get_args`

src/anndata/_core/merge.py
```diff
@@ -1750,33 +1906,64 @@ def concat(  # noqa: PLR0912, PLR0913, PLR0915
...
+        raise AssertionError(msg)
+
+    # Layers: `aligned_join` selects which layer keys appear; `reindexers`
+    # carries the off-axis alignment from X so each kept layer is aligned to
+    # the same alt-axis as X.
+    layer_mappings = [a.layers for a in adatas]
     layers = concat_aligned_mapping(
-        [a.layers for a in adatas], axis=axis, reindexers=reindexers
+        layer_mappings,
+        axis=axis,
```
ilan-gold, line 1928:
```suggestion 
layers = concat_aligned_mapping(
        [a.layers for a in adatas],
        axis=axis,
```

src/anndata/_core/merge.py
```diff
@@ -1090,13 +1194,23 @@ def outer_concat_aligned_mapping(
     concat_axis=None,
     fill_value=None,
     force_lazy: bool = False,
+    keys=None,
```
ilan-gold: Instead of adding another kwarg (which can have unintended side effects if the caller doesn't know the behavior), let's make this argument required by the caller so things are explicit and all the "default" logic is lifted up into one place

Ekin-Kahraman:
Thanks, addressed these review points:

- validation now uses the existing `Join_T` literal values via `JOIN_OPTIONS`
- `inner_concat_aligned_mapping` / `outer_concat_aligned_mapping` now require an explicit `keys=` argument, so the default key-selection logic is lifted to the call sites
- simplified the layers/obsm call site and removed the extra explanatory comments
- updated the `AnnCollection` obsm call site to pass `intersect_keys(...)` explicitly after making `keys` required

Local checks:

```bash
python -m pytest tests/test_concatenate.py -k aligned_axis_key_join
# 15 passed

python -m pytest tests/test_anncollection.py
# 5 passed

python -m ruff check src/anndata/_core/merge.py src/anndata/experimental/multi_files/_anncollection.py
# passed

python -m ruff format --check src/anndata/_core/merge.py src/anndata/experimental/multi_files/_anncollection.py
# passed
```

The remaining red checks on the PR appear to be maintainer-controlled triage gates: labels/milestone and GPU permission.

tests/test_concatenate.py
```diff
@@ -1891,3 +1891,422 @@ def test_1d_concat():
...
+    """`aligned_axis_key_join="inner"` intersects on-axis keys while leaving off-axis as outer."""
+    a, b = _adatas_with_partial_overlap_along_axis(axis_name)
+    axis = 0 if axis_name == "obs" else 1
+
+    res = concat([a, b], axis=axis, join="outer", aligned_axis_key_join="inner")
+    cols = list(getattr(res, axis_name).columns)
+    keys = list(getattr(res, f"{axis_name}m").keys())
+    assert cols == ["shared"]
+    assert keys == ["shared_m"]
+
```
ilan-gold, line 1990: These tests are basically identical, please dedup using parametrization

tests/test_concatenate.py
```diff
@@ -1891,3 +1891,422 @@ def test_1d_concat():
...
+    # New behaviour: outer on-axis columns, inner off-axis index.
+    r = concat(adatas, join="inner", aligned_axis_key_join="outer")
+    assert sorted(r.obs.columns) == ["1", "2", "3"]
+    assert list(r.var_names) == ["1"]
+
+    # Converse: inner on-axis columns, outer off-axis index.
+    r = concat(adatas, join="outer", aligned_axis_key_join="inner")
+    assert list(r.obs.columns) == ["1"]
+    assert sorted(r.var_names) == ["1", "2", "3"]
```
ilan-gold, line 2312: These should just be parametrizations

ilan-gold (commented):
Sorry for losing track of this - I look at my "requested review" very religiously, so feel free to smash that button when you're ready in the future.

PR #2418, opened 2026-05-05
fix: `unwriteable` was still `can_write` in all but name

ilan-gold:
When I made the change, I just `ctrl` + F'ed this, but forgot that we inverted the semantics from `can_write`. We also were never checking the index, which I discovered in #2379 

- [ ] Closes #
- [x] Tests added

- [x] Release note not necessary because: Unreleased feature fix


flying-sheep (approved):
Looks great. Sorry for not catching the missing logic inversion in the original PR.

PR #2368, opened 2026-03-18
feat: upgrade to zarr v3 sharding default

ilan-gold:
I am marking this 0.12.11 to force myself to backport this with only a warning about the upcoming change. For zarr v2, the auto sharding setting's docs had:

```
For v2 this setting will be ignored
```
always but we were not ignoring it, hence removing https://github.com/scverse/anndata/pull/2368/changes#diff-fc69accdad779387bbaf2b952adcff99c96d09fd6416011b319b6887d5a5498eL459-L461

- [x] Closes #2239
- [x] Tests added

- [ ] Release note not necessary because:

PR #2427, opened 2026-05-07
fix: warn for upcoming zarr autosharding + fix settings dependency

ilan-gold:
This PR also fixes the fact that `autoshard_zarr_v3` was dependent on the zarr version, which it should never have been as documented: https://anndata.readthedocs.io/en/stable/generated/anndata.settings.html#anndata.settings.auto_shard_zarr_v3

- [x] The 0.12.x verison of #2368 
- [ ] Tests added


tests/test_io_utils.py
```diff
@@ -97,6 +97,7 @@ class Foo:
         pass
 
     group = group_fn(tmp_path)
+    ad.settings.auto_shard_zarr_v3 = True
```
ilan-gold: Maybe this should be an `autouse` fixture actually?

flying-sheep: Aren’t you effectively basically doing that via `diskfmt` (which is used in many io tests)

https://github.com/scverse/anndata/blob/94336923af9410b4985115325d873facdfb36a11/tests/conftest.py#L46-L48

but yeah, looks a bit haphazard how it’s in some tests without clearly saying why.

maybe just autouse it and add a test that explicitly disables it?

ilan-gold: Yup that's exactly it.  I had the same thought as you about "effecitvely basically ..." but then before requesting your review had the same thought.

src/anndata/_io/specs/methods.py
```diff
@@ -140,6 +150,25 @@ def wrapper(
     return wrapper
 
 
+def suppress_autoshard_warning(func):
```
flying-sheep: that should take and return a `_WriteInternal[T]` right?

tests/test_io_elementwise.py
```diff
@@ -913,38 +913,25 @@ def test_h5_unchunked(
...
-    ],
-    ids=["shard_first", "write_format_first"],
-)
-def test_write_auto_sharded(tmp_path: Path, override: dict):
+def test_write_auto_sharded(tmp_path: Path):
     path = tmp_path / "check.zarr"
-    adata = gen_adata((1000, 100), **GEN_ADATA_NO_XARRAY_ARGS)
-    with ad.settings.override(**override):
+    adata = gen_adata((100, 10), **GEN_ADATA_NO_XARRAY_ARGS)
+    with ad.settings.override(auto_shard_zarr_v3=True, zarr_write_format=3):
```
flying-sheep, line 919: why here again?

ilan-gold: I wanted to have an explicit test not dependnt on the global settings

src/anndata/_io/specs/registry.py
```diff
@@ -114,7 +114,7 @@ def __init__(self) -> None:
 
     def register_write[S: StorageType, T: RWAble](
         self,
-        dest_type: type,
+        dest_type: type[S],
```
flying-sheep, line 117: ha, I missed that one!

src/anndata/_io/specs/methods.py
```diff
@@ -112,10 +118,32 @@ def zarr_v3_compressor_compat(dataset_kwargs: dict) -> dict:
...
+        )
+    elif auto_sharding:
         dataset_kwargs = {**dataset_kwargs, "shards": "auto"}
-    return dataset_kwargs
+    # Auto shard sizes are a relatively recent feature
+    supports_auto_shard_size = Version(version("zarr")) >= Version("3.1.4")
+    has_auto_shard_size = supports_auto_shard_size and isinstance(
+        zarr.config.get("array.target_shard_size_bytes"), int
+    )
+    # 1GB uncompressed shard size seems reasonable.
```
flying-sheep, line 140: I think you should elaborate here, magic numbers need good documentation.

Also please use underscores to format long ints like that (`…_000_000`)

src/anndata/_io/specs/methods.py
```diff
@@ -441,13 +501,15 @@ def write_basic(
...
-            f[k][...] = elem
+        with zarr_v3_sharding(
+            dataset_kwargs, format=f.metadata.zarr_format
+        ) as dataset_kwargs:
+            f.create_array(k, shape=elem.shape, dtype=dtype, **dataset_kwargs)
+            # see https://github.com/zarr-developers/zarr-python/discussions/2712
+            if isinstance(elem, ZarrArray | H5Array):
+                f[k][...] = elem[...]
+            else:
+                f[k][...] = elem
```
flying-sheep, line 516: why is this “actually write the data” block in the `with zarr_v3_sharding` block while the one in `write_vlen_string_array_zarr` isn’t?

tests/test_io_elementwise.py
```diff
@@ -913,38 +913,39 @@ def test_h5_unchunked(
...
-        with ad.settings.override(zarr_write_format=3):
-            with ad.settings.override(auto_shard_zarr_v3=True):
-                with ad.settings.override(zarr_write_format=2):
-                    pass
+def test_write_auto_sharded_default_warns(tmp_path: Path):
+    path = tmp_path / "check.zarr"
+    adata = gen_adata((100, 10), **GEN_ADATA_NO_XARRAY_ARGS)
+    ad.settings.reset("auto_shard_zarr_v3")
+    with (
+        ad.settings.override(zarr_write_format=3),
```
flying-sheep, line 945: why use a context manager for one modification and none for the other?

if the answer is “I’d use `ad.settings.reset(…)` as a context manager if it was one”, we should implement that!

----

PS: what was up with the weird `with ad.settings.override(zarr_write_format=3): with ad.settings.override(auto_shard_zarr_v3=True): …` stack? why not `with ad.settings.override(zarr_write_format=3, auto_shard_zarr_v3=True, …):`?

flying-sheep (commented):
Nice implementation!

One issue: I don’t understand when the sharding applies, just when creating an array or (also?) when writing data. Please clear that up in the code by either
- having the `g[...] = data` lines consistently inside or outside the `with zarr_v3_sharding` blocks, or
- adding a comment why there’s a difference


----

Also the commented-out code in `write_recarray_zarr` still uses the old pattern, please change that too.

ilan-gold:
> One issue: I don’t understand when the sharding applies, just when creating an array or (also?) when writing data. Please clear that up in the code by either

Yup, only applies at creation time

tests/test_io_elementwise.py
ilan-gold: > why use a context manager for one modification and none for the other?


I could have done `auto_shard_zarr_v3=None` to get the default but like to be extra explicit about these things. I don't really see the use-case for a context manager but am open to it!

> PS: what was up with the weird with ad.settings.override(zarr_write_format=3): with  ad.settings.override(auto_shard_zarr_v3=True): … stack? why not with ad.settings.override(zarr_write_format=3, auto_shard_zarr_v3=True, …):?

Where is this?

flying-sheep: I started implementing the context manager here: https://github.com/scverse/scverse-misc/pull/31

> Where is this?

deleted code, I’m just confused why it looked like this.

flying-sheep (approved):

tests/test_io_elementwise.py
ilan-gold: Ah it was for the dependency of the format on the shard setting, which was removed (and should have never been there in the first place)

Issue #2430, opened 2026-05-08
`layers` can be 3D on-disk


ilan-gold:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of anndata.
- [x] (optional) I have confirmed this bug exists on the main branch of anndata.

### Report

Code:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "anndata@git+https://github.com/scverse/anndata.git",
# ]
# ///

import anndata as ad
import numpy as np
adata = ad.AnnData(layers={"foo": np.arange(27).reshape((3, 3, 3))})
adata.write_h5ad("foo.h5ad")
```

Traceback:

No traceback but this is a violation of our spec: https://anndata.readthedocs.io/en/latest/fileformat-prose.html#anndata-specification-v0-1-0

cc @Zethson @eroell This is a pretty big deal that you guys are now relying on this.  When I heard that you were doing this, I assumed implicitly that it was checked and kosher, but @eroell approached me about something which prompted me to check this.

### Versions

anndata 0.12.14.dev1+g5d5858d89.d20260508
numpy   2.4.4
----    ----
xarray  2026.2.0
packaging       26.0
h5py    3.16.0
cloudpickle     3.1.2
pytz    2026.2
MarkupSafe      3.0.3
charset-normalizer      3.4.4
google-crc32c   1.8.0
pandas  2.3.3
PyYAML  6.0.3
legacy-api-wrap 1.5
python-dateutil 2.9.0.post0
pyarrow 23.0.1
donfig  0.8.1.post1
scipy   1.17.1
scverse-misc    0.0.3
msgpack 1.1.2
typing_extensions       4.15.0
psutil  7.2.2
dask    2026.1.2
tblib   3.2.2
numcodecs       0.16.5
session-info2   0.4
natsort 8.4.0
threadpoolctl   3.6.0
scanpy  1.12
awkward 2.9.0
fast-array-utils        1.3.1
Jinja2  3.1.6
awkward_cpp     52
six     1.17.0
toolz   1.1.0
fsspec  2026.2.0
zarr    3.1.6
----    ----
Python  3.14.3 (main, Feb  3 2026, 15:32:20) [Clang 17.0.0 (clang-1700.6.3.2)]
OS      macOS-15.1-arm64-arm-64bit-Mach-O
Updated 2026-05-08 12:45

Zethson:
[...]

PR #2432, opened 2026-05-10
fix: X shape check slices to first 2 shape values

eroell:
- [x] ~~Closes~~ One step towards #2430
- [x] Tests added

**Before**

higher-than-2D `X` was rejected at construction; you could not hold a higher-dimensional `X` in memory:

```python
import numpy as np
import anndata as ad
ad.AnnData(X=np.zeros((3, 4, 5)))
# ValueError: too many values to unpack (expected 2)
#   (from `n_obs, n_vars = X.shape`)

adata = ad.AnnData(X=np.zeros((3, 4)))
adata.X = np.zeros((3, 4, 5))
# FutureWarning: Automatic reshaping when setting X will be removed ...
# -> then ValueError, since (3, 4, 5) can't be reshaped to (3, 4)
```

higher-than-2D `layers` slipped through the in-memory shape check (only axes 0 and 1 are validated) and the writer happily wrote them to disk, silently violating the spec.

**After**

_In-memory_: higher-than-2D `X` and `layers` warns but still succeeds (the shape check uses `X.shape[:2]`).
_Writing_ a non-2D `X` / layer now hard-fails.
_Reading_ a non-conforming file warns but still succeeds:

```python
adata = ad.AnnData(X=np.zeros((3, 4, 5)))   # now OK in memory
# UserWarning: X must be 2-dimensional, but got an array with shape (3, 4, 5) (ndim=3). Storing higher-dimensional arrays in `X` or `layers` violates the AnnData specification, and cannot be written to disk.

adata.X = np.zeros((3, 4, 5))               # also OK
# UserWarning: X must be 2-dimensional, but got an array with shape (3, 4, 5) (ndim=3). Storing higher-dimensional arrays in `X` or `layers` violates the AnnData specification, and cannot be written to disk.

adata.layers["L"] = np.zeros((3, 4, 5))     # also OK
# UserWarning: Layer 'L' must be 2-dimensional, but got an array with shape (3, 4, 5) (ndim=3). Storing higher-dimensional arrays in `X` or `layers` violates the AnnData specification, and cannot be written to disk.

adata.write_h5ad("out.h5ad")
# ValueError: X must be 2-dimensional, but got an array with shape (3, 4, 5) (ndim=3). Storing higher-dimensional arrays in `X` or `layers` violates the AnnData specification, and cannot be written to disk.

ad.read_h5ad("legacy_non_conforming.h5ad")
# UserWarning: X must be 2-dimensional, but got an array with shape ...
# -> still returns the AnnData
```

Same applies to `layers["L"]` (error/warning message says `Layer 'L'` instead of `X`), and to the `zarr` IO path.


src/anndata/_core/storage.py
```diff
@@ -22,6 +22,66 @@
...
+    AnnData's on-disk specification requires ``X`` and every entry of
+    ``layers`` to be two-dimensional. Returns ``None`` for conforming values
+    (no ``shape`` attribute, ``None``, or ``ndim <= 2``).
+    """
+    if value is None:
+        return None
+    shape = getattr(value, "shape", None)
+    if shape is None:
+        return None
+    ndim = len(shape)
```
eroell, line 40: this also means that an array of shape `(m, n, 1)` can't be written to disk anymore, before this was ok.

Do you want to allow this or not?

eroell:
Seems I can't request reviews, but I'd be interested in your comments at this stage :)

src/anndata/_io/specs/methods.py
```diff
@@ -381,6 +386,10 @@ def read_anndata(elem: _GroupStorageType | H5File, *, _reader: Reader) -> AnnDat
     ]:
         if k in elem:
             d[k] = _reader.read_elem(elem[k])
+    # Older / non-conforming files may contain higher-dimensional `X` or
+    # `layers`. The on-disk spec forbids that; surface it as a warning so
+    # the user knows, but still construct the AnnData with what's there.
+    _warn_if_x_or_layers_3d_kwargs(d)
```
ilan-gold: I would move this warning into the construction of the `AnnData` object instead of reading (also in the other spots) so that people who do this in-memory will know their data is technically unwritable

src/anndata/_core/anndata.py
```diff
@@ -603,13 +603,13 @@ def X(self, value: _XDataType | None) -> None:
             msg = "The ability to set X with a scalar value will be removed in the future.  Initializing as an `np.array` with the shape of the current view."
             warn(msg, FutureWarning)
             value = np.full(self.shape, fill_value=value)
-        if hasattr(value, "shape") and value.shape != self.shape:
+        if hasattr(value, "shape") and value.shape[:2] != self.shape:
             msg = "Automatic reshaping when setting X will be removed in the future."
             warn(msg, FutureWarning)
             value = value.reshape(self.shape)
```
flying-sheep, line 625: I feel like the error thrown by `value.reshape(self.shape)` is probably ugly/misleading `if value.ndim > self.ndim`, no? Please check and if the error is confusing, and manually throw a clearer one in that case.

tests/test_x_layers_2d.py
```diff
@@ -0,0 +1,190 @@
...
+WhichAttr = Literal["X", "layers"]
+
+DISK_FORMATS = [
+    pytest.param("h5ad", id="h5ad"),
+    pytest.param("zarr", id="zarr"),
+]
+WHICH_ATTRS = [
+    pytest.param("X", id="X"),
+    pytest.param("layers", id="layers"),
+]
```
flying-sheep: these can also just be (parametrized) fixtures

flying-sheep (commented):
Not much to add to what Ilan said, thank you this looks clean!

src/anndata/_io/specs/methods.py
eroell: Adjusted to this, I updated the first PR comment accordingly to this behaviour. Assignment to `.X` and `.layers[<key>]` raise this warning now, too.

src/anndata/_core/anndata.py
eroell: Adjusted to this. The reshaping behaviour from before is maintained:

```py
adata = ad.AnnData(X=np.zeros((2, 4)))
adata.X = np.zeros((4, 2))
# FutureWarning: Automatic reshaping when setting X will be removed in the future.
```

while higher-than-2-dim is more strict about it, and throws a clearer error

```py
adata = ad.AnnData(X=np.zeros((2, 4)))
adata.X = np.zeros((4, 2, 1))
# ValueError: Cannot set `X` from an array of shape (4, 2, 1): its leading two dimensions (4, 2) do not match the AnnData shape (2, 4). Automatic reshaping is only supported for 2-D inputs.
```

src/anndata/_core/aligned_mapping.py
```diff
@@ -343,12 +343,33 @@ class LayersBase(AlignedMappingBase[TwoDIdx, str | None]):
...
+
 
 class Layers(AlignedActual[TwoDIdx, str | None], LayersBase):
     def __init__(self, parent: AnnData, *, store: MutableMapping[str | None, Value]):
         super().__init__(parent, store=store)
         self.isbacked = None not in self._data and self.parent.filename is not None
 
+    def __setitem__(self, key: str | None, value: Value) -> None:
+        super().__setitem__(key, value)
+        if key in self._data:
```
ilan-gold: Why do we need this check? Could you add a comment as to why?

src/anndata/_core/aligned_mapping.py
```diff
@@ -343,12 +343,33 @@ class LayersBase(AlignedMappingBase[TwoDIdx, str | None]):
     def __bool__(self) -> bool:
         return not self.keys() <= {None}
 
+    def _warn_if_spec_violation(self, key: str | None, val: Value) -> None:
```
ilan-gold: I think it's weird to make this a general method when it only does one specific thing.  There are lots of ways to violate the spec

ilan-gold: Maybe it makes sense to move this to `LayersBase` and rename it and `_spec_violation_message` to be a bit more clear

ilan-gold (commented):
Please add a release note and semantic commit title (see failing CI) :)

flying-sheep:
OK, this doesn’t actually close #2430 yet, so I changed that in the description.

I also adapted the title.

src/anndata/_core/aligned_mapping.py
eroell: I double-checked, and its not required - so I removed it, thanks for spotting!

src/anndata/_core/aligned_mapping.py
eroell: Renamed to `_non_2d_message`.

this warning here is already in `LayersBase` though?

flying-sheep: instead of modifying `__set__` and checking for `mapping`’s type, I think you could override `_validate_value` for `LayersBase`:

- rename `_warn_if_non_2d` to `_validate_value`, adapt the signature
- call `super()._validate_value(val, key)` in it
- undo your changes in `__set__`

maybe that’s what Ilan meant.

/edit: seems like `_validate_value` gets called when *getting* things, why?

tests/test_x_layers_2d.py
```diff
@@ -0,0 +1,206 @@
...
+    assert adata.X.shape == arr3d.shape
+
+    adata = ad.AnnData(X=arr2d)
+    with pytest.warns(UserWarning, match=r"Layer 'L'"):
+        adata.layers["L"] = arr3d
+    assert adata.layers["L"].shape == arr3d.shape
+
+    with pytest.warns(UserWarning, match=r"Layer 'M'"):
+        adata.layers = {"M": arr3d}
+    assert adata.layers["M"].shape == arr3d.shape
```
flying-sheep: this is 5 tests, not 1, please split up using `parametrize` or at least add `subtests`

src/anndata/_core/aligned_mapping.py
flying-sheep: OK, I changed validation so it only happens once. Since #1119, we validate on initialization, so this should be fine.

flying-sheep (approved):

flying-sheep:
OK @ilan-gold in 3ce3ead3ba162ed5ca5555a58c4adc4efaac85f7, I made it so validation only happens when setting aligned mappings, with the goal of allowing to raise warnings in there without the warnings being displayed on each access, and the side-benefit of increased performance by not re-validating everything on access.

This caused a very small breaking change illustrated by the change in 0e13d182a6b281bb764c01bdf59b95d01b6ddc6d: since we don’t go through all aligned mappings and re-validate everything whenever accessing one, we don’t aggressively re-sync `obs_names.name`→`obsm[*].index.name` on each `obsm` access. But since we don’t support intererior mutation of dataframes, and since I doubt that anyone even knows that this syncing happens, I think that’s more than fine.

src/anndata/_core/storage.py
```diff
@@ -22,6 +22,50 @@
...
+    if ndim <= 2:
+        return None
+    return (
+        f"{name} must be 2-dimensional, but got an array with shape "
+        f"{tuple(shape)} (ndim={ndim}). Storing higher-dimensional arrays "
+        f"in `X` or `layers` violates the AnnData specification, and cannot be written to disk."
+    )
+
+
+def _check_x_and_layers_are_2d_on_write(adata: AnnData) -> None:
```
ilan-gold, line 50: Please add this to the `AnnData.unwriteable` function and parametrize/write a test for that public API (there should already be a few tests for that function)

ilan-gold (approved):
Nice, just the one thing!

src/anndata/_core/storage.py
eroell: Added a check to `AnnData.unwriteable`.

Also added a test to check `AnnData.unwriteable` now evaluates to `True` for a non-2D `layer`/`X` in `tests/test_readwrite.py`, where most of the previous `AnnData.unwriteable` tests are located already.

ilan-gold (approved):

PR #2433, opened 2026-05-11
fix: correct IO types

flying-sheep:
before I start working on it, this stops the `_io/specs/methods.py` module from lighting up like a christmas tree.

“positional or keyword” arguments can take the role of either positional-only or keyword-only arguments in the protocol, which is how the `@_REGISTRY.register_write`-decorated functions still typecheck.

- [x] Release note not necessary because: types-only fix

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,4 +25,6 @@
     ],
     "python.terminal.activateEnvironment": true,
     "python.analysis.include": ["src/**/*", "ci/scripts/**/*", "tests/**/*"],
+    "python-envs.defaultEnvManager": "pypa.hatch:hatch",
+    "python-envs.defaultPackageManager": "pypa.hatch:hatch",
 }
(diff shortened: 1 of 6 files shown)


.vscode/settings.json
```diff
@@ -25,4 +25,6 @@
     ],
     "python.terminal.activateEnvironment": true,
     "python.analysis.include": ["src/**/*", "ci/scripts/**/*", "tests/**/*"],
+    "python-envs.defaultEnvManager": "pypa.hatch:hatch",
```
ilan-gold, line 28: 🎊

src/anndata/_io/specs/registry.py
```diff
@@ -112,13 +112,13 @@ def __init__(self) -> None:
         self.write = {}
         self.write_specs = {}
 
-    def register_write[T](
+    def register_write[S: StorageType, T: RWAble](
```
ilan-gold, line 115: Yup: https://github.com/scverse/anndata/pull/2427/changes#diff-bd6c3c0fd41d811b7bc18b35988156cd346ed084b0236e14867ea8e9a27976f0

ilan-gold (approved):

.vscode/settings.json
flying-sheep: The only remaining issue IMHO is that the ty and ruff extensions haven’t migrated to the “Python Environments” extension API yet, so they won’t respect the environment chosen yet.

But that’s in the works. I think once that hits, using this extension will be genuinely nicer than not.

PR #2439, opened 2026-05-12
feat: migrate to zarr v3 defaults

### LINKED ISSUE #2239: Flip zarr v3 `settings` to turned on

I think some or all of the relevant zarr v3 flags should be flipped on:

- [ ] zarr v3 by default might have to wait until strings are more of a solved quantity (currently writing it emits a warning), TODO: investigate
- [ ] I think auto-sharding can be turned on by-default once zarr v3 is turned on (independent of whether or not we have it the default)

ilan-gold:
- [ ] Closes #2239
- [ ] Tests added


src/anndata/_io/specs/registry.py
```diff
@@ -484,8 +484,9 @@ def read_elem_lazy(
     Reading a dense matrix from a zarr store lazily:
 
     >>> adata.layers["dense"] = ad.experimental.read_elem_lazy(g["layers/dense"])
-    >>> adata.layers["dense"]
-    dask.array<from-zarr, shape=(2700, 32738), dtype=float32, chunksize=(169, 2047), chunktype=numpy.ndarray>
+    >>> # TODO: remove ... once https://github.com/zarr-developers/zarr-python/issues/3602 becomes minimum zarr (3.1.6)
```
flying-sheep: this should make this not appear in the docs

```suggestion
    ..
    	TODO: remove “...” once https://github.com/zarr-developers/zarr-python/issues/3602 becomes minimum zarr (3.1.6)

    >>> adata.layers["dense"] = ad.experimental.read_elem_lazy(g["layers/dense"])
```

flying-sheep (approved):

src/anndata/_io/specs/registry.py
ilan-gold: Ah, good to know - I thought `#` doesn't appear in docs based on https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized but I guess the rendering and the output/running are different, thanks.

PR #2446, opened 2026-05-13
chore: migrate `settings` to `scverse_misc`

ilan-gold:
- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: dev change


src/anndata/_settings.py
flying-sheep, line 1: `_settings.pyi` is still there and now wrong!

src/testing/anndata/_pytest.py
```diff
@@ -38,7 +38,7 @@
 def setup_env() -> None:
     import anndata
 
-    anndata.settings.reset(anndata.settings._registered_options.keys())
+    anndata.settings.reset(*list(anndata.settings.model_fields_set))
```
flying-sheep: why wrapping in a list? everything iterable is unpackable!

src/anndata/_settings.py
```diff
@@ -1,532 +1,42 @@
...
 
-           {self.description}
-        """
-    else:
-        doc = f"""\
-        {self.option}: `{type_str}`
-            {self.description} (default: `{self.default_value!r}`).
-        """
-    return textwrap.dedent(doc)
+    min_rows_for_chunked_h5_copy: int = 1000
```
flying-sheep: should be `Annotated[int, Field(gt=0)]` or so, right?

src/anndata/_settings.py
```diff
@@ -1,532 +1,42 @@
...
+    """Whether or not to remove unused categories with :class:`~pandas.Categorical`."""
 
-class DeprecatedOption(NamedTuple):
-    option: str
-    message: str | None
-    removal_version: str | None
+    check_uniqueness: bool = True
+    """Whether or not to check uniqueness of the `obs` indices on `__init__` of :class:`~anndata.AnnData`."""
 
+    copy_on_write_X: bool = True
```
flying-sheep: needs deprecating: https://pydantic.dev/docs/validation/latest/concepts/fields/#deprecated-via-the-warningsdeprecated-decorator

we can of course use our own here like `Field(deprecated=deprecated(Deprecation("0.13", "... Removal in 0.14"))`

what’s we lose is the semantic `removal_version`, maybe we should add that to `scverse-misc`?

src/anndata/_settings.py
```diff
@@ -1,532 +1,42 @@
...
-    else:
-        doc = f"""\
-        {self.option}: `{type_str}`
-            {self.description} (default: `{self.default_value!r}`).
-        """
-    return textwrap.dedent(doc)
+    min_rows_for_chunked_h5_copy: int = 1000
+    """Minimum number of rows at a time to copy when writing out an H5 Dataset to a new location"""
 
+    disallow_forward_slash_in_h5ad: bool = False
```
flying-sheep: weird that this is not tested

```suggestion
    disallow_forward_slash_in_h5ad: bool = True
```

flying-sheep (changes requested):
Don’t we need to specify something so the env variable stuff works?

Otherwise please go over the defaults again, I found another one we flipped by now (shouldn’t that be caught by tests?)

ilan-gold:
> Don’t we need to specify something so the env variable stuff works?

https://scverse-misc.readthedocs.io/stable/api/settings.html Judging by this documentation, I would expect it to just "work"

src/anndata/_settings.py
ilan-gold: Oh I didn't think through the fact that things are typed now, that makes sense.

src/anndata/_settings.py
ilan-gold: We don't test the default behavior, it seems.  We only have an explicit parametrization!

src/anndata/_settings.py
ilan-gold: I went with a simple string because `pydantic` relies on implicit class behavior of the output of `deprecated`, which is undocumented: https://docs.python.org/3/library/warnings.html#warnings.deprecated

so it doesn't reallymake sense to use our custom deprecation

flying-sheep (approved):
The most beautiful diff stats! 

Image

I’m glad you implemented the original version though, the API informed `scverse-misc`’s a bunch!

ilan-gold:
> I’m glad you implemented the original version though, the API informed scverse-misc’s a bunch!

Your help was invaluable on the first pass - for one of my first "production-ready" python greenfields, I look back fondly :)

Issue #2036, opened 2025-07-17
feat: sensible chunk defaults for dask


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

#2033 at least makes it so that we are no longer creating arrays with very small chunks unnecessarily but from a performance perspective, mapping dask chunks to chunks on disk is not a good idea.  Especially with the codec pipeline in zarr being parallelized, the need for dask workers to be responsible for speeding up io is lessened and thus cramming more parallel reads within a worker is probably sensible.  I imagine the same is true for hdf5 even if it isn't parallel i.e., doing more/bigger reads per-dask-chunk.

See https://gist.github.com/lgarrison/ef2922fdb8ca78e8753c39758afedb56 for a related example of this issue where having few python process workers but utilizing the rust-based parallelization of `zarrs` improves throughput.

See https://github.com/zarrs/zarr_benchmarks/blob/9679f36ca795cce65adc603ae41147324208d3d9/scripts/zarr_dask_python_benchmark_roundtrip.py#L22 for using `shards` as the chunk size for dask (since shards contain chunks) as a starting point for implementing better defaults.

In general some todos:

- [ ] Use shards when applicable for zarr
- [ ] Create benchmarks for unsharded data in hdf5 and zarr to find the virtual vs. on-disk chunk size performance tradeoff
- [ ] Write some sort of general-purpose function that can be reused across zarr/hdf5

ehsanestaji:
I can take a first look at this and start with a small test-backed slice around the current Zarr/Dask chunk default behavior, then follow up with a focused PR if the scope looks right.

ilan-gold:
@ehsanestaji I'm not really sure where to start here is my issue.

I think it would be great, as a starting point, to maybe come up with a notebook highlighting the tradeoffs? As an example, you might compare dask with big chunks and small chunks (over different on-disk chunkings) on a "normal" workload like https://github.com/scverse/scanpy-tutorials/blob/main/dask.ipynb and profiling how workers/chunk sizes choices affects requirements.

Or maybe this is better done as a script you can run, and then the notebook analyzes the results of the grid search?

Do you have nay experience with using dask that you could bring here?

Thanks for jumping in!

ehsanestaji:
Thanks for the direction! A benchmark-first approach sounds good to me.

[...]

PR #2465, opened 2026-05-21
test: add Dask chunk grid benchmark scaffold

ehsanestaji:
## Summary

This adds an exploratory benchmark scaffold for #2036 so we can compare virtual Dask chunk choices against HDF5/Zarr on-disk chunk layouts before changing AnnData defaults.

The benchmark runner:

- creates dense `X` arrays with controlled HDF5/Zarr chunks and optional Zarr v3 shards
- reads `X` lazily through `anndata.experimental.read_elem_lazy`
- varies virtual Dask chunks, worker counts, and thread/process settings
- records runtime/package metadata, store size, task count, elapsed time, result shape/size, and coarse process/worker RSS readings
- includes array-level workloads and a Scanpy-style `scanpy_normalize_log1p` workload

This also adds a small notebook for summarizing the generated CSV and README instructions for smoke/larger-grid runs. Generated benchmark outputs are ignored under `benchmarks/results`.

## Local signal

A modest local grid (`3000x800`, HDF5/Zarr, on-disk chunks `250x800` and `1000x800`, default vs `1000x-1`, 1/2 workers, `sum_axis0` and `scanpy_normalize_log1p`) produced 32 rows. For small on-disk chunks (`250x800`), `1000x-1` reduced task counts and improved timings in the 1-worker Scanpy-style case by about 1.16x for HDF5 and 1.26x for Zarr.

These numbers are only an initial local smoke signal; the intent is to make the benchmark/review path available before proposing default behavior changes.

## Checks

- `ruff check benchmarks/scripts/dask_chunk_grid.py tests/test_dask_chunk_grid_script.py`
- `.venv/bin/python -m pytest tests/test_dask_chunk_grid_script.py -q`
- `python3 -m json.tool benchmarks/notebooks/dask_chunk_grid_analysis.ipynb`
- `git diff --check`
- tiny end-to-end benchmark smoke run wrote 8 CSV result rows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ __pycache__/
 benchmark/benchmarks/data
 benchmarks/benchmarks/data
 benchmarks/pkgs
+benchmarks/results
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -74,3 +74,44 @@ All benchmarks:
 ### View in the browser:
 
 You can view the benchmarks in the browser with `asv publish` followed by `asv preview`. If you want to include benchmarks of a local branch, I think you'll have to add that branch to the `"branches"` list in `asv.conf.json`.
+
+## Dask Chunk Default Exploration
+
+Issue [#2036](https://github.com/scverse/anndata/issues/2036) needs benchmark data for choosing more sensible virtual Dask chunk defaults when reading chunked HDF5/Zarr arrays lazily. The script below creates dense `X` arrays with controlled on-disk chunks, reads them through `anndata.experimental.read_elem_lazy`, runs a small set of Dask and Scanpy-style workloads, and writes one CSV row per grid point. Rows include runtime package versions, store size, task count, elapsed time, and coarse process/worker memory readings.
+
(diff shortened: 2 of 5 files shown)


ehsanestaji:
I updated the title to a semantic PR title. I don’t seem to have permission to add labels on this repo, but the current validation failures look like triage metadata rather than code failures. Could a maintainer please add the appropriate labels, likely `no milestone` and `skip-gpu-ci`? `benchmark`, `type: dask array`, and `performance 🐌` may also fit this benchmark-only PR.

ilan-gold:
> These numbers are only an initial local smoke signal; the intent is to make the benchmark/review path available before proposing default behavior changes.

So to this end, it might make sense for you @ehsanestaji to create a separate repo with benchmarks that can be run.  This repo would produce a graph/table at the end and then propose changes/reasons for those changes.  Does that sound reasonable? I don't really see why `anndata` itself should carry this code around.

Then there would be a PR to update the defaults, link to the benchmarking effort, and perhaps write a small documentation note/page explaining things here.

PR #2470, opened 2026-05-27
docs: Clarify anndata package status in R section

rcannood:
Updated the description of the anndata package for R to indicate it is superseded by anndataR.

- [x] Release note not necessary because: trivial docs change


Zethson (commented):
Do you want to keep it or should we even consider not mentioning it?

rcannood:
If possible I'd prefer to keep it to clear up any confusion between `library(anndata)` and `library(anndataR)`.

PR #2481, opened 2026-06-08
ci: bump codecov-action to v6 to fix signature verification

timtreis:
`codecov-action@v5` verifies the codecov CLI binary against Codecov's old Keybase key, which they bricked after migrating accounts. Uploads now fail with `Could not verify signature`. `@v6` (= 6.0.2) ships the updated key. Version pin bump.

Ref: https://github.com/codecov/codecov-action/issues/1956


ilan-gold:
Zizmor was added and handled this #2413

PR #2498, opened 2026-06-15
feat: allow `MuData` inside accessor

ilan-gold:
@ilia-kats Would this work? Feel free to push over this/edit as you want, just figured this seems like a pretty low-risk way to go.

- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: dev change (I think)

### DIFF
--- a/src/anndata/acc/__init__.py
+++ b/src/anndata/acc/__init__.py
@@ -63,7 +63,7 @@ __all__ = [
 ]
 
 
-class AdRef[I: Hashable]:
+class AdRef[I: Hashable, D: MuData | AnnData]:
     r"""A reference to a 1D or 2D array along one or two dimensions of an AnnData object.
 
     Examples
@@ -86,7 +86,7 @@ class AdRef[I: Hashable]:
 
     """
 
-    acc: RefAcc[Self, I]
+    acc: RefAcc[Self, I, D]
     r"""The accessor containing information about this array.
 
     See :term:`reference accessor`\ s for all possible types this can assume.
@@ -100,7 +100,7 @@ class AdRef[I: Hashable]:
 
     __match_args__: ClassVar = ("acc", "idx")
 
-    def __init__(self, acc: RefAcc[Self, I], idx: I) -> None:
+    def __init__(self, acc: RefAcc[Self, I, D], idx: I) -> None:
         self.acc = acc
         self.idx = idx
 
@@ -754,5 +754,6 @@ def _expand_idx2d_list(idx: Idx2DList) -> list[Idx2D]:
 
 
 if not TYPE_CHECKING:  # https://github.com/tox-dev/sphinx-autodoc-typehints/issues/580
-    R = AdRef[Hashable]
+    R = AdRef[Hashable, MuData | AnnData]
     I = Hashable
+    D = MuData | AnnData
--- a/src/anndata/acc/__init__.py
+++ b/src/anndata/acc/__init__.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, ClassVar, cast, overload
 import pandas as pd
 import scipy.sparse as sp
 
+from .. import AnnData
 from .._core.views import ArrayView
 from .._core.xarray import Dataset2D
 from ..compat import CupySparseMatrix, DaskArray, has_xp
@@ -25,7 +26,6 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Collection, Sequence
     from typing import Any, Literal, Self, TypeGuard
 
-    from .. import AnnData
     from .._core.aligned_mapping import AxisArrays
     from ..compat import XVariable
     from ..typing import InMemoryArray
@@ -150,7 +150,7 @@ class MapAcc[R: RefAcc](abc.ABC):
 
 
 @dataclass(frozen=True)
-class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
+class RefAcc[R: AdRef[I], I, D: MuData | AnnData](abc.ABC):  # type: ignore
     r"""Abstract base class for reference accessors.
 
     See :term:`reference accessor`\ s for all existing subclasses.
@@ -180,11 +180,11 @@ class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
         """Get a string representation of the index."""
 
     @abc.abstractmethod
-    def isin(self, data: AnnData | MuData, idx: I | None = None, /) -> bool:
+    def isin(self, data: D, idx: I | None = None, /) -> bool:
         """Check if the referenced array is in the AnnData object."""
 
     @abc.abstractmethod
-    def get(self, data: AnnData | MuData, idx: I, /) -> Array:
+    def get(self, data: D, idx: I, /) -> Array:
         """Get the referenced array from the AnnData object."""
 
     def _maybe_flatten(self, idx: I, a: Array) -> Array:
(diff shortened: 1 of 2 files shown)


src/anndata/acc/__init__.py
```diff
@@ -246,16 +248,16 @@ def __repr__(self) -> str:
...
-    def isin(self, adata: AnnData, idx: Idx2D | None = None) -> bool:
-        if adata.X is None if self.k is None else self.k not in adata.layers:
+    def isin(self, data: AnnData | MuData, idx: Idx2D | None = None) -> bool:
+        if data.X is None if self.k is None else self.k not in data.layers:
             return False
         for i, dim in zip(idx or (), ("obs", "var"), strict=False):
             if isinstance(i, str):  # at most one str
-                return i in getattr(adata, dim).index
+                return i in getattr(data, dim).index
         return True  # idx is None or [:, :]
```
ilia-kats, line 262: MuData doesn't have layers, so this one should stay AnnData-only.

src/anndata/acc/__init__.py
```diff
@@ -174,11 +176,11 @@ def idx_repr(self, idx: I, /) -> str:
         """Get a string representation of the index."""
 
     @abc.abstractmethod
-    def isin(self, adata: AnnData, idx: I | None = None, /) -> bool:
+    def isin(self, data: AnnData | MuData, idx: I | None = None, /) -> bool:
         """Check if the referenced array is in the AnnData object."""
```
ilia-kats, line 185: I'm not sure if this works if `LayersAcc` doesn't accept `MuData`.

ilia-kats (commented):
This just adjust the type annotation and we know that the implementation works with MuData, so it should be fine. There's just one caveat regarding `LayersAcc` (MuData doesn't have layers), otherwise looks good to me.

ilan-gold:
> we know that the implementation works with MuData

So went my thinking, seemed simpler than a protocol. Let's see why Phil says.

flying-sheep:
Sure, if it works!

src/anndata/acc/__init__.py
ilan-gold: Since it's the `abstractmethod`, is the typing wrong? I suppose this could be made generic and the `Layers` variant special-cased.

ilia-kats: Yes, overriding methods [can't have different types from the base](https://play.ty.dev/88db8e7d-8878-49dc-85c9-413e3665349d0). Making the base class generic works.

ilia-kats (approved):
Does it make sense to make all the `RefAcc` subclasses generic in `D`? I would have done the same that you're doing with `LayerAcc`, but using `AnnData | MuData` as the type argument to `RefAcc`. But I'm really not an expert on typing in Python.

Otherwise LGTM

ilan-gold:
> Does it make sense to make all the RefAcc subclasses generic in D? I would have done the same that you're doing with LayerAcc, but using AnnData | MuData as the type argument to RefAcc. But I'm really not an expert on typing in Python.



Oh much cleaner

ilan-gold:
Sorry I don't fully follow @ilia-kats actually - feel free to push

ilia-kats:
> Sorry I don't fully follow @ilia-kats actually - feel free to push

I don't seem to have push access here, but basically something like this:

```patch
@@ -150,7 +150,7 @@ class MapAcc[R: RefAcc](abc.ABC):
 
 
 @dataclass(frozen=True)
-class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
+class RefAcc[R: AdRef[I], I, D: AnnData | MuData](abc.ABC):  # type: ignore
     r"""Abstract base class for reference accessors.
 
     See :term:`reference accessor`\ s for all existing subclasses.
@@ -180,11 +180,11 @@ class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
         """Get a string representation of the index."""
 
     @abc.abstractmethod
-    def isin(self, data: AnnData | MuData, idx: I | None = None, /) -> bool:
+    def isin(self, data: D, idx: I | None = None, /) -> bool:
         """Check if the referenced array is in the AnnData object."""
 
     @abc.abstractmethod
-    def get(self, data: AnnData | MuData, idx: I, /) -> Array:
+    def get(self, data: D, idx: I, /) -> Array:
         """Get the referenced array from the AnnData object."""
 
     def _maybe_flatten(self, idx: I, a: Array) -> Array:
@@ -200,7 +200,7 @@ class RefAcc[R: AdRef[I], I](abc.ABC):  # type: ignore
 
 
 @dataclass(frozen=True)
-class LayerAcc[R: AdRef[Idx2D]](RefAcc[R, Idx2D]):
+class LayerAcc[R: AdRef[Idx2D]](RefAcc[R, Idx2D, AnnData]):
     r"""Reference accessor for arrays in layers (`A.`\ :attr:`~AdAcc.layers`).
 
     Examples
@@ -284,7 +284,7 @@ class LayerMapAcc[R: AdRef](MapAcc[LayerAcc]):
 
 
 @dataclass(frozen=True)
-class MetaAcc[R: AdRef[str | None]](RefAcc[R, str | None]):
+class MetaAcc[R: AdRef[str | None]](RefAcc[R, str | None, AnnData | MuData]):
     r"""Reference accessor for arrays from metadata containers (`A.`\ :attr:`~AdAcc.obs`/`A.`\ :attr:`~AdAcc.var`).
 
     Examples
@@ -361,7 +361,7 @@ class MetaAcc[R: AdRef[str | None]](RefAcc[R, str | None]):
 
 
 @dataclass(frozen=True)
-class MultiAcc[R: AdRef[int]](RefAcc[R, int]):
+class MultiAcc[R: AdRef[int]](RefAcc[R, int, AnnData | MuData]):
     r"""Reference accessor for arrays from multi-dimensional containers (`A.`\ :attr:`~AdAcc.obsm`/`A.`\ :attr:`~AdAcc.varm`).
 
     Examples
@@ -451,7 +451,7 @@ class MultiMapAcc[R: AdRef](MapAcc[MultiAcc]):
 
 
 @dataclass(frozen=True)
-class GraphAcc[R: AdRef[Idx2D]](RefAcc[R, Idx2D]):
+class GraphAcc[R: AdRef[Idx2D]](RefAcc[R, Idx2D, AnnData | MuData]):
     r"""Reference accessor for arrays from graph containers (`A.`\ :attr:`~AdAcc.obsp`/`A.`\ :attr:`~AdAcc.varp`).
 
     Examples
```

flying-sheep (approved):

src/anndata/acc/__init__.py
```diff
@@ -57,7 +63,7 @@
 ]
 
 
-class AdRef[I: Hashable]:
+class AdRef[I: Hashable, D: MuData | AnnData]:
```
flying-sheep, line 66: @ilan-gold @ilia-kats FYI: in a RC this is OK, but be aware that this is a breaking change.

E.g. when a class extends AdRef e.g. like this, `class Foo[T](AdRef[I]): ...`, it will raise an exception at module import time when AdRef adds a second generic parameter without a default.

ilia-kats:
Turns out that this PR actually breaks mudata: Since mudata depends on and imports anndata in lots of places, there are circular dependencies all over the place now, so trying to import mudata now gives

```python
ImportError: cannot import name 'MuData' from partially initialized module 'mudata' (most likely due to a circular import) (/data/ilia/mudata/src/mudata/__init__.py)
```

Apologies for not catching this earlier.

flying-sheep:
the MuData import should probably be in `if TYPE_CHECKING`.

Only `R = AdRef[Hashable, MuData | AnnData]` needs mudata at runtime, and that can become `R = AdRef["Hashable", "MuData | AnnData"]` I think.

PR #2506, opened 2026-06-17
perf: speed up backed sparse integer indexing

tippered1-debug:
## Summary

Integer indexing of backed sparse datasets currently reads every selected row or column separately, even when most of the indices form contiguous runs.

This PR groups those runs and reads them as slices. Shuffled indices and duplicates are restored to their original order after reading. Highly fragmented indexers continue to use the existing path, since slicing does not help in that case.

Related to #1224.
## Benchmark

Benchmarked on a `10000 x 10000` sparse matrix with density `0.01`. Each indexer contains `2048` elements. Results are medians of three runs.

### HDF5 CSR

* **Single run:** `7.10 ms` → `0.73 ms` — **9.7x faster**
* **Multiple runs:** `6.81 ms` → `0.74 ms` — **9.2x faster**
* **Clustered, shuffled:** `7.04 ms` → `1.03 ms` — **6.8x faster**
* **Clustered with duplicates:** `7.02 ms` → `0.83 ms` — **8.5x faster**

### Zarr CSR

* **Single run:** `16.60 ms` → `3.70 ms` — **4.5x faster**
* **Multiple runs:** `16.27 ms` → `13.92 ms` — **1.2x faster**
* **Clustered, shuffled:** `18.91 ms` → `14.30 ms` — **1.3x faster**
* **Clustered with duplicates:** `19.19 ms` → `8.53 ms` — **2.2x faster**

Fragmented indexers continue to use the existing path and stayed close to the previous timings.

- [ ] Closes #
- [x] Tests added

### DIFF
--- a/benchmarks/benchmarks/sparse_dataset.py
+++ b/benchmarks/benchmarks/sparse_dataset.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from types import MappingProxyType
 from typing import TYPE_CHECKING
 
+import h5py
 import numpy as np
 import pandas as pd
 import zarr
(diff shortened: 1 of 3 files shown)


ilan-gold (commented):
Can this reuse `subset_by_major_axis_mask`? It looks eerily familiar.  Or could that method be removed in favor of this code path? I'm not sure what the performance considerations here are but I think they are probably trivial i.e., operations on a small 1d array.

Thanks for the contribution!

PR #2509, opened 2026-06-18
fix: accessors: don't import mudata

### LINKED ISSUE #2511: Circular import between `anndata` and `mudata` when `mudata` is imported first

### Report

This is the reason for the current integration test failures of scirpy (https://github.com/scverse/scirpy/issues/716). 

When anndata is imported before mudata, this bug does not surface: 

- **`import anndata` first** → by the time `acc` imports, `AnnData` already exists, so
  `mudata`'s `from anndata import AnnData` completes, `mudata` finishes initializing, and
  `acc`'s `from mudata import MuData` succeeds. 
- **`import mudata` first** → `mudata` is stuck at line 3 importing `anndata`, which
  imports `acc`, which tries `from mudata import MuData` before `mudata` has defined it. 

Code:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     # anndata from the main branch (as installed by scverse/integration-testing).
#     # The released PyPI anndata does NOT have the `anndata.acc` subpackage and
#     # therefore does NOT reproduce this bug.
#     "anndata @ git+https://github.com/scverse/anndata",
#     "mudata",
# ]
# ///

# Import mudata FIRST (before anndata). This is what triggers the cycle:
#   mudata/__init__         ->  from anndata import AnnData
#   anndata/__init__        ->  from . import ... acc ...
#   anndata/acc/__init__    ->  from mudata import MuData   # mudata not finished yet -> ImportError
import mudata  # noqa: F401
```

Traceback:

```pytb
    Updated https://github.com/scverse/anndata (b7d260eef64059c02407e85c2930439f7bf7e79e)
      Built anndata @ git+https://github.com/scverse/anndata@b7d260eef64059c02407e85c2930439f7bf7e79e
Installed 26 packages in 17ms
  File "/home/sturm/projects/claude_sandbox/scirpy/anndata-mudata-repro/repro.py", line 32, in <module>
    import mudata  # noqa: F401
    ^^^^^^^^^^^^^
  File "/home/sturm/.cache/uv/environments-v2/repro-37997e8b40de5642/lib/python3.14/site-packages/mudata/__init__.py", line 3, in <module>
    from anndata import AnnData
  File "/home/sturm/.cache/uv/environments-v2/repro-37997e8b40de5642/lib/python3.14/site-packages/anndata/__init__.py", line 25, in <module>
    from . import abc, acc, experimental, io, types, typing  # isort: skip
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sturm/.cache/uv/environments-v2/repro-37997e8b40de5642/lib/python3.14/site-packages/anndata/acc/__init__.py", line 21, in <module>
    from mudata import MuData
ImportError: cannot import name 'MuData' from partially initialized module 'mudata' (most likely due to a circular import) (/home/sturm/.cache/uv/environments-v2/repro-37997e8b40de5642/lib/python3.14/site-packages/mudata/__init__.py)
```

### Versions

anndata main, mudata latest

ilia-kats:
#2498 introduced circular dependencies, such that `import mudata` no longer works, unless one imports `anndata` first. So

```python
import anndata
import mudata
```
works fine, however
```python
import mudata
```
results in 
```python
ImportError: cannot import name 'MuData' from partially initialized module 'mudata' (most likely due to a circular import) (/data/ilia/mudata/src/mudata/__init__.py)
```

The `else` branch in the `if TYPE_CHECKING` block is sufficient to generate correctly linked documentation.

- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: only affects RC

### DIFF
--- a/src/anndata/acc/__init__.py
+++ b/src/anndata/acc/__init__.py
@@ -6,7 +6,6 @@ import abc
 from collections.abc import Hashable
 from dataclasses import KW_ONLY, dataclass, field
 from functools import cached_property
-from importlib.util import find_spec
 from typing import TYPE_CHECKING, ClassVar, cast, overload
 
 import pandas as pd
@@ -17,7 +16,7 @@ from .._core.views import ArrayView
 from .._core.xarray import Dataset2D
 from ..compat import CupySparseMatrix, DaskArray, has_xp
 
-if TYPE_CHECKING or find_spec("mudata"):
+if TYPE_CHECKING:
     from mudata import MuData
 else:
     MuData = type("MuData", (), {"__module__": "mudata"})
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,7 @@ doc = [
     "sphinx-design",
     "anndata[dask]",
     "pandas>=3",
+    "mudata",
     # for unreleased changes
     { include-group = "dev-doc" },
 ]


src/anndata/acc/__init__.py
```diff
@@ -17,7 +16,7 @@
 from .._core.xarray import Dataset2D
 from ..compat import CupySparseMatrix, DaskArray, has_xp
 
-if TYPE_CHECKING or find_spec("mudata"):
+if TYPE_CHECKING:
     from mudata import MuData
 else:
     MuData = type("MuData", (), {"__module__": "mudata"})
```
flying-sheep, line 22: is that needed?

We could probably do `R = AdRef["Hashable", "MuData | AnnData"]`, right?

ilia-kats: I tried using strings, that doesn't link documentation correctly for subclasses (e.g. `GraphAcc[R: AdRef[Idx2D]](RefAcc[R, Idx2D, "MuData | AnnData"])` doesn't create hyperlinks to `MuData` or `AnnData` in the `Bases` section). But a
```python
class MuData:
    pass
```
here should work, I think.

flying-sheep: That’s what we have right now (except better since `__module__` is also correct)

ilan-gold: So this is essential? And we can merge?

ilan-gold (approved):

src/anndata/acc/__init__.py
ilia-kats: >So this is essential?

As far as I can tell, yes.

Issue #623, opened 2021-09-24
Setting adata.filename = None doesn't load .X to memory


Koncopd:
for a backed adata now `adata.filename=None` makes `adata.X` `None` instead of loading `.X` into memory.

flying-sheep:
That should be easy to fix!

PR #2517, opened 2026-06-27
test(backed): add regression for filename=None after read_h5ad

SID-6921:
## Summary
Adds regression coverage for switching from backed mode to memory mode when loading with 
ead_h5ad(..., backed='r+') and then setting data.filename = None.

## Details
- Adds a new parametrized test covering dense, CSR, and CSC X payloads.
- Verifies that after ilename=None:
  - isbacked is False
  - X is not None
  - data values are preserved

## Validation
- python -m pytest tests/test_backed_hdf5.py -k return_to_memory_mode -q

Refs #623

### DIFF
--- a/tests/test_backed_hdf5.py
+++ b/tests/test_backed_hdf5.py
@@ -319,15 +319,15 @@ def test_return_to_memory_mode(adata: ad.AnnData, backing_h5ad: Path):
 
 
 @pytest.mark.parametrize(
-    ("x", "is_sparse"),
+    "x",
     [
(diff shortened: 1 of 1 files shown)


SID-6921:
I pushed a follow-up commit to satisfy pre-commit linting on the new regression test. Current failing checks are triage gates (Triage: Check PR title, milestone, and labels and Triage: Check if GPU tests are allowed to run), which I can’t resolve from a fork. Could a maintainer please help set the required labels/milestone and triage permissions? Thanks!

PR #2528, opened 2026-07-02
feat: support `RefAcc` pathing for "full" arrays

ilan-gold:
- [ ] Closes #
- [x] Tests added

- [x] Release note not necessary because: Builds on an existing feature before a release? Do we need one?


ilan-gold:
@ilia-kats ping

docs/accessors.rst
```diff
@@ -35,9 +35,11 @@ For these purposes, they are
 
     or to extract the referenced vector:
 
-    >>> ref = A.obs["louvain"]
-    >>> adata[ref].categories[:2]  # doctest: +ELLIPSIS
+    >>> adata[A.obs["louvain"]].categories[:2]  # doctest: +ELLIPSIS
     Index(['CD4 T cells', 'CD14+ Monocytes'], dtype=...)
+    >>> adata[A.varm["PCs"]] # doctest: +ELLIPSIS
+    array([[...
+          ..., dtype=float32)
```
flying-sheep:
```suggestion
    array([[...]],
          shape=(1838, 50), dtype=float32)
```

src/anndata/_core/anndata.py
```diff
@@ -1049,16 +1058,25 @@ def _normalize_indices(
...
 
+    @overload
+    def __getitem__(
+        self, index: LayerAcc | MultiAcc | GraphAcc | MetaAcc
+    ) -> _XDataType | pd.DataFrame | Dataset2D | AwkArray: ...
     @overload
     def __getitem__(self, index: AdRef) -> Array: ...
     @overload
     def __getitem__(self, index: Index) -> AnnData: ...
     def __getitem__(self, index: Index | AdRef) -> AnnData | Array:
```
flying-sheep: you missed the other instance of `AdRef`.

Also why list `LayerAcc`, `GraphAcc`, … and not just the superclasses you actually check for? `AdRef | MapAcc | RefAcc`?

src/anndata/acc/__init__.py
```diff
@@ -184,7 +186,18 @@ def isin(self, data: D, idx: I | None = None, /) -> bool:
 
     @abc.abstractmethod
     def get(self, data: D, idx: I, /) -> Array:
-        """Get the referenced array from the AnnData object."""
+        """Get the indexed array from the AnnData object at `idx`."""
+
+    @abc.abstractmethod
+    def get_full_array(
```
flying-sheep: Ah so this is why we didn’t do it. Even more surface area.

And even more array types that we have no chance of keeping in sync with all the other ones. I don’t want to see `-> _XDataType | pd.DataFrame | Dataset2D | AwkArray` in here, only `-> SomeTypeAliasThatWeUseInOtherPlacesToo`.

Maybe instead of a whole new method, `get` should be changed instead so it acccepts `idx: I | None` (with `@overload`s) and the `None` overload returns the full one? Just an idea.

ilia-kats:
I've already released mudata 0.3.9 with basic accessor support, supporting this would be trivial but would require another release. Adapting the full accessor support (not yet released) should also be fine. Otherwise Phil has raised some valid points, but I'm not deep enough in AnnData internals to have an opinion, I think.

docs/accessors.rst
ilan-gold: Unfortunately, older versions of numpy don't print this info (i.e., our min job), it seems: https://github.com/scverse/anndata/actions/runs/28600131940/job/84806011278

ilan-gold:
Thanks @ilia-kats, was mostly for visiblity / clarity.

docs/accessors.rst
flying-sheep: ah, well, still, maybe we can get balanced brackets :sweat_smile:

src/anndata/_core/anndata.py
ilan-gold: > you missed the other instance of AdRef.

Hmmm, not sure I understand what you're referring to.

> Also why list LayerAcc, GraphAcc, … and not just the superclasses you actually check for? AdRef | MapAcc | RefAcc?

No good reason, but even better, per-class overloads

src/anndata/acc/__init__.py
ilan-gold: Oooo yes, absolutely.

docs/accessors.rst
ilan-gold: Done!

flying-sheep (approved):
Looks good!

Maybe give `_NoIdx` a nice `__repr__`:

Image

Without links, the overloads aren’t super helpful, but I guess that’s out of scope for this PR:

Image

I also noticed that we don’t document `AnnData.__getitem__`, we should do that in some other PR as well.

PR #2529, opened 2026-07-03
perf: try out new zarr codec pipeline

ilan-gold:
See https://zarr.readthedocs.io/en/latest/user-guide/examples/codec_pipeline_performance/#running

- [ ] Closes #
- [ ] Tests added


src/anndata/_io/zarr.py
```diff
@@ -33,10 +33,26 @@
...
+    # We are going to be "guinea pigs" for this new pipeline because it should be much faster
+    # and we're shortchanging our users otherwise.
+    # So we change the pipeline if it has not been changed by the user i.e.,
+    # it is the old BatchedCodecPipeline.
+    # This pipeline fully passes ours, zarr's, and zarr's downstream CI. - Ilan
+    old_pipeline = zarr.config.get("codec_pipeline.path")
     with (
+        zarr.config.set({
+            "codec_pipeline.path": "zarr.core.codec_pipeline.FusedCodecPipeline",
+            "codec_pipeline.max_workers": None,
```
flying-sheep, line 57: this changes the behavior if both zarr 3.3+ and zarrs is installed by setting `"codec_pipeline.max_workers": None`, is that OK?

maybe instead doing this makes more sense?

```py
if zarrs:
    ctx = ...
elif zarr_v3:
    ctx = ...
else:
    ctx = nullcontext()
```

ilan-gold: That would be cleaner yeah, but no, it has no effect: https://zarrs-python.readthedocs.io/en/latest/ is untouched by this

flying-sheep (approved):

PR #2530, opened 2026-07-06
refactor: use `sentinel` for `get` override

ilan-gold:
~~Using `Sentinel` for `RefAcc.get` means users can type-safely do their own overrides without needing our exact value.~~ Pipe dream, `Sentinel` is better than a random object, so now we export it and make it explicit

@ilia-kats sorry about this, @flying-sheep and I both dropped the ball on catching this

- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: live change to existing feature

### DIFF
--- a/docs/accessors.rst
+++ b/docs/accessors.rst
@@ -186,7 +186,7 @@ Extending accessors
 
 There are three layers of extensibility:
 
-#.  subclassing :class:`RefAcc` and creating a new :class:`AdRef` instance for creating them:
+#.  subclassing :class:`AdRef` and creating a new :class:`AdAcc` instance for creating them:
 
     ..  code-block:: python
 
@@ -204,9 +204,6 @@ There are three layers of extensibility:
 
         adata = sc.datasets.pbmc3k_processed()
         plt.scatter(*A.obsm["X_umap"][:, [0, 1]], c=A.obs["n_counts"], data=adata)
-
-    Note that for implementing :meth:`RefAcc.get`, you will need :class:`NO_IDX`.
-
 
 #.  subclass one or more of the :term:`reference accessor`\ s, and create a new :class:`AdAcc` instance:
 
@@ -226,6 +223,9 @@ There are three layers of extensibility:
     >>> A = AdAcc(ref_class=TwoDRef, meta_cls=MyMetaAcc)
     >>> A.obs[["a", "b"]]
     A.obs[['a', 'b']]
+
+    Note that for implementing :meth:`RefAcc.get` when subclassing, you will need :class:`NO_IDX`.
+
 
 #.  subclass :class:`AdAcc` to add new accessors:
 
--- a/docs/accessors.rst
+++ b/docs/accessors.rst
@@ -167,6 +167,7 @@ See :class:`AdAcc` for examples of how to use it to create :term:`reference`\ s 
         :toctree: generated/
 
         Idx2D
+        NO_IDX
 
 ..  toctree::
     :hidden:
@@ -176,6 +177,7 @@ See :class:`AdAcc` for examples of how to use it to create :term:`reference`\ s 
     generated/anndata.acc.MultiAcc
     generated/anndata.acc.GraphAcc
     generated/anndata.acc.Idx2D
+    generated/anndata.acc.NO_IDX
 
 .. _extending accessors:
 
@@ -202,6 +204,8 @@ There are three layers of extensibility:
 
         adata = sc.datasets.pbmc3k_processed()
         plt.scatter(*A.obsm["X_umap"][:, [0, 1]], c=A.obs["n_counts"], data=adata)
+
+    Note that for implementing :meth:`RefAcc.get`, you will need :class:`NO_IDX`.
 
 
 #.  subclass one or more of the :term:`reference accessor`\ s, and create a new :class:`AdAcc` instance:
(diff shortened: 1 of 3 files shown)


ilia-kats:
Sorry, but I don't see how that's better. To use this e.g. [here](https://github.com/scverse/mudata/blob/2a297803d96d018a077a49d3cfaf4acea38f3e0f/src/mudata/acc/__init__.py#L48) I'd still need to import `_NO_IDX`, wouldn't I?

ilan-gold:
> Sorry, but I don't see how that's better. To use this e.g. [here](https://github.com/scverse/mudata/blob/2a297803d96d018a077a49d3cfaf4acea38f3e0f/src/mudata/acc/__init__.py#L48) I'd still need to import _NO_IDX, wouldn't I?

I think the idea here is that you would create your own `Sentinel` object, which would type-check correctly because it is of type `Sentintel` but the actual default object being different would not matter.

ilia-kats:
`super().get()` in the linked method calls your implementation, and you have logic like `if idx is _NO_IDX`, so I think it has to be the same object. Unless I rewrite the mudata method to something like
```python
kwargs = {}
if idx is not _MU_NO_IDX:
    kwargs["idx"] = idx
super().get(..., **kwargs)
```

ilan-gold:
Ah I see now, I was thinking of the function signature and not the underlying implementation. I think what you have there would probably be the way to go unless @flying-sheep has something else to say

src/anndata/acc/__init__.py
```diff
@@ -3,11 +3,15 @@
...
 
 import abc
+import sys
 from collections.abc import Hashable
 from dataclasses import KW_ONLY, dataclass, field
 from functools import cached_property
 from typing import TYPE_CHECKING, ClassVar, cast, overload
 
+if sys.version_info < (3, 15):
+    from typing_extensions import Sentinel
```
flying-sheep: and `if sys.version_info >= (3, 15)`?

src/anndata/acc/__init__.py
```diff
@@ -205,7 +200,7 @@ def get(self, data: D, /) -> FullArray: ...
     @overload
     def get(self, data: D, idx: I, /) -> Array: ...
     @abc.abstractmethod
-    def get(self, data: D, idx: I | _NoIdx = _NO_IDX, /) -> Array | FullArray:
+    def get(self, data: D, idx: I | Sentinel = _NO_IDX, /) -> Array | FullArray:
```
flying-sheep: This should be `| _NO_IDX`, isn’t the whole point that the sentinel is both type and instance object?

ilan-gold: "Sentinels defined in this way **can** be used in type hints" (bold mine) - seems to signal you don't have to use it as both

If we did `| _NO_IDX` we would have to export that object as well for people implementing the `abstractmethod`?

src/anndata/acc/__init__.py
ilan-gold: It should just be available without importing, no?

flying-sheep: Ah, I forgot that it goes straight into `builtins`. Right!

src/anndata/acc/__init__.py
flying-sheep: I think the usual way to use `sentinel`s is that you define one as a constant and export it, as Ilia thought here too: https://github.com/scverse/anndata/pull/2530#issuecomment-4893123316

I don’t think “just any sentinel is accepted here” makes sense because you still check `if idx is _NO_IDX` also people would have to override all call sites when adding their own sentinel so that’s checked instead.

What speaks against exporting it? `def get(..., idx: I | NO_IDX = NO_IDX)` and mention it in the “overriding” section of the docs.

ilan-gold: Fine with exporting it, will do that then.  I'm maybe a bit too much of a "minimalist" in the number of things exported

docs/accessors.rst
ilan-gold: This docs section isn't actually subclassing `RefAcc` - I put the note about `NO_IDX` in this section since I'm not sure the intention of the section overall, but feel free to just edit this branch directly @flying-sheep

src/anndata/acc/__init__.py
flying-sheep: I get that, I just think allowing any sentinel is too confusing an alternative.

docs/accessors.rst
flying-sheep: > the intention of the section overall

Is to give guidance on how to extend.

E.g. `hv-anndata` just adds a new `A` that creates `AdRefs` that have a certain superclass, so the first list item applies

MuData needs to do more so a list item further down applies.

What’s unclear here? If you don’t get it, few people will.

flying-sheep (approved):

docs/accessors.rst
ilan-gold: > What’s unclear here? If you don’t get it, few people will.

"subclassing :class:`RefAcc`" but that is not what happens in the section, though.

> E.g. hv-anndata just adds a new A that creates AdRefs that have a certain superclass, so the first list item applies

Yes besides "subclassing :class:`RefAcc`", I understood what the section does.

flying-sheep: good, so this is just a small issue and not some bigger problem with the section!

I’ll clarify what I meant.

Issue #2414, opened 2026-04-30
Support `pairwise` for `concat_on_disk`


alam-shahul:
### Please describe your wishes and possible alternatives to achieve the desired result.

As requested in the [spec](https://anndata.readthedocs.io/en/latest/generated/anndata.experimental.concat_on_disk.html#anndata-experimental-concat-on-disk), I am opening a feature request for this. Would be nice to have.

PR #2558, opened 2026-07-15
fix(concat_on_disk): retain aligned-mapping keys under join="outer"

### LINKED ISSUE #2394: `concat_on_disk` with `join='outer'` doesn't retain all `.obsm` fields
Not sure if this is a bug or a missing feature, but `ad.concat` and `ad.concat_on_disk` seem to behave differently with `join='outer'` for `.obsm` fields that are not shared between all AnnData.

- [ ] (optional) I have confirmed this bug exists on the main branch of anndata.

### Report

Code:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///

import anndata
import numpy as np
a = ad.AnnData(
        X=np.ones((2, 10)),
        obsm={'X_emb': np.arange(4).astype(np.float64).reshape((2, 2))},
)
b = ad.AnnData(
        X=np.ones((4, 12)),
)

c = ad.concat([a, b], join="outer")
assert 'X_emb' in c.obsm.keys()

a.write_h5ad('a.h5ad')
b.write_h5ad('b.h5ad')

ad.experimental.concat_on_disk(['a.h5ad', 'b.h5ad'], out_file='d.h5ad', join="outer")

d = ad.read_h5ad('d.h5ad')
assert 'X_emb' in d.obsm.keys()

```

### Versions
`0.12.10`

alejandro-publius:
- [x] Closes #2394
- [x] Tests added

## Problem

`concat_on_disk` writes the aligned mappings (`.obsm`/`.varm` and `.layers`) over the **intersection** of their keys, regardless of `join`. Under `join="outer"` this silently drops any mapping key present in only some of the inputs — unlike in-memory `anndata.concat`, which keeps the union and fills the missing entries. When the retained keys additionally have differing shapes, the dense path raises `ValueError: Shapes do not align` instead of padding.

```python
from scipy import sparse

a = ad.AnnData(sparse.csr_matrix((3, 5)), obs=pd.DataFrame(index=[f"a{i}" for i in range(3)]),
               obsm={"X_emb": np.zeros((3, 2))})
b = ad.AnnData(sparse.csr_matrix((2, 5)), obs=pd.DataFrame(index=[f"b{i}" for i in range(2)]))
a.write_h5ad("a.h5ad"); b.write_h5ad("b.h5ad")

ad.concat([a, b], join="outer").obsm            # {'X_emb': (5, 2)}  ✅
ad.experimental.concat_on_disk(["a.h5ad", "b.h5ad"], "out.h5ad", join="outer")
ad.read_h5ad("out.h5ad").obsm                   # {}   ❌ X_emb silently dropped
```

## Fix

`_write_concat_mappings` now mirrors the in-memory `outer_concat_aligned_mapping`:

- takes the **union** of keys for `join="outer"` (intersection for `"inner"`, unchanged);
- generates outer reindexers so each key is padded along its free (feature) axis to the max width;
- fills keys missing from some objects with `fill_value` (promoting integer dtypes to float where the fill is `NaN`), by materializing just those keys through the same `concat_arrays` path the in-memory concat uses.

Keys present in every object that need no reindexing keep the existing memory-efficient streaming path, so the common case is unchanged. As a side effect this also fixes the inner-join case where differing `.obsm` widths previously raised.

## Tests

- `test_concat_on_disk_outer_mapping_missing_keys` — `.obsm` with dense / sparse / dataframe values, one key missing from an object and shared keys of differing widths.
- `test_concat_on_disk_outer_layers_missing_keys` — a `.layers` key missing from one object with a differing `var` axis (exercises the reindexed path).

Both assert the on-disk result equals in-memory `concat`. The full `tests/test_concatenate_disk.py` suite passes (498 tests).


src/anndata/experimental/merge.py
```diff
@@ -259,37 +263,97 @@ def write_concat_sparse(  # noqa: PLR0917
...
+):
+    """Concatenate a single mapping key that is absent from some objects (outer join).
+
+    Present elements are read into memory and objects missing the key contribute a
+    ``fill_value`` block, matching :func:`~anndata._core.merge.outer_concat_aligned_mapping`.
+    """
+    if reindexers is None:
+        reindexers = gen_outer_reindexers(elems, ns, new_index=index, axis=axis)
+    arrays = [
+        (elem if isinstance(elem, pd.DataFrame) else to_memory(elem))
```
ilan-gold: Why do these need to be brought into memory?

ilan-gold: Is it to avoid rewriting `_write_concat_sequence` to handle missing elements?

alejandro-publius: Yes—that was the reason in the original implementation, but I agree that materializing the present arrays undermined `concat_on_disk`.

I’ve since rewritten that path so it remains streaming: missing dense entries use lazy Dask fill blocks, missing sparse entries use shaped sparse fill blocks, and present arrays remain backed and flow through `_write_concat_sequence`. I also added coverage for custom fill values and axis-1 sparse layers while verifying the result still matches `anndata.concat`.

Does the revised approach address your concern?

alejandro-publius:
Sorry for the slow reply here. To answer the question directly, yes, that was the reason, and it was the wrong call. Bringing the present elements into memory was avoiding the work of teaching the write path to handle missing elements.

That has since been rewritten. `_concat_missing_mapping_key` and its `to_memory` call are gone. A key that is absent from an object now contributes a lazily built fill block that stacks straight onto the other reindexed blocks, so the objects that do have the key keep streaming exactly as they did before. `write_concat_sparse` and `write_concat_dense` handle the placeholder themselves, sparse via an empty matrix of the right shape and dense via `da.full`, and for the common `fill_value` of 0 the sparse block stores nothing at all, so a missing key never materialises a dense block regardless of width.

I have also just merged main in, since the mypy adoption in #2577 and the compat cleanup in #2574 had touched the same code paths. Two things there are worth a look, since they are judgement calls rather than mechanical conflict resolution.

`_write_concat_arrays` now decides between the sparse and dense writers using only the elements that are present, while still passing the full sequence, placeholders included, to whichever writer it picks. The narrowing list comprehensions added in #2577 would otherwise route the placeholders into the dense branch and lose alignment with `reindexers` and `ns`.

The `use_reindexing` fast path for `.{axis_name}m` is now conditioned on `use_reindexing or join == "outer"`. That optimisation predates outer joins over mapping keys, and taking it as is would hand back identity reindexers in exactly the case where per key reindexing is needed.

Locally `tests/test_concatenate_disk.py` is 502 passed and `tests/test_concatenate.py` is 1527 passed, with mypy clean on `merge.py`.

alejandro-publius:
@ilan-gold gentle ping on this one. Since my last comment the in-memory fallback is gone: an absent key now contributes a lazily built fill block that stacks onto the other reindexed blocks, so nothing is brought into memory. CI is green and it applies cleanly to `main`.

I have also opened #2636 (`pairwise=True` for `concat_on_disk`, closes #2414), which touches the same file but is independent; whichever lands second needs only a trivial rebase.

PR #2561, opened 2026-07-16
fix: fall back to asarray when DLPack export of a read-only index fails; buffer error fix

amalia-k510:
Was running into issues with certain scanpy tests when testing with jax arrays. Example:

```
FAILED tests/test_pca.py::test_mask_defaults[jax_array-float32] - BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack (supported by newer DLPack version).
FAILED tests/test_pca.py::test_mask_var_argument_equivalence[float32-jax_array] - BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack (supported by newer DLPack version).
FAILED tests/test_pca.py::test_mask_defaults[jax_array-float64] - BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack (supported by newer DLPack version).
FAILED tests/test_pca.py::test_mask_var_argument_equivalence[float64-jax_array] - BufferError: Cannot export readonly array since signalling readonly is unsupported by DLPack (supported by newer DLPack version).
```

This is the same problem we already handle in `__array__` a few lines above, where the comment notes we can't round-trip jax arrays through DLPack and should let numpy handle it in `asarray` instead.

The fix keeps `from_dlpack` as the fast path and only falls back to `asarray` if it actually fails. The array API spec says `from_dlpack` raises `BufferError` in exactly this case, so catching it is the intended way to handle it.

### DIFF
--- a/src/anndata/compat/__init__.py
+++ b/src/anndata/compat/__init__.py
@@ -87,7 +87,11 @@ class IndexManager:
             existing_xp = existing.__array_namespace__()
             if existing_xp is xp:
                 return existing
-            return xp.from_dlpack(existing)
+            try:
+                return xp.from_dlpack(existing)
+            except BufferError:
+                # numpy#20742: read-only arrays can’t be exported via DLPack
+                return xp.asarray(existing)
         self.add_array(xp.from_dlpack(src_arr, copy=True))
         return self._manager[device]
 


flying-sheep:
Ah yes, this is a JAX bug (https://github.com/jax-ml/jax/issues/40789), but we circumvented it until #2071 (0.13.0).

The spec is pretty clear: https://github.com/data-apis/array-api/issues/191

See **Notes** at the end of https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html:

> ```py
> try:
>     x.__dlpack__(max_version=(1, 0), ...)
>     # if it succeeds, store info from the capsule named "dltensor_versioned",
>     # and need to set the name to "used_dltensor_versioned" when we're done
> except TypeError:
>     x.__dlpack__(...)
> ```
> 
> This logic is also applicable to handling of the new `dl_device` and `copy` keywords.
> 
> DLPack 1.0 added a flag to indicate that the array is read-only (`DLPACK_FLAG_BITMASK_READ_ONLY`). A consumer that does not support read-only arrays should ignore this flag (this is preferred over raising an exception; the user is then responsible for ensuring the memory isn’t modified).

flying-sheep (approved):

Issue #2141, opened 2025-10-01
`filename` handling of zarr stores


ilan-gold:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of anndata.
- [x] (optional) I have confirmed this bug exists on the main branch of anndata.

### Report

Zarr stores have no consistent place for a root file handle and there are errors for in-memory stores when trying to access the "root path" (whatever that means)

Code:

```python
import zarr
from anndata._core.file_backing import filename

filename(zarr.group("foo.zarr"))
```

Traceback:

```pytb
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 4
      1 import zarr
      2 from anndata._core.file_backing import filename
----> 4 filename(zarr.group("foo.zarr"))

File ~/miniconda3/lib/python3.12/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

File ~/Projects/Theis/anndata/src/anndata/_core/file_backing.py:186, in _(x)
    183 @filename.register(ZarrArray)
    184 @filename.register(ZarrGroup)
    185 def _(x):
--> 186     return x.store.path

AttributeError: 'LocalStore' object has no attribute 'path'
```

xref: https://github.com/zarr-developers/zarr-python/issues/2266

### Versions

```python
zarr    3.1.3
anndata 0.13.0.dev41+g7883222dc
----    ----
PyYAML  6.0.3
parso   0.8.4
MarkupSafe      3.0.3
zstandard       0.25.0
asttokens       3.0.0
typing_extensions       4.15.0
Jinja2  3.1.6
xarray  2025.9.1
sphinxcontrib-htmlhelp  2.1.0
decorator       5.2.1
ipython 9.1.0
six     1.17.0
cloudpickle     3.1.1
tblib   3.1.0
wcwidth 0.2.13
psutil  7.1.0
setuptools      80.9.0
jedi    0.19.2
fast-array-utils        1.2.3
stack-data      0.6.3
sphinxcontrib-serializinghtml   2.0.0
pytz    2025.2
fsspec  2025.9.0
traitlets       5.14.3
packaging       25.0
scanpy  1.11.4
dask    2025.9.1
charset-normalizer      3.4.3
[...]

PR #2577, opened 2026-07-29
build: sync with cookiecutter-scverse template, adopt mypy

Zethson:
Syncs the repo config with [cookiecutter-scverse](https://github.com/scverse/cookiecutter-scverse) `main` (`707f3fa`, up from `v0.6.0`) and adopts the template's mypy hook.

**Template sync** — pre-commit hook set, `.editorconfig`, `biome.jsonc`, `Typing :: Typed`. pre-commit.ci is replaced by a `Pre-commit checks` CI job plus a `pre-commit` Dependabot ecosystem, which makes the standalone codespell workflow redundant. Deliberate deviation: taplo is kept over pyproject-fmt, since it also formats `hatch.toml`.

**Typing** — mypy over `src` and `tests`, configured in `[tool.mypy]`. Stubs for pandas, scipy, psutil, tqdm, jsonschema and boltons; installed-but-untyped dependencies are read from source via `follow_untyped_imports` rather than collapsing to `Any`; `ignore_missing_imports` is scoped to the optional dependencies absent at check time. No `# type: ignore` anywhere.

The rest of the commits are the fixes this surfaced. Notable real bugs: `filename()` raised `AttributeError` on every zarr node (`store.path` is gone in zarr v3); `_read_attr` raised `TypeError` whenever a default was passed; `obs_keys()`/`var_keys()` crashed on xarray-backed objects; `write_loom` emitted a `None`-named layer; `__sizeof__` crashed on COO in `uns`; `xp.ravel`/`xp.where(cond)`/`xp.array`/`xp.promote_types` are not in the array API standard; and four tests were asserting vacuously.


flying-sheep:
very cool! I think there are some parts that probably need to be refactored to support static typing properly, e.g. the aligned mapping stuff might be too cute for type checkers to understand. Mixins, associated classes which are generic and therefore would need higher-kinded typevars … maybe we can work around all this without a refactor, but maybe not.

I’ve ranted against Python’s type checker but watched this recently: https://www.youtube.com/watch?v=0BXuYlNrUmE

- Python’s type checking is still underpowered – having e.g. a useful numpy-style signature with `dtype` parameter is a huge PITA
- But maybe some code bases can be refactored to simply use simpler types

Zethson:
@flying-sheep glad you like it already! I actually had wanted to write a long message with some context so here we go:

I'm pushing for this because I am currently trying to ensure that all of our projects adhere to the latest template version. This is a pet peeve of mine and I just want to get it done. While doing so, I thought - why not move immediately to the upcoming version 1.0 with typing? Doing so for pertpy, I noticed that neither AnnData or scanpy are fully typed yet (or at least labeled as such via py.typed) making this effort a bit less meaningful. So my plan is to slowly get our fundamental packages typed (so also MuData & SpatialData soon) and then go from there.

This PR therefore mixes two concerns:
1. From template 0.old to 1.0.0 pre-release
2. Adopting mypy

If you insist, we can disentangle these two things into two PRs but I'm fine with just being pragmatic and solving everything in one go.

Implementation wise, this is mostly a **human guided** Claude Opus 5 high effort PR that cost a few hundred $ in tokens. The initial mypy run gave us ~2000 errors. I am worried that this would be too many errors to fix by hand and therefore this would have never been done. So I think it's a best effort approach to start with this LLM based PR.

@flying-sheep @ilan-gold As a result of that, I suggest that I'll have another look at the PR to fix anything I notice to be off and then hand this PR over to you. I don't want to be the "just did some LLM PR and then let others do the actual work aka reviewing it" but I think your advanced typing experience is a little bit beyond mine and you know the code base better.

I'd be happy if we could give this PR priority, please because it touches so many places. If we let this PR sit for too long, we'll keep fighting PR conflicts and end up wasting our time.

Does this sound okay with you?

I'll have a look now so after lunch, this PR will be ready for you.
Thanks!

Zethson:
Implementation notes:
1. This PR removes codespell because the template doesn't have it. Let me know if you insist on keeping it. No strong feelings.
2. pre-commit CI is expected to fail. After this PR is merged, we'll disable it.

docs/api.md
```diff
@@ -147,6 +147,7 @@ In particular, for pytorch-based models.
    :toctree: generated/
 
    experimental.AnnCollection
+   experimental.AnnCollectionView
```
Zethson: Do we want to expose this? Probably not right?

src/anndata/_core/aligned_mapping.py
```diff
@@ -24,22 +39,87 @@
...
 
 
-OneDIdx = Sequence[int] | Sequence[bool] | slice
-TwoDIdx = tuple[OneDIdx, OneDIdx]
+OneDIdx: TypeAlias = tuple[Idx1D]  # noqa: UP040
+TwoDIdx: TypeAlias = tuple[Idx1D, Idx1D]  # noqa: UP040
+# everything `coerce_array` lets through; `None` stands for a missing `.X` in `layers`
 # TODO: pd.DataFrame only allowed in AxisArrays?
-Value = pd.DataFrame | CSMatrix | CSArray | np.ndarray
+Value: TypeAlias = (  # noqa: UP040
```
Zethson: This seems to be a bit too generous for me as it's basically the AnnData equivalent of "everything"

flying-sheep:
> or at least labeled as such via py.typed

anndata is since very shortly ago!

> If you insist, we can disentangle these two things into two PRs but I'm fine with just being pragmatic and solving everything in one go.

I’m fine with that too, the typing change is the biggest part of the new template update.

> Does this sound okay with you?

Yeah. I’m happy with it because despite LLMs producing especially subtle logic errors compared to humans, I always get tripped up by “what’s a mypy bug and what’s an actual typing issue” myself, so maybe Claude did a better job than I could ever have done (even if I had spent weeks focusing on this).

If we find typing issues, we can fix them.

> I'll have a look now so after lunch, this PR will be ready for you.

great, just request my review once you’re ready.

src/anndata/_core/aligned_mapping.py
```diff
@@ -196,7 +280,7 @@ def __delitem__(self, key: K) -> None:
         with view_update(self.parent, self.attrname, ()) as new_mapping:
             del new_mapping[key]
 
-    def __contains__(self, key: K) -> bool:
+    def __contains__(self, key: object) -> bool:
```
Zethson, line 261: This looks a bit fishy. Also below where K just got replaced by object

src/anndata/_core/aligned_mapping.py
```diff
@@ -419,7 +504,7 @@ def __len__(self) -> int:
             return data_length + 1
         return data_length
 
-    def __contains__(self, key: str | None) -> bool:
+    def __contains__(self, key: object) -> bool:
```
Zethson, line 491: Losing some precision here. Double check

src/anndata/_core/aligned_mapping.py
```diff
@@ -235,7 +323,7 @@ def __setitem__(self, key: K, value: Value):
         else:
             self._data[key] = value
 
-    def __contains__(self, key: K) -> bool:
+    def __contains__(self, key: object) -> bool:
```
Zethson, line 305: Not sure about this one.

src/anndata/_core/file_backing.py
```diff
@@ -191,7 +199,8 @@ def _(x: h5py.Group | h5py.Dataset) -> str:
 
 @filename.register(zarr.Array | zarr.Group)
 def _(x: zarr.Array | zarr.Group) -> str:
-    return x.store.path
+    # Zarr v3 stores have no generic path, but their `str` is a unique URL
+    return str(x.store)
```
Zethson: Is this a change in behavior? Careful

src/anndata/_core/merge.py
```diff
@@ -682,27 +686,26 @@ def _apply_to_array_api(
...
-    ) -> CSMatrix:
+    ) -> CSMatrix | CSArray:
         if isinstance(el, CupySparseMatrix):
             from cupyx.scipy import sparse
         else:
             from scipy import sparse
-        import array_api_compat
 
-        xp = array_api_compat.array_namespace(el.data)
+        xp = el.data.__array_namespace__()
```
Zethson: Does this change behavior? Check

src/anndata/_core/sparse_dataset.py
```diff
@@ -378,75 +430,59 @@ def shape(self) -> tuple[int, int]:
...
-        NotImplementedError
-            If the matrix to append is not one of :class:`~scipy.sparse.csr_array`, :class:`~scipy.sparse.csc_array`, :class:`~scipy.sparse.csr_matrix`, or :class:`~scipy.sparse.csc_matrix`.
-        ValueError
-            If both the on-disk and to-append matrices are not of the same format i.e., `csr` or `csc`.
-        OverflowError
-            If the underlying data store has a 32 bit indptr, and the new matrix is too large to fit in it i.e., would cause a 64 bit `indptr` to be written.
-        AssertionError
-            If the on-disk data does not have `csc` or `csr` format.
-        """
+    def append(  # noqa: PLR0912, PLR0915
```
Zethson, line 485: Why did the docstring get deleted here?

src/anndata/_core/views.py
```diff
@@ -165,11 +181,14 @@ def convert_all(arrs: Iterable[np.ndarray]) -> Iterable[np.ndarray]:
...
         return results[0] if len(results) == 1 else results
 
-    def keys(self) -> KeysView[str]:
+    def keys(self) -> tuple[str, ...]:
         # it’s a structured array
-        return self.dtype.names
+        return self.dtype.names or ()
 
-    def copy(self, order: str = "C") -> np.ndarray:
+    # deliberately not `Self`: copying a view has to stop it being one
```
Zethson:
```suggestion
```

src/anndata/_core/xarray.py
```diff
@@ -214,7 +217,10 @@ def __getitem__(
         self, key: Mapping[Any, Any] | Hashable | Iterable[Hashable]
     ) -> Dataset2D | XDataArray:
         ret = self.ds.__getitem__(key)
-        if is_empty := (len(key) == 0 and not isinstance(key, tuple)):  # empty Dataset
+        # empty Dataset
```
Zethson:
```suggestion
```

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -118,22 +120,22 @@ def read_sparse_as_dask(
     elem: h5py.Group | zarr.Group,
     *,
     _reader: LazyReader,
-    chunks: tuple[int, ...] | None = None,  # only tuple[int, int] is supported here
+    chunks: tuple[int | None, ...]
+    | None = None,  # only tuple[int, int] is supported here
```
Zethson: This comment was here b4 but idk whether it makes sense because the typing should tell us the truth and not such a comment

src/anndata/_io/specs/methods.py
```diff
@@ -154,7 +148,7 @@ def _to_cpu_mem_wrapper(write_func):
     def wrapper(
         f,
         k,
-        cupy_val: CupyArray | CupyCSMatrix,
+        cupy_val,
```
Zethson: why did typing get removed here?

src/anndata/_io/specs/methods.py
```diff
@@ -345,9 +338,12 @@ def write_anndata(
             continue
         _check_has_no_slash_key(sub_key, elem)
         if sub_key == "layers":
-            if None in elem:
-                _writer.write_elem(g, "X", elem[None], dataset_kwargs=dataset_kwargs)
-            elem = {k: v for k, v in elem.items() if k is not None}
+            assert isinstance(elem, MutableMapping)
+            # `layers` is keyed by `str | None`, where `None` is `X`
```
Zethson:
```suggestion
```

src/anndata/_io/specs/methods.py
```diff
@@ -1285,23 +1283,21 @@ def write_nullable(
...
+for store_type, array_type in product(_STORE_TYPES, PANDAS_STRING_ARRAY_TYPES):
     _REGISTRY.register_write(
         store_type, array_type, IOSpec("nullable-string-array", "0.1.0")
     )(write_nullable)
 
 
 class _BaseMaskedArray(Protocol):
     def __call__(
-        self, values: NDArray[np.number], /, *, mask: NDArray[np.bool_]
+        self, values: NDArray[Any], /, *, mask: NDArray[np.bool_]
```
Zethson, line 1302: Can we avoid the Any here?

src/anndata/_io/specs/methods.py
```diff
@@ -1391,11 +1387,14 @@ def write_scalar_zarr(
     # and error out in v3
     dataset_kwargs = _remove_scalar_compression_args(dataset_kwargs)
 
+    filters: list[VLenUTF8] | None
+    dtype: np.dtype
+    fill_value: str | None
     match f.metadata.zarr_format, value:
         case 2, str():
-            filters, dtype, fill_value = [VLenUTF8()], VariableLengthUTF8(), ""
+            filters, dtype, fill_value = [VLenUTF8()], np.dtypes.StringDType(), ""
```
Zethson: Is this a change in behavior?

src/anndata/_io/read.py
```diff
@@ -426,11 +428,11 @@ def _read_text(  # noqa: PLR0912, PLR0915
     first_column_names: bool | None,
     dtype: str,
 ) -> AnnData:
-    comments = []
-    data = []
+    comments: list[str] = []
```
Zethson, line 425: Can this also be a Sequence? Or Iterable?

src/anndata/_io/read.py
```diff
@@ -426,11 +428,11 @@ def _read_text(  # noqa: PLR0912, PLR0915
...
+    rows: list[np.ndarray] = []
     lines = _iter_lines(f)
-    col_names = []
-    row_names = []
+    col_names: list[str] = []
```
Zethson, line 428: Can this also be a Sequence? Or Iterable?

src/anndata/_io/read.py
```diff
@@ -426,11 +428,11 @@ def _read_text(  # noqa: PLR0912, PLR0915
...
+    row_names: list[str] = []
```
Zethson, line 429: Can this also be a Sequence? Or Iterable?

src/anndata/acc/__init__.py
```diff
@@ -555,26 +569,32 @@ def isin(self, data: MuData | AnnData, idx: Idx2D | None = None) -> bool:
...
     ) -> _XDataType | InMemoryArray:
         full: _XDataType = getattr(data, f"{self.dim}p")[self.k]
         if idx is NO_IDX:
             return full
-        df = cast("pd.DataFrame", getattr(data, self.dim))
-        # TODO: remove wrapping in [] when dropping scipy <1.14
-        iloc = tuple([df.index.get_loc(i)] if isinstance(i, str) else i for i in idx)
-        return self._maybe_flatten(idx, full[iloc])
+        index: pd.Index = getattr(data, self.dim).index
+        # TODO: remove wrapping in an array when dropping scipy <1.14
```
Zethson: We've already dropped support for that right? So we can get rid of that I guess

src/anndata/experimental/merge.py
```diff
@@ -180,9 +208,9 @@ def _df_index(df: zarr.Group | h5py.Group) -> pd.Index:
 def write_concat_dense(  # noqa: PLR0917
     arrays: Sequence[zarr.Array | h5py.Dataset],
     output_group: zarr.Group | h5py.Group,
-    output_path: zarr.Group | h5py.Group,
+    output_path: str,
```
Zethson, line 213: Is this really supposed to be str and not zarr.Group | h5py.Group like we had b4?

src/anndata/experimental/merge.py
```diff
@@ -316,54 +343,58 @@ def _write_concat_arrays(  # noqa: PLR0913, PLR0917
...
+        raise NotImplementedError(msg)
 
 
 def _write_concat_sequence(  # noqa: PLR0913, PLR0917
     arrays: Sequence[
         pd.DataFrame | BaseCompressedSparseDataset | h5py.Dataset | zarr.Array
     ],
     output_group: zarr.Group | h5py.Group,
-    output_path: str | Path,
+    output_path: str,
```
Zethson, line 380: Should we keep Path support here?

src/anndata/tests/helpers.py
```diff
@@ -203,7 +215,7 @@ def gen_random_column(  # noqa: PLR0911
 
 def gen_typed_df(
     n: int,
-    index: pd.Index[str] | None = None,
+    index: pd.Index[Any] | None = None,
```
Zethson: Can we have something more specific here than Any?

.cruft.json
```diff
@@ -1,43 +1,44 @@
...
-                "trim_blocks": true
-            },
-            "_template": "https://github.com/scverse/cookiecutter-scverse",
-            "_commit": "d383d94fadff9e4e6fdb59d77c68cb900d7cedec"
-        }
-    },
-    "directory": null
+  "template": "https://github.com/scverse/cookiecutter-scverse",
+  "commit": "707f3faab836f9272912f92188b63b31da70c064",
+  "checkout": "main",
```
Zethson, line 4: This should probably not be main but 0.8.0 or something for now. There'll just be less merge conflicts later when the 1.0.0 sync runs

.pre-commit-config.yaml
```diff
@@ -1,34 +1,56 @@
...
   - repo: https://github.com/ComPWA/taplo-pre-commit
     rev: v0.9.3
     hooks:
       - id: taplo-format
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.16.0
+    hooks:
+      - id: ruff-check
+        types_or: [python, pyi, jupyter]
+        args: [--fix, --exit-non-zero-on-fix]
```
Zethson, line 21: @flying-sheep I personally also have unsafe fixes enabled and haven't run into issues but that's not a big deal

pyproject.toml
```diff
@@ -266,6 +278,30 @@ strict = true
 max-args = 7
 max-positional-args = 5
 
+[tool.mypy]
+mypy_path = "src"
```
Zethson: Do we need this? We have some config in the pre-commit file

docs/api.md
flying-sheep: I never really engaged with AnnCollection … I’d say that if the idea is that `AnnCollection` and `AnnCollectionView` basically have the same API but one is a view of the other, we could maybe just add a manual stub instead of autogenerating it, like:

```markdown
:::{class} experimental.AnnCollectionView

A view of {class}`experimental.AnnCollection` with most of the same functions and attributes.
:::
```

src/anndata/_core/aligned_mapping.py
flying-sheep: No this is correct if the implementation actually deals with any object. Think:

```py
d: set[str] = {1, 2, 3}

"x" in d  # not a TypeError, just False
```

It doesn’t make much *sense* to do that check as it’s statically known to be False, but it’s not technically wrong either.

src/anndata/_core/aligned_mapping.py
flying-sheep: see above

src/anndata/_core/file_backing.py
flying-sheep: this is #2141, this `@register` is simply broken and unused.

We should probably remove it while fixing the type of `def filename` to not include `zarr.Group`.

flying-sheep (changes requested):
needs the setup from https://github.com/scverse/cookiecutter-scverse/pull/540

after this PR also `strict = true`

Zethson:
@flying-sheep you're making these changes now right? Then I won't also change the setup. I was just about to address your last request.

docs/api.md
Zethson: Honestly, I think in the near future, we should just completely remove it and point people to annbatch (as we've started doing). We're just dragging this around

src/anndata/_core/aligned_mapping.py
flying-sheep: IIRC we allow slightly different unions in `{obs,var}m` vs `layers`, but maybe not.

This is certainly more correct than before, my main concern is why this isn’t integrated with the data types in `typing.py`.

Ideally this is just `anndata.typing._ArrayDataStructureTypes` which we could now make public.

flying-sheep:
I’m on it, getting a little hands-on helps me understand the changes better anyway.

you could do the AnnCollection thing that’s pretty self-contained.

.github/workflows/codespell.yml
flying-sheep, line 1: why? this did sometimes find typos.

Zethson: I had commented on that above. It's not a part of the template but we can keep it if you like. No problem.

src/anndata/_core/aligned_mapping.py
```diff
@@ -24,26 +28,69 @@
...
+
+
+OneDIdx: TypeAlias = tuple[Idx1D]  # noqa: UP040
+TwoDIdx: TypeAlias = tuple[Idx1D, Idx1D]  # noqa: UP040
+
+
+def _copy_value(v: AlignedArray) -> AlignedArray:
+    # awkward arrays have immutable buffers, and on-disk elements aren’t copyable
+    if isinstance(v, AwkArray | NoneType | ON_DISK_TYPES):
+        return copy(v)
```
flying-sheep: so this changed behavior from `a.copy()` → `copy(a)` for on-disk types. What’s the impact of this? Is there a new test for this?

src/anndata/_core/aligned_mapping.py
```diff
@@ -24,26 +28,69 @@
...
+    if isinstance(v, pd.DataFrame | Dataset2D | XDataArray):
+        return v.copy()
+    return _copy_array(v)
+
+
+def _copy_array(a: _XDataType) -> InMemoryArray:
+    """An in-memory copy of an array element."""
+    if isinstance(a, ON_DISK_TYPES):
+        return to_memory(a)
```
flying-sheep: this is never hit right?

src/anndata/_core/aligned_mapping.py
```diff
@@ -280,12 +335,11 @@ def to_df(self) -> pd.DataFrame:
         """Convert to pandas dataframe."""
         df = pd.DataFrame(index=self.dim_names)
         for key in self.keys():
-            value = self[key]
-            for icolumn, column in enumerate(value.T):
+            for icolumn, column in enumerate(asarray(self[key]).T):
```
flying-sheep, line 350: if `self[key]` is a DataFrame already, this will nuke all column types right?

flying-sheep (commented):
OK, reviewed aligned_mapping

src/anndata/_core/anndata.py
```diff
@@ -1308,9 +1358,7 @@ def to_df(self, layer: str | None = None) -> pd.DataFrame:
             raise ValueError(msg)
         else:
             X = self.X
-        if issparse(X):
-            X = X.toarray()
-        return pd.DataFrame(X, index=self.obs_names, columns=self.var_names)
+        return pd.DataFrame(asarray(X), index=self.obs_names, columns=self.var_names)
```
flying-sheep: again, what if X is already a data frame here?

src/anndata/_core/file_backing.py
```diff
@@ -66,29 +67,36 @@ def __repr__(self) -> str:
         else:
             return f"Backing file manager of file {self.filename}."
 
+    @property
+    def _open_file(self) -> h5py.File:
+        if self._file is None:
+            msg = "Backing file is not open."
+            raise ValueError(msg)
```
flying-sheep: ValueError? Which value?

src/anndata/_core/aligned_mapping.py
Zethson: You're saying it'll never be hit because we won't have on-disk types in memory anyways?

src/anndata/_core/aligned_mapping.py
Zethson: You are correct. Good catch!

src/anndata/_core/anndata.py
Zethson: Yup, same issue.

src/anndata/_core/file_backing.py
Zethson: IIRC h5py also throws a ValueError in such cases.
I mean we can also throw a Runtime error or so. A custom Exception might be even better.

src/anndata/_core/aligned_mapping.py
Zethson: I think behavior won't change, because none of the on-disk types define `.copy()`. Before, we just raised an `AttributeError`.

src/anndata/acc/__init__.py
```diff
@@ -29,12 +32,23 @@
...
 
+    class NoIdx(Enum):
+        """Type of the :data:`NO_IDX` sentinel."""
+
+        NO_IDX = "NO_IDX"
+
+elif sys.version_info >= (3, 15):
+    from typing import sentinel as NoIdx
+else:
+    from typing_extensions import sentinel as NoIdx
```
flying-sheep: nope, that was perfect as-is.

sentinels are both types and values, and the code should stay that way.

src/anndata/_core/file_backing.py
flying-sheep: I’d say RuntimeError for now, and we introduce `AnnDataError` (including one or more `RuntimeError` subclasses) later

src/anndata/_core/aligned_mapping.py
flying-sheep: no I’m saying there are no call sites where this branch can be hit logically. I’d bet that coverage for this branch is missing, because unless I misinterpreted something, it can’t ever possibly be hit.

src/anndata/_core/aligned_mapping.py
flying-sheep: and now we won’t, or does `copy` internally also call `.copy()`? I’m confused.

src/anndata/_core/aligned_mapping.py
Zethson: I think we're using different words but mean the same thing.

src/anndata/_core/aligned_mapping.py
Zethson: `copy` doesn't call `.copy()`. So it might e.g. fall back to the pickle protocol, and h5py can't do that. So we could instead just return v for on-disk types.

src/anndata/_io/specs/lazy_methods.py
```diff
@@ -375,8 +398,12 @@ def read_nullable(
     )
 
 
-for dtype in ["integer", "boolean", "string-array"]:
-    for group_type in [zarr.Group, h5py.Group]:
-        _LAZY_REGISTRY.register_read(group_type, IOSpec(f"nullable-{dtype}", "0.1.0"))(
-            partial(read_nullable, encoding_type=f"nullable-{dtype}")
+_NULLABLE_ENCODING_TYPES: list[
+    Literal["nullable-integer", "nullable-boolean", "nullable-string-array"]
+] = ["nullable-integer", "nullable-boolean", "nullable-string-array"]
```
flying-sheep: @Zethson FYI: this is commonly written as `CONST: Final = ("a", "b", "c")`

src/anndata/_io/specs/methods.py
flying-sheep: I wondered the same and since you didn’t answer it here: no, because `pd.arrays.IntegerArray` and friends specify a plain `np.ndarray` here which is `NDArray[Any]`.

src/anndata/_io/read.py
flying-sheep: no, these aren’t parameters, they are variables, and for these you want concrete types.

(as you do for return values)

src/anndata/acc/__init__.py
```diff
@@ -555,26 +570,35 @@ def isin(self, data: MuData | AnnData, idx: Idx2D | None = None) -> bool:
...
+            case str() as i, slice() as j:
+                sub = full[np.array([index.get_loc(i)]), j]
+            case slice() as i, str() as j:
+                sub = full[i, np.array([index.get_loc(j)])]
+            case slice() as i, slice() as j:
+                sub = full[i, j]
+            case _:  # pragma: no cover
+                msg = f"Invalid index: {idx}"
+                raise TypeError(msg)
+        return self._maybe_flatten(idx, sub)
```
flying-sheep, line 596: I don’t quite get that change and how the comment is related

pyproject.toml
```diff
@@ -289,6 +289,7 @@ max-positional-args = 5
 [tool.mypy]
 mypy_path = [ "src" ]
 explicit_package_bases = true
+warn_redundant_casts = true
```
Zethson, line 292: Ohh, I didn't know this option exists. Super nice!

tests/accessors/test_accessors.py
```diff
@@ -336,7 +340,7 @@ def test_resolve_matrix(
         pytest.param("obsp.g[c1,:]", False, id="obsp-vec-as-matrix"),
     ],
 )
-def test_resolve_vec_mismatch(spec: str, *, vec: bool) -> None:
+def test_resolve_vec_mismatch(spec: str, *, vec: Literal[True, False]) -> None:
```
flying-sheep, line 343: great job mypy, what’s the 3rd bool? 😄

pyproject.toml
flying-sheep: not that it works properly! https://github.com/python/mypy/issues/21796

.github/workflows/codespell.yml
flying-sheep: yes let’s. why should we remove functionality just because it’s not in the template?

src/anndata/_core/aligned_mapping.py
flying-sheep: ah, it can actually be hit because Raw calls `self._X = _copy_array(adata.X)`

.github/workflows/codespell.yml
flying-sheep: ah, it’s a pre-commit hook anyway, so this was duplicate.

docs/api.md
flying-sheep: I agree, let’s do it in a different PR though

flying-sheep (approved):

flying-sheep:
@Zethson you forgot autofix, could you please implement that?

Zethson:
@flying-sheep do you mean the ruff config option?

PR #2584, opened 2026-08-03
fix: remove hard-coded scheduler from `da.store`

joshua-gould:
`test_dask_distributed_write `tests writing to zarr store using dask so no new tests added


flying-sheep (changes requested):
The change you’re reverting was made intentionally here: https://github.com/scverse/anndata/pull/2183

You can see there which dask versions have the bug it circumvents. Makes sense to not hard-code it for dask versions where it’s fixed, but you should make it so that `threads` is still hardcoded for afflicted dask version.

joshua-gould:
The bug was caused by the anndata lock, not by the scheduler so it should be safe to remove the scheduler as the lock has been removed.

flying-sheep:
Not only that, there’s more, see test failures. You could probably make this work by leaving the hardcoded scheduler in place only for `is_h5 or Version(version("dask")) < ...` re-introducing the version gate from the PR I linked.

Then when using modern dask and zarr, the user scheduler is respected.

joshua-gould:
Sorry, I didn't realize this code block was used for saving h5 files as well. I updated the pull request to use threaded scheduler with h5. Thanks.

flying-sheep (dismissed):
still missing the dask version gate, with that I can merge it

joshua-gould:
Added dask version gate. Thanks.

flying-sheep:
OK, looks good to me now, but @ilan-gold said that the threading solution seems safer.

So now that we no longer support zarr 2, maybe that’s obsolete, but let’s ask Ilan.

src/anndata/_io/specs/methods.py
```diff
@@ -586,7 +586,11 @@ def write_basic_dask_dask_dense(
             dataset_kwargs, format=f.metadata.zarr_format
         ) as dataset_kwargs:
             g = f.require_array(k, shape=elem.shape, dtype=elem.dtype, **dataset_kwargs)
-    da.store(elem, g, scheduler="threads")
+    # use threaded scheduler with dask<=2025.3.0 avoid "Could not serialize object of type HighLevelGraph" error
+    if isinstance(f, h5py.Group) or Version(version("dask")) <= Version("2025.3.0"):
+        da.store(elem, g, scheduler="threads")
+    else:
+        da.store(elem, g)
```
ilan-gold, line 593: Yeah, so with zarr sharding and processes, AFAIK, this is not safe.  I'm not sure `"threads"` is particularly safe either, but I think it's better.  @d-v-b maybe knows? Or @melonora ?

joshua-gould: Maybe the xarray developers would have some insight? I checked their code and they do not hardcode the scheduler.

joshua-gould: Doesn't https://github.com/dask/dask/blob/ba5045e001d67decd6a83a0f50f666a5e3845250/dask/array/core.py#L1196 ensure safety across processes or threads?

joshua-gould: Also see https://github.com/dask/dask/pull/12105 and https://github.com/zarr-developers/zarr-python/issues/3514.

melonora: Not certain why I only see a notification just now. To reply https://github.com/dask/dask/blob/ba5045e001d67decd6a83a0f50f666a5e3845250/dask/array/core.py#L1196, as far as I am aware this actually locks the whole array and not individual blocks. I ran into this when implementing the direct writing of sharding with `da.to_zarr`. Whether it is safe depends on whether shards and chunks are aligned, but I believe it should throw an error in any case when that is not the case.

flying-sheep: I have no experience here. @melonora could you please check? If your belief being true is all that’s needed here we’re close to wrapping this up.

ilan-gold: Any updated here @melonora ? Maybe @joshua-gould could just post a script a verifying one way or the other that this safety check exists i.e., the error is thrown when not aligned?

joshua-gould: Here's an example that shows when chunks or shards are not aligned, data is corrupted:
```
import dask.array as da

a = da.arange(100).reshape((10, 10)).rechunk((2, 2))
da.to_zarr(a, "test1.zarr", overwrite=True)
print((a == da.from_zarr("test1.zarr")).all().compute())  # works

da.to_zarr(a, "test2.zarr", chunks=(4, 4), overwrite=True)
print((a == da.from_zarr("test2.zarr")).all().compute())  # corrupted

da.to_zarr(a, "test3.zarr", shards=(4, 4), overwrite=True)
print((a == da.from_zarr("test3.zarr")).all().compute())  # corrupted
```

joshua-gould: This is a bug in da.to_zarr (see https://github.com/dask/dask/issues/12583). Using anndata write_zarr, which creates a zarr array and then uses da.store works as expected. Example:

```
import dask.array as da
import zarr

a = da.arange(100).reshape((10, 10)).rechunk((2, 2))

da.store(a, zarr.create_array("test2.zarr", shape=a.shape, dtype=a.dtype, chunks=(4, 4), overwrite=True))
print((a == da.from_zarr("test2.zarr")).all().compute())  # works

da.store(a, zarr.create_array("test3.zarr", shape=a.shape, dtype=a.dtype, chunks=(2, 2), shards=(4, 4), overwrite=True))
print((a == da.from_zarr("test3.zarr")).all().compute())  # works
```

ilan-gold: Interesting, I tried this also with adding 

```python
cluster = dd.LocalCluster(n_workers=2)
client = dd.Client(cluster)
```

so it seems to work with processes as well. I'm pretty confident about this given your bug reports there which seem reasonable and the proposed fix in https://github.com/dask/dask/pull/12588.  Please keep an eye on this @joshua-gould (you seem to be) and try to keep anndata up to date here :) Thanks!

ilan-gold (approved):

PR #2589, opened 2026-08-05
docs: Fix two docstring parameter names

karpovantonme:
Two docstrings name a parameter the signature does not have.

`_io/specs/methods.py`, `zarr_v3_compressor_compat`: documents `dataset_kwarg`, the parameter is `dataset_kwargs`. A missing `s`.

`_types.py`, `ReduceFunc.__call__`: documents `ref_acc` while the third keyword argument is `attr_name`. The old description ("A reference to help uses distinguish where they are in the `AnnData` object") does not quite fit `attr_name`, so I reworded it to say what that argument is. If the wording is off, tell me and I will use yours.

Documentation only, no behaviour change.

- [ ] Closes #
- [ ] Tests added
- [x] Release note not necessary because: docstring text only, no user-visible change.

### DIFF
--- a/src/anndata/_io/specs/methods.py
+++ b/src/anndata/_io/specs/methods.py
@@ -106,7 +106,7 @@ def zarr_v3_compressor_compat(dataset_kwargs: dict) -> dict:
 
     Parameters
     ----------
-    dataset_kwarg
+    dataset_kwargs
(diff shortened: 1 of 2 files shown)


flying-sheep:
thanks!

PR #2594, opened 2026-08-06
fix: handle sparse consistently

flying-sheep:
We should have run GPU tests for the typing bug, but this way I fixed things a bit more robustly.

- [x] Release note not necessary because: fix for unreleased bug

### DIFF
--- a/src/anndata/_io/zarr.py
+++ b/src/anndata/_io/zarr.py
@@ -69,7 +69,7 @@ def write_zarr(
     ) -> None:
         if (
             chunks is not None
-            and not isinstance(elem, sparse.spmatrix)
+            and not isinstance(elem, sparse.sparray | sparse.spmatrix)
             and elem_name.lstrip("/") == "X"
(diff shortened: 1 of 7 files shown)


src/anndata/_io/zarr.py
```diff
@@ -69,7 +69,7 @@ def callback(
     ) -> None:
         if (
             chunks is not None
-            and not isinstance(elem, sparse.spmatrix)
+            and not isinstance(elem, sparse.sparray | sparse.spmatrix)
```
flying-sheep, line 72: Unclear why this wasn’t the case. Any idea @ilan-gold?

flying-sheep:
this fixes GPU tests so I’m merging it, but please still take a look in case there are issues.

src/anndata/_io/zarr.py
ilan-gold: Just a mistake!

PR #2599, opened 2026-08-11
test: sort parametrized set values for pytest-xdist

### LINKED ISSUE #2585: Error running tests with hatch
### Question

I get the following error when I try to run tests using hatch
```
hatch test -p
ERROR gw1 - Different tests were collected between gw2 and gw1. The difference is:
ERROR gw0 - Different tests were collected between gw2 and gw0. The difference is:
ERROR gw3 - Different tests were collected between gw2 and gw3. The difference is:
ERROR gw6 - Different tests were collected between gw2 and gw6. The difference is:
ERROR gw4 - Different tests were collected between gw2 and gw4. The difference is:
ERROR gw5 - Different tests were collected between gw2 and gw5. The difference is:
ERROR gw7 - Different tests were collected between gw2 and gw7. The difference is:
```

joshua-gould:
- Closes #2585 

- Release note not necessary because: only fixes running tests using pytest-xdist

See https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#order-and-amount-of-test-must-be-consistent

### DIFF
--- a/tests/test_readwrite.py
+++ b/tests/test_readwrite.py
@@ -996,7 +996,7 @@ def test_h5py_attr_limit(tmp_path):
 
 
 @pytest.mark.parametrize(
-    "elem_key", set(get_literal_members(AnnDataElem)) - {"raw", "X"}
+    "elem_key", sorted(set(get_literal_members(AnnDataElem)) - {"raw", "X"})
 )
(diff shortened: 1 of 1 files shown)


flying-sheep:
thanks!

PR #2602, opened 2026-08-13
style: use pyproject-fmt

### LINKED ISSUE #2337: use pyproject-fmt

we should do this right before the 0.13 release since it changes the pyproject.toml a lot, and backporting wouldn’t be fun afterwards

just making this issue and adding it to the milestone so we don’t forget.

flying-sheep:
- [x] Closes #2337
- [x] Tests added


pyproject.toml
```diff
@@ -257,41 +187,51 @@ ignore = [
...
+# base64 blobs (e.g. images in notebook outputs)
+ignore-regex = "[A-Za-z0-9+/]{100,}={0,2}"
+skip = ".git,*.pdf,*.svg"
+
+[tool.pyproject-fmt]
+indent = 4
+table_format = "short"
+expand_tables = [
+    "tool.ruff.lint.flake8-tidy-imports.banned-api",
+]
```
flying-sheep, line 227: this is the only manual edit to this file

PR #2605, opened 2026-08-13
fix: stricter accessor validation

### LINKED ISSUE #2604: some problems regarding the string shortcuts for accessors
### Question

Hi,

So I have some questions and issues related to the string format

## Early X return
- For the special case of X we should just not return early as I mentioned here: https://github.com/scverse/scanpy/pull/4288#discussion_r3767471388

## Dot as a separator

This is a preference but if I had an objection it would be because seurat v5 uses `.` as a separator but also we'd be asking for the user to again modify their files like in `/`. While we still have `/` unavailable, most reasonable approach for me would be use that as a separator.

 ## Positional vs df index handling in obsm/varm

We currently have 
```python
from anndata.acc import A

adata = ad.AnnData(np.zeros((3, 2), dtype="f4"))
adata.obsm["df"] = pd.DataFrame({"PC1": [1, 2, 3]}, index=adata.obs_names)

ref = A.obsm["df"][:, 0]
assert ref in adata   # passes
adata[ref]            # pandas.errors.InvalidIndexError: (slice(None, None, None), slice(0, 1, None))

```
If we don't want to support this, we should give better errors and return False for  `ref in adata`.

Something like this might be a possible solution

```
obsm.arr.0     → position 0
obsm.df.PC1    → the column named PC1
```

flying-sheep:
- [x] Closes #2604
- [x] Tests added

- [x] Release note not necessary because: small internal change

### DIFF
--- a/tests/accessors/test_accessors.py
+++ b/tests/accessors/test_accessors.py
@@ -415,6 +415,18 @@ def test_resolve_vec_mismatch(spec: str, *, vec: Literal[True, False]) -> None:
 
 
 @pytest.mark.parametrize(
+    ("spec", "pat"),
+    [
+        pytest.param("Xenon", r"Unknown accessor name", id="x-like"),
+        pytest.param("Xenon[:,:]", r"Unknown accessor name", id="x-vec-like"),
+    ],
+)
+def test_resolve_errors(spec: str, *, pat: str) -> None:
+    with pytest.raises(ValueError, match=pat):
+        A.resolve(spec)
+
+
+@pytest.mark.parametrize(
     ("spec", "vec"),
     [("X", True), ("X[:,:]", False)],
     ids=["as-vec", "as-matrix"],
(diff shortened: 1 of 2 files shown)


tests/accessors/test_accessors.py
selmanozleyen, line 215:
```suggestion
pytest.param(lambda a: a.obsm["a"][True], id="obsm-bool"),
```
I think this one should fail right? But it doesn't in the json branch I think. So even if this is covered it would be nice to add this as a test

selmanozleyen: turns out this is because bool is a subclass of int in
```
 @staticmethod
  def process_idx(i: object, /) -> int:
      if not isinstance(i := _drop_multi_slice(i), int):
          msg = f"Unsupported index {i!r}"
          raise TypeError(msg)
      return i
```

flying-sheep: addressed in #2614

Issue #1505, opened 2024-05-22
`concat_on_disk` merge strategies are untested/not implemented


ilan-gold:
### Please describe your wishes and possible alternatives to achieve the desired result.

We should implement them so that they work properly.   I am not really sure if this is a bug since `concat_on_disk` is `experimental` and reading through the old PR, I don't see any discussion of it or any tests for it.

Here's an MVCE for `first` with default arguments, although by adding a `merge_type` argument to the current test suite for `concat_on_disk` you can see a full list of problems:

```python
from anndata.tests.helpers import (
    assert_equal,
    gen_adata,
)
import anndata as ad
import numpy as np
from scipy import sparse
import pandas as pd

GEN_ADATA_OOC_CONCAT_ARGS = dict(
    obsm_types=(
        sparse.csr_matrix,
        np.ndarray,
        pd.DataFrame,
    ),
    varm_types=(sparse.csr_matrix, np.ndarray, pd.DataFrame),
    layers_types=(sparse.spmatrix, np.ndarray, pd.DataFrame),
)

adata_1 = gen_adata((100, 200), **GEN_ADATA_OOC_CONCAT_ARGS)
adata_2 = gen_adata((50, 60), **GEN_ADATA_OOC_CONCAT_ARGS)
adata_1.write_h5ad('test_1.h5ad')
adata_2.write_h5ad('test_2.h5ad')
ad.experimental.concat_on_disk(['test_1.h5ad', 'test_2.h5ad'], 'merged.h5ad', merge="first")
```
raises:
```python
IORegistryError: No method registered for writing <class 'pandas.core.series.Series'> into <class 'h5py._hl.group.Group'>
Error raised while writing key 'var_cat' of <class 'h5py._hl.group.Group'> to /var
```
Here's a full list of the tests that fail from `test_anndatas_with_reindex` when `merge` is tested:

<details>
  <summary>Errors + tests</summary>

```python
FAILED tests/test_concatenate_disk.py::test_anndatas_with_reindex[0-array-inner-zarr-10-unique] - AssertionError: Error raised from element 'varp'.
FAILED tests/test_concatenate_disk.py::test_anndatas_with_reindex[0-array-inner-h5ad-10-same] - AssertionError: Error raised from element 'varp'.
[...]

PR #2609, opened 2026-08-15
fix: promote bool columns to nullable boolean when reindexing alt-axis annotations under outer join

srikarjy:
Part of #1505.

### The bug

With `join="outer"` and any union merge strategy (`first`/`unique`/`same`/`only`), the alternate-axis dataframe is reindexed to the union index. Reindexing a plain numpy `bool` column introduces missing values, which pandas promotes to `object` dtype holding `[True, False, nan]`. The IO registry then dispatches that column to the string writer, which fails with `TypeError: Can't implicitly convert non-string objects to strings` (h5py) or `expected unicode string, found True` (zarr) — the outer-join `TypeError` cluster from #1505's failure list.

```python

a = ad.AnnData(np.ones((3, 2)), var=pd.DataFrame({"flag": [True, False]}, index=["g1", "g2"]))
b = ad.AnnData(np.ones((3, 2)), var=pd.DataFrame({"flag": [True, True]}, index=["g2", "g3"]))
m = ad.concat([a, b], join="outer", merge="first")
m.write_h5ad("out.h5ad")  # TypeError before this PR
```

Notably this is not specific to `concat_on_disk`: in-memory `concat` produces the same unwritable `object` column, so the fix lives in the shared merge layer rather than `experimental/merge.py`.

### The fix

- `Reindexer._apply_to_df_like` promotes numpy `bool` columns to pandas nullable `boolean` before reindexing, only when the reindex actually introduces missing values (so inner joins / aligned indices keep plain `bool`, matching the behavior tested in `test_bool_promotion`). This reuses the previously-unused `np_bool_to_pd_bool_array` helper.
- `merge_dataframes` now routes its reindexing through `Reindexer` instead of a bare `df.reindex`, so `var`/`obs` and `varm`/`obsm` share the same path.

This mirrors how the concat axis already handles the same promotion (#1001), but the resulting dtype change (`bool` → nullable `boolean` in merged alt-axis annotations under outer join) is user-visible — happy to switch to a different fill policy if preferred.

### DIFF
--- a/docs/release-notes/2609.fix.md
+++ b/docs/release-notes/2609.fix.md
@@ -1,0 +1,1 @@
+Fix writing the result of `concat`/`concat_on_disk` with `join="outer"` and a union `merge` strategy when alternate-axis annotations contain `bool` columns, by promoting them to nullable `boolean` on reindexing instead of `object` {user}`srikarjy`
--- a/src/anndata/_core/merge.py
+++ b/src/anndata/_core/merge.py
@@ -593,6 +593,10 @@ class Reindexer:
     ):
         if fill_value is None:
             fill_value = np.nan
+        # Reindexing with missing values promotes numpy bool columns to object,
+        # which cannot be written to disk, so use pandas' nullable boolean dtype instead.
+        if isinstance(el, pd.DataFrame) and -1 in self.idx:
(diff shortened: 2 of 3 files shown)


flying-sheep:
thank you!

PR #2614, opened 2026-08-24
feat: allow JSON (de)serializing accessors

flying-sheep:
Backport to patch release since it’s a missing part

- [ ] Closes #
- [x] Tests added

### DIFF
--- a/src/anndata/acc/_parse_json.py
+++ b/src/anndata/acc/_parse_json.py
@@ -67,14 +67,12 @@ def to_json(  # noqa: PLR0911
     ref: AdRef | LayerAcc | MultiAcc | GraphAcc,
 ) -> list[str | int | None]:
     match ref:
-        case AdRef(LayerAcc(k), idx):
-            return ["layers", k, *_idx_2d_ser(idx)]
+        case AdRef(LayerAcc() | GraphAcc() as acc, idx):
+            return [*to_json(acc), *_idx_2d_ser(idx)]
         case AdRef(MetaAcc(dim), idx):
             return [dim, idx]
-        case AdRef(MultiAcc(dim, k), i):
-            return [f"{dim}m", k, i]
-        case AdRef(GraphAcc(dim, k), idx):
-            return [f"{dim}p", k, *_idx_2d_ser(idx)]
+        case AdRef(MultiAcc() as acc, i):
+            return [*to_json(acc), i]
         case LayerAcc(k):
             return ["layers", k]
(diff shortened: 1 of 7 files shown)


src/anndata/acc/acc-schema-v1.json
```diff
@@ -1,71 +1,107 @@
 {
+    "$id": "https://anndata.scverse.org/en/stable/acc-schema-v1.json",
```
selmanozleyen, line 2: since we are making these changes don't we need to bump the version? I am not familiar with what guarantees we will give so I am not sure about this one.

src/anndata/acc/_parse_json.py
```diff
@@ -46,16 +63,24 @@ def _idx_2d_ser(idx: Idx2D) -> tuple[str | None, str | None]:
...
+) -> list[str | int | None]:
+    match ref:
+        case AdRef(LayerAcc(k), idx):
+            return ["layers", k, *_idx_2d_ser(idx)]
+        case AdRef(MetaAcc(dim), idx):
+            return [dim, idx]
+        case AdRef(MultiAcc(dim, k), i):
+            return [f"{dim}m", k, i]
+        case AdRef(GraphAcc(dim, k), idx):
+            return [f"{dim}p", k, *_idx_2d_ser(idx)]
```
selmanozleyen:
```suggestion
        case AdRef(LayerAcc() | GraphAcc() as acc, idx):
            return [*to_json(acc), *_idx_2d_ser(idx)]
        case AdRef(MetaAcc(dim), idx):
            return [dim, idx]
        case AdRef(MultiAcc() as acc, i):
            return [*to_json(acc), i]
```
why not this to avoid bugs in future?

selmanozleyen (approved):
Looks good! Only one code dedup I mentioned. Something I found here exposed another problem but I mentioned that in the other PR here: https://github.com/scverse/anndata/pull/2605

src/anndata/acc/_parse_json.py
flying-sheep: awesome, yeah!

src/anndata/acc/acc-schema-v1.json
flying-sheep: hmm, the root schema got more permissive, but it now accepts a strict superset of what it accepted before …

I don’ think it’s worth a v2 so soon after releasing it, I doubt anyone is using it yet.

PR #2628, opened 2026-09-01
chore: disable zarr v2 writing (at least as far as we can control it)

ilan-gold:
2 reasons to do this:

1. Fewer tests mean faster completion
2. No one should be writing v2 data anymore given that all reasonable data is in-spec writable and spatial data has upgraded (and anndataR can now read v3 data as well: https://anndatar.scverse.org/reference/write_zarr.html?q=v3#zarr-format)

I'm setting this as a 0.13.4 to remind myself to backport the functionality by deprecating instead of removing the setting. I will then add a test for that as well as a sensible release note for the 0.13.4 release.

- [ ] Closes #
- [ ] Tests added


tests/test_readwrite.py
```diff
@@ -477,26 +476,17 @@ def check_compressed(key, value):
...
-        compressor = Blosc(cname="zstd", clevel=3, shuffle=Blosc.BITSHUFFLE)
-    else:
-        from zarr.codecs import ZstdCodec
-
-        # Don't use Blosc since it's defaults can change:
-        # https://github.com/zarr-developers/zarr-python/pull/3545
-        compressor = ZstdCodec(level=3, checksum=True)
+    # Don't use Blosc since it's defaults can change:
+    # https://github.com/zarr-developers/zarr-python/pull/3545
+    compressor = ZstdCodec(level=3, checksum=True) if use_compression else None
```
flying-sheep, line 489: FYI: When resolving merge conflicts, I fixed this too, the `if ...` part was missing and `use_compression` did nothing.

Please check if that makes sense, or if we e.g should just remove `use_compression`.

src/anndata/_io/zarr.py
```diff
@@ -247,14 +245,6 @@ def read_dataframe(group: zarr.Group | zarr.Array) -> RWAble:
         return read_elem(group)
 
 
-def open_write_group(
-    store: StoreLike, *, mode: AccessModeLiteral = "w", **kwargs
-) -> zarr.Group:
-    if "zarr_format" not in kwargs:
-        kwargs["zarr_format"] = settings.zarr_write_format
-    return zarr.open_group(store, mode=mode, **kwargs)
```
flying-sheep, line 255: I took the liberty to remove this, without zarr v2 this is just `open_group` with `mode="w"` as default.

flying-sheep (approved):
Looks good, and simple enough!

PR #2634, opened 2026-09-04
chore: use memory store in tests

ilan-gold:
I try to resolve the problem "what if our zarr tests didn't take so long and we also didn't write SO much data during testing" by moving to memory stores where possible - there is still a test explicitly for local storage in `tests/test_readwrite.py::test_readwrite_equivalent_h5ad_zarr` (or where appropriate to not change) but otherwise, this should provide a big speedup!

- [ ] Closes #
- [ ] Tests added

- [x] Release note not necessary because: testing change

### DIFF
--- a/src/anndata/experimental/merge.py
+++ b/src/anndata/experimental/merge.py
@@ -13,6 +13,7 @@ import numpy as np
 import pandas as pd
 import zarr
 from scipy.sparse import csc_matrix, csr_matrix
+from zarr.abc.store import Store
 
 from .._core.file_backing import to_memory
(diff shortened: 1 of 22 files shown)


ilan-gold:
Not too shabby! https://github.com/scverse/anndata/actions/runs/33877901352/job/101039244161 for example took 13min but now it runs in 9min!

flying-sheep (approved):
very nice! I tried doing the same for HDF5, but it seems to give no advantage, so I just backported the helper from there.

PR #2642, opened 2026-09-08
feat: compat profiles for writing

### LINKED ISSUE #2057: Compatibility profiles for writing

The idea would be that the `write_{h5ad,zarr}` functions can be made to write backwards-compatible files. 

This would prevent issues like #2039 and (the `.uns` part of) #2052.

flying-sheep:
OK, let’s get this going, I think it’s nicer for users than individual feature gates.

This is a rather low-engineering version:

- no integration with the registry (only `settings.override`)
- individual features don’t live in the data structure, but rather there’s a version comparison wherever something’s gated on
- therefore the docs are manually written and not autogenerated

we might go with this and refactor as we go or make it more sophisticated from the start, what do you think @ilan-gold?

----

- [x] Closes #2057
- [x] Tests added


flying-sheep:
@ilan-gold could you please take a look?

docs/write-compat.md
```diff
@@ -0,0 +1,67 @@
+# Write compatibility profiles
+
+Every so often, `anndata` changes what it writes to disk:
+it starts using an encoding that older versions can’t read,
+or it starts rejecting output that violates the {doc}`on-disk spec <fileformat-prose>`.
+
+{attr}`anndata.settings.write_compat` selects which anndata version’s write behavior to target,
```
ilan-gold: What's the use-case of making this a setting globally?

docs/write-compat.md
```diff
@@ -0,0 +1,67 @@
...
+  - Relaxes both restrictions added in 0.13:
+    3-dimensional {attr}`~anndata.AnnData.X`/{attr}`~anndata.AnnData.layers` ({issue}`2430`) and
+    `/` in `h5ad` keys of `obs`, `var` and `uns` ({issue}`2039`) can be written again.
+* - `"0.13"`
+  - 2026-07-07
+  - No relaxations. This is the default.
+```
+
+Both relaxations produce files that violate the spec – anndata wrote such files before 0.13
+and still reads them, but other implementations may not.
```
ilan-gold:
```suggestion
and still reads them, but other older implementations may not.
```
right? no just any other implementation (like in rust)

docs/release-notes/2642.feat.md
```diff
@@ -0,0 +1,3 @@
+Add write compatibility profiles: {attr}`anndata.settings.write_compat` (and a `compat` argument on the write functions and {meth}`~anndata.AnnData.unwriteable`) selects which anndata version’s write behavior to target, see {doc}`/write-compat`.
+Passing `compat="0.12"` re-allows writing 3-dimensional {attr}`~anndata.AnnData.X`/{attr}`~anndata.AnnData.layers` ({issue}`2430`) and `/` in `h5ad` keys ({issue}`2039`).
+{attr}`anndata.settings.disallow_forward_slash_in_h5ad` is deprecated in favor of it {smaller}`P Angerer`
```
ilan-gold: I don't think we should go backwards here - the former violates our spec and the second is pretty absurd, but has a solution (the setting). What do you think? I thought this PR would be restricted to a method of introducing changes to our on-disk spec incrementally and more easily.

ilan-gold: For example, introducing the JSON-in-uns update without having to completely bump our full spec i.e., element-level changes

PR #2652, opened 2026-09-17
fix: raise on float index arrays with any non-integer value

### LINKED ISSUE #2651: Float index arrays are silently truncated unless every value is a non-integer

### Report

Follow-up to #1735 / #1746. A float index array only raises when every value is a non-integer. If at least one value is a whole number, the rest get truncated silently, e.g. `adata[np.array([0.5, 2.0])]` returns observations 0 and 2.

The check in `_from_array` (`src/anndata/_core/index.py`) is

```python
if xp.all((indexer - indexer_int) != 0):
    raise IndexError(...)
```

I think that should be `xp.any`. The 0.11.2 release note says "Error out on floating point indices that are not actually integers", and numpy raises for any float index. `test_index_float_sequence_raises_error` only uses arrays where all values are non-integers, so it doesn't catch this.

Whole-number floats like `[0.0, 2.0]` still work with `any`, so that part wouldn't change. Happy to open a PR.

Code:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///

import anndata as ad
import numpy as np

adata = ad.AnnData(np.arange(12, dtype="float32").reshape(4, 3))

for index in ([0.5, 1.5], [0.5, 2.0], [0.9, 1.0]):
    try:
        print(index, "->", list(adata[np.array(index)].obs_names))
    except IndexError as e:
        print(index, "-> IndexError:", e)
```

Output:

```
[0.5, 1.5] -> IndexError: Indexer array([0.5, 1.5]) has floating point values.
[0.5, 2.0] -> ['0', '2']
[0.9, 1.0] -> ['0', '1']
```

Same output on 0.13.3.post0 from PyPI.

### Versions

| Package | Version                 |
| ------- | ----------------------- |
| anndata | 0.14.0.dev56+g4ded337f9 |
| numpy   | 2.5.3                   |
| pandas  | 3.0.5                   |
| scipy   | 1.18.1                  |

| Component | Info                                                                        |
| --------- | --------------------------------------------------------------------------- |
| Python    | 3.13.13 (main, Apr  7 2026, 18:19:01) [Clang 21.0.0 (clang-2100.0.123.102)] |
| OS        | macOS-26.6.2-arm64-arm-64bit-Mach-O                                         |
| Updated   | 2026-09-17 08:47                                                            |

haeganm:
- [x] Closes #2651
- [x] Tests added

`_from_array` only raised for a float index array when every value was a non-integer (`xp.all`). An array like `[0.5, 2.0]` passed the check and was truncated to `[0, 2]`. Switched to `xp.any`, which matches numpy and the 0.11.2 release note for #1746.

Added partially-integral cases to `test_index_float_sequence_raises_error` (they fail on `main`), plus a small test that whole-number float arrays like `[1.0, 3.0]` still index as before.


ilan-gold (approved):
Big oops, thanks for the catch!

